PR #8, opened 2021-09-28
add example express-js-live-cursors

ofoucherot:


examples/express-javascript-live-cursors/static/app.js
```diff
@@ -0,0 +1,1977 @@
+(() => {
```
GuillaumeSalles, line 1: This file should not be committed. Is it because the .gitignore is missing?

ofoucherot: Fixed

GuillaumeSalles (approved):

PR #18, opened 2021-10-15
v4 server api integration

GuillaumeSalles:
Server always create an empty root so we useDefaultStorage root only if some keys are missing

### DIFF
--- a/packages/liveblocks-react/src/index.tsx
+++ b/packages/liveblocks-react/src/index.tsx
@@ -347,11 +347,20 @@ export function useStorage<TRoot extends Record<string, any>>(): [
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react/src/index.tsx
```diff
@@ -385,10 +394,21 @@ export function useMap<TKey extends string, TValue>(
 }
 
 /**
- * Returns the LiveList associated to the provided key. If the LiveList does not exists, a new empty LiveList will be created.
+ * Returns the LiveList associated to the provided key. If the LiveList does not exists, a new LiveList will be created.
```
ofoucherot, line 397:
```suggestion
 * Returns the LiveList associated with the provided key. If the LiveList does not exist, a new LiveList will be created.
```

GuillaumeSalles: Done :)

PR #24, opened 2021-11-03
Ignore getStorage promise in useStorage if hook is unmounted

GuillaumeSalles:
`useStorage` hook is calling an asynchronous function to initialize the storage. We need to ignore the promise callback to make sure that a state change is not triggered once a components is unmounted.

For more information:

https://overreacted.io/a-complete-guide-to-useeffect/


GuillaumeSalles:
Thank you @bluematter for finding the root cause 🙇

ofoucherot (approved):

PR #26, opened 2021-11-05
Unified batching / undo / redo between presence and storage

GuillaumeSalles:
### Batching 
It's now possible to batch modifications to the storage and presence with `room.batch`. 

If you modify the same item multiple times during a batch, the subscribers are called only once.

``` typescript
room.batch(() => {
  animals.delete(0);
  animals.push("🐱");
  animals.push("🐶");
});
// We notify animals subscribers only once.
```

It also waits for the end of the batch to notify all the subscribers. It is useful when you want to ensure state consistency between different parts of the storage and presence. 

Let's consider a design tool (like Figma). When drawing a rectangle, the shape is added to the storage and is selected at the same time. The code would look something like this:

``` typescript
room.batch(() => {
  const newRectangle = { id: uuid(), color: "blue", x: 0, y: 0, height: 100, width: 100 };
  shapes.push(newRecangle);
  room.updatePresence({ selectedId: newRectangle.id });
});
```

The batch ensures that the storage and presence modifications are done in the same frame for the local users and the other users in the room.

Batching also impacts the undo/redo history. If you batch multiple operations at once, it will add a single entry to the undo/redo history.

### Undo/redo for presence and storage

`room.history.undo` and `room.history.redo` lets you control the undo / redo stack for local operations. Undo/redo does not impact the modifications made by other users.

It's also possible to add the current user presence to the history with an additional argument on `room.updatePresence`. 

For example, if we consider a design tool (like Figma), the selection is added to the undo history. You can do it with: 
```typescript
room.updatePresence({ selectedId: "rectangle-id" }, { addToHistory: true });
```

It's also possible to ignore modifications with `room.history.pause` and room.history.resume.

To continue the parallel with a design tool, when you move a rectangle, you usually don't want to add the rectangle positions to the undo/redo stack while the pointer is still pressed.

```typescript
function onPointerDown(event) {
  room.history.pause();
  // start draging
}

function onPointerMove(event) {
  // drag rectangle
}

function onPointerUp(event) {
  // finish draging
  room.history.resume();
  // If you undo
}
```

#### room.batch
Batches modifications made during the given function.
All the modifications are sent to other clients in a single message.
All the subscribers are called only after the batch is over.
All the modifications are merged in a single history item (undo/redo).

#### room.history.undo
Undoes the last operation executed by the current client. It does not impact operations made by other clients.

#### room.history.redo
Redoes the last operation executed by the current client. It does not impact operations made by other clients.

#### room.history.pause
All future modifications made on the Room will be merged together to create a single history item until `resume` is called.

#### room.history.resume
Resumes history. Modifications made on the Room are not merged into a single history item anymore.

#### room.updatePresence options
`room.updatePresence` accepts a new optional arguments to add the presence modifications to the undo/redo stack. See new features explained above.

#### useRoom
Returns the room associated with the current RoomProvider.

#### useUndo
Returns a function that undoes the last operation executed by the current client.
It does not impact operations made by other clients.

#### useRedo
Returns a function that redoes the last operation executed by the current client.
It does not impact operations made by other clients.

#### useHistory
Returns `room.history` of the room associated with the current RoomProvider.

### Breaking changes from 0.12.3

`subscribe`, `subscribeDeep`, `unsubscribe` and `unsubscribeDeep` have been deprecated on LiveObject, LiveMap, and LiveList. Instead, we use `room.subscribe` to object storage data structures. Subscribe now returns a callback to unsubscribe.

Example with a subscription on a LiveObject

#### Before
```typescript
function onChange() {};
liveObject.subscribe(onChange);
liveObject.unsubscribe(onChange);
```

#### After
```typescript
const unsubscribe = room.subscribe(liveObject, () => { });
unsubscribe();
```

Example with a deep subscription on a LiveObject

#### Before
```typescript
function onChange() {};
liveObject.subscribeDeep(onChange);
liveObject.unsubscribeDeep(onChange);
```

#### After
```typescript
const unsubscribe = room.subscribe(liveObject, () => { }, { isDeep: true });
unsubscribe();
```

Example when `subscribe` was used in the context of React.

### Before
```typescript
const [layerData, setLayerData] = useState(layer.toObject());
useEffect(() => {
  function onChange() {
    setLayerData(layer.toObject());
  }
  layer.subscribe(onChange);
  return () => layer.unsubscribe(onChange);
}, [layer]);
```

### After
```typescript
const [layerData, setLayerData] = useState(layer.toObject());
const room = useRoom();
useEffect(() => {
  return room.subscribe(layer, () => setLayerData(layer.toObject()));
}, [layer]);
```


packages/liveblocks/src/LiveObject.ts
```diff
@@ -0,0 +1,380 @@
...
+      reverse = [
+        { type: OpType.DeleteObjectKey, id: this._id!, key: key as string },
+      ];
+    } else {
+      reverse = [
+        {
+          type: OpType.UpdateObject,
+          id: this._id!,
+          data: { [key]: previousValue },
+        } as any, // TODO
```
ofoucherot, line 142: Do we want to merge the TODO comment?

examples/react-multiplayer-drawing-app/public/index.html
```diff
@@ -0,0 +1,43 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
+    <meta
+      name="description"
+      content="Web site created using create-react-app"
+    />
+    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
```
ofoucherot, line 12: We should use the Livelbocks logo

examples/react-multiplayer-drawing-app/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "react-multiplayer-drawing-app",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@liveblocks/client": "^0.12.0",
+    "@liveblocks/react": "^0.12.0",
+    "@testing-library/jest-dom": "^5.14.1",
```
ofoucherot, line 8: We should move these packages to dev dependencies

ofoucherot (approved):
small comments but lgtm

PR #59, opened 2022-02-08
V0.15: zustand / redux middlewares and better integration with React immutability model

GuillaumeSalles:
## Motivation

- Reduce the friction as much as possible to transform an existing react single player app into a multiplayer one
- Reduce the impact of Liveblocks on an existing code base
- Better integration with React immutability model
- Meet developers where they are and integrate with existing state management libraries. 

## Solution

Create utilities to sync an immutable state with existing Liveblocks data structures. Reuse these utilities to build middleware for [zustand](https://github.com/pmndrs/zustand) and [redux](https://redux.js.org/), and create better react hooks.
Expose underlying room to let the developers create multiplayer-specific features (cursors, undo/redo, offline, etc).

## How to sync immutable state with Liveblocks data structure

Screen Shot 2022-02-08 at 11 10 34 AM

If the developer updates the color of the first shape, the new immutable state will look like this:

Screen Shot 2022-02-08 at 11 10 44 AM

The resulting operation on Liveblocks data structure will be:

`root.get("shapes").get(0).set("color", "green")`.

For complex operations like `LiveList.move`, it's still possible to access the underlying data structure and use it directly to make sure the conflicts are solved in a predictable way.

### How it works

Screen Shot 2022-02-07 at 11 26 59 PM

Current trade-off: Developer can only sync one room at a time.

Open question: should we create a specific middleware for the multiroom to keep the API of most common case clean or should we try to make a more flexible mapping / type option?

### What's missing for a beta

TODO


a-type:
I'm curious about what you mean for multi-room use cases. Is there anything preventing the user from creating multiple Zustand stores to connect to each Liveblocks room they want to use at once?

a-type:
Also, it would be nice to have a way to bind `others` to storage as well so we can react to changes in others presence state.

a-type:
> Were you looking for something else?

Ah, no, I did end up finding that, thanks!

I did some cursory investigations into using this yesterday. Ultimately, though, I found the ergonomics of Zustand didn't fit well with how our data is structured. I do think this integration is really cool though, just won't be able to provide much in-depth feedback since I don't plan to adopt it. I can give you some notes on my experience if you're interested, but I'm not sure our app is the target for this feature after all - the hooks work well enough for us.

PR #63, opened 2022-02-09
feat: @liveblocks/client works in non-dom environments

martolini:
Inspired by `yjs` and `hocuspocus`, one should be able to interact with the same storage interface from a node environment. This allows for a `WebSocket` polyfill to be passed in when entering a room.

`yarn add @liveblocks/client ws`

```ts
import { createClient } from '@liveblocks/client';

const client = createClient({ publicApiKey: 'your-key' });
const room = client.enter('room-id', { WebSocketPolyfill: require('ws') })

const storage = await room.getStorage();
const map = storage.getMap('mymap');
map.set('a', 1);
```

### DIFF
--- a/packages/liveblocks/src/client.ts
+++ b/packages/liveblocks/src/client.ts
@@ -49,7 +49,8 @@ export function createClient(options: ClientOptions): Client {
     options: {
       defaultPresence?: Presence;
       defaultStorageRoot?: TStorageRoot;
-    } = {}
+      WebSocketPolyfill: typeof WebSocket
+    } = {WebSocketPolyfill: WebSocket}
   ): Room {
     let internalRoom = rooms.get(roomId);
     if (internalRoom) {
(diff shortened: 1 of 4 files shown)


GuillaumeSalles:
@martolini Sorry for not getting back to you sooner. That's awesome! Thank you so much for taking the time to make this PR. 

I'm going to merge it and do some adjustments on my end to not make you wait too much. I'll add another option for the fetch polyfill, do some e2e tests, and add error messages if no polyfills are detected in a non-dom env. I'll release a new version in the next 48h.

martolini:
Sounds great. Looking forward to it. 🚀

GuillaumeSalles:
cc @martolini @Pondorasti

Just to let you know, that I merged a [PR](https://github.com/liveblocks/liveblocks/pull/74) to support fetch and WebSocket polyfills. It was a bit more complicated than anticipated due to potential breaking changed with Next.js SSR.

You can see one of our e2e test that use the client in a node environment as an example: https://github.com/liveblocks/liveblocks/blob/main/e2e/node-sandbox/index.test.js

I'm going to write some documentation and release 0.14.1 tomorrow morning. Thanks again for setting up the initial PR!

PR #71, opened 2022-02-17
fix: LiveObject conflicts when creating/deleting key

ofoucherot:
To prevent flickering, we don't apply a remote op if a local op on the same key has not been acknowledged yet.
Missing pieces in that mechanism cause an issue where clients' states are not consistent when local and remote delete/create operations on a LiveObject happen at the same time.

This PR:
- properly update the local acknowledgment cache when a key is deleted.
- properly update the local acknowledgment cache when a crdt is added as a nested element.
- add e2e fuzzy tests for LiveObject with undo/redo and nested objets.


packages/liveblocks/src/LiveObject.ts
```diff
@@ -136,12 +136,36 @@ export class LiveObject<
...
+
+      return { modified: false };
+    }
+
+    if (isLocal) {
+      this.#propToLastUpdate.set(key as string, opId);
+    } else if (this.#propToLastUpdate.get(key as string) === undefined) {
+      // Remote operation with no local change => apply operation
+    } else if (this.#propToLastUpdate.get(key as string) === opId) {
+      // Acknowlegment from local operation
```
ofoucherot, line 161: It should never be possible to get here.

PR #84, opened 2022-03-16
Update issue templates

captaindev404:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
```
GuillaumeSalles, line 15: It would be nice if the default template was a bit more edited. We're mainly providing APIs so the "Go to", "Scroll down to" are not really in context.

I would change this section by something like this:

> For example, here are some steps that would be useful to us:
> 1. Initialize storage
> 2. Stop internet connection
> 3. Restore internet connection
> 4. The storage is not in sync

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
```
GuillaumeSalles, line 24: Maybe screenshots or videos?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Environment (please complete the following information):**
```
GuillaumeSalles, line 26: I would also add packages used [e.g @liveblocks/react and @liveblocks/client]

.github/ISSUE_TEMPLATE/example-bug-report.md
```diff
@@ -0,0 +1,35 @@
+---
```
GuillaumeSalles, line 1: What's the difference between this file and `bug_report`?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
```
GuillaumeSalles, line 1: I think this feature request template would be a bit better if it was more related to APIs and not GUI app

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+name: Feature request
+about: Suggest an idea for this project
+title: "[FEATURE_REQUEST]"
+labels: feature request
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
GuillaumeSalles, line 11: Related to the comment above:

A clear and concise description of what the problem is. Ex. Almost every time I make use of `useOthers` I need to check if [...]

GuillaumeSalles (commented):
Thanks for setting that up! I would make the default template a bit more related to Liveblocks but other than that, I think it will be useful to new contributors!

.github/ISSUE_TEMPLATE/example-bug-report.md
captaindev404: This file allows creating issues on our Liveblocks examples.
I can delete it if we don't want to split the issue declaration between the liveblocks packages and examples.

.github/ISSUE_TEMPLATE/bug_report.md
captaindev404: I'll improve this one.

.github/ISSUE_TEMPLATE/example-bug-report.md
captaindev404: I've removed this template. 
It can be confusing for our users to have a specific template for liveblocks examples.

.github/ISSUE_TEMPLATE/bug_report.md
captaindev404: I added the real-world examples with version: @liveblocks/react@0.15.1

.github/ISSUE_TEMPLATE/bug_report.md
captaindev404: I added the proposal for videos and a tips to upload directly the file in the issue.

.github/ISSUE_TEMPLATE/bug_report.md
captaindev404: I've changed the template using a single scenario for all the paragraphs.

.github/ISSUE_TEMPLATE/feature_request.md
captaindev404: I've updated the template with the `Room.id` scenario to illustrate what kind of information we are waiting for.

.github/ISSUE_TEMPLATE/feature_request.md
captaindev404: Do you have specific questions in mind?

GuillaumeSalles (approved):
It's way better! Thank you for making all these adjustments! Giving more steps adapted to Liveblocks will help user understand what we're expecting from new issues :) 

Thank you @captaindev404 !

.github/ISSUE_TEMPLATE/feature_request.md
GuillaumeSalles: It's more clear now than it's showing an example related to one of our APIs

PR #92, opened 2022-03-23
fix: retry policy on Websocket errors

ofoucherot:
- Always retry connecting to websocket server, except if code is 4999 (custom Liveblocks code).
- If code is >= 4000 and <= 4100 (application server errors), retries are spaced.


packages/liveblocks-client/src/room.ts
```diff
@@ -952,30 +953,38 @@ See v0.13 release notes for more information.
...
-          console.error(
-            `Connection to Liveblocks websocket server closed. Reason: ${error.message} (code: ${error.code})`
-          );
-        }
         listener(error);
       }
-    } else if (event.wasClean === false) {
+
+      const delay = getRetryDelay(true);
       state.numberOfRetry++;
```
ofoucherot, line 967: numberOfRetry should be incremented after

Issue #97, opened 2022-03-25
@liveblocks/client and @liveblocks/react does not work Vite


GuillaumeSalles:
**Describe the bug**

Using @liveblocks/react and @liveblocks/client with Vite does not work. Vite bundles `@liveblocks/client` in `@liveblocks/react` during pre-bundling which causes duplicated classes in final bundle. Vite behavior is different from webpack configuration in Next or `create-react-app`.

**To Reproduce**

- Create a new Vite project with react
- Setup liveblocks
- Try using `useList`, `useMap` or `useObject`
- It will throw the obscure error `[object Object] is not a valid event name`

**Expected behavior**

`@liveblocks/client` and `@liveblocks/react` should work with Vite by default

**Environment (please complete the following information):**

- `@liveblocks/react@0.15.5` `@liveblocks/client@0.15.5`
- Vite

GuillaumeSalles:
It seems that adding this to Vite config fixes the issue. Still need to dig more if it's an issue with our bundling or with Vite interpretation of peer dependencies.

```js
optimizeDeps: {
  exclude: ["@liveblocks/react"],
},
```

GuillaumeSalles:
Fixed by #98

PR #98, opened 2022-03-26
@liveblocks/react: esm and mjs build to fix vite integration

GuillaumeSalles:
Update packaging to build `@liveblocks/react` for ESM. 

Vite pre-building bundle `@liveblocks/client` CJS in `@liveblocks/react` even if it's declared as peer dependency (is it a Vite issue?). Duplicated `@liveblocks/client` introduces issues with `instanceOf`. 

Having both `@liveblocks/client` and `@liveblocks/react` as ESM fixes the issue.

### DIFF
--- a/packages/liveblocks-react/rollup.config.js
+++ b/packages/liveblocks-react/rollup.config.js
@@ -1,19 +1,108 @@
-import typescript from "rollup-plugin-typescript2";
+/**
+ * This configuration is heavily inspired from https://github.com/pmndrs/zustand/blob/main/rollup.config.js
+ */
+import path from "path";
 import resolve from "@rollup/plugin-node-resolve";
+import babelPlugin from "@rollup/plugin-babel";
+import typescript from "@rollup/plugin-typescript";
+import esbuild from "rollup-plugin-esbuild";
+import dts from "rollup-plugin-dts";
+import { promises } from "fs";
+const createBabelConfig = require("./babel.config");
 
-import pkg from "./package.json";
+const extensions = [".tsx"];
 
-export default {
-  input: "src/index.tsx",
-  output: [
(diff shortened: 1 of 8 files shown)


GuillaumeSalles:
Fixes #97

packages/liveblocks-react/rollup.config.js
```diff
@@ -1,19 +1,108 @@
...
+  await promises.rm(`./lib`, {
+    recursive: true,
+    force: true,
+  });
+
+  return [
+    ...createDeclarationConfig("src/index.tsx", "index"),
+    createCommonJSConfig("src/index.tsx", "index.js"),
+    createESMConfig("src/index.tsx", "esm/index"),
+  ];
```
nvie, line 107: Hmm. Not sure what's going on here without scrutinizing the exact changes here. This makes the `rollup.config.js` diverge a lot from the other 4 packages though, which is a bit unfortunate. I've kept all of the rest 99% the same. One thing I'd love to try is to use `tsup` for bundling our packages. It supposedly ships with great community defaults and best practices for publishing TypeScript projects as ESM/CJS in a near-zero config way, is super fast to build, and is compatible with a proper NPM monorepo setup, which would also be on my wish list (aka bye bye `link-liveblocks.sh`). It would allow us to get rid of all this manual rollup and plugin configuration we are currently juggling.

nvie: @GuillaumeSalles From just looking at this PR's diff, I don't understand the cause of the Vite bug and/or the fix for it. I would probably be able to dig in by setting up a Vite project myself and diving in from there, but I'll spare myself the effort and time. When you're back from your holiday, I'd love to better understand the cause/fix made here, though.

Issue #87, opened 2022-03-16
Remove private field usage to simplify bundling


GuillaumeSalles:
We're currently using [private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#private_fields) and [private methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#private_methods). 

While this is better for encapsulation, they are not widely supported (less than 90% based on Caniuse). Babel injects a lot of code on hot paths which arguably outweighs the benefit of encapsulation. Removing private fields and methods in favor of `@internal` to hide them from auto-completion would improve our final bundle.

nvie:
This was addressed in #106.

PR #106, opened 2022-03-31
@liveblocks/client: Remove private methods and fields

GuillaumeSalles:
#87


packages/liveblocks-client/src/LiveList.ts
```diff
@@ -20,15 +20,17 @@ type LiveListItem = [crdt: AbstractCrdt, position: string];
  */
 export class LiveList<T> extends AbstractCrdt {
   // TODO: Naive array at first, find a better data structure. Maybe an Order statistics tree?
-  #items: Array<LiveListItem> = [];
+  private _items: Array<LiveListItem>;
```
nvie, line 23: Minor nit: can we settle on using a consistent prefix? I think a `_` prefix works just fine, but noticed a double underscore `__` used in the other file.

packages/liveblocks-client/src/AbstractCrdt.ts
```diff
@@ -20,41 +20,41 @@ export interface Doc {
...
 
 export abstract class AbstractCrdt {
-  #parent?: AbstractCrdt;
-  #doc?: Doc;
-  #id?: string;
-  #parentKey?: string;
+  private __parent?: AbstractCrdt;
+  private __doc?: Doc;
+  private __id?: string;
+  private __parentKey?: string;
```
nvie, line 26: The rationale for this, as explained in #87, makes a lot of sense 👍

nvie: Do you think it’s worth to set up a custom ESLint rule for this, so future devs will automatically get a rejection when using private fields/methods, and see this explanation as the reason?

When publishing `decoders` v2.0.0, I forbid myself to use specific costly Babel constructs going forward, and got a [big drop in bundle size](https://github.com/nvie/decoders/blob/c3d0b413ac7f06adab8a83ee26ad97b169724831/docs/assets/size-comparison%402x.png) from it.

I enforced this upon myself by setting up an [ESLint rule](https://github.com/nvie/decoders/blob/51abb4050216d7fa96ba4e7c4dc77d42d4d5b7b7/.eslintrc.js#L75-L81) that forbids the use of the `for..of` loops in library code, and replaced those with `.forEach()` in most cases. A bit uglier, of course, but users benefit from smaller bundles (and a faster runtime).

Codifying this decision hugely helps enforce this in the future, and acts as a reminder of the rationale for the disallowance at the same time, because the reason is explained in the error message.

nvie (approved):
Nice!

packages/liveblocks-client/src/LiveList.ts
GuillaumeSalles: The double `__` is ugly :/ 

There are collisions with some getters in `AbstractCrdt` and I didn't find a better prefix. 

Maybe there is a way to get rid of those protected getters but for the scope of the PR, I just wanted to remove babel extra code. I can create another task to find a better pattern for the base class to remove these double underscore for good.

packages/liveblocks-client/src/AbstractCrdt.ts
GuillaumeSalles: I like the idea of using Eslint rules to prevent some features that would increase the final bundle size! Definitely worth it as the team is growing.

packages/liveblocks-client/src/LiveList.ts
nvie: > _"There are collisions with some getters in AbstractCrdt and I didn't find a better prefix."_

Aha, I see. That wasn't clear to me at first — thanks for clarifying 👍

GuillaumeSalles:
@nvie @ofoucherot I'll push a fix tomorrow to fix the integrations tests

PR #107, opened 2022-03-31
Use `useRerender()` hook

nvie:
Changes the "increment hidden state var" trick into a new hook called `useRerender()`, which makes the intent behind this trick clearer, and call sites more minimal.

At first, I kept the `x => x + 1` logic in there, but then worried about potentially hitting the `MAX_SAFE_INTEGER` limit for long running apps, after which increments will no longer functionally work as intended:

Screen Shot 2022-03-31 at 09 24 32@2x

Instead, in this new implementation, I'm simply assigning a new empty object instance every time, which won't run into a max limit (and [isn't at all slower](https://jsben.ch/0fT7s)).

### DIFF
--- a/packages/liveblocks-react/src/index.tsx
+++ b/packages/liveblocks-react/src/index.tsx
@@ -10,6 +10,7 @@ import {
   BroadcastOptions,
 } from "@liveblocks/client";
 import * as React from "react";
+import useRerender from "./useRerender";
 
 type LiveblocksProviderProps = {
   children: React.ReactNode;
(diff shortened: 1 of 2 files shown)


GuillaumeSalles:
The useRerender hook makes it more clear 👍 

For the `MAX_SAFE_INTEGER` limit, it would take 285616 years at one increment per millisecond. Liveblocks and the client session would be older than humanity itself :D 

For the performance, my guess is that a number vs object assignment is negligible compared to what's happening if an update happens (Liveblocks & react). So, I don't mind if we use one or the other, you can choose the one that you find the more readable :)

nvie:
> For the `MAX_SAFE_INTEGER` limit, it would take 285616 years at one increment per millisecond. Liveblocks and the client session would be older than humanity itself :D

LOL, good point 🙈 

I reverted the implementation back to increments in 7c6437c2eb36bbc79bd58a7fada95cde002592ae, since I think those are definitely more readable 👍

ofoucherot (approved):

PR #109, opened 2022-03-31
Make `getOthers` return a standard array interface

nvie:
This changes the API of `getOthers()` to return a standardized (but readonly) array directly, rather than the custom array proxy object.

This means that:

```tsx
const others = useOthers();

others.toArray().some(u => u.presence?.isVIP);
//    ^^^^^^^^^^ No longer needed

others.some(u => u.presence?.isVIP);  // All array methods are directly available
```

Also, `others.count` can just become `others.length`.

The non-standard APIs `count` and `toArray()` are still retained by this change, but could be marked as `@deprecated` going forward, and be removed in a future release.

### DIFF
--- a/packages/liveblocks-client/src/types.ts
+++ b/packages/liveblocks-client/src/types.ts
@@ -128,7 +128,7 @@ export type AuthenticationToken = {
 // reasons only. We want to eventually deprecate these APIs in favor of
 // `length` and just accessing the array directly here.
 //
-type ReadonlyArrayish<T> = readonly T[] & {
+type ReadonlyArrayWithLegacyMethods<T> = readonly T[] & {
   // @deprecated
   readonly count: number;
   // @deprecated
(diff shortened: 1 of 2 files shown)


packages/liveblocks-client/src/types.ts
```diff
@@ -123,27 +123,24 @@ export type AuthenticationToken = {
   info?: any;
 };
 
+//
+// NOTE: The `count` and `toArray()` methods are here for backward-compatible
+// reasons only. We want to eventually deprecate these APIs in favor of
+// `length` and just accessing the array directly here.
+//
+type ReadonlyArrayish<T> = readonly T[] & {
```
ofoucherot, line 131: What do you think of the name: `ReadonlyArrayWithLegacyMethods` ?

nvie: Good suggestion, adopted in b6d6990.

ofoucherot (approved):
Nice improvement!

PR #113, opened 2022-04-06
Set all package targets to ES2015

nvie:
This explicitly sets all TypeScript compile targets to `es2015` (= alias of `es6`). In the case of `@liveblocks/react`, this means an upgrade from `es5` to `es2015`. This upgrade is necessary to avoid TypeScript errors like the one below:

Screen Shot 2022-04-06 at 09 39 37@2x

To avoid errors like above, the targeted version of `@liveblocks/react` (and other secondary packages) should never be higher than the targeted ES version of `@liveblocks/client`.

In most other packages, the targeted version wasn't specified explicitly, i.e. `esnext`, which may make the compiled code not useable for developers that cannot use the latest ES version yet. This PR effectively “downgrades” the targeted ES version in those packages.

Whether we pick `es2015`, or perhaps want to target a higher runtime is a trade-off between compiled bundle size and support in more code bases.

I therefore think `es2015` is a good pick for now, but lmk if you think we should aim higher here!

### DIFF
--- a/packages/liveblocks-react/tsconfig.json
+++ b/packages/liveblocks-react/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2015",
     "module": "esnext",
     "outDir": "lib",
     "rootDir": "src",
--- a/packages/liveblocks-react/tsconfig.json
+++ b/packages/liveblocks-react/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2015",
-    "module": "esnext",
+    "module": "es2020",
     "outDir": "lib",
     "rootDir": "src",
     "lib": ["es6", "dom", "es2016", "es2017"],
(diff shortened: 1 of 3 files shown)


GuillaumeSalles:
Targeting `es2015` makes sense! Thanks for adjusting that everywhere. 

I didn't verify but does it impact the bundle size? We're using babel for cjs and esbuild for esm and I don't think they're using tsconfig `target` option.

I didn't find great best practices for modern library bundling so most of it is inspired by https://github.com/pmndrs/zustand/blob/main/rollup.config.js

It's still not clear to me if our bundle should include some polyfills or we should assume that developers will setup global polyfill based on their target.

nvie:
Oh wow, you're right, this change doesn't have any impact on the build output at all. Seems like this setting is completely ignored from the bundle's perspective. However, the TypeScript error above goes away for me locally, more of an editor setting when interpreting the compiled `client/lib/index.d.ts` files when working inside `react/src/index.ts` locally then, I guess.

> I didn't find great best practices for modern library bundling so most of it is inspired by https://github.com/pmndrs/zustand/blob/main/rollup.config.js

Yeah, there are just too many options for bundling NPM packages. Each with so many settings and config files that it's dazzling.

nvie:
I guess it's fine to merge this in then, since there is no impact on the produced bundle here at all. It just fixes TypeScript errors in my local editor.

PR #122, opened 2022-04-07
Make `getOthers` return a standard array interface

nvie:
Take two for getting #109 merged.

See ad1413decb726b6775b417459d38ec88ef58939a for the necessary fix!

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -15,9 +15,15 @@ jobs:
         with:
           repository: liveblocks/liveblocks
           ref: refs/heads/${{ github.head_ref }}
-      - run: npm install
-      - run: npm run build
-      - run: npm run test-ci
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Ensure TypeScript compiles
+        run: npm run build
+
+      - name: Run unit tests
+        run: npm run test-ci
 
   e2e-tests:
     timeout-minutes: 60
--- a/packages/liveblocks-client/src/room.ts
+++ b/packages/liveblocks-client/src/room.ts
@@ -78,20 +78,19 @@ function makeOthers<T extends Presence>(presenceMap: {
     return publicKeys;
   });
 
-  return {
-    get count() {
-      return array.length;
-    },
-    [Symbol.iterator]() {
-      return array[Symbol.iterator]();
-    },
-    map(callback) {
-      return array.map(callback);
-    },
-    toArray() {
-      return array;
-    },
-  };
+  const arrayish: Others<T> = Object.assign(
+    array,
+
+    // NOTE: We extend the array instance with custom `count` and `toArray()`
+    // methods here. This is done for backward-compatible reasons. These APIs
+    // will be deprecated in a future version.
+    {
+      count: array.length,
+      toArray: () => array,
+    }
+  );
+
+  return Object.freeze(arrayish);
 }
 
 function log(...params: any[]) {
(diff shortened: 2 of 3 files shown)


GuillaumeSalles:
I'm a bit late on that one. It makes me realize that I've never documented why `Others` is not an array.

My initial thought was that transforming the internal map of users (`{ [connectionId: string]: User }`) into an array might impact performance for a high number of users moving their cursors. Updating users' internal map when a user updates their cursor position is `O(1)`. Transforming the map into an array makes it `O(n)` where `n` is the number of users in the room. Our examples are re-rendering all cursors every time a cursor is updated. Using the current `room.subscribe("others")` interface, it's possible to re-render a single cursor at a time in `O(1)`. The only missing piece is lazily building the array of users in `toArray` instead of eagerly. **_But hey, it might be a premature optimization; I didn't measure anything, and since `n` is always pretty small, it might not be an issue._** I would love to do some tests with 200 cursors to see if it has an impact or not. If not, I think what you're proposing is way more ergonomic for our users.

Another potential solution would be to return an array only for `useOthers` but not for `room.getOthers`. It might feel weird at first, but now that I'm thinking about it, `useOthers` will always trigger a re-render whenever any user is updated by the nature of how hooks work. And `room.getOthers` would be the escape path for some performance optimization (a similar trade-off made by our storage hooks).

Sorry for not bringing all that context sooner. Let me know what you think.

nvie:
> My initial thought was that transforming the internal map of users (`{ [connectionId: string]: User }`) into an array might impact performance for a high number of users moving their cursors. Updating users' internal map when a user updates their cursor position is `O(1)`. Transforming the map into an array makes it `O(n)` where `n` is the number of users in the room. Our examples are re-rendering all cursors every time a cursor is updated. Using the current `room.subscribe("others")` interface, it's possible to re-render a single cursor at a time in `O(1)`. The only missing piece is lazily building the array of users in `toArray` instead of eagerly.

Oh, I think I see now. My thinking went like this: [any time you call `makeOthers()`](https://github.com/liveblocks/liveblocks/blob/80e5dedd2b111d97a998297d4329a091e6c326fd/packages/liveblocks-client/src/room.ts#L393), you [immediately loop over all of its items anyway](https://github.com/liveblocks/liveblocks/blob/80e5dedd2b111d97a998297d4329a091e6c326fd/packages/liveblocks-client/src/room.ts#L395), so there would not be no gain in building that array lazily. But now that I'm looking more closely, I see that it's not `state.others` that's getting looped over, but `others`, which is a list of _events_! 🙈

Still, I'm reading [these lines](https://github.com/liveblocks/liveblocks/blob/80e5dedd2b111d97a998297d4329a091e6c326fd/packages/liveblocks-client/src/room.ts#L392-L393) as implying that `makeOthers()` is getting called any time there is an cursor update from one of the connected users already. So if this would have a perf impact, we would already perceive that problem today, not?

> **_But hey, it might be a premature optimization; I didn't measure anything, and since `n` is always pretty small, it might not be an issue._** I would love to do some tests with 200 cursors to see if it has an impact or not. If not, I think what you're proposing is way more ergonomic for our users.

Would love to see the measurement of the perf impact here 👍 

> Another potential solution would be to return an array only for `useOthers` but not for `room.getOthers`. It might feel weird at first, but now that I'm thinking about it, `useOthers` will always trigger a re-render whenever any user is updated by the nature of how hooks work. And `room.getOthers` would be the escape path for some performance optimization (a similar trade-off made by our storage hooks).

I like that option as a fallback, for sure!

Will revert this merge for now, and make sure to update some of the internal variable names here that caused my confusion.

PR #124, opened 2022-04-07
Automate publishing to NPM

### LINKED ISSUE #118: Automate package publishing to NPM
Publishing the packages to NPM involves [a lot of subtle steps](https://www.notion.so/liveblocks/How-to-074d3dfa5b944adb986fe176572d6887#9a10b3d58f014982a513f667ca5d67bd) that are currently carried out manually. There's a lot of room for error if you miss a subtle step or make a mistake.

By automating this process, we can perform various checks upfront, and help the package maintainers to more quickly release a version to NPM in the future.

nvie:
Fixes #118.

This adds a script that performs all the steps from [these instructions](https://www.notion.so/liveblocks/How-to-074d3dfa5b944adb986fe176572d6887#9a10b3d58f014982a513f667ca5d67bd).

This needs to be tested against more edge cases, but it's built in a way where failures of certain steps will not cause harm. The script will end and you can try running it again without issues.

The script is written in a self-checking way, so it will verify its own assumptions before taking action, and fail with a friendly error message otherwise. For example, it won't continue if you have local changes in your worktree, or if you're not up-to-date with the latest upstream version, or if you enter an incorrect version number, etc.

Once I have push access to NPM, I can give this script a spin!

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -15,6 +15,10 @@ err () {
     echo "$@" >&2
 }
 
+is_valid_version () {
+    echo "$1" | grep -qEe "^[0-9]+[.][0-9]+[.][0-9]+(-[[:alnum:].]+)?$"
+}
+
 usage () {
     err "usage: publish.sh [-V <version> [-t <tag>] [-h]"
     # err "usage: publish.sh [-Vtnh]"
@@ -155,6 +159,12 @@ check_npm_stuff_is_stable () {
 }
 
 check_all_the_things () {
+    if [ -n "$VERSION" ] && ! is_valid_version "$VERSION"; then
+        # Check for typos early on
+        err "Invalid version: $VERSION"
+        exit 2
+    fi
+
     check_git_toolbelt_installed
     check_jq_installed
     check_moreutils_installed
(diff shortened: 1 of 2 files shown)


scripts/publish.sh
```diff
@@ -0,0 +1,332 @@
...
+    if [ "$CURRENT_VERSION" != "$VERSION"; then
+        if ! git modified | grep -qEe package-lock.json; then
+            err "Hmm. package-lock.json wasn\'t affected by the version bump. This is fishy. Please manually inspect!"
+            exit 5
+        fi
+    fi
+}
+
+build_pkg () {
+    rm -rf lib
```
GuillaumeSalles, line 234: We're cleaning `lib` in every `build` scripts so I think we can make the assumptions that every build script should clean every previously built artifacts.

GuillaumeSalles:
That's amazing! Bash is not my cup of tea and I still find your script pretty clear! 

Would it be possible to add a `beta` flag? Could be useful for the first time you're going to publish using this script

GuillaumeSalles (approved):

scripts/publish.sh
nvie: Fantastic. Updated in 1fc1bb442bd49a74748ff617c5f76454c306dc2c.

nvie:
> That's amazing! Bash is not my cup of tea and I still find your script pretty clear!

Glad about that! There is a strict subset of the shell language that I have learned over the years that is safe to use, is platform independent, and avoids bad practices (like not wrapping all uses of variables in quotes, etc.). It doesn't make the scripts more readable, though 🙈 

> Would it be possible to add a `beta` flag? Could be useful for the first time you're going to publish using this script

You mean publish the package under a `beta` tag on NPM? Absolutely, will do! 👍

PR #127, opened 2022-04-07
[@liveblocks-react]: support react 18

ofoucherot:


packages/liveblocks-react/package.json
```diff
@@ -34,7 +34,7 @@
   "license": "Apache-2.0",
   "peerDependencies": {
     "@liveblocks/client": "0.15.10",
-    "react": "^16.14.0 || ^17"
+    "react": ">=16.14.0 || ^17 || ^18"
```
nvie, line 37: Yay for supporting React 18!

nvie: Btw, I think using `>=16.14.0` makes the remainder of the line moot, because all versions higher than 16 also match when using `>=`, including 17, 18, and even (in the future) 19. Is there a reason you changed to `>=` here that I'm missing?

nvie (approved):
Just left one comment about the version range spec, but otherwise looks good to me! Nice work 👍

packages/liveblocks-react/package.json
ofoucherot: It was my initial change and I forgot to revert it. IT's fixed now!

PR #128, opened 2022-04-08
Support nested `.envrc` files

nvie:
In recent PRs, I've included `.envrc` files in some of the project's directories. For those unfamiliar, these are config files for the amazing [direnv](https://direnv.net/) tool. When you use direnv on your machine, it will automatically load and unload settings when you `cd` into or `cd` out of a directory. I use it for all my projects, and it's amazing.

For example, when you `cd` into `packages/liveblocks-client`, it reads `layout node` which will put `packages/liveblocks-client/node_modules/.bin` on the `$PATH` for you. This means you can easily run locally-installed packages like `eslint`.

Also, when you `cd` out of that directory and into another directory, it will undo those `$PATH` changes automatically.

Can totally recommend it!

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,3 @@
+# Automatically put scripts/ in your $PATH, as long as you
+# are inside this repo
 PATH_add ./scripts/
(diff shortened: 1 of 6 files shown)


GuillaumeSalles (approved):
I didn't now direnv before you mentioned it. It looks pretty useful, I'll give it a try for sure!

PR #129, opened 2022-04-08
Automate local linking of packages

nvie:
This automates the linking step a bit to make local development a bit easier. As I'm going over the package sources to update some of the TypeScript types, I found it easy to get them linked to the incorrect versions. For example, when updating a type inside the `src/` in the client package, those changes won't show up in the React package, unless I would:

- Rebuild the client package
- Link them with NPM

This script automates that process.

To use it, simply `cd` into one of the secondary package directories, and run `../../scripts/link.sh`. (If you use direnv and have the set up from #128, you can simply run `link.sh` without the `../../scripts/` prefix from anywhere—can recommend!)

What this script will do:

1. Checks to see if the client package needs to be rebuilt. If determines that by looking at the file modification timestamps in `client/src` and `client/lib`. All files in `lib/` must be older than any file in `src/`. If not, it rebuilds things.
2. Checks to see if the current package has been linked locally correctly. If not, it will link it up.

Basically, this makes `link.sh` an idempotent script that you can run at any time and it will check your local setup and fix it if necessary.

## Future Work
I would love to generalize this script in another pass, to make it work with any of our local packages, meaning you could run this inside any example directory as well, and it would check and link _all_ Liveblocks packages that that project needs.


GuillaumeSalles (approved):
Pretty cool! We did a few scripts to link our packages to examples, maybe this could be unified in the future 🤔 

The tricky part is the peer dependencies and duplicated React packages: 
https://github.com/liveblocks/liveblocks/blob/main/examples/nextjs-whiteboard/scripts/link-liveblocks.sh

nvie:
> The tricky part is the peer dependencies and duplicated React packages: https://github.com/liveblocks/liveblocks/blob/main/examples/nextjs-whiteboard/scripts/link-liveblocks.sh

Yeah, I would love to automate that process by autodetecting what the "current" project needs are from looking at its `package.json` file, and doing The Right Thing™ automatically. That way, we wouldn't have to copy this script into each project either - you just `cd` into a project, run the link script, and you're good.

Will see if I can add that in before merging this PR in. Would be nice to have a single way to do it, rather than to have a mix of approaches between packages and projects.

ofoucherot (approved):
🚀

nvie:
It now works reliably with the CI build, too! 🙌 
I've gone ahead and updated the instructions here too: https://www.notion.so/liveblocks/How-to-074d3dfa5b944adb986fe176572d6887#0e1c03df0bee49e78d62dedfa4bf9856

PR #131, opened 2022-04-11
Let Rollup builds fail if there are TypeScript errors

nvie:
I noticed that sometimes the Rollup could throw TypeScript errors, but would happily continue. This adds a setting that will let those builds fail if such TypeScript errors are found, so we can fix them.


ofoucherot (approved):
🚀

PR #137, opened 2022-04-12
Improve type coverage (first pass)

nvie:
This PR makes a start at improving our type coverage. This is by far not done yet, but these updates are the low-hanging fruit that could be replaced without any issues, and remove some noise from the updates to come in upcoming PRs.

Also note the introduction of `fixme` alias as a temporary helper type. This is used in cases where `any` should not be used (because it disables the typechecker), but `unknown` isn't the _intended_ type and this can be further improved upon in a future pass.

In cases where `unknown` actually is the proper type to use, I've used `unknown`.

### DIFF
--- a/packages/liveblocks-client/src/immutable.ts
+++ b/packages/liveblocks-client/src/immutable.ts
@@ -32,7 +32,7 @@ function liveListToJson(value: LiveList<any>) {
   return value.toArray().map(liveNodeToJson);
 }
 
-export function liveNodeToJson(value: any): any {
+export function liveNodeToJson(value: unknown): any {
   if (value instanceof LiveObject) {
     return liveObjectToJson(value);
   } else if (value instanceof LiveList) {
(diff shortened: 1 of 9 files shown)


packages/liveblocks-client/src/immutable.ts
```diff
@@ -32,7 +32,7 @@ function liveListToJson(value: LiveList<any>) {
   return value.toArray().map(liveNodeToJson);
 }
 
-export function liveNodeToJson(value: any): any {
+export function liveNodeToJson(value: unknown): any {
```
nvie, line 35: Here, `unknown` is used intentionally. The point of this function is to take any value and as such its type is unknown (but it won't disable typechecking inside this function, like `any` does!).

packages/liveblocks-client/src/room.ts
```diff
@@ -1233,7 +1236,7 @@ See v0.13 release notes for more information.
   }
 
   function broadcastEvent(
-    event: any,
+    event: fixme,
```
nvie, line 1239: Here, the type of this param is known. I haven't figured out what it is exactly (yet), but in the meantime, let's at least not disable type checking on it. Hence the `fixme` type (which is aliased to `unknown`). It signifies that I will come back to this later, and add a proper type annotation. In the meantime, though, let's avoid the `any` here.

nvie:
Going to assume none of this is controversial, and taking the liberty of merging this in now. I have some local branches that all rely on these basic updates. Please do let me know if you disagree with anything in here, though! Happy to adjust after the fact. 🙏

PR #140, opened 2022-04-13
Make `getOthers()` API return an array directly

nvie:
This simplifies the API for the `getOthers()` call to not return a custom array proxy, but a real array. This was attempted to be merged in #109 and #122, but led to some debate about potentially closing off a future optimization path, see https://github.com/liveblocks/liveblocks/pull/122#issuecomment-1092444391.

I'm now reopening this PR as it stood to continue that discussion, because I think the API simpliciation would still be a better DX.


nvie:
Closing as this has landed in the 0.18.0 beta.

PR #141, opened 2022-04-13
Make running mass commands on all packages/examples easier to run

nvie:
This PR adds three helper scripts:

- `for-all-examples.sh <your cmd>`
- `for-all-packages.sh <your cmd>`
- `for-all-e2e.sh <your cmd>`

Which you can use as follows:

```sh
$ for-all-examples.sh npm install
```

Or to run prettier en-masse:

```sh
$ for-all-examples.sh prettier --write src
```

You can provide any command and this script will first `cd` into each directory and then run your command. The current working directory from where you invoke this script isn't relevant, it will run fine from anywhere in the repo.

By default, if one command fails, it immediately aborts the script. If you want to allow and ignore failures and just continue with the remainder of the directories, you can provide the `-f` (force) argument. Full usage:

```
$ for-all-examples.sh -h
usage: for-all-examples.sh [-qfh] <cmd> [<arg1> [<arg2> ...]]

Runs the given script inside all of our examples directories.

Options:
-q    Don't print the current directory
-f    Continue, even after an error occurs
-h    Show this help
```

### DIFF
--- a/scripts/for-all-e2e.sh
+++ b/scripts/for-all-e2e.sh
@@ -1,0 +1,57 @@
+#!/bin/sh
+set -eu
+
+# Ensure this script can assume it's run from the repo's
+# root directory, even if the current working directory is
+# different.
+ROOT="$(git rev-parse --show-toplevel)"
+if [ "$(pwd)" != "$ROOT" ]; then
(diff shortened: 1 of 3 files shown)


ofoucherot (approved):
Great!

PR #144, opened 2022-04-13
Bump NPM packages with security issues in all examples

nvie:
Like #136 , but now for all the example projects. This is the effect of running `npm audit fix` in all example projects.


ofoucherot (approved):
🔒

PR #146, opened 2022-04-13
fix: patchList when new array is more than 2 elements less than previous array

ofoucherot:
The issue is only impacting the Redux and Zustand integrations.
In your redux or Zustand state, if you try to replace an array of N elements with an array of N - 2 elements, it throws an error.
Example: [1,1,1] => [2].


nvie (approved):
Great find! Also thanks for the helpful explainer on Slack, that was really instrumental in understanding this change 👍

PR #150, opened 2022-04-14
Improve types of Live storage APIs

nvie:
This PR improves many TypeScript types, removes a lot of implicit/explicit `any`s in favor of more properly typed APIs to match the actual expectations.

There are a couple of notable introductions and changes, which I'll describe here.

### Introduction of `Json` type
First of all, I've added a proper `Json` type, which represents any legal JSON value (i.e. something that can reliably be sent to `JSON.stringify()`, or be returned from `JSON.parse()`). This is a recursive data structure, so this works no matter how deep your JSON tree is.

Similarly, there also exists `JsonArray` or `JsonObject`, which are like `Json`, but their outer value is an array or object respectively. Especially the latter is quite useful.

### Introduction of `Lson` type 😄 
Similar to the `Json` type above, I'm introducing a new sibling type which I've branded `Lson`[^1]. An `Lson` value is just like a `Json` value, except that it may also contain nested Live storage data structures (`LiveMap`, `LiveObject`, `LiveList`, etc.).

For example:
- `{ a: 1, b: 42, c: [{ d: null, e: true }] }` is both a legal `Json` and a legal `Lson` value
- `{ a: 1, b: 42, c: new LiveList([new LiveObject({ d: null, e: true })]) }` is NOT a `Json` value, but is a legal `Lson` value

### Live storage datastructure API types

The existing live storage APIs are now typed as follows:

- `class LiveList<TItem extends Lson = Lson>` (instead of `class LiveList<T>`)
- `class LiveObject<T extends LsonObject = LsonObject>` (instead of `class LiveObject<T extends Record<string, any> = Record<string, any>>`)
- `class LiveMap<TValue extends Lson = Lson>` (instead of `class LiveMap<TKey extends string, TValue>`, see note on breaking changes below)

### Breaking Changes
The `LiveMap<TKey extends string, TValue>` signature is now typed as `LiveMap<TValue extends LiveData>` — in other words, (1) the `TKey` type argument is removed and you are forced to use `string` as keys, and (2) the value provided must be an `Lson` shape. Put differently, this removes support for type definitions with string literals like `LiveMap<'foo' | 'bar', string>`. This simplifies the internals a lot. If such cases exist in the wild, users will need to change to `LiveMap<string, string>` or switch to use a LiveObject, like `LiveObject<{ foo: string, bar: string }>`. I think this matches the intended use, and may better illustrate the difference between these Live storage data structures: you use `LiveObject` for heterogeneous data, and `LiveMap` for homogeneous data (and are forced to use generic string keys).

[^1]: Finding a name to describe this type was hard 😅 ! I initially went with `LiveData` and `LiveObjectData`, but found those super confusing when seeing alongside the other types we have. For example, try to spot these types in a list like `LiveObject`, `LiveMap`, `LiveData`, `LiveObjectData`, `LiveRegister`, etc. Since they serve a different purpose than the `LiveXxx` data structures (which are CRDTs), I wanted the name to be clearly different. So I instead branded it as `Lson`, to make it more obvious they're more akin to `Json` than they are to `LiveXxx` instances. I realize that technically "LON" would be a more accurate acronym than "LSON" (because the "S" is from "Script" in the original acronym), but removing that "S" would completely destroy the association with "JSON", which is the point 😄).


packages/liveblocks-client/src/immutable.test.ts
```diff
@@ -172,12 +158,9 @@ describe("2 ways tests with two clients", () => {
...
     test("add item to array", async () => {
-      const { storage, state, assert } = await prepareStorageImmutableTest<
-        {
-          syncList: LiveList<string>;
-        },
-        { syncList: string[] }
-      >(
+      const { storage, state, assert } = await prepareStorageImmutableTest<{
+        syncList: LiveList<string>;
+      }>(
```
nvie, line 163: The (JSON) state now gets automatically derived from the provided (LSON) storage type. This means, further down on L172, `state.syncList` will get inferred as `string[]` automatically, from the provided `LiveList<string>` type here.

packages/liveblocks-client/src/index.ts
```diff
@@ -9,13 +9,17 @@ export type {
...
 
+export type { Json, JsonObject } from "./json";
+export type { Lson, LsonObject } from "./lson";
+
 export { createClient } from "./client";
 
 import {
   liveObjectToJson,
-  liveNodeToJson,
+  lsonToJson,
```
nvie, line 22: Changing the name for consistency here.

packages/liveblocks-client/src/LiveMap.ts
```diff
@@ -13,22 +13,23 @@ import {
...
+import { Lson } from "./lson";
 
 /**
  * The LiveMap class is similar to a JavaScript Map that is synchronized on all clients.
  * Keys should be a string, and values should be serializable to JSON.
  * If multiple clients update the same property simultaneously, the last modification received by the Liveblocks servers is the winner.
  */
-export class LiveMap<TKey extends string, TValue> extends AbstractCrdt {
-  private _map: Map<TKey, AbstractCrdt>;
+export class LiveMap<TValue extends Lson = Lson> extends AbstractCrdt {
```
ofoucherot, line 24: I don't have a strong opinion about it but I feel like it's more explicit when you have to declare the type of the key and value. Like it's clear that it's a map.

ofoucherot:
What do you think of `LiveJson`?

nvie:
> What do you think of `LiveJson`?

I considered that name, too! Like I explained [here](https://github.com/liveblocks/liveblocks/pull/150#user-content-fn-1-997633013506fc6eef19a183699a22ba), naming was hard initially and I found any name starting with `Live` to be confusing in combination with the CRDTs. I think it would be nice to avoid that confusion and keep the `Live`-prefix exclusively for the CRDTs. I liked the `Lson` name because it’s short (which is always nice for super common types), associates nicely with `Json`, and was a bit fun/playful.

That said… definitely open to name suggestions!

packages/liveblocks-client/src/LiveMap.ts
nvie: Yeah. The symmetry with `Map<K, V>` is definitely nice, and it initially feels like a weird sacrifice. However, with `Map<K, V>` you can use anything as keys (object instances, strings, numbers, etc).

But with a `LiveMap` you cannot, and this new type honestly reflects that. Since `LiveMap` keys can only really be `string`s (implied currently anyway by the `TKey extends string` constraint), as a user you must already always explicitly type/provide `string` as the first type param. If you have a case for limiting `TKey` to only certain string literals (think `LiveMap<'foo' | 'bar', X>`), I think a `LiveObject` would be a better fit by design.

Another goal here is to help users better pick which one to use, because right now it's a little unclear for what cases you would pick one over the other. The documentation doesn't help to make that choice, because both descriptions are the same:

Screen Shot 2022-04-15 at 19 47 51@2x

Hopefully this can help make it a bit clearer. I will definitely also update the docs to better help illustrate the differences and for what use cases to pick which data structure. Hope this helps to make this feel a little less weird.

ofoucherot:
> 
I will try to think of alternatives but I'm good with `Lson`!

packages/liveblocks-client/src/LiveMap.ts
ofoucherot: Yeah I agree that our API/documentation doesn't correctly explain when to use LiveMap vs LiveObject.

I'm still not sure about that breaking change. Do you think we could ask the community on Discord what they think about it?

GuillaumeSalles: I remember wanting to remove the "key" param at some point 😄 Here is the TODO I put a long time ago:

https://github.com/liveblocks/liveblocks/commit/608382dae3de7f0db0a762dca073337d762f3ba4#diff-273c9393df39cefe5414e6fc728a8da73aee11cce8507a7288ac8d22ea89a63dR523

For the backstory, the first implementation of the storage didn't include `LiveMap`. It was only `LiveObject` (called `LiveRecord` at the time)  and `LiveList`. I had to help a few early users that wanted `LiveObject` with dynamic keys (not fully known at compile time). I quickly realized `LiveObject<{ [key: string]: string }>` or `LiveObject<Record<string, string>>` were not intuitive for a significant percentage of typescript developers. Hence I created `LiveMap` for dynamic keys.

So why I finally decided to keep the first argument?

- Symmetry with `Map` in typescript, as mentioned by @ofoucherot. We're only missing the `clear` method (could be implemented but might require a special server op) and iterability-in-order (will probably never be implemented). 
- We might want to support key as `number` in the future. It's been a while since I tested this but I think it should be relatively easy to implement and wouldn't introduce a breaking change.
- We support template literal types. For example, it could be useful for a spreadsheet.

```
type SpreadsheetKey = `${Uppercase<string>}:${number}`
const livemap = new LiveMap<SpreadsheetKey, string>();

livemap.set("C:2", "Hey");
livemap.set("invalidKey", "Hey");
```

https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html


- Subjectively, having 2 generic arguments for the `LiveMap` and 1 for the `LiveObject` make `LiveMap.get` vs `LiveObject.get` or `LiveMap.set` vs `LiveObject.set` signatures intuitive to me without looking at it.
- Related to the first point, all the Map implementations I could find in different languages take 2 generic arguments. (Sometimes called HashMap or Dictionnary).

I'm a bit reluctant to introduce this breaking change because it might impact a significant percentage of our users, but happy to brainstorm the pros and cons if it makes the API more intuitive for future users. :)

packages/liveblocks-client/src/lson.ts
```diff
@@ -0,0 +1,73 @@
...
+import type { LiveMap } from "./LiveMap";
+import type { LiveObject } from "./LiveObject";
+import type { LiveRegister } from "./LiveRegister";
+import type { Json } from "./json";
+
+/**
+ * Think of Lson as a sibling of the Json data tree, except that the nested
+ * data structure can contain a mix of Json values and LiveXxx instances.
+ */
+export type Lson =
```
GuillaumeSalles, line 11: I love that! 😍 

Easy to understand and catchy name! So useful to make sure no one is passing a function or other non serializable values! There is probably a way to also use it on zustand mapping to make sure no one is trying to synchronize a function on the store.

packages/liveblocks-client/src/live.ts
```diff
@@ -52,7 +54,7 @@ export type UserLeftMessage = {
 export type EventMessage = {
   type: ServerMessageType.Event;
   actor: number;
-  event: any;
+  event: JsonObject;
```
GuillaumeSalles, line 57: To be tested but I think it could be a `Json`

packages/liveblocks-client/src/live.ts
```diff
@@ -83,7 +85,7 @@ export enum ClientMessageType {
 
 export type ClientEventMessage = {
   type: ClientMessageType.ClientEvent;
-  event: any;
+  event: JsonObject;
```
GuillaumeSalles, line 88: To be tested but I think it could be a `Json`

packages/liveblocks-client/src/immutable.ts
```diff
@@ -3,36 +3,54 @@ import { LiveList } from "./LiveList";
...
+function lsonListToJson(value: Lson[]): Json[] {
+  return value.map(lsonToJson);
+}
+
+function liveListToJson(value: LiveList<Lson>): Json[] {
+  return lsonListToJson(value.toArray());
 }
 
-export function liveNodeToJson(value: unknown): any {
+export function lsonToJson(value: Lson | AbstractCrdt): Json {
```
GuillaumeSalles, line 48: The new `lson` concept has a great side effect on naming! Really cool

GuillaumeSalles (commented):
Incredible work! Like I said in another comment, I love the Lson concept! Simple to understand a improve a lot our external API. I think we could even create a `Lson` to `Json` mapping type (with a lot of conditionals 😅 ) for our future immutable APIs.

I'm less sure about the `LiveMap` breaking change but happy to brainstorm :) 

Thanks so much for tackling this, it's a huge step forward to not use `any` anymore.

packages/liveblocks-client/src/LiveMap.ts
nvie: I think I went in a bit too aggressively there, and I will undo that change by reintroducing the `TKey extends string` type param there 👍 The arguments for retaining symmetry, enabling string literal types and future support for numeric keys make total sense, and I definitely did not intend to close those roads off!

nvie:
> I think we could even create a `Lson` to `Json` mapping type (with a lot of conditionals 😅 ) for our future immutable APIs.

You mean [this type](https://github.com/liveblocks/liveblocks/blob/bf26b554250b8a30ccda4454d464f477c70aeeb7/packages/liveblocks-client/src/lson.ts#L30-L73) 😉 ? It’s what I’m using for avoiding the double argument in our immutable tests, see also [this comment](https://github.com/liveblocks/liveblocks/pull/150/files#r851257066).

packages/liveblocks-client/src/lson.ts
nvie: > There is probably a way to also use it on zustand mapping to make sure no one is trying to synchronize a function on the store.

Can figure out, will be nice to include for sure!

packages/liveblocks-client/src/live.ts
nvie: This is fixed in 7b28f2542c7aaa2f68374cbcbce990b7dc31cf01. Also, I added an E2E test to test that broadcasting any `Json` value (not just objects) works. I couldn't find an existing test for the broadcast/event listener APIs, so I had to add a new one for it myself (see bd5dee884812a527adac5a95c0e4e46bd150b669).

packages/liveblocks-client/src/LiveMap.ts
nvie: Restored the key type in e45c468c7a107308d7ebab5ea567089351538581 — thx for the feedback 👍 !

GuillaumeSalles:
> > I think we could even create a `Lson` to `Json` mapping type (with a lot of conditionals 😅 ) for our future immutable APIs.
> 
> You mean [this type](https://github.com/liveblocks/liveblocks/blob/bf26b554250b8a30ccda4454d464f477c70aeeb7/packages/liveblocks-client/src/lson.ts#L30-L73) 😉 ? It’s what I’m using for avoiding the double argument in our immutable tests, see also [this comment](https://github.com/liveblocks/liveblocks/pull/150/files#r851257066).

I completely missed that type 🙈 Yes it's exactly what I had in mind! :chefkiss:

packages/liveblocks-react/src/index.tsx
```diff
@@ -15,6 +15,7 @@ import {
 } from "@liveblocks/client";
 import * as React from "react";
 import useRerender from "./useRerender";
+export type { Json, JsonObject } from "@liveblocks/client";
```
GuillaumeSalles, line 18: Why are we re exporting these types?

nvie: I added these for convenience reasons, mostly. I found that in some projects, like our e2e tests, you already had an `import { … } from '@liveblocks/react'`, it was a bit annoying to have to add a new import line to import `Json` from `'@liveblocks/client'` instead of simply adding `Json` to the existing list of imports. Any objection to doing this?

nvie:
Will add the fix to disallow nesting Lson data underneath Json nodes tomorrow, like we talked about in our call today, and then get this merged in tomorrow 👍

nvie:
The last few promised updates are now in. After talking to Guillaume yesterday, I realized that this type definition would allow illegal LSON values, namely ones where LiveStructure instances would live inside "normal" lists/objects, for example:

```ts
const items: Lson = new LiveList<{
  foo: { a: new LiveObject({ b: 42 }) },
  bar: [ new LiveMap<string, number>([]) ],
}>;
```

Here, the values under keys `foo` and `bar` are both illegal. In order to use a `LiveObject` or `LiveMap` underneath these keys, the values themselves would need to be LiveStructure instances themselves. This is now enforced by the type system in the latest updates.

To only legal equivalents of the above are now:

```ts
const items: Lson = new LiveList<{
  foo: new LiveObject({ a: new LiveObject({ b: 42 }) }),
  bar: new LiveList([new LiveMap<string, number>([])]),
}>;
```

Or, if you don't need nested live stuff:

```ts
const items: Lson = new LiveList<{
  foo: { a: { b: 42 } },
  bar: [new Map<string, number>()],
}>;
```

PR #152, opened 2022-04-15
Fix TypeScript errors in `@liveblocks/zustand`

nvie:
When running `tsc` inside the `liveblocks-zustand` package, I kept seeing these TypeScript errors:

Screen Shot 2022-04-15 at 08 36 33@2x

Those were all coming from the `fakeApp.ts` module, which I found no reference too. So I'm deleting it in this PR to get rid of the errors.

Please let me know if this is a mistake!


nvie:
Going to get this merged to get rid of the `tsc` errors, but please let me know if this is a mistake! 🙏

PR #155, opened 2022-04-20
Allow "for-all" scripts to run shell cmds as well

nvie:
Previously, you could run, for example:

```
$ for-all-examples.sh npm install
```

But you could not run more than one command in each directory easily. This PR changes that, by supporting a `-c` flag, which allows you to specify a _shell_ command to run (it uses `sh -c 'your command here'`) to run it, instead of directly running `your command here`.

This has the added benefit that you can use `&&` to run multiple commands:

```
$ for-all-examples.sh -c 'npm install && link-liveblocks.sh && npm run build'
```

### DIFF
--- a/scripts/for-all-e2e.sh
+++ b/scripts/for-all-e2e.sh
@@ -15,20 +15,23 @@ err () {
 }
 
 usage () {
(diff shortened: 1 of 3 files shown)


nvie:
Non-invasive PR, so merging this in.

PR #156, opened 2022-04-20
Improve type safety across the entire client API

nvie:
This PR adds a lot more type definitions, both internal and external, so all the APIs know about the developer-provided Presence and Storage types.

It's basically made possible by a massive trickle-down of `TPresence` and `TStorage` type variables, to tie all the APIs to know about those developer-provided types.

This work now includes [the "factory" function that I proposed in this Gist](https://gist.github.com/nvie/eaf806820d4825c12609932a405c9530), now named `createHooks`[^1].

### Still Todo

- [ ] Settle on final name for `createHooks()`
- [x] Rename `TStorageRoot` → `TStorage` for brevity
- [x] Remove default type assignments for Live structures (e.g. don't allow to use `LiveMap` or `LiveMap<>` as types, always force users to provide the type arguments, i.e. `LiveMap<string, number>`). This behavior is safer, and matches `Array<...>`, `Map<...>` and object types.
- [ ] ~Can we force others to only show up once their Presence is available?~
- [x] Enforce initial value in `new LiveObject()` constructor (`LiveMap` and `LiveList` can use "empty" as defaults, but `LiveObject` cannot reliably do that!)
- [x] Move initialization of state to the root level:
  - [ ] Allow initial presence or storage to be `null` to effectively disable it (for users that don't want to use presence/storage, it makes no sense to be forced to initialize it)
  - [ ] If storage/presence initialization is not done at the room level, disable it with `null`
  - [x] Disallow initial values at the hook levels
- [x] Rewrite all `TPresence extends JsonObject` to `P extends Presence` for brevity?
- [x] Rewrite all `TStore extends LsonObject` to `S extends Storage` for brevity?
- [ ] ~Deprecate `defaultPresence` in favor of `initialPresence`~
- [ ] ~Deprecate `defaultStorageRoot` in favor of `initialStorage`~
- [x] Expose `useStorageValue()` (to replace existing hooks)
- [x] Polish the `useList`, `useMap`, `useObject` hook APIs, deprecate them
- [ ] Write blog post to describe the changes
- [ ] Release as a private beta version, ask for community feedback

### Later

- [ ] Experiment with _not_ enforcing JSON compatibility statically, but at runtime
- [ ] Use "schema definitions" to infer the types _and_ provide automatic schema validation at the same time (would also avoid the `interface` problem)

[^1]: Still looking for a better name here. I didn't like `makeRoomContext()` because it sounds cryptical. Technically `createHooks()` is inaccurate, because it also creates the `RoomProvider` which isn't a hook. Perhaps something a bit vaguer like `setup()` might be better. Ideas?


GuillaumeSalles (approved):
Incredible work! 👏 

I played a bit with the `createHooks` and the DX is so much better! I'm still not sure what would be the best way to introduce this API. For non-typescript users, it might feel like an unnecessary indirection, but at the same time, it would simplify documentation to have a single API. 🤔 

I'm optimistic that the community might help us figure out the last naming trade off!

Thank you so much for championing this typescript refactoring! The DX will be at completely different level when this land! Let's discuss tomorrow how we could release it.

nvie:
Closing this PR. I've been cherry-picking smaller bits from this to land in 0.16.x (for non-breaking refactors) and 0.17 (for breaking API changes), and will eventually still make all of these changes, but not in a single, giant PR.

PR #157, opened 2022-04-21
Prepwork: split React package into modules

nvie:
## What does this PR do?
This PR just performs a little bit of organization to help make with the upcoming TypeScript improvements. Why do this?

First of all, this just nicely splits up client and room related logic into separate modules, which is just nice to do anyway.

More importantly though, I'm [experimenting with a way to import the React based hooks differently](https://gist.github.com/nvie/eaf806820d4825c12609932a405c9530) to make them type-safe all the way through. In that reality, [these APIs](https://github.com/liveblocks/liveblocks/blob/80745add4204d26644fda251bb3c5d85d0863d32/packages/liveblocks-react/src/index.tsx#L2-L20) would need to pass through that factory function. Putting them in a separate module will make it easier (and more obvious) to make that change.


nvie:
Getting this merged in because it's just a code split/move, and it will make follow-up PRs more readable (less noisy). No new code is introduced with this PR.

PR #159, opened 2022-04-21
Add code comments for posterity

nvie:
This PR leaves a few comments in places that were not obvious to me, but which Guillaume clarified for me on our pairing session earlier this week. Documenting these insights for posterity here.

### DIFF
--- a/packages/liveblocks-client/src/room.ts
+++ b/packages/liveblocks-client/src/room.ts
@@ -178,10 +178,6 @@ export type Effects = {
 };
 
 type Context = {
-  /**
-   * @deprecated Use `roomId` instead.
-   */
-  room: string;
   roomId: string;
   throttleDelay: number;
   fetchPolyfill?: typeof fetch;
--- a/packages/liveblocks-client/src/room.ts
+++ b/packages/liveblocks-client/src/room.ts
@@ -178,7 +178,11 @@ export type Effects = {
 };
 
 type Context = {
+  /**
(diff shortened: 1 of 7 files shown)


packages/liveblocks-client/src/room.ts
```diff
@@ -178,7 +178,11 @@ export type Effects = {
 };
 
 type Context = {
+  /**
+   * @deprecated Use `roomId` instead.
+   */
   room: string;
+  roomId: string;
```
nvie, line 185: I read `context.room` somewhere, and assumed it was of type `room: Room`, but actually it was a `room: string` when I looked more closely, so changing the name. I'm duplicating the field here for now and deprecating it, but I'm not sure if that's even necessary. Seems to be a private field used only by the client itself?

nvie, line 185: I read `context.room` somewhere, and assumed it was of type `room: Room`, but actually it was a `room: string` when I looked more closely, so changing the name. I'm duplicating the field here for now and deprecating it, but I'm not sure if that's even necessary. Seems to be a private field used only by the client itself?

GuillaumeSalles: I didn't verify but I'm 99% sure that it's internal so it could be renamed.

nvie: Fixed in ea71ceaf7b9ecf24da748996b8cb243688904f61 👍

PR #160, opened 2022-04-22
Fix ESLint complaints

nvie:
This PR tweaks some settings in tsconfig and ESLint so that they're consistent about reporting unused args, and addresses those issues.


packages/liveblocks-client/src/room.test.ts
```diff
@@ -1095,7 +1095,7 @@ describe("room", () => {
         ],
       ];
 
-      await reconnect(2, newInitStorage);
+      reconnect(2, newInitStorage);
```
nvie, line 1098: This didn't return a promise, so no need to await here.

GuillaumeSalles (approved):
👌

PR #161, opened 2022-04-25
Avoid trusting unverified inputs

nvie:
This PR plugs a potential security risk where untrusted inputs (via the result of `JSON.parse()`) are used (partially) unverified, allowing arbitrary data to trickle into our system which won't match the TypeScript types at runtime. There is no concrete risk, but in general you want to avoid this pattern, because it's a potential attack vector that malicious users could exploit.

This PR adds explicit validation of these inputs. How? By first [disallowing the use of `JSON.parse()`](https://github.com/liveblocks/liveblocks/blob/a8e8bc5eea5a0db30b216ad23b519b63bb54d52d/packages/liveblocks-client/.eslintrc.js#L68-L73), because in TypeScript its result is `any` (and it can throw!). Instead, we use the type-safe `parseJson()` helper that I added here, which will return `Json`. This way, TypeScript forces us to handle all the assumptions we're making explicitly first. You can see this reflected in the extended conditional checks.

## Future Work
A similar issue [exists here still](https://github.com/liveblocks/liveblocks/blob/a8e8bc5eea5a0db30b216ad23b519b63bb54d52d/packages/liveblocks-client/src/room.ts#L852-L853), where the assumed shape of the TypeScript object should be verified at runtime instead of assumed using an `as` statement. This issue isn't addressed yet. For big object structures like this, using `if`-statements becomes unwieldy. Of course, there [exists a library to help express this more elegantly 😉](https://github.com/nvie/decoders), but I didn't want to introduce a dependency here yet.

### DIFF
--- a/packages/liveblocks-client/src/json.ts
+++ b/packages/liveblocks-client/src/json.ts
@@ -25,3 +25,15 @@ export function parseJson(rawMessage: string): Json | undefined {
     return undefined;
   }
 }
+
+export function isJsonArray(data: Json): data is JsonArray {
+  return Array.isArray(data);
+}
+
+export function isJsonObject(data: Json): data is JsonObject {
+  return data !== null && typeof data === "object" && !isJsonArray(data);
+}
+
+export function isJsonScalar(data: Json): data is JsonScalar {
+  return typeof data !== "object";
+}
--- a/packages/liveblocks-client/src/json.ts
+++ b/packages/liveblocks-client/src/json.ts
@@ -33,7 +33,3 @@ export function isJsonArray(data: Json): data is JsonArray {
 export function isJsonObject(data: Json): data is JsonObject {
   return data !== null && typeof data === "object" && !isJsonArray(data);
 }
-
-export function isJsonScalar(data: Json): data is JsonScalar {
-  return typeof data !== "object";
-}
--- a/packages/liveblocks-client/src/json.ts
+++ b/packages/liveblocks-client/src/json.ts
@@ -19,6 +19,7 @@ export type JsonObject = { [key: string]: Json };
  */
 export function parseJson(rawMessage: string): Json | undefined {
   try {
+    // eslint-disable-next-line no-restricted-syntax
     return JSON.parse(rawMessage);
   } catch (e) {
     return undefined;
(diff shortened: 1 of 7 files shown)


packages/liveblocks-client/src/json.ts
```diff
@@ -12,3 +12,28 @@ export type Json = JsonScalar | JsonArray | JsonObject;
...
+
+export function isJsonArray(data: Json): data is JsonArray {
+  return Array.isArray(data);
+}
+
+export function isJsonObject(data: Json): data is JsonObject {
+  return data !== null && typeof data === "object" && !isJsonArray(data);
+}
+
+export function isJsonScalar(data: Json): data is JsonScalar {
```
GuillaumeSalles, line 37: I don't think it's used right now

packages/liveblocks-client/src/room.ts
```diff
@@ -196,9 +196,10 @@ export function makeStateMachine(
       auth: (room: string) => Promise<AuthorizeResponse>,
       createWebSocket: (token: string) => WebSocket
     ) {
-      if (isTokenValid(state.token)) {
-        const parsedToken = parseToken(state.token!);
-        const socket = createWebSocket(state.token!);
+      const token = state.token;
+      if (token && isTokenValid(token)) {
```
GuillaumeSalles, line 200: Nice! thanks for cleaning that up at the same time!

packages/liveblocks-client/src/utils.test.ts
```diff
@@ -6,8 +6,25 @@ import {
...
   isTokenValid,
+  compact,
 } from "./utils";
 
+describe("compact", () => {
+  it("compact w/ empty list", () => {
+    expect(compact([])).toEqual([]);
+  });
+
+  it("icompact removes nulls and undefined values", () => {
```
GuillaumeSalles, line 17:
```suggestion
  it("compact removes nulls and undefined values", () => {
```

Tiny typo

nvie: Good catch! 🦅👁️ Fixed in 44b63ba1400454046ef86cf2df2634ec5f3ac66f 👍

packages/liveblocks-client/src/json.ts
nvie: You're right! Removed in a8e7a14fa7cf1a26275161780af020efb2ce4c6c 👍

ofoucherot (approved):
🚀

PR #165, opened 2022-04-26
Fix to make `JsonObject` type partial

nvie:
This fixes the `JsonObject` type to recognize that keys could be validly `undefined`.

Wrong behavior:

```tsx
const obj: JsonObject = ...;
obj.whatever;   // => Json ❌
```

New/correct behavior:

```tsx
const obj: JsonObject = ...;
obj.whatever;   // => Json | undefined ✅
```

### DIFF
--- a/packages/liveblocks-client/src/immutable.ts
+++ b/packages/liveblocks-client/src/immutable.ts
@@ -13,7 +13,10 @@ function lsonObjectToJson<O extends LsonObject>(
 ): { [K in keyof O]: Json } {
   const result: { [K in keyof O]: Json } = {} as any;
   for (const key in obj) {
-    result[key] = lsonToJson(obj[key]);
+    const val = obj[key];
+    if (val !== undefined) {
+      result[key] = lsonToJson(val);
+    }
   }
   return result;
 }
@@ -294,7 +297,10 @@ function patchImmutableNode(
 
         for (const key in update.updates) {
           if (update.updates[key]?.type === "update") {
-            newState[key] = lsonToJson(update.node.get(key));
+            const val = update.node.get(key);
+            if (val !== undefined) {
+              newState[key] = lsonToJson(val);
+            }
           } else if (update.updates[key]?.type === "delete") {
             delete newState[key];
           }
--- a/packages/liveblocks-client/src/json.ts
+++ b/packages/liveblocks-client/src/json.ts
@@ -11,7 +11,7 @@
 export type Json = JsonScalar | JsonArray | JsonObject;
 export type JsonScalar = string | number | boolean | null;
 export type JsonArray = Json[];
-export type JsonObject = { [key: string]: Json };
+export type JsonObject = Partial<{ [key: string]: Json }>;
 
 /**
  * Alternative to JSON.parse() that will not throw in production. If the passed
--- a/packages/liveblocks-client/src/json.ts
+++ b/packages/liveblocks-client/src/json.ts
@@ -11,7 +11,7 @@
 export type Json = JsonScalar | JsonArray | JsonObject;
 export type JsonScalar = string | number | boolean | null;
 export type JsonArray = Json[];
-export type JsonObject = Partial<{ [key: string]: Json }>;
+export type JsonObject = { [key: string]: Json | undefined };
 
 /**
  * Alternative to JSON.parse() that will not throw in production. If the passed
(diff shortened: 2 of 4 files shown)


GuillaumeSalles:
My initial thought was that we were supposed to do the opposite, but maybe I'm wrong 🤔 

From what I understand, `undefined` is not serializable to json, so we shouldn't be able to set have undefined values on a `LiveObject`. `UpdateObjectOp.data` should be a `JsonObject` or a `Record<string, Json>` instead of a `Partial<JsonObject>`? Sorry about the confusion, I'm still not 100% of what's the correct type

GuillaumeSalles:
My initial thought was that we were supposed to do the opposite, but maybe I'm wrong 🤔 

From what I understand, `undefined` is not serializable to json, so we shouldn't be able to set have undefined values on a `LiveObject`. `UpdateObjectOp.data` should be a `JsonObject` or a `Record<string, Json>` instead of a `Partial<JsonObject>`? Sorry about the confusion, I'm still not 100% of what's the correct type

nvie:
Yeah, its been confusing for sure. My wrong thinking before comes from the fact that `undefined` does not have a JSON _representation_, but you can still work with them in TypeScript land on the input and output side. I hope the following examples help to illustrate that.

On the serializing side, not allowing `undefined` in inputs was too restrictive. It’s okay to send in `{ a: 1, b: undefined }`, which is the same as sending in `{ a: 1 }`, because they both produce the same string when using `JSON.stringify()`.

On the parsing side, `{ [key: string]: Json }` means that all potential thinkable keys on an object are always going to be `Json` in TypeScript. They almost never are! For example: `JSON.parse("{ a: 1 }").someRandomProperty` is definitely going to be `undefined`.

#### The old/wrong way

```ts
const x: JsonObject = { a: 1, b: undefined };
//                            ^ ❌ Errors here
JSON.stringify(x);  // Even though this call is fine!
```

Also:

```ts
const y: JsonObject = JSON.parse("{ a: 1 }");
const a: Json = y.a;  // Correct
const b: Json = y.b;  // ❌ Incorrect! It’s actually `undefined`, but TypeScript misses that truth
```

#### The new/correct way

```ts
// ✅ Fine now
const x: JsonObject = { a: 1, b: undefined };
JSON.stringify(x);
```

Also:

```ts
const y: JsonObject = JSON.parse("{ a: 1 }");
const a = y.a;
const b = y.b;
//    ^ ✅ Now correctly infers that these could be Json | undefined
```

nvie:
tl;dr `undefined` itself isn’t a legal `Json` value, but a `JsonObject` can still have implicitly-missing or explicitly-missing (through `undefined`) entries.

nvie:
Tying this back to your `UpdateObjectOp.data` use case: I think from that perspective, we should simply ignore any keys with `undefined` values ourselves (skipping over them when looping). Because depending on where the JSON value is generated, it’s possible that there are explicit-undefined keys (and thus more entries to loop over), which should be treated as implicit-undefined keys (which would not pop up when looping all entries).

I hate the fact that implicit-missing and explicit-missing values are observable (e.g. when looping over entries), but this is the most honest reflection of the possible realities that are the consequence of JS’ language design.

GuillaumeSalles:
Thank you for the detailed explanation! I now understand why you implemented the fix that way. The new `JsonObject` type is a better representation of a javascript object created by a `JSON.parse`.

> I hate the fact that implicit-missing and explicit-missing values are observable (e.g. when looping over entries), but this is the most honest reflection of the possible realities that are the consequence of JS’ language design.

Yep, we're on the same page. "The looping over entries" is what I was thinking about when I mentioned serialization. My intuition is that if we allow `LiveObject<{ x?: number }>`, we might give the impression that the LiveObject keys will be the same on all clients after `liveObject.set("x", undefined)`. But at the same time, it could help us refine the type of `LiveObject.delete` by allowing only keys where the value can be `undefined`?

GuillaumeSalles (approved):

nvie:
> My intuition is that if we allow `LiveObject<{ x?: number }>`, we might give the impression that the LiveObject keys will be the same on all clients after `liveObject.set("x", undefined)`. But at the same time, it could help us refine the type of `LiveObject.delete` by allowing only keys where the value can be `undefined`?

Ohh... great point! I didn't even consider that use case yet, but I think that this now indeed naturally solves for that. In `LiveObject.set(x, y)`, `y` should never be undefined (or better… it should implicitly translate to `LiveObject.delete(x)` instead), and indeed: not allowing `.delete()` on keys that aren't declared to be optional would be an amazing DX if you ask me :)

PR #167, opened 2022-04-27
Make packages subpath exports compatible with typescript

GuillaumeSalles:
Update our packages subpath exports to be compatible with typescript.

[Subpath exports](https://nodejs.org/api/packages.html#subpath-exports) are supported by node.js but typescript (and tsc) only support it with the latest beta version (and `moduleResolutions: "node12"`).

To work around this limitation, we update the subpath exports to have the exact same name as the js file they map to. That way typescript can understand the imports while more modern tools like `esbuild` will use them if necessary.

We copy the package.json and README file to the `lib` folder after a build and assume that link and publish will be done from this subfolder.  (Technique shamelessly stolen from zustand : https://github.com/pmndrs/zustand/blob/main/package.json#L68)

### DIFF
--- a/packages/liveblocks-redux/package.json
+++ b/packages/liveblocks-redux/package.json
@@ -3,17 +3,17 @@
   "version": "0.16.0",
   "sideEffects": false,
   "description": "A store enhancer to integrate Liveblocks into Redux stores.",
-  "main": "./lib/index.js",
-  "types": "./lib/index.d.ts",
+  "main": "./index.js",
+  "types": "./index.d.ts",
   "files": [
-    "lib/"
+    "**"
   ],
   "exports": {
     ".": {
-      "types": "./lib/index.d.ts",
-      "module": "./lib/esm/index.js",
-      "import": "./lib/esm/index.mjs",
-      "default": "./lib/index.js"
+      "types": "./index.d.ts",
+      "module": "./esm/index.js",
+      "import": "./esm/index.mjs",
+      "default": "./index.js"
     }
   },
   "keywords": [
@@ -25,7 +25,7 @@
     "collaborative"
   ],
   "scripts": {
-    "build": "rollup -c",
+    "build": "rollup -c && cp ./package.json ./README.md ./lib",
     "test": "jest --watch",
     "dtslint": "dtslint --localTs node_modules/typescript/lib --expectOnly types"
   },
--- a/scripts/link-liveblocks.sh
+++ b/scripts/link-liveblocks.sh
@@ -180,8 +180,11 @@ prep_liveblocks_deps () {
 
             rebuild_if_needed
 
-            # Built package is located in the lib folder, including the package.json
-            cd "./lib"
+            # If a `package.json` exists in the dist folder, consider it the
+            # root of the package.
+            if [ -f "./lib/package.json" ]; then
+                cd "./lib"
+            fi
 
             # Register this link
             npm_link
--- a/scripts/link-liveblocks.sh
+++ b/scripts/link-liveblocks.sh
@@ -180,6 +180,9 @@ prep_liveblocks_deps () {
 
             rebuild_if_needed
 
+            # Built package is located in the lib folder, including the package.json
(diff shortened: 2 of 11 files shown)


scripts/publish.sh
```diff
@@ -270,7 +270,7 @@ publish_to_npm () {
         read -p "OTP token? " OTP
     done
 
-    npm publish --tag "${TAG:-latest}" --otp "$OTP"
```
GuillaumeSalles, line 273: This is the only thing that I didn't test it manually. Do you see a better way to do it @nvie ?

packages/liveblocks-redux/package.json
```diff
@@ -25,7 +25,7 @@
     "collaborative"
   ],
   "scripts": {
-    "build": "rollup -c",
+    "build": "rollup -c && cp ./package.json ./README.md ./lib",
```
nvie, line 28: Ah yes!

For what it's worth, I'm doing something similar in the build script for decoders. For example, I'm also [copying some files manually](https://github.com/nvie/decoders/blob/main/bin/build.sh#L72), changing [all references to `dist/`](https://github.com/nvie/decoders/blob/main/bin/build.sh#L89) from the original `package.json`, and [removing keys from `package.json` that don't carry value on NPM](https://github.com/nvie/decoders/blob/main/bin/build.sh#L82-L87). The result is a beautiful package though, for someone who ever inspects it. I'm not using rollup, as I could never figure out how to make it work with all the environments at once (TypeScript, Flow, ES modules, CJS, etc.), so I'm basically creating 4 versions of each original source file (`.ts`, `.js.flow`, `.mjs`, and `.js`, respectively). For the `.mjs` files, that means having to re-write the imports inside those files, [using this hack](https://github.com/nvie/decoders/blob/main/bin/build.sh#L57) which rewrites `import ... from 'foo'` to `import ... from 'foo.mjs'`.

It's basically very manually crafting the contents of what you'd see on the "Explore" tab on https://www.npmjs.com/package/decoders. It's a shame that this is so hard to fully control, and unclear what best practices are. It took me a lot of beta releases of 2.0 to get the final structure right, but the end result is beautiful, works in all environments, and is fully tree-shakable.

Part of the trick was realizing that `package.json` does too many things. It's used for local development, but also acts as the "index" for what gets published on NPM.

Eventually I'd be in favor of wrapping this up into a light generic build script like with decoders, where we can do more tweaks/cleanups pre-publishing.

scripts/link-liveblocks.sh
```diff
@@ -180,6 +180,9 @@ prep_liveblocks_deps () {
 
             rebuild_if_needed
 
+            # Built package is located in the lib folder, including the package.json
+            cd "./lib"
```
nvie, line 184: This may not be true always, so I'd make this conditional. If a `lib/package.json` exists, then use that, otherwise don't `cd`. This will make it work with packages that don't use a specially crafted
package.json for distribution, or older packages. Can add this in a commit!

scripts/publish.sh
nvie: I've never seen this syntax, but it may work. For the decoders build script, I'm just `cd`ing into that dist directory, and then using this very same `npm publish` command, where it will pick up the local `package.json` file to publish from. Can add!

nvie (approved):
Looking great! Thanks for addressing this. This stuff is hard to figure out, but happy to help keep iterating on this as I've already got some scar tissue from publishing my own open source packages for multiple environemtns already.

scripts/link-liveblocks.sh
nvie: Added in 18d278a3b438d24ec21d1abdbc499ca4f8b9797a!

scripts/publish.sh
nvie: Added in 43180fcc0cefccb564701f23f4e18185aaec979a!

GuillaumeSalles:
Thanks for adjusting the scripts and the review! @nvie

PR #169, opened 2022-04-27
Fix React 18 StrictMode

GuillaumeSalles:


ofoucherot (approved):

packages/liveblocks-react/src/rooms.tsx
```diff
@@ -58,20 +58,28 @@ export function RoomProvider<TStorageRoot>({
 
   const client = useClient();
 
+  const [room, setRoom] = React.useState(
+    client.enter(id, {
+      defaultPresence: defaultPresence ? defaultPresence() : undefined,
+      defaultStorageRoot,
+      DO_NOT_USE_withoutConnecting: typeof window === "undefined",
+    } as any)
```
nvie, line 66: Would it be problematic to keep calling `client.enter()` on every render? I realize the result of doing so would be discarded if there already is state, but since `client.enter()` has side effects, it seems we should wrap this in a `() => client.enter(...)` call to only actually enter this on the first render?

nvie: (Previously this was guarded by the `client.getRoom(id) || ...` check.)

nvie: Proposed fix: https://github.com/liveblocks/liveblocks/pull/171

PR #174, opened 2022-05-02
Fix bug

nvie:
Following up on https://github.com/liveblocks/liveblocks/pull/156#discussion_r860409862, and getting this fixed now first, before the beta lands.

### DIFF
--- a/packages/liveblocks-client/src/room.test.ts
+++ b/packages/liveblocks-client/src/room.test.ts
@@ -541,15 +541,20 @@ describe("room", () => {
     room.authenticationSuccess({ actor: 0 }, ws);
     ws.open();
 
+    expect(state.buffer.presence).toEqual(null);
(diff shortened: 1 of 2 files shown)


ofoucherot (approved):
Lgtm 🚀 

It would be nice to have a unit test for this issue.

nvie:
> It would be nice to have a unit test for this issue.

Agreed, I should have added that in the first place. Done now in c27581da4982cad5849c676a9f6adc0d649329cd (you can see the test failing there), and then applied the fix in 0d0a2b6765b2d0262b2a60429b9b1177cc844fb, after which you can see the test succeed again!

nvie:
Forgot to also update the CHANGELOG, now done in b6849ce4496a3cd0a9e89751539d4a014d9fa819.

PR #175, opened 2022-05-02
Fix types in our whiteboard example

nvie:
I found these type issues after trying to migrate one example to the new typed APIs I'm working on. Yay, I guess? 🙌

### DIFF
--- a/examples/nextjs-whiteboard/src/index.tsx
+++ b/examples/nextjs-whiteboard/src/index.tsx
@@ -50,7 +50,8 @@ export default function Room() {
       id={"multiplayer-canvas"}
       defaultPresence={() => ({
         selection: [],
-        penPoints: null,
+        cursor: null,
+        pencilDraft: null,
         penColor: null,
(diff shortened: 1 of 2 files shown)


marcbouchenoire (approved):
Glad to see us already reaping the benefits from the new API. 🙂

Btw I'm almost done with my examples branch (https://github.com/liveblocks/liveblocks/compare/feature/examples), if you're planning on bigger changes like moving them to the new typed API for example we should make sure to coordinate.

nvie:
> Btw I'm almost done with my examples branch (https://github.com/liveblocks/liveblocks/compare/feature/examples), if you're planning on bigger changes like moving them to the new typed API for example we should make sure to coordinate.

No worries! I'm not touching the examples at all (yet). This was just something I found while locally testing the waters against a random example, and figured it would be nice to port these now-known bugs back. Don't worry about me, and change the examples as you need to. I can make changes to the examples later on after you're done.

PR #178, opened 2022-05-03
Lint complaints, various cleanups

nvie:
This PR bundles a few relatively small changes. ESLint fixes, TypeScript hints, name cleanups, etc. None of these changes should have a behavioral impact.

---

**Update:** I've moved a subset (= the trivial parts) of this PR into #182, and got those merged in. What remains in this PR is just the `async`/`await` stuff.

### DIFF
--- a/packages/liveblocks-client/src/client.ts
+++ b/packages/liveblocks-client/src/client.ts
@@ -37,11 +37,11 @@ export function createClient(options: ClientOptions): Client {
     return internalRoom ? internalRoom.room : null;
   }
 
-  function enter<TStorageRoot>(
+  function enter<TStorage>(
     roomId: string,
     options: {
       defaultPresence?: Presence;
-      defaultStorageRoot?: TStorageRoot;
+      defaultStorageRoot?: TStorage;
       /**
        * INTERNAL OPTION: Only used in a SSR context when you want an empty room to make sure your react tree is rendered properly without connecting to websocket
        */
--- a/packages/liveblocks-client/src/room.ts
+++ b/packages/liveblocks-client/src/room.ts
@@ -462,6 +462,8 @@ export function makeStateMachine(
         presence: false,
       },
     };
+
+    const createdNodeIds = new Set<string>();
     for (const op of item) {
       if (op.type === "presence") {
         const reverse = {
@@ -492,16 +494,30 @@ export function makeStateMachine(
         }
         const applyOpResult = applyOp(op, isLocal);
         if (applyOpResult.modified) {
-          result.updates.storageUpdates.set(
-            applyOpResult.modified.node._id!,
-            mergeStorageUpdates(
-              result.updates.storageUpdates.get(
-                applyOpResult.modified.node._id!
-              ) as any, // FIXME
-              applyOpResult.modified
-            )
-          );
-          result.reverse.unshift(...applyOpResult.reverse);
+          const parentId = applyOpResult.modified.node._parent?._id!;
+
+          // If the parent was created in the same batch, we don't want to notify
+          // storage updates for the children.
+          if (!createdNodeIds.has(parentId)) {
+            result.updates.storageUpdates.set(
+              applyOpResult.modified.node._id!,
+              mergeStorageUpdates(
+                result.updates.storageUpdates.get(
+                  applyOpResult.modified.node._id!
+                ) as any, // FIXME
+                applyOpResult.modified
+              )
+            );
+            result.reverse.unshift(...applyOpResult.reverse);
+          }
+
+          if (
+            op.type === OpType.CreateList ||
+            op.type === OpType.CreateMap ||
+            op.type === OpType.CreateObject
+          ) {
+            createdNodeIds.add(applyOpResult.modified.node._id!);
+          }
         }
       }
     }
--- a/packages/liveblocks-client/src/room.ts
+++ b/packages/liveblocks-client/src/room.ts
@@ -553,8 +553,6 @@ export function makeStateMachine(
         return parent._attachChild(op, isLocal);
       }
     }
-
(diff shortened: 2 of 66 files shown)


packages/liveblocks-client/src/client.ts
```diff
@@ -37,11 +37,11 @@ export function createClient(options: ClientOptions): Client {
     return internalRoom ? internalRoom.room : null;
   }
 
-  function enter<TStorageRoot>(
+  function enter<TStorage>(
```
nvie, line 40: This is a consistency pass, so all the type params are named the same everywhere.

packages/liveblocks-client/src/room.ts
```diff
@@ -1228,15 +1226,11 @@ See v0.13 release notes for more information.
   let _getInitialStatePromise: Promise<void> | null = null;
   let _getInitialStateResolver: (() => void) | null = null;
 
-  function getStorage<TRoot extends LsonObject>(): Promise<{
-    root: LiveObject<TRoot>;
+  async function getStorage<TStorage extends LsonObject>(): Promise<{
```
nvie, line 1229: This is the result of TypeScript convert this function to an async function.

packages/liveblocks-client/src/room.ts
```diff
@@ -1636,45 +1629,51 @@ function prepareAuthEndpoint(
...
+    );
+  }
 
-      return authResponse;
-    });
+  if (!isJsonObject(raw)) {
+    throw new AuthenticationError(
+      `Expected a JSON object when doing a POST request on "${endpoint}".`
+    );
+  }
```
nvie, line 1666: Here is the only small behavioral change I added: it checks to see that the `res.json()` call actually returns a JSON object (not any other JSON value).

packages/liveblocks-client/src/room.ts
```diff
@@ -553,8 +553,6 @@ export function makeStateMachine(
         return parent._attachChild(op, isLocal);
       }
     }
-
-    return { modified: false };
```
nvie, line 557: This removes a line of unreachable code. The added benefit here is that if we extend the `OpType` enum, TypeScript will force us to get back to this function and deal with it. In other words, not exhaustively handling these cases will be a TypeScript error.

nvie:
I noticed an issue where running `jest` from secondary packages was now failing, after switching the `Promise` to `async`/`await` in bdc3d918117fba9fec7a26301a476aca8248f8a3 and 93b6e877120604aacaf1de82ae0d2f84a8fbc180. This is because we've configured TypeScript to [target es2015](https://github.com/liveblocks/liveblocks/blob/b6849ce4496a3cd0a9e89751539d4a014d9fa819/packages/liveblocks-client/tsconfig.json#L3) in its builds (so the regenerator runtime boilerplate doesn't have to be shipped). This is no problem for other TypeScript projects importing from `@liveblocks/client`, because we assume those projects are modern enough (= [at least have `es2015` or higher in their libs](https://github.com/liveblocks/liveblocks/blob/b6849ce4496a3cd0a9e89751539d4a014d9fa819/packages/liveblocks-react/tsconfig.json#L7)).

This means we should be able to use `async`/`await` without issues in `@liveblocks/client`. (If not, we should declare its target to be `es5` again, in exchange for a larger bundle.)

However, `jest` won't transpiling the contents from `node_modules`. We could either switch to a slower tool like `ts-jest`, or… just require the `regenerator-runtime` here, which is what I've done in ffe8682514c0ca02e79bd63a4dab2e0bdc322737.

nvie:
Follow-up thought! Given the previous comment about `jest` failling for our secondary packages, and seeing these two green checkmarks…

Screen Shot 2022-05-03 at 11 34 05@2x

…does that mean we're not running `jest` in the other projects, and only running it for `liveblocks-client` on CI?

packages/liveblocks-client/src/room.ts
GuillaumeSalles: I remember doing the exact opposite because regenerator runtime was causing an issue in specific contexts, but I don't remember why 🤔 

Do you remember @ofoucherot? Was it for some SSR or node issues? Or just to reduce bundle size?

https://github.com/liveblocks/liveblocks/pull/59/commits/8a0e092bad2e748bef6d73db9771908b2dd618e0

GuillaumeSalles (commented):
LGTM, but I would wait to see if @ofoucherot remember why we decided to remove async/await from `@liveblocks/client`

packages/liveblocks-client/src/room.ts
nvie: Ohhh, I see where I made the mistake in my reasoning, I think. I must have momentarily gotten my JS history wrong, and mistakenly thought `async`/`await` were part of the `es2015`, but they aren't 🤦

The correct history, for reference (I just looked it up 😉):

- `es5` had nothing fancy
- `es2015` (aka `es6`) introduced the Promise API, and generator functions
- `es2017` (aka `es7`) introduced `async`/`await` keywords

Since we're targeting `es2015` for our CJS bundles, TypeScript should transpile away any use of `async`/`await` and indeed emit the regenerator runtime code in our bundle. That would definitely increase the bundle size for the CJS part!

I agree that using language patterns that needlessly pump the bundle size should be avoided if possible, so in this case, it makes perfect sence to use the Promise API directly and avoid the use of `async`/`await`, because then TypeScript can just emit the code without transpiling.

I think we can forbid usage of certain language constructs with a simple ESLint rule, and provide exactly this rationale there. For public libraries that makes perfect sence!

nvie: Will close this PR in favor of one that adds an ESLint rule to forbid the use of `async`/`await` keywords, and provide this rationale.

PR #180, opened 2022-05-03
Add `Resolve<T>` helper type

nvie:
This type is a useful helper when building libraries with type definitions. Trick comes from https://effectivetypescript.com/2022/02/25/gentips-4-display/, and when used strategically it can really improve the developer experience in IDEs, see [this playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAShDOB7ANgNwgHgCoD4oF4osoIAPYCAOwBN4oAxAV0oGNgBLRSqAfiKgBcUAN4BtANJR23ANYQQiAGZEAukKwSVAXwDcAKD2hIsBI2TBsefHqhQAPiKiIZQ4ACdGEADRQwAQxBkRD9qdSgtG3tHZyFFP2R4aB8INzcAW3gAcyF4d2lM8P09ZAhgKDcARiE4eDMLXLd80RUcfQB6NtsAPR4DErK3ACZqhBR0DBq6jAamlta9Du7eoA):

Screen Shot 2022-05-03 at 12 09 53@2x

(Not all types have to be resolved. There is a place for both of these type display variants. This utility lets us control it a bit though!)

I'm going to be using this in the next PR.

### DIFF
--- a/packages/liveblocks-client/src/index.ts
+++ b/packages/liveblocks-client/src/index.ts
@@ -4,7 +4,6 @@ export { LiveList } from "./LiveList";
 export type {
   Others,
   Presence,
-  Resolve,
   Room,
   Client,
   User,
--- a/packages/liveblocks-client/src/index.ts
+++ b/packages/liveblocks-client/src/index.ts
@@ -4,6 +4,7 @@ export { LiveList } from "./LiveList";
 export type {
   Others,
   Presence,
+  Resolve,
   Room,
   Client,
   User,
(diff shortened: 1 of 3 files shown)


GuillaumeSalles:
Wow I've never seen such a trick and it's pretty neat! 

Could we export it via "@liveblocks/client/internal" to reduce the noise for external users?

GuillaumeSalles (approved):

nvie:
> Could we export it via "@liveblocks/client/internal" to reduce the noise for external users?

Great point, done in 9cdb6584fa28a653a017eba4716f3bfd8616bce7 👍

PR #183, opened 2022-05-04
Disallow use of `async` functions

nvie:
Following up on https://github.com/liveblocks/liveblocks/pull/178#discussion_r864522989, this now formally disallows the use of `async` inside `liveblocks-client`, because it will inflate the bundle size (only the CommonJS output).


nvie:
Now I still need to figure out why in my local setup the TypeScript language server keeps complaining about this to me. I haven't found a way to ignore this line yet.

Screen Shot 2022-05-04 at 09 39 43@2x

GuillaumeSalles (approved):
Thank you for enforcing this!

PR #185, opened 2022-05-04
Add general purpose `deprecate()` helper

nvie:
This adds a `deprecate()` and `deprecateIf()` helper utilities, and exports them as internals, so we can use them from all libraries. These runtime functions will throw a console warning at runtime. Use these whenever you also use a `@deprecated` comment annotation.

Only has an effect in dev/test envs. In production, this function will be a no-op.

Every unique deprecation message will be thrown only once, even if deprecated APIs are called multiple times, or from multiple places. Only the first warning will get printed, the rest of them will be no-ops.

## Uniqueness key
Normally you won't need the optional `key` param, and it defaults to the message content itself. But you can use it to control uniqueness, which is useful for the deprecation I want to use in the in upcoming 0.17 release.

For example, we want to deprecate the `defaultXxx` arguments here:

- `createRoom(defaultPresence, defaultStorageRoot)` (lowest-level, internal API)
  - called by `client.enter(defaultPresence, defaultStorageRoot)` (public API)
    - called by `<RoomProvider defaultPresence={...} defaultStorageRoot={...} />` (public API)

But where to point the developer towards? If we throw the error at the deepest level (from `createRoom()`), then the developer will have no idea. They're never manually calling that after all! They call `client.enter()` or use the `RoomProvider` APIs. To be maximally helpful to developers, we'll want to speak the language in the error message that the developer is familiar with.

That's where the `key` arg comes in. We can put a deprecation message in _all_ public APIs. So we show the deprecation warning in `client.enter()` _and_ inside `RoomProvider`, but use the same message key. That way, if they are using the `RoomProvider`, they'll see a message indicating they should rename the `RoomProvider` props. Then, when the RoomProvider internally invokes `client.enter()`, it tries to display an error, but sees that an error for this key has already been emitted, and it will not nag about those internals that don't make sense to the developer.

And otherwise, if they use `client.enter()` directly, we'll display a warning specific to that API.

### DIFF
--- a/packages/liveblocks-client/src/internal.ts
+++ b/packages/liveblocks-client/src/internal.ts
@@ -7,6 +7,8 @@
  * time and without announcement. This module purely exists to share code
  * between the several Liveblocks packages.
  *
+ * But since you're so deep inside Liveblocks code... we're hiring!
+ * https://join.team/liveblocks ;)
  */
 
 export * from "./live";
--- a/packages/liveblocks-client/src/internal.ts
+++ b/packages/liveblocks-client/src/internal.ts
@@ -7,8 +7,6 @@
  * time and without announcement. This module purely exists to share code
  * between the several Liveblocks packages.
  *
- * But since you're so deep inside Liveblocks code... we're hiring!
- * https://join.team/liveblocks ;)
  */
 
 export * from "./live";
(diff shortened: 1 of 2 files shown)


GuillaumeSalles (approved):
Really cool! I love that we more and more tools to improve the overall DX!

PR #189, opened 2022-05-06
Fix the way packages are built

### LINKED ISSUE #187: Build internals.d.ts "on top" of public index.d.ts
The current way the public vs internal libraries are built make it their definitions potentially conflicting and incompatible when you're using and mixing both in secondary packages.

For example:

- We have a centrally defined public type, e.g. `Json`
- In the internals module, we define a function that _uses_ this `Json` type, say `privateFn(value: Json)`

The way rollup is configured now is that it duplicates the `Json` definition _and_ the private function and rolls them all up in `internal.d.ts`. This is wrong.

Why?

In an external package, suppose we have:

```tsx
import { Json } from '@liveblocks/client';
import { privateFn } from '@liveblocks/client/internal';

const value: Json = ...;
privateFn(value);
//             ^^^^^ Error!
```

To fix this, we can:

```tsx
import { privateFn, Json } from '@liveblocks/client/internal';

const value: Json = ...;
privateFn(value);
//             ^^^^^ OK!
```

Why is this? It's because technically TypeScript cannot be sure that the types are the same.

```tsx
import { Json } from '@liveblocks/client';
import { Json } from '@liveblocks/client/internal';
import { Json } from 'whatever';
```

To TypeScript, all of the above imports might be completely different types. We'll need to have a way of telling Typescript that the public and internal `Json` types are the same! One way to do that would be to not duplicate all those public APIs, but instead let rollup build this output in a way where the internals module locally imports from `"./index.d.ts"` somehow.

nvie:
This changes our Rollup build to let our public and internal modules share the same internals.

## Loom video
[](https://www.loom.com/share/1762462e9c4e4b948c06c275125df5fb)

## Results
Bundlephobia is [happy about this change](https://bundlephobia.com/package/@liveblocks/client@0.16.4-beta1), too:

As you can see, it now gets recognized as the first release that has official tree-shaking support, and the overall bundle size shrunk.

## The problem
Currently, in secondary libraries when we import code from both:

```ts
import { foo } from '@liveblocks/client';
import { bar } from '@liveblocks/client/internal';
```

Then, if `bar` also relies on `foo`, it will not be the same `foo`. This has consequences for our TypeScript types, and for module globals that should be singletons (like package-level caches). Importing from the public and internal APIs effectively creates two instances of those package-level caches, but there should be only one source.

## The cause
The cause is the way we currently build our packages with Rollup. For example, just looking at the ESM modules, we make separate build outputs [for the public module](https://github.com/liveblocks/liveblocks/blob/10f863eb447c3902c7c70e3d373b5e33a29f1d60/packages/liveblocks-client/rollup.config.js#L107) and [the internal module](https://github.com/liveblocks/liveblocks/blob/10f863eb447c3902c7c70e3d373b5e33a29f1d60/packages/liveblocks-client/rollup.config.js#L111). This does not allow Rollup to use code splitting/sharing.

## The solution
Generate all ESM files in a single build step, and allow Rollup's code splitting/sharing to emit three files per "category":

- `index.js`
- `internals.js`
- `shared.js`

The `index` and `internals` get "exposed" in the package as entry points, but most of the code will be sitting in `shared`, which they both import from.

This way, all types and module globals are ultimately coming from the same place, without sacrificing tree-shakability. Even shrinking our bundle size, and the bundle size of all customers!

## Future work
This PR changes the build for `@liveblocks/client` only. However, I think it would be nice to use the same setup and generated output structure for our other packages.

Fixes #187.


packages/liveblocks-client/package.json
```diff
@@ -59,8 +46,8 @@
     "msw": "^0.39.1",
     "node-fetch": "2.6.7",
     "rollup": "^2.68.0",
+    "rollup-plugin-command": "^1.1.3",
     "rollup-plugin-dts": "^4.1.0",
-    "rollup-plugin-esbuild": "^4.8.2",
```
nvie, line 63: Not really sure what we needed `esbuild` for here. I started trying to get this to work with the normal typescript plugin, and assumed we only needed to use `esbuild` for speed, but the build is as fast now as it is on `main`, so I figured if we don't need `esbuild`, having fewer build tool in the mix is better. (I think following the same reasoning, perhaps we can get rid of Babel here too and let the TypeScript plugin output those CJS.)

Please let me know if I'm missing the rationale for `esbuild` or whether I'm overlooking something 🙏

packages/liveblocks-react/rollup.config.js
```diff
@@ -1,20 +1,43 @@
-/**
- * This configuration is heavily inspired from https://github.com/pmndrs/zustand/blob/main/rollup.config.js
- */
-import path from "path";
-import resolve from "@rollup/plugin-node-resolve";
 import babelPlugin from "@rollup/plugin-babel";
-import typescript from "@rollup/plugin-typescript";
-import esbuild from "rollup-plugin-esbuild";
+import commandPlugin from "rollup-plugin-command";
```
nvie, line 2: Please ignore all the noise caused here. If you've reviewed `rollup.config.js` in the client project, it's now 99.999% the same for the other projects. Only differences are the input file params — think `['index.ts', 'internal.ts']` vs `['index.tsx']`!

packages/liveblocks-client/package.json
GuillaumeSalles: We currently use `esbuild` because the last time I refactored our build system, I took inspiration from : https://github.com/pmndrs/zustand/blob/main/rollup.config.js

Performance is not an issue for our small bundle, so let's remove it if the only impact it does not impact the output

GuillaumeSalles (approved):

PR #192, opened 2022-05-06
Simplify internal imports

nvie:
Following #189, we can now import these in a non-hacky way, too.


GuillaumeSalles (approved):
Thanks for cleaning that up!

PR #194, opened 2022-05-09
Link script improvements

nvie:
Tweaks the link script a bit to make it simpler and more reliable. Previously, the check to see if the build is still fresh was relying on comparing source vs build file timestamps, which is unreliable. In this PR, I'm switching that check to verify that using SHA1 hashes on the input files.

Also, this level of robustness completely removes the need for the `-f` flag, which I've made optional now. You can now safely change any invocation of `link-liveblocks.sh -f` to just `link-liveblocks.sh`.

While I was at it, I've made the logs it emits a bit more readable too.


nvie:
Going to get this merged in to improve our builds. If you have comments though, please lmk and I will adjust in retrospect!

PR #195, opened 2022-05-09
Only export internals that are needed/used

nvie:
This PR changes the export declarations inside `internals.ts` to only export what's actually used by secondary packages. The use of `*` will default to exporting too much in the bundle, and thus potentially makes the bundle larger than strictly needed.

By only exporting what's actually used, rollup will automatically tree-shake off any dead code, if we have any. If we ever find ourselves in need of exposing more internal APIs, we can simply add those to the list of exports, and the bundle will adjust.

### DIFF
--- a/packages/liveblocks-client/src/internal.ts
+++ b/packages/liveblocks-client/src/internal.ts
@@ -11,8 +11,44 @@
  * https://join.team/liveblocks ;)
  */
 
-export * from "./live";
-export * from "./position";
+export type {
+  ClientEventMessage,
+  ClientMessage,
+  ClientMessageType,
+  CrdtType,
+  CreateListOp,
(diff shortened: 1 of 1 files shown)


packages/liveblocks-client/src/internal.ts
```diff
@@ -11,18 +11,14 @@
  * https://join.team/liveblocks ;)
  */
 
-export * from "./live";
-export * from "./position";
```
nvie, line 15: Turns out nothing from `position.ts` was used anywhere…?

nvie:
Getting this merged in to rebase the deprecation work on top of that I'd love to push out today. As always, please do review and comment in retrospect! 🙏

PR #196, opened 2022-05-09
Use type-only imports where possible

nvie:
This is a small clean-up pass that utilizes TypeScript's ["type-only imports" and exports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export). By enabling this flag in `tsconfig.json`, TypeScript will throw errors if you're only importing types from a specific module, and remind you to declare the imports as `import type` instead. By doing so, you statically declare that all of those imports will all be types. This is useful to tools like Babel as they can know for sure that those imports are intended to strictly be types. I also find it makes code more readable to me as a developer. Reading `import { Json } from './json'`, you instantly know that it's just a type, and not a class, for example.

This can also help reduce chances of import cycles in some cases. If an import cycles is only needed at the type level, then using `import type` instead removes the import cycle at runtime, so tools like Node, Babel, or bundlers have an easier time.

### DIFF
--- a/packages/liveblocks-client/src/AbstractCrdt.ts
+++ b/packages/liveblocks-client/src/AbstractCrdt.ts
@@ -1,5 +1,6 @@
-import { CreateOp, Op, OpType, SerializedCrdt } from "./live";
-import { StorageUpdate } from "./types";
+import { OpType } from "./live";
+import type { CreateOp, Op, SerializedCrdt } from "./live";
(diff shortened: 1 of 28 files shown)


nvie:
Getting this merged in to rebase the deprecation work on top of that I'd love to push out today. As always, please do review and comment in retrospect! 🙏

PR #197, opened 2022-05-09
Restore `exports` setting

nvie:
Restores the exports setting. I found a case where this export is still needed.

If you have a (modern) project depending on Redux, and you're importing from `@liveblocks/redux` somewhere using ES modules, then that will resolve to the file `liveblocks-redux/lib/index.mjs`, which contains this line of code:

```jsx
import { patchLiveObjectKey, lsonToJson, patchImmutableObject } from '@liveblocks/client/internal';
```

However, in `.mjs` files, it's crucial to import using file extensions. We can fix this in two ways:

- Either this compiled module should import from `@liveblocks/client/internal.mjs` instead; or
- We set up this subexport, so this `.mjs` file can find the source

This PR does the latter, to avoid having to add hacks to rewrite those imports in generated `.mjs` files.

### DIFF
--- a/packages/liveblocks-client/package.json
+++ b/packages/liveblocks-client/package.json
@@ -5,6 +5,16 @@
   "main": "./index.js",
   "module": "./index.mjs",
   "types": "./index.d.ts",
+  "exports": {
+    "./internal": {
+      "require": "./internal.js",
+      "import": "./internal.mjs"
+    },
+    ".": {
+      "require": "./index.js",
+      "import": "./index.mjs"
+    }
+  },
   "files": [
     "**"
   ],


packages/liveblocks-client/rollup.config.js
```diff
@@ -186,7 +186,10 @@ export default async () => {
   const external = [
     ...Object.keys(pkgJson?.dependencies ?? {}),
     ...Object.keys(pkgJson?.peerDependencies ?? {}),
-  ];
+  ].flatMap((dep) =>
+    // Also include `@liveblocks/.../internal` as an external reference, by convention
+    dep.startsWith("@liveblocks/") ? [dep, dep + "/internal"] : [dep]
+  );
```
nvie, line 192: After restoring this `"exports"` section, I noticed the CommonJS build started suffering from an issue, where `@liveblocks/client/internal` would no longer be recognized as an "external" reference any longer. This somehow wasn't a problem for the ESM build.

Therefore, I have added this "fix" in our Rollup builds, to consider both `@liveblocks/client` _and_ `@liveblocks/client/internal` as external refs.

This had the following effect on the build output:
https://gist.github.com/nvie/938f21ae8bb8c5daca1287553a816605

nvie:
Getting this merged now, because it may start breaking projects that suffer from the use case mentioned in the PR description once we release 0.16.4.

PR #201, opened 2022-05-09
Fix broken examples

nvie:


nvie:
> Could you do a quick `./scripts/for-all-examples.sh npx prettier --write --ignore-path .gitignore '**/*.{js,jsx,ts,tsx,json,md,mdx,yml,css,vue,svelte,html}'`?

Amazing, thx for calling that out! 😍

PR #208, opened 2022-05-10
Remove deprecated `Room.unsubscribe()` APIs

nvie:
Removes the `Room.unsubscribe()` APIs that have been deprecated since 0.13. There should not be any usage of this in the wild anymore. Removing these now, prior to continuing on the TypeScript improvements.

Docs are removed in https://github.com/liveblocks/liveblocks.io/pull/207.


GuillaumeSalles (approved):

nvie:
Merging this in, now that `main` is targeting the 0.17 release next week.

PR #209, opened 2022-05-10
Fix: Nuxt and Svelte examples

marcbouchenoire:
This PR mostly fixes the `nuxtjs-live-avatars` example currently being deployed as a static site without its authentication endpoint.


nvie (approved):
I don't have enough experience in this area to give a valuable review here, but I'll take your word for it! 🚀

PR #210, opened 2022-05-10
Add custom script to smartly skip Vercel builds

nvie:
This morning, @marcbouchenoire and I paired on our Vercel setup, and wanted the best of these both worlds:

1. Rebuild/redeploy the examples if they changed, or if any of the Liveblocks packages changed.
2. Reduce the number of (concurrent) builds we issue on every commit, to save precious time and $$$$$.

This PR adds a script we can run for each example to decide smartly whether to trigger a Vercel deployment or not, based on the contents of the Git diff.

It works as follows.

```
$ should-example-rebuild-on-vercel.sh -v nextjs-live-cursors
Example "nextjs-live-cursors" unaffected by change.
```

You give it the name of an example, and it returns exit code 0 (no rebuild needed) or 1 (please rebuild). It determines that value in a smart way: if any of the files in the example itself have changed (compared to `origin/main`, that is), it requests a rebuild. But it will also read the dependencies from `package.json` and figure out which Liveblocks packages this relies on dynamically. When so, it will also check if any of the files in that package have changed.

Some files, like `*.md` files, hidden dotfiles, or `*.test.*` files will be ignored and won't trigger a build.

This command is intended to be suitable to run on Vercel, specifically for this config: https://vercel.com/docs/concepts/projects/overview#ignored-build-step

### DIFF
--- a/scripts/should-example-rebuild-on-vercel.sh
+++ b/scripts/should-example-rebuild-on-vercel.sh
@@ -1,0 +1,125 @@
+#!/bin/sh
+set -eu
+
+# Ensure this script can assume it's run from the repo's
+# root directory, even if the current working directory is
+# different.
+ROOT="$(git rev-parse --show-toplevel)"
+if [ "$(pwd)" != "$ROOT" ]; then
+    ( cd "$ROOT" && exec "$0" "$@" )
+    exit $?
+fi
+
+err () {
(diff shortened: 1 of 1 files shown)


marcbouchenoire (approved):
Lovely. 🔥

PR #213, opened 2022-05-11
Allow skipping Vercel builds by using specific tags in commit messages

nvie:
Examples of tags you can put in commit messages that will cause example builds to be skipped for those PRs/branches:

- `[no deploy]`
- `[skip Vercel]`
- `[skip examples]`

Basically any combination of: `[{no,skip} {deploy,examples,vercel}]` is recognized (case-insensitive).

Once such a commit lands on a PR or branch, there is no way to "undo" that, unless you rewrite the Git history by removing the commit or changing the commit message.

### DIFF
--- a/scripts/should-example-rebuild-on-vercel.sh
+++ b/scripts/should-example-rebuild-on-vercel.sh
@@ -90,6 +90,14 @@ get_all_changed_files () {
         # on Vercel environments
         echo '
         const data = require("./diff-since-main.json");
+
+        // If some commits on this branch have the phrase "[no vercel]" or
+        // "[skip examples]" or some combination like that in their message, skip
+        // this build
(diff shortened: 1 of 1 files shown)


nvie:
@marcbouchenoire As you can see, this PR is a self-documenting example. All of the Vercel builds in this PR have been skipped :)

marcbouchenoire (approved):
Is this planning ahead or has this use-case already come up within the first day? 😅

nvie:
> Is this planning ahead or has this use-case already come up within the first day? 😅

Haha! Planning ahead, kind of. Also, I have some PRs that are longer-living to be merged in later (for a later release), so I figured having a way to disable these would be nice either way. I'll definitely use this feature from time to time :)

PR #217, opened 2022-05-11
Upkeep: Example READMEs

marcbouchenoire:
This PR adds badges linking to (upcoming) live previews for each example, and fixes broken relative links.


nvie (approved):
Nice! Thank you for cleaning these up and making them more consistent! 🙌

PR #242, opened 2022-05-13
Uncertainty cleanups

nvie:
This PR avoids using the non-null assertion operator (`!`) in TypeScript, because they can mask real problems with the code. The steps I've taken here are pretty mechanical refactorings:

- I've added a `nn()` helper, which stands for "non-nullable". By making this a method, we can let it throw a runtime error during development if ever such assertion turns out not to be true. In production, this will be a no-op, just like the non-null assertion (`!`) operator is.
- Next, I've mechanically refactored all call sites using that pattern, and replaced those with `nn()` calls. In some cases, I was able to refactor the code in a different way which avoided the operator altogether.

Running our unit test suite pointed out two places in the code where actually seem to rely on the fact that we inject/use `undefined` values at runtime. Protecting against those values actually makes the test suite fail. I've annotated those places clearly, and may need some help to interpret the edge cases.

Also, there can be more cases like this that aren't covered by our unit tests of course!

### DIFF
--- a/packages/liveblocks-client/src/room.ts
+++ b/packages/liveblocks-client/src/room.ts
@@ -1,5 +1,6 @@
 import type { ApplyResult } from "./AbstractCrdt";
 import { AbstractCrdt, OpSource } from "./AbstractCrdt";
+import { nn } from "./assert";
 import { LiveList } from "./LiveList";
 import type { LiveMap } from "./LiveMap";
 import { LiveObject } from "./LiveObject";
@@ -607,7 +608,7 @@ export function makeStateMachine<TPresence extends JsonObject>(
 
         const applyOpResult = applyOp(op, source);
         if (applyOpResult.modified) {
-          const parentId = applyOpResult.modified.node._parent?._id!;
+          const parentId = nn(applyOpResult.modified.node._parent?._id);
 
           // If the parent was created in the same batch, we don't want to notify
           // storage updates for the children.
--- a/packages/liveblocks-client/src/room.ts
+++ b/packages/liveblocks-client/src/room.ts
@@ -264,8 +264,7 @@ export type State<TPresence extends JsonObject> = {
       storageUpdates: Map<string, StorageUpdate>;
     };
   };
-  offlineOperations: Map<string | undefined, Op>;
-  //                              ^^^^^^^^^ NOTE: Bug? Unintended?
+  offlineOperations: Map<string, Op>;
 };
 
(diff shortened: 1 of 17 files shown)


packages/liveblocks-client/src/room.ts
```diff
@@ -561,7 +563,8 @@ export function makeStateMachine<TPresence extends JsonObject>(
       },
     };
 
-    const createdNodeIds = new Set<string>();
+    const createdNodeIds = new Set<string | undefined>();
+    //                                      ^^^^^^^^^ NOTE: Bug? Unintended?
```
nvie, line 567: Here's one case where the unit test suite pointed out a case where we annotate that this Set only ever contains strings, but in reality it also relies on `.add(undefined)` and `.has(undefined)` checks. This may need some investigation.

packages/liveblocks-client/src/room.ts
```diff
@@ -258,7 +259,8 @@ export type State<TPresence extends JsonObject> = {
       storageUpdates: Map<string, StorageUpdate>;
     };
   };
-  offlineOperations: Map<string, Op>;
+  offlineOperations: Map<string | undefined, Op>;
+  //                              ^^^^^^^^^ NOTE: Bug? Unintended?
```
nvie, line 263: Here's another case where the unit test suite pointed out a case where we annotate that this Map only ever contains string keys, but in reality it also relies on `.set(undefined, ...)` and `.get(undefined)` calls. This may need some investigation.

ofoucherot: Operations created by LiveList.clear were missing an `opId`.

packages/liveblocks-client/src/room.ts
ofoucherot: Parent can be undefined if it's the root. Thank you for finding this!

packages/liveblocks-client/src/room.ts
nvie: Ah! Glad we found this bug. Should `opId` be a mandatory property on those operations? Or are there valid reasons for an `opId` to be missing?

GuillaumeSalles: Operations sent over the network always contain an `opId` but they do not while inside the undo/redo stack. We assign an `opId` to operations from the undo/redo stack just before sending them to the server. There is probably a better way to model it (tuple `[opId, Op]`?) while reworking the op structure.

ofoucherot (approved):
🚀

PR #246, opened 2022-05-16
Export missing internals

nvie:
This restores some of the exports that were removed before, but turn out to be used by our Cloudflare codebase.


nvie:
Going to merge this into main, and also roll a 0.16.7 release for this.

PR #249, opened 2022-05-17
Split `live.ts` module into smaller modules

### LINKED ISSUE #247: Internal naming cleanups for messages and ops
This is prepwork for https://github.com/liveblocks/liveblocks-cloudflare/issues/53.

Proposed (internal) naming cleanup:

- Split up `live.ts` file into multiple smaller modules to separate client messages, server messages, and operations.
- Export all of those internals again, so the Cloudflare repo can use all of those internal types and write runtime validations for that that remain in sync over time, as those types evolve (or even support multiple concurrent versions of the protocol at the same time)
- Abbreviate long type names by using `Msg` instead of `Message` (because this is used so often)
  - `ServerMessage` → `ServerMsg`
  - `ClientMessage` → `ClientMsg`
- For consistency, make all members of such unions end with `ServerMsg` or `ClientMsg` suffixes
- Rename enums with op and message codes, and capitalize all members so they are more easily recognized as consts/enum members elsewhere. (It took me a while to distinguish the union members from enum members.)
  - `OpType` → `OpCode`
  - `ServerMessageType` → `ServerMsgCode`  
    Example: `ServerMessageType.UpdatePresence` → `ServerMsgCode.UPDATE_PRESENCE`
  - `ClientMessageType` → `ClientMsgCode`
    Example: `ClientMessageType.FetchStorage` → `ClientMsgCode.FETCH_STORAGE`

Because none of these types are "public", we can easily migrate them, since there will be no difference at the runtime or at the protocol level.

nvie:
This splits the various definitions from `src/live.ts` into several, smaller, modules and moves them into `src/types/*.ts`. While at it, this also moves the `json.ts` and `lson.ts` modules into `src/types/Json.ts` and `src/types/Lson.ts` respectively, and moves `src/types.ts` to `src/types/index.ts`.

There are more types we can split out this way inside `src/types/index.ts` itself, so that eventually that index file is going to be nothing more than a re-exporting index file for convenience.

There are no functional changes made in this PR, only code moving around.

Fixes #247 .


nvie:
Changes approved [in Slack](https://liveblocks.slack.com/archives/C02PZL7QAAW/p1652714485571889?thread_ts=1652712658.836539&cid=C02PZL7QAAW). Since this is mechanical, going to get this merged in now, but please let me know if you have feedback — happy to adjust.

PR #251, opened 2022-05-18
Be precise about Crdt types

nvie:
This PR tightens the type security around some of our most basic SerializedCrdt union type. The key addition here is the notion of an explicit `SerializedRootObject` (which is like a "normal" `SerializedObject`, but without parent references).

Also note the addition of the `NodeMap` and `ParentToChildNodeMap` helper types, which will make cleaning up these types in our Cloudflare backend a lot easier, too.


packages/liveblocks-client/src/types/SerializedCrdt.ts
```diff
@@ -8,15 +8,23 @@ export enum CrdtType {
...
   | SerializedList
   | SerializedMap
   | SerializedRegister;
 
+export type SerializedRootObject = {
+  type: CrdtType.OBJECT;
+  parentId?: undefined;
+  parentKey?: undefined;
+  data: JsonObject;
+};
```
nvie, line 22: NOTE: This is almost the same as a "normal" SerializedObject. Please note that this is NOT a new Op code or something like that. Just codifies at the type level that either both keys are set, or both are not.

nvie:
Going to get these type improvements in, as they should not have an impact on clients. They do enable other improvements in our Cloudflare backend, though.

PR #252, opened 2022-05-20
Improve accuracy of Op types

nvie:
Like #251 , but applies the same treatment to `Op` type as well. By treating the root object as the edge case (where parent id/key are missing), we get a lot more certainty in all the other cases, simplifying all the remaining code massively.


packages/liveblocks-zustand/src/index.test.ts
```diff
@@ -279,17 +275,15 @@ describe("middleware", () => {
...
-      );
+      expect(JSON.parse(socket.sentMessages[0])).toEqual([
+        {
+          type: ClientMsgCode.UPDATE_PRESENCE,
+          data: { cursor: { x: 0, y: 0 } },
+        },
+        {
+          type: ClientMsgCode.FETCH_STORAGE,
+        },
+      ]);
```
nvie, line 286: Sorry about the prettier-induced noise here. The structural change made here is:

```expect(xxx).toEqual(JSON.stringify(yyy))```

→

```
expect(JSON.parse(xxx!)).toEqual(yyy)
```

So that these tests are more robust in the light of field ordering.

PR #254, opened 2022-05-23
Slay some `any`s

nvie:


e2e/next-sandbox/test/storage/list.test.ts
```diff
@@ -117,7 +117,7 @@ test.describe("Storage - LiveList", () => {
     await assertContainText(pages, "0");
   });
 
-  test("fuzzy with undo/redo push delete and move", async () => {
+  test.skip("fuzzy with undo/redo push delete and move", async () => {
```
nvie, line 120: @GuillaumeSalles Not sure if this is related to the recent list bug fixes, but this test seems flaky again on this PR. I cannot for the life of me figure out how the changes made here can make this test fail, since they're only type-level changes, and one removal of a method that is dead code as far as I can tell.

PR #257, opened 2022-05-24
Require explicit annotations at module boundaries

nvie:
This configures ESLint to require explicit type annotations at all the module boundaries. Specifically, function return types. TypeScript is good at inferring return types for functions, but sometimes those are implicitly `any`, and as a result, using those functions leads to a false sense of safety.

### DIFF
--- a/packages/liveblocks-redux/test/utils.ts
+++ b/packages/liveblocks-redux/test/utils.ts
@@ -1,3 +1,4 @@
+import type { Json } from "@liveblocks/client";
 import type {
   IdTuple,
   SerializedList,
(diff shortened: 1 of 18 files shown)


packages/liveblocks-redux/test/utils.ts
```diff
@@ -138,7 +139,7 @@ export function register(
   id: string,
   parentId: string,
   parentKey: string,
-  data: any
+  data: Json
```
nvie, line 142: Side-fix.

packages/liveblocks-zustand/test/utils.ts
```diff
@@ -138,7 +139,7 @@ export function register(
   id: string,
   parentId: string,
   parentKey: string,
-  data: any
+  data: Json
```
nvie, line 142: Side-fix.

PR #259, opened 2022-05-24
More explicit state management for parent links

nvie:
This PR makes the state management around "parent links" more explicit inside the AbstractCrdt base class.


ofoucherot (approved):

nvie:
I added a few more cleanups after your review, @ofoucherot! Mostly just to further improve readability, but nothing very interesting was added.

PR #263, opened 2022-05-25
Upgrade TypeScript to 4.7

nvie:


packages/liveblocks-client/src/types/_compat.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * In TypeScript 4.6, VisibilityState got renamed to
+ * DocumentVisibilityState. There's no good way to conditionally
+ * (based on the TypeScript version our users are using) read one
+ * or the other, so we're defining the type locally ourselves
+ * here.
+ */
+
+export type DocumentVisibilityState = "hidden" | "visible";
```
nvie, line 9: ☝️ This is a small but cheap solution to support all TypeScript versions, both <4.6, or >=4.6.

PR #269, opened 2022-05-26
Test helper type improvements

nvie:
Some random type improvements to our test helpers, to make it easier to catch incorrect usages.

### DIFF
--- a/packages/liveblocks-client/src/immutable.ts
+++ b/packages/liveblocks-client/src/immutable.ts
@@ -2,7 +2,14 @@ import { LiveList } from "./LiveList";
 import { LiveMap } from "./LiveMap";
 import { LiveObject } from "./LiveObject";
 import { LiveRegister } from "./LiveRegister";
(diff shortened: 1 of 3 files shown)


packages/liveblocks-client/src/immutable.ts
```diff
@@ -181,11 +188,11 @@ export function patchLiveList<T extends Lson>(
   }
 }
 
-export function patchLiveObjectKey<O extends LsonObject>(
+export function patchLiveObjectKey<O extends LsonObject, J extends Json>(
```
nvie, line 191: Oh lol, this is catching a lot of issues in the Zustand package. Sweet, will have to take a look at fixing those first then, I guess.

GuillaumeSalles (approved):
I'm assuming this goes with #288 ?

LGTM!

nvie:
> I'm assuming this goes with #288 ?

Yeah, the issues from this PR have been fixed by #288 indeed, so tests are now all green 👍

PR #274, opened 2022-05-30
Remove `intent` field from root object's `Op` type

nvie:
This PR fixes the PR feedback from https://github.com/liveblocks/liveblocks-cloudflare/pull/80#discussion_r884374547. `intent` (and `deletedId`) are always set together and are only set on "child" ops, never for the root object.

While changing this, I found more type issues that could be solved by being more precise in the `_serialize()` type signature, which was a nice side benefit.

### DIFF
--- a/packages/liveblocks-client/src/LiveList.ts
+++ b/packages/liveblocks-client/src/LiveList.ts
@@ -1314,7 +1314,7 @@ function sortListItem(items: LiveNode[]) {
  * serializing a LiveStructure should not know anything about intent
  */
 function addIntentAndDeletedIdToOperation(
-  ops: Op[],
+  ops: CreateChildOp[],
   deletedId: string | undefined
 ) {
   if (ops.length === 0) {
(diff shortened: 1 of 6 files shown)


packages/liveblocks-client/src/LiveList.ts
```diff
@@ -1324,17 +1324,6 @@ function addIntentAndDeletedIdToOperation(
...
-  if (
-    firstOp.type !== OpCode.CREATE_LIST &&
-    firstOp.type !== OpCode.CREATE_OBJECT &&
-    firstOp.type !== OpCode.CREATE_REGISTER &&
-    firstOp.type !== OpCode.CREATE_MAP
-  ) {
-    throw new Error(
-      "Internal error. Serialized LiveStructure first op should be CreateOp"
-    );
-  }
```
nvie, line 1336: This is now caught at the type level 👍

packages/liveblocks-client/src/LiveObject.ts
```diff
@@ -55,14 +56,20 @@ export class LiveObject<O extends LsonObject> extends AbstractCrdt {
...
    * @internal
    */
-  _serialize(parentId?: string, parentKey?: string, doc?: Doc): Op[] {
+  _serialize(parentId: string, parentKey: string, doc?: Doc): CreateChildOp[];
+  _serialize(
+    parentId?: undefined,
+    parentKey?: undefined,
+    doc?: Doc
+  ): CreateOp[];
+  _serialize(parentId?: string, parentKey?: string, doc?: Doc): CreateOp[] {
```
nvie, line 65: Overloaded here, only for `LiveObject`. When `parentId` and `parentKey` aren't provided, this will return a `CreateOp[]`, otherwise, all the entries in here will be `CreateChildOp[]` (which excludes the possibility of having a "root object" op in there).

PR #276, opened 2022-05-30
Fixes LiveList.clear updates and refactor a bit test updates

### LINKED ISSUE #265: LiveList.clear does not generate the right updates
`LiveList.clear` does not generate the right updates. This break immutability patch in zustand/redux.

GuillaumeSalles:
This PR fixes #265 and create a new testing function for testing update.

### DIFF
--- a/packages/liveblocks-client/e2e/list.test.ts
+++ b/packages/liveblocks-client/e2e/list.test.ts
@@ -1,10 +1,5 @@
 import "regenerator-runtime/runtime";
 import { LiveList } from "../src/LiveList";
-import {
-  listUpdate,
-  listUpdateInsert,
-  listUpdateMove,
-} from "../test/updatesUtils";
 import { prepareTestsConflicts } from "./utils";
 
 describe("LiveList conflicts", () => {
@@ -15,7 +10,7 @@ describe("LiveList conflicts", () => {
         {
           list: new LiveList(),
         },
-        async ({ root1, root2, wsUtils, assert, updates1, updates2 }) => {
+        async ({ root1, root2, wsUtils, assert }) => {
           root1.get("list").push("A");
           root2.get("list").push("B");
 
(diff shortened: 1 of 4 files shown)


packages/liveblocks-client/e2e/list.test.ts
```diff
@@ -26,16 +21,6 @@ describe("LiveList conflicts", () => {
           await wsUtils.flushSocket2Messages();
 
           assert({ list: ["A", "B"] });
-
```
GuillaumeSalles, line 29: I removed these assertions to simplify the tests.

Updates are not strictly validated. They are applied on an immutable state, then we assert that this immutable state is consistent with the LiveStructure.

packages/liveblocks-client/src/LiveList.test.ts
```diff
@@ -108,38 +138,10 @@ describe("LiveList", () => {
       });
 
       items.push(new LiveObject({ a: 0 }));
-      assert(
```
GuillaumeSalles, line 111: I removed updates assertion in this unit tests because it already is doing a lot. Instead I created a `prepareStorageUpdateTest` function that help us test only tests updates.

packages/liveblocks-client/src/LiveList.ts
```diff
@@ -1015,10 +1014,10 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
             ...item._serialize(nn(this._id), item._getParentKeyOrThrow())
           );
 
-          updateDelta.push(deleteDelta(i));
+          // Index is always 0 because updates are applied one after another
+          // when applied on an immutable state
+          updateDelta.push(deleteDelta(0));
```
ofoucherot, line 1019: Ah yes!

ofoucherot (approved):
🚀 
Testing the updates by applying them with immutability is 💯

PR #284, opened 2022-05-31
React native/atob

vinodsantharam:
Hi guys.

After a large brainstorming with Guillaume, we have figured out a solution to prompt a message to developers who are in a JS environment without atob operator. 
So we add a new property to the ClientOptions in order to let the dev use a polyfill for the operator atob.


packages/liveblocks-client/src/client.ts
```diff
@@ -52,6 +52,16 @@ export function createClient(options: ClientOptions): Client {
   const clientOptions = options;
   const throttleDelay = getThrottleDelayFromOptions(options);
 
+  if (typeof atob == "undefined") {
+    if (options.atobPolyfill == undefined) {
+      throw new Error(
+        "You need to polyfill atob operator. Please follow the instructions at https://liveblocks.io/docs/errors/liveblocks-client/atob-polyfill"
+      );
+    }
+    // At this point, atob does not exist so we either on React Native or on Node<16, hence global is available.
```
GuillaumeSalles, line 61:
```suggestion
    // At this point, atob does not exist so we are either on React Native or on Node < 16, hence global is available.
```

packages/liveblocks-client/src/client.node.test.ts
```diff
@@ -135,3 +141,38 @@ describe("createClient", () => {
...
+
+  test("should throw error if atob polyfill is not set", () => {
+    expect(() => {
+      const client = createClient({
+        WebSocketPolyfill: MockWebSocket,
+        fetchPolyfill: fetchMock,
+        atobPolyfill: undefined,
+      } as ClientOptions);
+    }).toThrowError(
+      "You need to polyfill atob operator. Please follow the instructions at https://liveblocks.io/docs/errors/liveblocks-client/atob-polyfill"
```
nvie, line 165: I love these error detail links, and would love to add more of these. They're super helpful to developers. Also, one small suggestion:

```suggestion
      "You need to polyfill the atob function. Please follow the instructions at https://liveblocks.io/docs/errors/liveblocks-client/atob-polyfill"
```

packages/liveblocks-client/src/client.ts
```diff
@@ -52,6 +52,16 @@ export function createClient(options: ClientOptions): Client {
   const clientOptions = options;
   const throttleDelay = getThrottleDelayFromOptions(options);
 
+  if (typeof atob == "undefined") {
+    if (options.atobPolyfill == undefined) {
+      throw new Error(
+        "You need to polyfill atob operator. Please follow the instructions at https://liveblocks.io/docs/errors/liveblocks-client/atob-polyfill"
```
nvie, line 58:
```suggestion
        "You need to polyfill the atob function. Please follow the instructions at https://liveblocks.io/docs/errors/liveblocks-client/atob-polyfill"
```

nvie (approved):
I love this solution and I love that the error automatically shows in the relevant context, and by prompting the dev with helpful steps! Nice work! 💯

GuillaumeSalles (approved):
Perf! Thank you for pushing this through!

packages/liveblocks-client/src/types/index.ts
```diff
@@ -264,6 +264,7 @@ export type ClientOptions = {
   throttle?: number;
   fetchPolyfill?: any;
   WebSocketPolyfill?: any;
+  atobPolyfill?: (data: string) => string;
```
nvie, line 267: Thinking out loud… would this be a good time to restructure the options a bit, into this shape:

```ts
export type Polyfills = {
  fetch?: any;
  WebSocket?: any;
  atob?: (data: string) => string;
};

export type ClientOptions = {
  throttle?: number;
  polyfills?: Polyfills;
} & (
  | { publicApiKey: string; authEndpoint?: never }
  | { publicApiKey?: never; authEndpoint: AuthEndpoint }
);
```

I think that might make it a bit more readable and obvious what these are for. If we do so, we should definitely keep the existing two (`fetchPolyfill` and `WebSocketPolyfill`) on the ClientOptions in 0.17, but we could deprecate them and remove them in 0.18.

nvie (approved):
This is great! Thanks for plowing through this! 💯

packages/liveblocks-client/src/types/index.ts
nvie: This won't have to be part of this PR, btw! We could do that in a separate one.

nvie: Since we'll want to include this change with the 0.17 release, I think it'll be nice to also include this suggested reorganization into a `polyfills` sub-object. I have a item on my 0.17 checklist for that (see #303), so we won't forget about it.

PR #288, opened 2022-06-01
Type `patchLiveObjectKey` more correctly

nvie:
This had a few consequences where TypeScript now correctly pointed out some edge cases in the Redux and Zustand packages.

### DIFF
--- a/packages/liveblocks-client/src/immutable.ts
+++ b/packages/liveblocks-client/src/immutable.ts
@@ -71,22 +71,29 @@ function isPlainObject(obj: unknown): obj is { [key: string]: unknown } {
   );
 }
 
-function anyToCrdt(obj: unknown): any {
-  //                              ^^^ AbstractCrdt?
-  if (obj == null) {
-    return obj;
-  }
-  if (Array.isArray(obj)) {
-    return new LiveList(obj.map(anyToCrdt));
-  }
-  if (isPlainObject(obj)) {
+/**
+ * Deeply converts all nested lists to LiveLists, and all nested objects to
+ * LiveObjects.
+ *
+ * As such, the returned result will not contain any Json arrays or Json
+ * objects anymore.
+ */
+function deepLiveify(value: Lson | LsonObject): Lson {
+  if (Array.isArray(value)) {
+    return new LiveList(value.map(deepLiveify));
+  } else if (isPlainObject(value)) {
     const init: LsonObject = {};
-    for (const key in obj) {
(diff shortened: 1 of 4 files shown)


packages/liveblocks-client/src/immutable.ts
```diff
@@ -71,22 +71,29 @@ function isPlainObject(obj: unknown): obj is { [key: string]: unknown } {
...
-  }
-  if (isPlainObject(obj)) {
+/**
+ * Deeply converts all nested lists to LiveLists, and all nested objects to
+ * LiveObjects.
+ *
+ * As such, the returned result will not contain any Json arrays or Json
+ * objects anymore.
+ */
+function deepLiveify(value: Lson | LsonObject): Lson {
```
ofoucherot, line 81: I like the name :D

GuillaumeSalles: Ahah same

packages/liveblocks-client/src/immutable.ts
```diff
@@ -71,22 +71,29 @@ function isPlainObject(obj: unknown): obj is { [key: string]: unknown } {
...
+  if (Array.isArray(value)) {
+    return new LiveList(value.map(deepLiveify));
+  } else if (isPlainObject(value)) {
     const init: LsonObject = {};
-    for (const key in obj) {
-      init[key] = anyToCrdt(obj[key]);
+    for (const key in value) {
+      const val = value[key];
+      if (val === undefined) {
```
GuillaumeSalles, line 88: We're removing undefined keys which I think is ok, but what is the rationale behind it? Is it simplify `deepLiveify` signature?

GuillaumeSalles (approved):
👏 Thank you for improving these already tricky types

packages/liveblocks-client/src/immutable.ts
nvie: It's a remnant of the way we'd have to define `JsonObject` (see [this discussion](https://github.com/liveblocks/liveblocks/pull/165#issuecomment-1110586644) as a refresher). This means we can treat all JSON objects equally: with potentially explicitly-`undefined` values. In most loops, it means that if you skip over explicit-`undefined`s, then you get the behavior you really want, namely only keys that _do_ have a value. In the ideal world, looping over all entries in an object would not return explicitly-`undefined` keys, just as it doesn't return the implicitly-`undefined` keys.

In this case we have to because `init[key] = undefined` (on L91) is not a legal assignment in TypeScript.

PR #291, opened 2022-06-01
Deprecate concrete `Presence` type

### LINKED ISSUE #289: Deprecate the concrete `Presence` type as exposed by the client package
The [`Presence` type](https://github.com/liveblocks/liveblocks/blob/cd45ea70f05e796d24a75ce211dc36f72330bf94/packages/liveblocks-client/src/types/index.ts#L253) is currently defined as a concrete type, but presence should become a type _variable_, so users can bring their own types. Removing this public API is going to be a breaking change, so we'll start deprecating this type in 0.17, and remove it entirely in 0.18.

nvie:
**PR Stack:**

- #291  👈 This PR!
  - #316

---

In 0.17, we want to deprecate the use of the built-in concrete `Presence` type. This PR marks the `Presence` type deprecated, and stops using it in our own packages entirely, in favor of `JsonObject` constraints.

Users using the `Presence` type should have no problem adopting. If they keep using the deprecated type, it will keep working (because it's an alias for `JsonObject`). If they roll their own type definition, that's fine, or if they switch to use the opaque `JsonObject` type deliberately, that's also fine. The only case in which things will break is if their Presence data contains fields that aren't JSON-serializable that got accepted before.

Fixes #289 .

### 🏁 Checklist

Things to-do:

- [x] Go over each public API that has seen changes here
- [x] Go over each private API, and clean those up (no breaking changes)

### DIFF
--- a/packages/liveblocks-client/.eslintrc.js
+++ b/packages/liveblocks-client/.eslintrc.js
@@ -85,6 +85,16 @@ module.exports = {
         selector: 'TSTypeReference[typeName.name="AbstractCrdt"]',
         message: "Don't refer to AbstractCrdt as a type. Use LiveNode instead.",
       },
+
+      // Ensure we always specify all the type params in this code base. Even
+      // though many of these are (temporarily) optional, they are only
+      // optional for our end users, not internally.
+      {
+        selector:
+          "TSTypeReference[typeName.name=/^(Room|Machine|State|Effects|InternalRoom|User|OthersEvent|MyPresenceCallback|OthersEventCallback)$/][typeParameters.params.length != 1]",
+        message: "Missing type params.",
+      },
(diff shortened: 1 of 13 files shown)


packages/liveblocks-client/.eslintrc.js
```diff
@@ -85,6 +85,16 @@ module.exports = {
         selector: 'TSTypeReference[typeName.name="AbstractCrdt"]',
         message: "Don't refer to AbstractCrdt as a type. Use LiveNode instead.",
       },
+
```
GuillaumeSalles, line 88: Really nice :)

GuillaumeSalles (approved):
LGTM! Let's do the same for storage 💪

PR #294, opened 2022-06-03
Centralize, DRY up, and refine logic around token parsing

nvie:
Fixes the first part of https://github.com/liveblocks/liveblocks-cloudflare/issues/98. The manual type-checking in here is quite laborious since we cannot use decoders for this, but… this should be enough so that our private repos can import from and share the same types and implementation for this.


packages/liveblocks-client/src/utils.test.ts
```diff
@@ -281,30 +280,6 @@ describe("findNonSerializableValue", () => {
   );
 });
 
-describe("isTokenValid", () => {
```
GuillaumeSalles, line 284: Any reasons for deleting expiration unit tests?

GuillaumeSalles (approved):
Amazing 👏  so much cleaner!

packages/liveblocks-client/src/room.test.ts
```diff
@@ -87,70 +80,14 @@ describe("room / auth", () => {
     consoleErrorSpy.mockRestore();
   });
 
-  test("should reuse token after reconnect", async () => {
```
GuillaumeSalles, line 90: The two following tests are not validating what they think they're validating (and never have) for a few reasons.

`withDateNow` does not accept a promise as a callback.
`withDateNow` expects a number in millisecond and not in second.
`withDateNow` breaks `waitFor` because it use `Date.now` to retry.

We should use the state machine to have a better control of time to really validate token reuse. I'll create another ticket just to reintroduce those tests properly.

nvie: > `withDateNow` does not accept a promise as a callback.
> `withDateNow` expects a number in millisecond and not in second.
> `withDateNow` breaks `waitFor` because it use `Date.now` to retry.

Woah! 🤯 Thank you for digging in here!

PR #307, opened 2022-06-08
DRY up logic in test cases

nvie:
This PR DRYs up some of the logic in our test cases around setup of machines. This is a prefactoring for the upcoming #291, in which it would otherwise be super tedious to be annotating the type params everywhere. That can now be done in a single place, namely in `setupStateMachine()`, and the returns triplet of helpers will all have the same type params.


GuillaumeSalles (approved):
Nice little cleanup 👌

PR #308, opened 2022-06-08
Remove `any`-typed response from `fetchAuthEndpoint`

nvie:
This type-checks the response from the server more carefully, and returns a known type.


GuillaumeSalles (approved):
:chefkiss: Even better error message! A lot of users are still struggling with the custom auth endpoint so thank you for improving this pain point!

PR #312, opened 2022-06-09
Prettify rollup build output

nvie:
This PR sets up our Rollup builds to run prettier over the build output. Since we expect the packages to be included and [processed by a bundler](https://liveblocks.io/docs/guides/troubleshooting#process-not-defined), we don't ship minified builds ourselves. As such, we might as well prettify the output to make it easier to inspect what's in there.


ofoucherot (approved):
🌟

PR #316, opened 2022-06-09
Add `TStorage` type params

nvie:
Like #291, but this time focus on all the TStorage types, and passing those around internally correctly.


GuillaumeSalles (approved):
Everything make sense! There are probably things that I don't see but migrating examples will probably uncover some stuff

PR #326, opened 2022-06-10
Lift `TPresence` and `TStorage` up to the factory level

### LINKED ISSUE #324: Lift `TPresence` and `TStorage` up to the `create()` factory level

nvie:
**PR stack:**

- #326  👈 This PR
  - #333

---

Fixes #324.


packages/liveblocks-react/src/index.tsx
```diff
@@ -1,5 +1,7 @@
 export { LiveblocksProvider, useClient } from "./client";
 
+export { configureRoom } from "./factory";
```
nvie, line 3: @GuillaumeSalles After some rounds of renaming, I settled on `configureRoom()` instead of `create()`. So far, I like this name the best, because I thought `create()` was a bit too abstract/vague. Also, any names with "hooks" in it, I found confusing regarding `RoomProvider`. I like `configureRoom` because it's about configuring/setting up, and it implies that it's for a specific room (type), and therefore it makes sense when used multiple times as well, once for each room type an app needs.

PR #333, opened 2022-06-14
Deprecate storage hooks

nvie:
**PR stack:**

- #326
  - #333  👈 This PR

---

This build further on the code-generated compatibility layer introduced in #326. It extends it by allowing the factory (aka new-style) layer to export both `useMap` and `deprecated_useMap`, and if such a `deprecated_`-prefixed export is found there, the compat layer will be generated from _those_ functions instead.

This allows us to start changing the type params for these functions slightly. For example, when users are using classic `useMap` (from a direct import) and use type annotations in those call sites, then those remain working, but when the switch is made to the new factory-style `useMap`, then those type annotations will start to error, and the new returned type will automatically become `TStorage[TKey]` for those calls. The fix then is to remove those type annotations.

### DIFF
--- a/packages/liveblocks-client/package-lock.json
+++ b/packages/liveblocks-client/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@liveblocks/client",
-  "version": "0.16.13",
+  "version": "0.17.0-test1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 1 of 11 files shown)


nvie:
Approved while on a Zoom call, and we decided to get both of these PRs merged, and iterate on there while we go over examples and update documentation 👍

PR #336, opened 2022-06-14
fix(createClient): throw error on SK use

### LINKED ISSUE #330: Throw a descriptive error message if secret key is used instead of the public key

FlowFlorent:
Fixes #330

### DIFF
--- a/packages/liveblocks-client/src/client.node.test.ts
+++ b/packages/liveblocks-client/src/client.node.test.ts
@@ -53,6 +53,19 @@ describe("createClient", () => {
     ).not.toThrow();
   });
 
+  test("should throw if secret key is used instead of the public key", () => {
+    expect(() =>
+      createClientAndEnter({
+        publicApiKey: "sk_xxx",
+        WebSocketPolyfill: MockWebSocket,
+        fetchPolyfill: fetchMock,
+        atobPolyfill: atobPolyfillMock,
+      })
+    ).toThrowError(
+      "Invalid key. You are using the secret key which is not supported. Please use the public key instead. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientPublicKey"
+    );
+  });
+
   test("should not throw if WebSocketPolyfill is set", () => {
     expect(() => {
       createClientAndEnter({
--- a/packages/liveblocks-client/src/client.node.test.ts
+++ b/packages/liveblocks-client/src/client.node.test.ts
@@ -49,17 +49,17 @@ describe("createClient", () => {
       "authEndpoint must be a string or a function. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientAuthEndpoint",
     ],
     [
-      "sk_test_123",
+      "sk_xxx",
       undefined,
       "Invalid key. You are using the secret key which is not supported. Please use the public key instead. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientPublicKey",
     ],
     [
-      "pk_test_123",
+      "pk_xxx",
       "/api/auth",
       "You cannot use both publicApiKey and authEndpoint. Please use either publicApiKey or authEndpoint, but not both. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClient",
     ],
     [
-      "pk_test_123",
+      "pk_xxx",
       authEndpointCallback,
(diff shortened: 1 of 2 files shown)


packages/liveblocks-client/src/client.ts
```diff
@@ -181,6 +181,10 @@ function getThrottleDelayFromOptions(options: ClientOptions): number {
 function prepareAuthentication(clientOptions: ClientOptions): Authentication {
   // TODO: throw descriptive errors for invalid options
   if (typeof clientOptions.publicApiKey === "string") {
+    if (clientOptions.publicApiKey.startsWith("sk_"))
```
FlowFlorent, line 184: Wasn't sure wether to use `ifError` instead, let me know if it should be updated

FlowFlorent:
There are two ways to go about fixing this problem: 
1. Exclusion -> if `sk_` then throw
2. Inclusion -> `if `pk_` then do not throw

I've implemented 1., let me know if this works 🙏

packages/liveblocks-client/src/client.ts
nvie: No, this is fine I think. `ifError` will only error in development mode, but if you use an `sk_` key in production in a public key position, that will never work anyway, so better to catch it in the client, always, even (especially?) in production.

nvie: I'd probably catch both positive and negative cases here, aka structure it like:

```ts
const { publicApiKey, authEndpoint } = clientOptions;

if (publicApiKey !== undefined) {
  // A-ha! They're trying to use a public key! Let's check a few more things then…

  // First of all, ensure that they're not *also* setting the authEndpoint key!
  if (authEndpoint !== undefined) {
    // Throw friendly error explaining you cannot use both, and need to chose either of the two
  }

  if (typeof publicApiKey !== 'string') {
    // Throw friendly error that it should be a string, not whatever they used. Perhaps also hint that they should look for a key starting with `pk_` (and provide a link to the dashboard in there?)
  }

  if (startsWith('sk_')) {
    // A-ha! User provided the *secret* key! A common mistake! Give a friendly/helpful error message that they need to use the other one (and provide a link to the dashboard in there?)
  } else if (!publicApiKey.startsWith('pk_')) {
    // This is neither a public or a secret key, but another string. Tell them to look for their `pk_` key (and provide a link to the dashboard in there?)
  }

  // We made it! By now this is a (syntactically) valid PK, and we can continue
}
```

nvie (approved):
This is looking good, but I think we can  take the error messaging a step further and provide even more useful checks. Doing all these checks really [helps to not make our users think](https://docs.temporal.io/blog/error-message-design/) while setting up Liveblocks in their projects.

packages/liveblocks-client/src/client.ts
FlowFlorent: Great comment, indeed, might as well make it bulletproof, all these use cases will end up happening at some point

FlowFlorent: @nvie I've addressed your comments, let me know if this works 🙏

packages/liveblocks-client/src/client.node.test.ts
```diff
@@ -31,6 +31,55 @@ function createClientAndEnter(options: ClientOptions) {
...
+    ],
+    [
+      undefined,
+      null,
+      "authEndpoint must be a string or a function. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientAuthEndpoint",
+    ],
+    [
+      "sk_xxx",
+      undefined,
+      "Invalid key. You are using the secret key which is not supported. Please use the public key instead. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientPublicKey",
```
GuillaumeSalles, line 54: I would replace `key` by `publicApiKey` to give instant feedback on which option is used incorrectly.

```suggestion
      "Invalid publicApiKey. You are using the secret key which is not supported. Please use the public key instead. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientPublicKey",
```

GuillaumeSalles (approved):
Awesome job! Really eager to see if it lowers the setup friction!

packages/liveblocks-client/src/client.node.test.ts
FlowFlorent: Good improvement! Will adjust

PR #337, opened 2022-06-15
Upgrade examples to 0.17 patterns

GuillaumeSalles:


examples/react-whiteboard/src/App.js
```diff
@@ -45,7 +45,7 @@ function Canvas({ shapes }) {
 
   const insertRectangle = () => {
     batch(() => {
-      const shapeId = Date.now();
+      const shapeId = Date.now().toString();
```
GuillaumeSalles, line 48: On v6 protocol, we enforce that a parentKey should be a string, but our whiteboard example/tuto actually use a number. This one is actually quite bad... I'm to tired to think of a solution but it's something we need to figure out before releasing 0.17. cc @ofoucherot @nvie

nvie: Ah, good catch. Curious how you figured out what went wrong with this. I assume these messages would be silently dropped/ignored by the backend, causing weird behavior in the app? Was it immediately obvious to you what went wrong? Without TypeScript support, these may be hard to catch indeed!

The solution that would solve the issue at the core is to introduce a runtime type check in the place where we generate the `Op` instance, and possibly allow `LiveMap.set()` to take a number key as well, in which case we would simply convert it to a string before building the `Op` instance.

I can make that fix!

nvie: Small update here: I went looking to fix this in the client, but quickly realized this a much deeper rabbit hole, and warrants a separate quality pass of its own.

This is not the only case that we need to address. We can tackle this particular issue as a one-off fix now, sure, but the _class_ fix for this would be to make sure that _all_ public APIs always perform runtime type checking on their arguments. (We wouldn't have to do this for internal APIs; we have TypeScript for that.)

This would be very maintenance unfriendly and error prone to be enforcing manually, so I'd suggest to enforce this using some kind of build step or (custom) rollup plugin.

For now, though, the fix for this #339.

examples/nextjs-live-cursors-chat/liveblocks.config.ts
```diff
@@ -0,0 +1,32 @@
...
+
+type Presence = {
+  cursor: {
+    x: number;
+    y: number;
+  } | null;
+  message: string;
+};
+
+type Storage = {};
```
nvie, line 16: Let's add a comment here to illustrate that this is deliberately not using Storage APIs:

```suggestion
// Storage APIs are not used in this example app
type Storage = {};
```

examples/nextjs-live-cursors-scroll/liveblocks.config.ts
```diff
@@ -0,0 +1,24 @@
...
+type Cursor = {
+  x: number;
+  y: number;
+};
+
+type Presence = {
+  cursor: Cursor | null;
+};
+
+type Storage = {};
```
nvie, line 17:
```suggestion
// Storage APIs are not used in this example app
type Storage = {};
```

examples/nextjs-live-cursors/liveblocks.config.ts
```diff
@@ -0,0 +1,24 @@
...
+type Cursor = {
+  x: number;
+  y: number;
+};
+
+type Presence = {
+  cursor: Cursor | null;
+};
+
+type Storage = {};
```
nvie, line 17:
```suggestion
// Storage APIs are not used in this example app
type Storage = {};
```

examples/nextjs-live-selection/liveblocks.config.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const client = createClient({
+  publicApiKey: process.env.NEXT_PUBLIC_LIVEBLOCKS_PUBLIC_KEY!,
+});
+
+type Presence = {
+  selectedId: string | null;
+};
+
+type Storage = {};
```
nvie, line 12:
```suggestion
// Storage APIs are not used in this example app
type Storage = {};
```

nvie (approved):
I love all of these updates, so good to see all the type annotations/casting gone! 😍 
Thanks for going through all of these and getting them fixed! 🙏

examples/react-native-todo-list/liveblocks.config.js
```diff
@@ -0,0 +1,13 @@
+import { createClient } from "@liveblocks/client";
+import { createRoomContext } from "@liveblocks/react";
+import { decode } from "base-64";
+
+const PUBLIC_KEY = "pk_YOUR_PUBLIC_KEY";
+
+const client = createClient({
+  publicApiKey: PUBLIC_KEY,
+  atobPolyfill: decode,
```
nvie, line 9: @vinodsantharam As part of 0.17, I removed this API, and changed it to:

```suggestion
  polyfills: {
    atob: decode,
  },
```

PR #339, opened 2022-06-15
Address issue with parentKey potentially being sent in as a non-string

### LINKED ISSUE #338: Enforce `parentKey` to always be a string
See https://github.com/liveblocks/liveblocks/pull/337#discussion_r897511528 for context.

nvie:
Fixes https://github.com/liveblocks/liveblocks/pull/337#discussion_r897511528.

As mentioned in the https://github.com/liveblocks/liveblocks/pull/337#discussion_r897511528 thread already, this only fixes this at the surface of a much deeper rabbit hole, which warrants a separate quality pass of its own later.

This is not the only case that we need to address. The class fix for this would be to make sure that all public APIs always perform runtime type checking on their arguments. (We wouldn't have to do this for internal APIs; we have TypeScript for that.)

This would be very maintenance unfriendly and error prone to be enforcing manually, so I'd suggest to enforce this using some kind of build step or (custom) rollup plugin.

Made a ticket for this larger task, see #340.

### DIFF
--- a/packages/liveblocks-client/src/LiveMap.ts
+++ b/packages/liveblocks-client/src/LiveMap.ts
@@ -248,6 +248,8 @@ export class LiveMap<
    * @returns The element associated with the specified key, or undefined if the key can't be found in the LiveMap.
    */
   get(key: TKey): TValue | undefined {
+    key = typeof key === "string" ? key : (String(key) as TKey); // For non-TS users
+
     const value = this._map.get(key);
     if (value == undefined) {
       return undefined;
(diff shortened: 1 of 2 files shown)


ofoucherot (approved):

nvie:
@GuillaumeSalles Did you get the chance to go over the existing storage data and see if there is data that needs to be migrated? Are you OK with this fix, or should we approach this more generically in your opinion?

GuillaumeSalles (commented):
Waiting for more real world usage to see how we're tackling this issue

nvie:
This PR is now 3 months old. I think we can just close this one now, @GuillaumeSalles ?

nvie:
Closing this old PR now.

PR #344, opened 2022-06-15
feat(secretKey): handle error

FlowFlorent:
Ensures that only the secret key should be used in the node client

### DIFF
--- a/packages/liveblocks-node/src/index.ts
+++ b/packages/liveblocks-node/src/index.ts
@@ -33,6 +33,18 @@ export async function authorize(
   options: AuthorizeOptions
 ): Promise<AuthorizeResponse> {
   try {
+    const { room, userId, userInfo, secret } = options;
+
+    if (secret.startsWith("pk_")) {
+      throw new Error(
+        "Invalid key. You are using the public key which is not supported. Please use the secret key instead. For more information: https://liveblocks.io/docs/api-reference/liveblocks-node#authorize"
+      );
+    } else if (!secret.startsWith("sk_")) {
+      throw new Error(
+        "Invalid key. Please use the secret key instead. For more information: https://liveblocks.io/docs/api-reference/liveblocks-node#authorize"
+      );
+    }
+
     const result = await fetch(
       (options as AllAuthorizeOptions).liveblocksAuthorizeEndpoint ||
         "https://liveblocks.io/api/authorize",
@@ -43,9 +55,9 @@ export async function authorize(
           "Content-Type": "application/json",
         },
         body: JSON.stringify({
-          room: options.room,
-          userId: options.userId,
-          userInfo: options.userInfo,
+          room,
+          userId,
+          userInfo,
         }),
       }
     );
@@ -61,11 +73,11 @@ export async function authorize(
       status: 200,
       body: await result.text(),
     };
-  } catch (er) {
+  } catch (error) {
     return {
       status: 403,
       body: 'Call to "https://liveblocks.io/api/authorize" failed. See "error" for more information.',
-      error: er,
+      error,
     };
   }
 }
--- a/packages/liveblocks-node/src/index.ts
+++ b/packages/liveblocks-node/src/index.ts
@@ -29,22 +29,32 @@ type AuthorizeResponse = {
   error?: Error;
 };
 
+const unknownSecretKeyError = new Error(
+  'We expect a secret key ("sk_") here, but we found an unknown key instead. Hint: You can find your secret key at https://liveblocks.io/dashboard/apikeys. For more information: https://liveblocks.io/docs/api-reference/liveblocks-node#authorize'
+);
+
 export async function authorize(
   options: AuthorizeOptions
 ): Promise<AuthorizeResponse> {
+  const { room, userId, userInfo, secret } = options;
+
+  if (!(typeof secret === "string" && secret.length > 0)) {
+    return unauthorized(unknownSecretKeyError);
+  }
+
+  if (secret.startsWith("pk_")) {
+    return unauthorized(
+      new Error(
+        'We expect a secret key ("sk_") here, but we found a public key ("pk_") instead. Hint: You can find your secret key at https://liveblocks.io/dashboard/apikeys. For more information: https://liveblocks.io/docs/api-reference/liveblocks-node#authorize'
+      )
+    );
+  }
+
+  if (!secret.startsWith("sk_")) {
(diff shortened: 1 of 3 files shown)


packages/liveblocks-node/src/index.ts
```diff
@@ -33,6 +33,18 @@ export async function authorize(
   options: AuthorizeOptions
 ): Promise<AuthorizeResponse> {
   try {
+    const { room, userId, userInfo, secret } = options;
+
+    if (secret.startsWith("pk_")) {
```
FlowFlorent, line 38: wondering if these predicates should go above the `try...catch` block. 

If it's inside the try...catch, then the liveblocks-client user would know that something is faulty in the auth server from the 403 response error message.

If it's above, then there's the risk that the auth server emits a 500, which might not be ideal. 

What do you think would be the best DX?

FlowFlorent: Thinking about it, this could also be handled by the authorize endpoint response directly, and we could not have anything here

GuillaumeSalles: > wondering if these predicates should go above the try...catch block.
> If it's inside the try...catch, then the liveblocks-client user would know that something is faulty in the auth server from the 403 response error message.
> If it's above, then there's the risk that the auth server emits a 500, which might not be ideal.
> What do you think would be the best DX?

I think these validations should trow as soon as possible. We can't really know how they're wrapping `authorize` and I think it's expected to have a 500 in this case since it's code issue more than a authorization issue. What do you think @ofoucherot @nvie?

packages/liveblocks-node/src/index.ts
```diff
@@ -33,6 +33,18 @@ export async function authorize(
...
+      throw new Error(
+        "Invalid key. You are using the public key which is not supported. Please use the secret key instead. For more information: https://liveblocks.io/docs/api-reference/liveblocks-node#authorize"
```
nvie, line 40: I've been thinking about the language we use when we throw errors, and I think phrases like "Invalid key" or "You are […] which is not supported" may unnecessarily reinforce that the user made a mistake. To me, the golden standard of error messages is the Elm compiler, which is always very friendly and feels like a friend that helps you while setting things up or trying things out—and I think it's a great model to follow with all errors on Liveblocks.

In that style, we could consider erroring as follows:

_“We expect a secret key ("sk\_") here, but we found a public key ("pk\_") instead. Hint: You can find your secret key at https://liveblocks.io/dashboard/apikeys.”_

This doesn't put the blame on the user, but simply mentions what our expectations are, and hints at what the user could do next to fix this error.

I absolutely realize that our other existing error messages don't do this yet, but since we'd have to start somewhere, perhaps this is a good moment to start doing that here with these new error messages.

nvie (approved):
The changs look good to me. Thx for improving our error messages 👍

packages/liveblocks-node/src/index.ts
FlowFlorent: Since it's a code issue, wouldn't it be better to throw an error? A 4XX is a user error, a 5XX is a server error

Additionally a 403 has two effects:
- "leaks" how the backend is done by telling what's going on there
- makes it the `@liveblocks-client` user's responsibility to flag that it is indeed a server issue. 

With these reasons I would opt towards throwing outside the `try...catch` the errors. 

What do you think @nvie @GuillaumeSalles ?

FlowFlorent: And on the other hand this would be a valid 403 https://github.com/liveblocks/liveblocks/pull/346 as it's the client that failed to pass a valid room.

nvie: > Since it's a code issue, wouldn't it be better to throw an error?

Yeah, this is an interesting hybrid case. You can look at this as a code issue because it's just a function invocation, but you could also look at this as a local HTTP wrapper around the real Liveblocks auth endpoint… Hmm 🤔 

In either case, we should never throw a 5xx, because those are for _unexpected_ errors only — aka stuff that _we_ must fix if they ever happen. It's very much expected that some users will send in a missing or incorrect auth token, though.

> Additionally a 403 has two effects:
> * "leaks" how the backend is done by telling what's going on there

Agree with you on leaking implementation details in responses is not a great practice. I also don't think this is leaking any secret info – it's well understood/documented that this requires a secret key to authorize, and as such I'm looking at those error messages as an on-the-fly document snippet/reminder. To me, that's different from leaking that "a user with this email address isn't found", which actually provides a hacker with information.

> * makes it the `@liveblocks-client` user's responsibility to flag that it is indeed a server issue.

I think to handle this, we may also add this same layer of checks to the client itself. So that the client self-checks before even making this request.

---

Taking a step back, though, is this project the right place where we need to add extra helpful checks in the first place? We have the following possible enforcement points:

(1.) Client, before making the request (self-check)
(2.) and (3.) When receiving token in `@liveblocks/node` (aka what this PR adds)
(4.) The real check.

```
   ┌───────────────────────┐          ┌─────────────────────┐           ┌─────────────────────┐
   │                       │ ────────►│                     │ ────────► │                     │
   │  Client               │(1)    (2)│  Customer backend   │(3)     (4)│  Liveblocks server  │
   │                       │          │                     │           │                     │
   │  @liveblocks/client   │          │  @liveblocks/node   │           │  liveblocks.io      │
   │                       │          │                     │           │                     │
   └───────────────────────┘          └─────────────────────┘           └─────────────────────┘
```

Of course, (4) is not up for discussion, and (1) seems a good idea to do, to help clients/devs catch config errors as early as possible.

But the more I think about this, the more I'm inclined to think that the checks in `@liveblocks/node` don't add that much? Wouldn't it be more helpful if those would simply relay whatever they receive, and let the liveblocks.io backend decide what's ok or not?

**tl;dr** If we already have (1) and (4), do we really need to worry about (2)/(3) at all?

FlowFlorent: Yes, I agree that it's better to keep @liveblocks/node dumb & ensure that the validation happens on liveblocks.io. Another advantage is that people using the `PK` with the direct HTTP call will also be made aware of the error.

FlowFlorent: Thanks for the detailed and amazing feedback. 

As a result, closing this PR in favour of https://github.com/liveblocks/liveblocks.io/pull/265

FlowFlorent:
Closing in favour of https://github.com/liveblocks/liveblocks.io/pull/265

PR #345, opened 2022-06-15
Lift `TUserMeta` and `TEvent` up to the `createRoomContext()` level

### LINKED ISSUE #342: Lift `TUserMeta` and `TEvent` up to the `createRoomContext()` level
Talked about this on the call with @GuillaumeSalles. We'll also introduce two optional type params at the `createRoomContext()` level: `TUserMeta` and `TEvent`, where:

- Make `TStorage` optional
- `TUserMeta` is a special object of shape `{ id?: string, info?: Json }`, so it matches our backend and makes it possible to do type overriding, defaults to `{}`
- `TEvent` is just JSON, but defaulting to `never`

nvie:
This PR also lifts the `TUserMeta` and `TEvent` type arguments to the highest level, like we discussed. Fixes #342.

### DIFF
--- a/packages/liveblocks-client/e2e/utils.ts
+++ b/packages/liveblocks-client/e2e/utils.ts
@@ -13,11 +13,19 @@ import {
   patchImmutableObject,
 } from "../src/immutable";
 import type { LiveObject } from "../src/LiveObject";
-import type { JsonObject, LsonObject, ToJson } from "../src/types";
+import type {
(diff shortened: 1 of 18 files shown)


nvie:
This was reviewed and approved by @GuillaumeSalles through a video call + Slack, outside of GitHub. Will merge this, and get the examples updated in #337 .

Issue #331, opened 2022-06-13
Improve error message if custom authentication endpoint return is not structured properly


GuillaumeSalles:
https://discord.com/channels/913109211746009108/913158542565994606/984868180365279312

PR #346, opened 2022-06-15
feat(authorize): provide meaningful error when room is missing

FlowFlorent:
Solves part of https://github.com/liveblocks/liveblocks/issues/331

Next step: throw a meaningful error when result of auth in `createClient` is missing `{token: "xxx"}`


FlowFlorent:
Additionally, a check could be added in `@liveblocks-client` for a missing `actor` in the token

nvie (approved):
Nice! Thx for improving our error messages! 👌

nvie (approved):

PR #347, opened 2022-06-15
feat(token): ensure auth provides correct response & token has actor

FlowFlorent:
Related PR: https://github.com/liveblocks/liveblocks.io/pull/261

- Ensures auth provides correct response format by including the token
- Improves the error message when validating that the token is for a room
- Adds related tests

### DIFF
--- a/packages/liveblocks-client/src/AuthToken.test.ts
+++ b/packages/liveblocks-client/src/AuthToken.test.ts
@@ -63,7 +63,7 @@ describe("parseRoomAuthToken", () => {
     } catch (error) {
       expect(error).toEqual(
         new Error(
-          "Authentication error: we expected a room token but did not get one. Hint: if you are using a callback, ensure the room is passed when creating the token. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientWithCallback"
+          "Authentication error: we expected a room token but did not get one. Hint: if you are using a callback, ensure the room is passed when creating the token. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientCallback"
         )
       );
     }
--- a/packages/liveblocks-client/src/AuthToken.test.ts
+++ b/packages/liveblocks-client/src/AuthToken.test.ts
@@ -1,5 +1,5 @@
 import type { JwtMetadata } from "./AuthToken";
-import { isTokenExpired } from "./AuthToken";
+import { isTokenExpired, parseRoomAuthToken } from "./AuthToken";
 
 describe("isTokenExpired", () => {
   const MINUTES = 60 * 1000;
@@ -31,5 +31,41 @@ describe("isTokenExpired", () => {
     };
 
     expect(isTokenExpired(futureToken)).toBe(true);
+  });
+});
+
+describe("parseRoomAuthToken", () => {
+  const roomToken =
+    "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2MTY3MjM2NjcsImV4cCI6MTYxNjcyNzI2Nywicm9vbUlkIjoiazV3bWgwRjlVTGxyek1nWnRTMlpfIiwiYXBwSWQiOiI2MDVhNGZkMzFhMzZkNWVhN2EyZTA5MTQiLCJhY3RvciI6MCwic2NvcGVzIjpbIndlYnNvY2tldDpwcmVzZW5jZSIsIndlYnNvY2tldDpzdG9yYWdlIiwicm9vbTpyZWFkIiwicm9vbTp3cml0ZSJdfQ.IQFyw54-b4F6P0MTSzmBVwdZi2pwPaxZwzgkE2l0Mi4";
+  const apiToken =
+    "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2MTY3MjM2NjcsImV4cCI6MTYxNjcyNzI2NywiYXBwSWQiOiI2MDVhNGZkMzFhMzZkNWVhN2EyZTA5MTQiLCJzY29wZXMiOlsicm9vbTpyZWFkIiwicm9vbTp3cml0ZSJdfQ.7Wplt6YV_YbpPAcAFyC8pX8tk5BGNy53GdoH1_u8sjo";
+
+  test("should parse a valid token", () => {
+    const parsedToken = parseRoomAuthToken(roomToken);
+    expect(parsedToken).toEqual({
+      iat: 1616723667,
+      exp: 1616727267,
+      roomId: "k5wmh0F9ULlrzMgZtS2Z_",
+      appId: "605a4fd31a36d5ea7a2e0914",
+      actor: 0,
+      scopes: [
+        "websocket:presence",
+        "websocket:storage",
+        "room:read",
+        "room:write",
+      ],
+    });
+  });
+
+  test("should throw if token is not a room token", () => {
+    try {
+      parseRoomAuthToken(apiToken);
+    } catch (error) {
+      expect(error).toEqual(
+        new Error(
+          "Authentication error: we expected a room token but did not get one. Hint: if you are using a callback, ensure the room is passed when creating the token. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientWithCallback"
+        )
+      );
+    }
   });
 });
--- a/packages/liveblocks-client/src/AuthToken.ts
+++ b/packages/liveblocks-client/src/AuthToken.ts
@@ -136,10 +136,13 @@ export function parseRoomAuthToken(
     } = data;
     return token;
   } else {
-    throw new Error("Authentication error: invalid room auth token");
+    throw new Error(
+      "Authentication error: we expected a room token. Hint: if you are using a callback, ensure the room is passed when creating the token. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClient"
+    );
   }
 }
 
+// TODO: unused in all packages & not accessible from @liveblocks/client users, remove?
 export function parseAuthToken(token: string): AuthToken & JwtMetadata {
   const data = parseJwtToken(token);
   if (data && isAuthToken(data)) {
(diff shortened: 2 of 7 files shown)


packages/liveblocks-client/src/AuthToken.ts
```diff
@@ -125,6 +126,7 @@ function parseJwtToken(token: string): JwtMetadata {
 export function parseRoomAuthToken(
   tokenString: string
 ): RoomAuthToken & JwtMetadata {
+  // TODO: check if token is indeed a string here
```
FlowFlorent, line 129:
```suggestion
```

packages/liveblocks-client/src/room.ts
```diff
@@ -320,21 +320,26 @@ export function makeStateMachine<
...
-            }
-            const parsedToken = parseRoomAuthToken(token);
-            const socket = createWebSocket(token);
-            authenticationSuccess(parsedToken, socket);
-            state.token = token;
-          })
-          .catch((er: unknown) =>
-            authenticationFailure(
-              er instanceof Error ? er : new Error(String(er))
+        // token is expired or invalid
```
GuillaumeSalles, line 323: or does not exist

packages/liveblocks-client/src/room.ts
```diff
@@ -792,6 +797,7 @@ export function makeStateMachine<
       return null;
     }
 
+    // auth prepared here, a Promise, but not resolved yet.
```
GuillaumeSalles, line 800: Technically, it returns a function that returns a Promise. We "create" a promise only when `auth` is called.

packages/liveblocks-client/src/room.ts
```diff
@@ -320,21 +320,26 @@ export function makeStateMachine<
...
-            state.token = token;
-          })
-          .catch((er: unknown) =>
-            authenticationFailure(
-              er instanceof Error ? er : new Error(String(er))
+        // token is expired or invalid
+        // therefore we authenticate
+        return (
+          auth(context.roomId)
+            // TODO: provide a meaningful error message when token is undefined here
```
GuillaumeSalles, line 327: It might be easier to throw an error from here https://github.com/liveblocks/liveblocks/blob/eb286928cbdf86472169d30c64162e3ee1891b53/packages/liveblocks-client/src/room.ts#L1752
by wrapping `authentication.callback`. That way the token will never be `undefined` at line 329.

packages/liveblocks-client/src/AuthToken.ts
```diff
@@ -93,6 +93,7 @@ export function isRoomAuthToken(data: JsonObject): data is RoomAuthToken {
   return (
     typeof data.appId === "string" &&
     typeof data.roomId === "string" &&
+    // TODO: throw specific error if actor is not a number
```
nvie, line 96: FYI, this function is a [TypeScript predicate function](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) and as such it should never throw. Other projects (e.g. liveblocks-cloudflare) also rely on this implementation. That said, I agree that throwing targeted error messages is a Good Idea™, just saying that the throwing cannot happen inside a predicate function.

packages/liveblocks-client/src/room.ts
FlowFlorent: done!

packages/liveblocks-client/src/client.ts
```diff
@@ -103,6 +103,8 @@ export function createClient(options: ClientOptions): Client {
           // TODO Patch this using public but marked internal fields?
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
           (clientOptions as any)?.liveblocksServer || "wss://liveblocks.net/v6",
+        // authentication is prepped here, but it's not used until the room is created
+        // passed as the context for `createRoom`
```
FlowFlorent, line 107:
```suggestion
```

packages/liveblocks-client/src/room.test.ts
```diff
@@ -60,7 +60,7 @@ function setupStateMachine<
   return { machine, state, effects };
 }
 
-describe("room / auth", () => {
+describe.only("room / auth", () => {
```
FlowFlorent, line 63:
```suggestion
describe("room / auth", () => {
```

packages/liveblocks-client/src/room.ts
```diff
@@ -320,21 +320,26 @@ export function makeStateMachine<
...
-            state.token = token;
-          })
-          .catch((er: unknown) =>
-            authenticationFailure(
-              er instanceof Error ? er : new Error(String(er))
+        // token is expired, invalid, or does not exist
+        // therefore we authenticate
+        return (
+          auth(context.roomId)
+            // TODO: provide a meaningful error message when token is undefined here
```
FlowFlorent, line 327:
```suggestion
```

packages/liveblocks-client/src/room.ts
```diff
@@ -320,21 +320,26 @@ export function makeStateMachine<
...
-            const parsedToken = parseRoomAuthToken(token);
-            const socket = createWebSocket(token);
-            authenticationSuccess(parsedToken, socket);
-            state.token = token;
-          })
-          .catch((er: unknown) =>
-            authenticationFailure(
-              er instanceof Error ? er : new Error(String(er))
+        // token is expired, invalid, or does not exist
+        // therefore we authenticate
```
FlowFlorent, line 324:
```suggestion
```

packages/liveblocks-client/src/room.ts
FlowFlorent, line 800:
```suggestion
```

packages/liveblocks-client/src/room.ts
```diff
@@ -802,6 +808,7 @@ export function makeStateMachine<
     );
 
     updateConnection({ state: "authenticating" });
+    // auth passed in authenticate effect
```
FlowFlorent, line 811:
```suggestion
```

packages/liveblocks-client/src/AuthToken.ts
FlowFlorent: I've actually not thrown an error but rather made the error message easier to understand https://github.com/liveblocks/liveblocks/pull/347/files#diff-2f8a3dfc5fdb3180961ecbac8995ec77c35765b7bdbd9957e4c9af62f620bb01R140

From what I understand, if we reach this point, the only way for it to fail is that the token is from a custom callback which was misused (room was not provided)

packages/liveblocks-client/src/AuthToken.ts
```diff
@@ -135,10 +136,13 @@ export function parseRoomAuthToken(
...
     return token;
   } else {
-    throw new Error("Authentication error: invalid room auth token");
+    throw new Error(
+      "Authentication error: we expected a room token. Hint: if you are using a callback, ensure the room is passed when creating the token. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClient"
+    );
   }
 }
 
+// TODO: unused in all packages & not accessible from @liveblocks/client users, remove?
```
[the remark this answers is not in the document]
nvie: Ha! Great catch, I think you're right. It's good to also always check usage of these APIs in https://github.com/liveblocks/liveblocks-cloudflare and https://github.com/liveblocks/liveblocks.io, because those might also import these APIs using `import { parseAuthToken } from "@liveblocks/client/internal"`, but I just checked them and they're also not using this anymore either!

I think what happened is that at the end of some refactorings, we ended up with only usages of `parseRoomAuthToken`, and never this one anymore, and I forgot to clean this one up. I think it would indeed be safe to remove this — nice! ✨🧽✨

packages/liveblocks-client/src/AuthToken.ts
```diff
@@ -135,10 +136,13 @@ export function parseRoomAuthToken(
     } = data;
     return token;
   } else {
-    throw new Error("Authentication error: invalid room auth token");
+    throw new Error(
+      "Authentication error: we expected a room token but did not get one. Hint: if you are using a callback, ensure the room is passed when creating the token. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientWithCallback"
```
nvie, line 140: I noticed the hash was incorrect here?

```suggestion
      "Authentication error: we expected a room token but did not get one. Hint: if you are using a callback, ensure the room is passed when creating the token. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientCallback"
```

packages/liveblocks-client/src/AuthToken.ts
nvie: Nice 👍 ! You can remove this TODO comment here.

nvie (approved):
Nice work, @FlowFlorent ! 🙌

packages/liveblocks-client/src/AuthToken.ts
nvie: Also perhaps say "room ID" in the error message, not just "room"? This might also be something to mention in those linked docs, because it's also not clear from the docs what exactly that `room` param is that the callback receives: a string? or an object?

FlowFlorent: Indeed, thanks for the 🦅 eyes! 

Will adjust both, good catches 😄

packages/liveblocks-client/src/AuthToken.ts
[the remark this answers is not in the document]
FlowFlorent: Oh you're right, I'll do a sanity check in those codebases as well from now on. Will clean it up then :)

packages/liveblocks-client/src/AuthToken.test.ts
```diff
@@ -33,3 +33,39 @@ describe("isTokenExpired", () => {
...
+    });
+  });
+
+  test("should throw if token is not a room token", () => {
+    try {
+      parseRoomAuthToken(apiToken);
+    } catch (error) {
+      expect(error).toEqual(
+        new Error(
+          "Authentication error: we expected a room token but did not get one. Hint: if you are using a callback, ensure the room is passed when creating the token. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientWithCallback"
```
FlowFlorent, line 66:
```suggestion
          "Authentication error: we expected a room token but did not get one. Hint: if you are using a callback, ensure the room is passed when creating the token. For more information: https://liveblocks.io/docs/api-reference/liveblocks-client#createClientCallback"
```

PR #348, opened 2022-06-16
Mark some APIs as `@internal` that were missing it

nvie:
This adds some missing `@internals`. The effects on the outer observable API are pretty significant.

Compare:
- The API diff we would (accidentally) ship if we [were to release `main` as 0.17 today](https://gist.github.com/nvie/0ddda98cee4e0d2242b2648c462b5b31)
- The API diff [with this PR applied](https://gist.github.com/nvie/c4d8ca61648bff584ddb2701d19acf36)


FlowFlorent (approved):
lgtm!

PR #350, opened 2022-06-16
Fix: remove a leftover from refactoring

nvie:
Removes an old type param position that I forgot to remove after refactoring. Found this by inspecting [the outer API diff output](https://gist.github.com/nvie/c4d8ca61648bff584ddb2701d19acf36#file-v0-16-17-vs-v0-17-diff-L547-L555) :)


FlowFlorent (approved):
👌 🧽 ✨

PR #354, opened 2022-06-20
Cleanup TypeScript definitions for `subscribe` overloads

### LINKED ISSUE #343: Fix bugs in `subscribe` overloaded type definitions
There is one particular `any` in there, see https://github.com/liveblocks/liveblocks/blob/acfdba97599a2e87a4d30f39088e8b7f4a2e9ae4/packages/liveblocks-client/src/room.ts#L735

This type is currently hiding a few bugs in the type definition of `subscribe`, mostly related to exactly what payload is sent into each callback.

nvie:
Fixes #343 , correcting the type bug that previously were hidden by the `any` in one of the overloads, and restructures the internals to be a bit simpler to follow. Does not contain any functional changes.

### DIFF
--- a/packages/liveblocks-client/src/room.ts
+++ b/packages/liveblocks-client/src/room.ts
@@ -3,8 +3,6 @@ import { OpSource } from "./AbstractCrdt";
 import { nn } from "./assert";
 import type { RoomAuthToken } from "./AuthToken";
 import { isTokenExpired, parseRoomAuthToken } from "./AuthToken";
-import type { LiveList } from "./LiveList";
-import type { LiveMap } from "./LiveMap";
 import { LiveObject } from "./LiveObject";
 import type {
   Authentication,
@@ -24,7 +22,6 @@ import type {
   JsonObject,
   LiveNode,
   LiveStructure,
-  Lson,
   LsonObject,
   MyPresenceCallback,
   NodeMap,
@@ -35,7 +32,8 @@ import type {
   ParentToChildNodeMap,
   Polyfills,
   Room,
-  RoomEventCallbackMap,
+  RoomEventCallback,
+  RoomEventCallbackFor,
   RoomEventName,
   RoomInitializers,
   RoomStateServerMsg,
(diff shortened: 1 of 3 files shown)


packages/liveblocks-client/src/types/index.ts
```diff
@@ -505,113 +519,41 @@ export type Room<
...
+     * const liveMap = new LiveMap();  // Could also be LiveList or LiveObject
+     * const unsubscribe = room.subscribe(liveMap, (updates) => { }, { isDeep: true });
      * unsubscribe();
      */
-    <TItem extends Lson>(
-      liveList: LiveList<TItem>,
-      callback: (updates: LiveListUpdates<TItem>[]) => void,
+    <L extends LiveStructure>(
+      liveStructure: L,
+      callback: StorageCallback,
```
nvie, line 556: There's a fix here. When using `room.subscribe()` with `{ isDeep: true }` on, say, a LiveList, then `updates` will not be a `LiveListUpdate` array per se. It can also contain other updates, because `isDeep` will include all nested updates. This bug was pointed out by TypeScript correctly when I removed the `any`.

packages/liveblocks-client/src/room.ts
```diff
@@ -730,69 +707,54 @@ export function makeStateMachine<
...
+
+    if (isLiveNode(first)) {
+      const node = first;
+      if (options?.isDeep) {
+        const storageCallback = second as StorageCallback;
+        return subscribeToLiveStructureDeeply(node, storageCallback);
+      } else {
+        const nodeCallback = second as (node: L) => void;
+        return subscribeToLiveStructureShallowly(node, nodeCallback);
+      }
```
nvie, line 740: By splitting up this helper into two versions, they're both much easier to type and implement. See their implementation to see how much simpler this code is to read and reason about now :)

packages/liveblocks-client/src/types/index.ts
nvie: And, for the record, the other bug that TypeScript pointed out here was fixed in https://github.com/liveblocks/liveblocks/pull/352/files#diff-c687f5217f04bdc741f7e57853e0c381ba3ffded12e1763d36dfb783b9ea642aR105 already.

packages/liveblocks-client/src/room.ts
GuillaumeSalles: So much cleaner! 😮

packages/liveblocks-client/src/types/index.ts
GuillaumeSalles: Can't believe that we missed those 2 typings issues...

GuillaumeSalles (approved):
Looks great!

PR #355, opened 2022-06-20
feat(liveblocks/node): add to automatic releases

FlowFlorent:


packages/liveblocks-node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@liveblocks/node",
-  "version": "0.3.0",
+  "version": "0.17.0-test1",
```
FlowFlorent, line 3: current version for all packages on `main`

nvie (approved):
Nice! 👍

I think this simplifies the package versioning/distribution a lot, despite the fact that `@liveblocks/node` will likely have many exactly-the-same versions. I don't think that's a problem.

packages/liveblocks-node/package.json
```diff
@@ -1,6 +1,6 @@
...
   "description": "A server-side utility that lets you set up a Liveblocks authentication endpoint.",
   "main": "./lib/cjs/index.js",
   "module": "./lib/esm/index.js",
```
nvie, line 6: Oh, btw. If we're going to do this, it may be nice to invest a bit more in the build process and use the same build process we use for the other packages, so it's all unified and there's no different setup for this package anymore. Not sure if there is a reason this package uses different build steps. Perhaps it's out of necessity but it might as well be legacy? If the latter, let's keep it on par with the rest while we're at it.

To clarify, what I mean is using Rollup to produce the build (instead of TypeScript), and output these in the same "flattened" package output (no different ESM/CJS directories)—just like all the other packages.

I think the `rollup.config.js` files from the other packages are generic enough to be copy/pastable in here.

FlowFlorent: Thanks for the tips! I've adjusted the PR to reflect that, `npm run build` works succesfully with `rollup` now. 
I re-used the config from `liveblocks-client` to achieve this and slightly modified it to fit the needs for `liveblocks-node`.

packages/liveblocks-node/rollup.config.js
```diff
@@ -0,0 +1,217 @@
...
+    declaration: true,
+    outDir,
+    tsconfig: "./tsconfig.build.json",
+    noEmitOnError: true, // Let rollup build fail if there are TypeScript errors
+  });
+}
+
+function getBabelOptions(extensions, targets) {
+  return {
+    ...babelConfig,
```
FlowFlorent, line 52: a simple object spread unlike for liveblocks-client, where the config is a function

nvie (approved):
Re-approval! I reviewed your latest changes this morning already, but forgot to re-approve explicitly. This looks beautiful. Thanks for cleaning it up!

PR #364, opened 2022-06-24
Upkeep: Apply guidelines to React Native example

marcbouchenoire:
I was looking at how we could integrate this example within the upcoming detail pages but I haven't found a way that makes sense yet.

While I was there, I ran Prettier and made a few tweaks to bring this example closer to others, following the guidelines we have in Notion.


vinodsantharam:
Sorry @marcbouchenoire I missed your request. Just wanted to say that I migrated this example to Typescript. So I have the feeling that is outdated...

PR #374, opened 2022-06-28
Also upgrade `@liveblocks/node` in all examples from now on

nvie:
Previously, we didn't version `@liveblocks/node` alongside the other packages, but since 0.17, we do. This reflects that in the helper `upgrade-examples.sh` script, and updates all examples we have that rely on `@liveblocks/node`.

### DIFF
--- a/scripts/upgrade-examples.sh
+++ b/scripts/upgrade-examples.sh
@@ -44,8 +44,7 @@ list_all_projects () {
 
 list_liveblocks_deps () {
     jq -r '(.dependencies // {})|keys[]' package.json \
-        | grep -Ee '^@liveblocks' \
-        | grep -vEe '@liveblocks/node'
+        | grep -Ee '^@liveblocks'
 }
 
 list_install_args () {
--- a/e2e/next-sandbox/package-lock.json
+++ b/e2e/next-sandbox/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@liveblocks/client": "^0.17.1",
-        "@liveblocks/node": "0.3.0",
+        "@liveblocks/node": "^0.17.1",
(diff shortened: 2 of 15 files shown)


scripts/upgrade-examples.sh
```diff
@@ -44,8 +44,7 @@ list_all_projects () {
 
 list_liveblocks_deps () {
     jq -r '(.dependencies // {})|keys[]' package.json \
-        | grep -Ee '^@liveblocks' \
-        | grep -vEe '@liveblocks/node'
```
nvie, line 48: No longer skipping our `node` package with a `grep -v` here 😄

nvie:
@CTNicholas @marcbouchenoire Any idea what may be up with these Sveltekit builds on Vercel? AFAIK, nothing significant changed in between 0.3 and 0.17 for `@liveblocks/node`. Not something related to node engines at least I think?

FlowFlorent (approved):
nice! I had definitely missed this part in my PR, thanks for doing it 😄

CTNicholas:
> @CTNicholas @marcbouchenoire Any idea what may be up with these Sveltekit builds on Vercel? AFAIK, nothing significant changed in between 0.3 and 0.17 for `@liveblocks/node`. Not something related to node engines at least I think?

Just had a quick look and recognised it immediately, SvelteKit requires Node 16, but we had Node 14 set. There's another error now though, not sure what it is. Look into it more soon!

nvie:
I think this was a false alarm—or at least the error message wasn't pointing at the problem directly. I think this was caused by the "empty" NPM package bug that [was reported on Discord](https://discord.com/channels/913109211746009108/913109211746009112/991402332015308951) yesterday, and which was fixed by 4546cf9bbe325ff988bf8922ef486667b9edaabd. After upgrading to 0.17.1, these deployment errors for Vercel also went away.

PR #365, opened 2022-06-24
Advanced avatars and cursors examples

CTNicholas:
## Info
I've added two advanced Next.js examples to the repo, one for _live cursors_ and one for _live avatars_. These two examples use `framer-motion` to create smoother animations.

### `nextjs-live-avatars-advanced`
Contains a spring-animated avatar stack, and also a reusable `Avatar` component, allowing for multiple types of avatars.

### `nextjs-live-cursors-advanced`
Contains spring-animated live cursors, and also a reusable `LiveCursors` component that allows you to pass it a cursor panel (React `ref` containing an element). From there it will accurately calculate cursor positions from the element's scroll position, the position of the element on the page etc.

_To be released after our new examples page is ready_

PR #379, opened 2022-07-01
Upgrade examples to use DRYed up `createRoomContext()` re-exports

marcbouchenoire:


marcbouchenoire:
@CTNicholas Let's not forget to do the same in https://github.com/liveblocks/liveblocks/pull/365 before merging.

nvie (approved):

PR #380, opened 2022-07-01
Suspense support in Storage APIs

nvie:
This PR is a minimal proof of concept to add Suspense support to our Storage APIs:

- `useStorage()`
- `useMap()`
- `useList()`
- `useObject()`

Each of these now have an overload on this branch that will allow you to pass in a `useMap('mykey', { suspense: true })` config option to use the function overload that will never return `null`. Instead, these hooks _suspend_ instead of returning `null`.

A quick refresher on how Suspense works with custom hooks. Such hooks will throw a promise (instead of an error) up the call stack. Then the `<Suspense>` boundary component will catch that thrown promise and render the fallback instead. Then, much later, once the thrown promise resolves, Suspense will rerender the component tree, by which time it should be able to resolve synchronously without suspending.

Ultimately, we add this in preparation for better/simpler Storage hooks. I'm not proposing per se to include these APIs as such in the `useMap` & friends hooks. Perhaps we'll decide that only a new `useSelector()` API will have support for Suspense. All the existing `useMap` & friends overloads are already pretty overwhelming to maintain. I just added them in here for now, because these are the ones we have today.

### DIFF
--- a/examples/nextjs-whiteboard-advanced/src/index.tsx
+++ b/examples/nextjs-whiteboard-advanced/src/index.tsx
@@ -94,7 +94,19 @@ export default function Room() {
       }}
     >
       <div className={styles.container}>
-        <ClientSideSuspense fallback={<h1>hahaha</h1>}>
+        {
+          //
+          // XXX Not sure what is the best way to make Suspense and SSR best
+          // friends. The new NextJS and React 18 might have a better way to
+          // offer this.
+          //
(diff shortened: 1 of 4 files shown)


nvie:
Closing this one, will re-open this PR later to target for merging into 0.18. It's _way_ easier to add this against 0.18, after we've cleaned up all the deprecated APIs.

Issue #286, opened 2022-06-01
Properly type out the `patchLiveObjectKey` helper family


nvie:
These helpers are "somewhat" typed out, but there still is a lot of manual type casting going on in there, which is a smell. Since this is such an important core feature of the client, and relied on by the Zustand and Redux packages as well, this deserves a 100% type safety.

Related to #279 and #280.

Issue #233, opened 2022-05-12
Move tests to separate folder


nvie:
Rather than having tests live "among" source files, like:

<img src="https://user-images.githubusercontent.com/83844/168039962-f44b1a46-e383-4201-b377-f19e6b161184.png" width="500" />

I'd propose to move these to the following structure:

<img src="https://user-images.githubusercontent.com/83844/168040724-f39adc81-d866-4c27-9cd1-582f840292aa.png" width="500" />

(Here, the helper file `test/utils.ts` is moved to `src/__tests__/_helpers.ts`.)

Moving unit tests to a separate folder has a few benefits:

1. This is still a standard Jest setup: it will still find and discover all the tests.
2. It's easier to look at source files or tests separately.
3. It's easy to glob all source files, i.e. `src/*.ts` previously included tests, and it's hard to write a pattern that excludes tests.
4. Ignore patterns are now easier.

When we do, let's use this same organization for _all_ packages.

Issue #223, opened 2022-05-11
Refactor away `AbstractCrdt`


nvie:
Get rid of the abstract base class and stop using inheritance in favor of using a tagged union. This will simplify the code by making it more readable for newcomers, and help with further TypeScript improvements in the next release.

Targets of refactorings:
- [x] Define `LiveStructure` interface, which is roughly equal to `AbstractCrdt` class (but has no runtime component to it)
- [ ] Inline all `AbstractCrdt` base methods in each implementation (this duplicates a little bit of code in exchange for improved readability)

PR #384, opened 2022-07-04
Release 0.18

nvie:
This PR is the gathering place for all future breaking changes that will need to land in our next major version: ✨ **0.18** ✨. This is a long-running PR that can be used as the merge target for other PRs that introduce breaking changes.

> **Note** This PR will be continually rebased to keep up to date with the latest `main`.

### Checklist (after release)
- [ ] [Update whiteboard tutorial](https://github.com/liveblocks/liveblocks.io/pull/395)

### Checklist (todo before release)
- [ ] Tidy up the changelog
- [x] [Documentation, tutorials, examples, guides](https://github.com/liveblocks/liveblocks.io/issues/375)

### Shipping in final release
- [x] #492 

### Shipped in 0.18.0-beta4
- [x] Make final call on the others hook family
  - [x] Rename `useOthersWithData` to its final name: `useOthersMapped`
  - [x] Change its return type to tuples
  - [x] Rename `useConnectionIds` to `useOthersConnectionIds`?
- [x] Rename `{ root }` to `{ storage }` in mutation context (to distinguish it clearly from the "immutable" `root` in the useStorage selector)
- [x] Put in final documentation URLs
- [x] #486
- [x] #482 

### Shipped in 0.18.0-beta3
- [x] #476 (see [feedback](https://liveblocks.slack.com/archives/C01TDLHJ9R6/p1662733879465849))
- [x] #478 (see [feedback](https://github.com/liveblocks/liveblocks/pull/384#discussion_r967249977))
- [x] #479 (see [feedback](https://github.com/liveblocks/liveblocks/pull/384#discussion_r967264141))
- [x] #473
- [x] #474
- [x] #475
- [x] ~Have a special Suspense version of `RoomProvider`? (see [feedback](https://liveblocks.slack.com/archives/C01TDLHJ9R6/p1662839631365749))~

### Shipped in 0.18.0-beta2
- [x] #471
- [x] #472

### Shipped in 0.18.0-beta1
- [x] #467
  - [x] ~Implement `useOtherIds()` hook~ Note: this is split and renamed to `useConnectionIds()` and `useOthersWithData()` in beta3
  - [x] Implement `useOther()` singular hook
- [x] #465
- [x] #464
- [x] #466

### Shipped in 0.18.0-beta0
- [x] #447
- [x] #451
- [x] #452 
- [x] #460 
- [x] #453
- [x] Restore `useStorageRoot()` in 0.18 for backward compat reasons
- [x] #401
- [x] #404
  - [x] ~Clean up `patchImmutable*` helpers — no longer needed?~ Still needed to support Redux/Zustand implementations for now. Maybe later.

### Internals / cleanups
- [x] In `useMutation`, throw instead of showing console warning
- [x] Put back backward compatible `useStorageRoot()` hook
- [x] #443
- [x] #449 
- [x] #450
- [x] #233 
- [x] Remove deprecated APIs
  - [x] #385 
  - [x] #389 
  - [x] #390
  - [x] #410

### Pushed to later (in 0.18.x)
- [ ] #454 
- [ ] #223 
- [ ] #286 
- [ ] #462

### Pushed to later (in 0.19.x)
- [x] ~Drop support for zustand v3 in favor of v4~


packages/liveblocks-react/src/factory.tsx
```diff
@@ -5,30 +5,54 @@ import type {
...
+
+const EMPTY_OTHERS =
+  // NOTE: asArrayWithLegacyMethods() wrapping should no longer be necessary in 0.19
+  asArrayWithLegacyMethods([]);
+
+function getEmptyOthers() {
+  return EMPTY_OTHERS;
 }
 
+export type MutationContext<
```
GuillaumeSalles, line 45: Would it make sense to add "getMyPresence" or the current user in this context?

In the drawing board example, we often have to pass the selection as a dependency that comes from the presence.

```ts
const selection = useSelf(user => user.presence.selection);
return useMutation({ root } => { /*...*/ }, [selection]);
```

This invalidate a callback reference even when not strictly necessary. Something like this 👇 could be a slight improvement.

```ts
return useMutation({ root, self } => { /*...*/ }, []);
```

packages/liveblocks-react/src/factory.tsx
```diff
@@ -433,54 +548,159 @@ export function createRoomContext<
...
+      selector ?? (identity as (others: Others<TPresence, TUserMeta>) => T),
+      isEqual
     );
   }
 
-  function useOthers(): Others<TPresence, TUserMeta> {
-    const room = useRoom();
-    const rerender = useRerender();
+  function useOtherIds(): readonly number[]; // TODO: Change to ConnectionID for clarity?
+  function useOtherIds<T>(
```
GuillaumeSalles, line 593: `useOtherIds` relies on `User.connectionId` which is confusing because we have an 'id' property on `User`. Not sure if simplifying the name is worth the potential confusion. 🤔

GuillaumeSalles (commented):
I love where this is going! So excited to release this to our users, such a huge milestone 🚀 

Thank you so much for your amazing work <3

packages/liveblocks-react/src/factory.tsx
nvie: Agreed, I’ve missed this in a couple places too, and wished it was there on the context for convenience! If we add this one in though, should we also include all other contexts? I’d be cheap to add those in I think? For example, [here](https://github.com/liveblocks/liveblocks/pull/384/files#diff-a154e780519dd9c827a011031e35442f7be80f0ed2f78ef1a83ab84c30728826R357) I needed access to the immutable storage value. I can aLeo imagine situations where you want to mutate something based on the state of `others`.

Would it make sense to include all of the following?

- `root`
- `setMyPresence`
- `me`
- `others`
- `immutableRoot` (equiv to `root.toImmutable()` so not strictly needed)

packages/liveblocks-react/src/factory.tsx
nvie: I love this hook, but still hate the name. Also dislike `useOthersConnectionIds`. Looking for a name that stresses the iterating/mapping/subselection nature—while pairing it with the connection IDs.

Any of these coming closer to a good name?
- useOtherForEach
- useEachOther (funny English ruins this one, unfortunately)
- useOthersMapped
- useMapOverOthers
- useSelectOthers
- useOthersData
- usePickFromOthers

nvie: Naming proposal:

- We ditch the `useOtherIds()` hook that's now in 0.18.0-beta2.
- We split its two overloads into two new hooks:
  - `useConnectionIds(): number[]`
  - `useOthersMapped(itemSelector, isEqual): { connectionId: number; data: T }[]`

ofoucherot: I like the split in two hooks. Personally, I'm not a fan of the name `userOthersMapped`. But I don't have good ideas for a better name. The best name for me is `useOthersData`.

To give more ideas:
- useOthersAsIterable
- useSelectorOthers

nvie: usePluckOthers

nvie: How about `useOthers2` and in the doc string issue a naming contest? 😅

ofoucherot: You have my vote for `useOthersData` !

GuillaumeSalles: Not easy to find the right name for this hook 😬 

My 2 cents on the proposed solutions:
- Not showing the word "connectionId" in the name forces developers to look at the return type or the documentation to understand what the hook is doing. I think this is not great for discoverability reasons, especially because we re-export them. I understand that putting such a long word in a name might not be appealing, but I think explicitness is better in this case.
- IMO, `selector` and `mapped` words are not explicit enough because `useOthers` is also taking a selector.

Taking these points into consideration, I would go with `useOthersWithConnectionId`. A tad too long but I think it's worth it to avoid the confusion with `useOthers(selector)`.

I also considered using a special selector like this:

```ts
useOthers(withConnectionId(user => user.presence.cursor))
```

But even if more composable, I think it's too indirect.

nvie: > _Taking these points into consideration, I would go with `useOthersWithConnectionId`._

I can see that argument. My gripe with that name is that (1) `useOthers` also returns a value that has connection IDs, so initial reactions may be "why the `WithConnectionId` part" and (2) to me, although it _also_ returns connection IDs, the main purpose is the derived `data` you want to select. That's why it exists mainly. The connection ID is only there to support typical React usage of it, but it isn't the main thing. That's why I've been leaning towards `useOthersWithData` more. Especially now that we're splitting off the `useConnectionIds()` as a separate (and clearly named!) hook.

So, the ultimate compromise here between your both proposals might actually be the name I prefer most…

`useOthersData` + `useOthersWithConnectionId` = `useOthersWithData` ❤️ ?

CTNicholas: What about `useConnectionIds` and `useConnectionData`? The connection feels a little more obvious there, and I prefer this. My second choice is `useConnectionIds` and `useOthersWithConnectionId`, even if it's long, because it's quite clear what it is.

---

Alternatively, what if we had a hook that returned selectors, instead of ids:
```
function Parent() {
  // List of selector functions for each other user
  const selectors = useOthersSelectors();

  // Functions are stable and can be used as map keys (is this allowed in React?)
  return (
    {selectors.map((useOther) => 
      <Child key={useOther} useOther={useOther} />
    )}
  );
}


function Child({ useOther }) {
  const cursor = useOther((other) => other.presence.cursor));
}
```

Could also return ids and selectors:
```
function Parent() {
  // List of { id, selector } objects for each other user
  const selectors = useOthersSelectors();

  // Use id as map key
  return (
    {selectors.map((other) => 
      <Child key={other.id} useOther={other.selector} />
    )}
  );
}


function Child({ useOther }) {
  const cursor = useOther((other) => other.presence.cursor));
}
```

GuillaumeSalles: @CTNicholas 💡! That's really clever! I really like that we don't have to transfer any "connectionId" from the parent to the children components. I have never seen a single library where you pass a hook as a property so I'm curious to know how you got that idea. Is it coming from you or did you see this pattern somewhere else?

CTNicholas: @GuillaumeSalles Thanks—I just enjoy coming up with unconventional API ideas, I've never seen it before! Definitely wondering if there's a reason we haven't seen it.

PR #386, opened 2022-07-04
Upgrade shoe example to 0.17 patterns

GuillaumeSalles:


marcbouchenoire (approved):
Too bad I got stuck on the `react-whiteboard` bug because [I did the same thing this morning](https://github.com/liveblocks/liveblocks/commit/2e364c4b4640bf42d593ac15505d29d7dd6a254e) but didn't open or push it while I was looking for a fix.

marcbouchenoire:
Since the `react-whiteboard` bug was actually already fixed in the end I'll open a PR for this right away!

PR #395, opened 2022-07-06
Update tests to make it easier to work with test IDs

nvie:
This restructures the "test ID" helpers a bit, so we're able to use plain string literals (which I think are easier to read and debug), and structure it so that TypeScript will prevent us from making typos.

### DIFF
--- a/packages/liveblocks-react/src/index.test.tsx
+++ b/packages/liveblocks-react/src/index.test.tsx
@@ -17,6 +17,16 @@ import * as React from "react";
 
 import { createRoomContext } from "./factory";
 
+type TestID = "me-x" | "increment" | "othersJson" | "liveObject" | "unmount";
+
+function testId(testId: TestID) {
+  return testId;
+}
+
+function element(testId: TestID) {
+  return screen.getByTestId(testId);
+}
+
 /**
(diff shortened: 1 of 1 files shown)


packages/liveblocks-react/src/index.test.tsx
```diff
@@ -17,6 +17,16 @@ import * as React from "react";
 
 import { createRoomContext } from "./factory";
 
+type TestID = "me-x" | "increment" | "othersJson" | "liveObject" | "unmount";
+
+function testId(testId: TestID) {
+  return testId;
+}
```
nvie, line 24: Essentially a no-op, but by using this wrapper, a type check is enforced.

packages/liveblocks-react/src/index.test.tsx
```diff
@@ -17,6 +17,16 @@ import * as React from "react";
...
+
+function element(testId: TestID) {
+  return screen.getByTestId(testId);
```
nvie, line 27: Light wrapper around `.getByTestId()` that enforces a type check on the literal.

GuillaumeSalles (approved):

PR #396, opened 2022-07-06
Upgrade all examples to 0.17.5

nvie:


nvie:
I guess that's the consequence of [this decision](https://liveblocks.slack.com/archives/C02PZL7QAAW/p1656951292381109), not?

PR #400, opened 2022-07-07
Fix: Examples on mobile

marcbouchenoire:
This PR fixes various styles/behaviors in examples to improve them on mobile:
- [x] Prevent scrolling outside of iframes from cursor examples
- [x] Fix styling issues


nvie (approved):
Tried out a few, and they feel much better on mobile now. Thanks for improving those! 🙌

PR #403, opened 2022-07-11
Abstract out "insort" logic in LiveList

nvie:
This PR extracts out some common logic into a helper method called `_insortItem(newItem)`, which (a) pushes a new child item into the internal array of items, and (b) in-place sorts that list. This PR isn't intended to optimize this algorithm, but to DRY up these operations which are repeated a few times in the current implementation.

Simply put:

- `this._items.push(xxx); sortListItem(this._items)` → `this._insortItem(xxx)`
- `sortListItem(this._items)` → `this._sortItems()`

This will make adding the soon-to-come invalidation logic for cached JSON a lot easier to implement, as there will be fewer places that would need to mark the JSON cache as invalid. (That upcoming work, in turn, supports an efficient `useSelector()` implementation.)

### DIFF
--- a/packages/liveblocks-client/src/LiveList.ts
+++ b/packages/liveblocks-client/src/LiveList.ts
@@ -80,7 +80,7 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
       const child = deserialize([id, crdt], parentToChildren, doc);
 
       child._setParentLink(list, crdt.parentKey);
-      list._insortItem(child);
+      list._insertAndSort(child);
     }
 
     return list;
@@ -115,7 +115,7 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
    *
    * Adds a new item into the sorted list, in the correct position.
    */
-  _insortItem(item: LiveNode): void {
+  _insertAndSort(item: LiveNode): void {
     this._items.push(item);
     this._sortItems();
   }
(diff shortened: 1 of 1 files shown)


packages/liveblocks-client/src/LiveList.ts
```diff
@@ -104,6 +110,21 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
     return ops;
   }
 
+  /**
+   * @internal
+   *
+   * Adds a new item into the sorted list, in the correct position.
+   */
+  _insortItem(item: LiveNode) {
```
nvie, line 118: "Insort" is not a typo. I took the name from Python's equivalent stdlib function: [`insort()`](https://docs.python.org/3/library/bisect.html#bisect.insort) (from the `bisect` module). That `bisect` module from Python offers other interesting helper functions for maintaining an always-sorted list internally, which we might use to avoid having to sort the items list every time we change an item. (An optimization for later, though. Not for now.)

ofoucherot (approved):
lgtm 🚀

packages/liveblocks-client/src/LiveList.ts
nvie: Okay, that name may actually make more sense. Updated in c63ad5fcd7953b2bf20c135d0a1f7ade887aae4e! 🙏

Issue #367, opened 2022-06-26
Better storage hooks for @liveblocks/react


GuillaumeSalles:
The goal of this issue is to fix the major drawbacks of `@liveblocks/react` storage hooks.

This will require:
- [ ] Preparing a tech design and presenting it to the team
- [ ] Find a way to collect feedback from the community as early as possible
- [ ] Plan implementation, release and documentation

_**Create additional sub-tasks and link them here to track the overall progress.**_

# Storage hooks drawbacks

## Handling nested data structures is not intuitive

`useList`, `useMap`, `useObject` only let you access direct children of the storage root. 

Let's consider this storage root type:

```typescript
type StorageRoot = {
  child: LiveObject<Child>
}

type Child = {
  grandChild: LiveObject<GrandChild>
}

type GrandChild = {
  x: number
}
```

It's not possible to get `grandChild` with `useObject`. To access `grandChild`, you need to do:

```tsx
function Component() {
  const child = useObject("child");

  if(child == null) {
    return null;
  }

  const grandChild = child.get("grandChild");
}
```

Re-rendering your component when grandChild is changing is even more awkward

```tsx
function Component() {
  const child = useObject("child");

  if(child == null) {
    return null;
  }

  return <OtherComponent grandChild={child.get("grandChild")} />
}

function OtherComponent({ grandChild }) {
  const room = useRoom();
  const [{ x }, setData] = useState(grandChild.toObject());

  useEffect(() => {
    return room.subscribe(grandChild, () => {
       setData(grandChild.toObject());
    });
  }, [room, grandChild]);
}
```

## `useList`, `useObject`, `useMap` return mutable data structures

Our LiveStructures are mutable for performance reasons and to fit better with non-react environments.

This can lead to unintuitive behavior when mixed with react hooks dependencies or `React.memo`.

We had several questions like this one:

```tsx
function Component() {
  const object = useObject("key");
  useEffect(() => {
    // Why is this effect not triggered when `object` is modified??

    // Answer: object is mutable, the reference never changes so this effect will not be triggered
  }, [object]);
}
```

## Re-rendering when a computed value of multiple storage parts has been updated

We need to access the underlying room to achieve this.

```tsx
function Component() {
  const objA = useObject("a");
  const objB = useObject("b");

  const room = useRoom();
  const [computedValue, setComputedValue] = useState();

  useEffect(() => {
    function onChange() {
      setComputedValue(objA.get("x") + objB.get("x"));
    }

[...]

PR #404, opened 2022-07-11
Add `useSelector()` hook

nvie:
This PR proposes and implements a new hook in `@liveblocks/react` that makes working with data from Storage a lot easier: `useSelector()`. This proposed solution solves many of the problems stated in #367.

## Introducing `useSelector()`
The `useSelector()` API is a common abstraction in state management libraries for React, like Zustand, or Redux. People familiar with those libraries should have little trouble understanding this hook.

For those unfamiliar, here is an example usage:

```ts
function MyComponent() {
  const data = useSelector(root => root.x.y.z);
  if (data === null) {
    return <p>Loading...</p>;
  }

  return <pre>{JSON.stringify(data)}</pre>;
}
```

This simple example illustrates a few different new behaviors compared to our current `useMap`, `useList`, `useObject` hooks:

1. The `root => root.x.y.z` is called a "selector" function.
2. Via `root`, you have access to your entire Storage, but as immutable "normal" JavaScript data structures. No more dealing with Live data structures when reading data from storage. For example, if your Storage is a `LiveObject<{ x: LiveList<string> }>`, then root will be of type `{ x: string[] }` here.
3. This function is re-evaluated any time something changes in `Storage`.
4. You can return anything you like from this function.
5. If the returned value from this function changes, your component will get re-rendered.

This API solves a lot of use cases that were previously more difficult to solve with Liveblocks APIs.

## Replacing `useList`, `useMap`, `useObject`
The following examples can trivially be replaced:

```ts
// Before
const animals   = useList('animals')?.toArray();
const usersById = useMap('usersById')?.toMap();
const thing     = useObject('thing')?.toObject();

// After
const animals   = useSelector(root => root.animals);    // string[]
const usersById = useSelector(root => root.usersById);  // Map<string, User>
const thing     = useSelector(root => root.thing);      // Thing
```

There's no more need for using `.toArray()`, `.toObject()` etc constructions to convert Live values to their "normal" JavaScript equivalents. With `useSelector()` all you have to deal with is simple/normal JavaScript datastructures.

## Accessing nested data
This also solves the problem of accessing nested data. Deeply nested live objects are now trivially accessed:

```ts
// Before
const child = useObject('child');
if (child == null) {
  return <p>Loading...</p>;
}

const grandChild = child.get('grandChild').toObject();
```

Now:

```ts
const grandChild = useSelector(root => root.child.grandChild);
if (grandChild == null) {
  return <p>Loading...</p>;
}
```

Benefits:
1. Accessing the grandChild is more natural with a normal dotted path selection. The selector function won't get invoked until storage has fully loaded.
2. Additionally, you don't have to deal with the `.get()` and `.toObject()` methods anymore, because it's all normal JSON.
3. The "before" example won't re-render automatically when `grandChild` or any of its nested properties change, but using the `useSelector()` API, they will. No more awkward subscribing to the `room`, or setting up manual watchers.

## Re-rendering when either of multiple storage parts change

Previously impossible with the existing hooks, but now almost trivial:

```ts
function Component() {
  const computedValue = useSelector(root => root.x + root.y);
}
```

This component will automatically re-render if either `x` or `y` changes.

## Efficient re-rendering out of the box
Notice the `root => root.child.grandChild` selector in the example above? The selector function is **re-revaluated** anytime something changes inside storage—so consider your selector functions a hot code path.

But the component will only get **re-rendered** if the returned value from the selector changes! All JSON values underneath `root` will be referentially identical if nothing about them (or their children) have changed. That's why `root => root.child.grandChild` can be written without any performance overhead. If you would have written `root => ({ ...root.child.grandChild })`, then the component _would_ re-render anytime something changes in storage, even if elsewhere.

This has implications if you want to filter a LiveList value, for example. For example, suppose `root.numbers` is a live list with numbers, and you want to only show even numbers in the list. You could naively use:

```ts
const evens = useSelector(root => root.numbers.filter(n => n % 2 === 0));
```

Since `.filter()` returns a new array instance every time, this is not a stable function, and as a result, this will trigger a re-render of your component anytime _something_ changes in Liveblocks Storage, even stuff unrelated to the `'numbers'` key.

To account for this, you'll need to use a shallow comparison function to compare the returned results between evaluations. To do this, use:

```ts
import { shallow } from "@liveblocks/client";

const evens = useSelector(root => root.numbers.filter(n => n % 2 === 0), shallow);
//                                                                       ^^^^^^^
```

Now, this component will automatically re-render if (and only if!) an _even_ number gets added to the LiveList. If other users add _odd_ numbers to the list, this component will know it doesn't have to re-render.


packages/liveblocks-client/src/EventQueue.ts
```diff
@@ -0,0 +1,37 @@
...
+    _observers.add(callback);
+    return () => _observers.delete(callback);
+  }
+
+  function notify(event: T) {
+    _observers.forEach((callback) => callback(event));
+  }
+
+  return [subscribe, notify];
+}
```
nvie, line 37: This is a simple helper to encapsulate subscription/notification updates. By creating an event pair this way, all the code to manage those event callbacks get abstracted away, leading to a lot less code (and supporting variables) in the Room. In addition, the split between subscribe and notify means you can publicly expose the `subscribe` part to the world, in an outer API, to allow external subscribers, while keeping the `notify` part for internal-to-the-room use only (so external parties cannot call `notify` themselves).

I would love to restructure all the internal `subscribe()`-related support code in the Room using this abstraction too, but it would be beyond the scope of this PR (and I wouldn't get this done before my holiday anyway).

I'll make a ticket to generalize this implementation clean up the Room internals using this little helper abstraction.

packages/liveblocks-client/src/immutable.ts
```diff
@@ -271,25 +271,32 @@ function getParentsPath(node: LiveNode): Array<string | number> {
...
 
-export function patchImmutableObject<S extends JsonObject>(
+//
+// TODO: Remove `patchImmutableObject`!
+//
+// This helper is now only used internally, to support our Zustand and
+// Redux packages. We should be able to reimplement those using the new
+// `.toImmutable()` APIs.
+//
+export function legacy_patchImmutableObject<S extends JsonObject>(
```
GuillaumeSalles, line 281: One thing we should consider for the zustand/redux integration is whether to keep the incoming immutable reference or always use the one generated from the room's storage.

For example, with a zustand store like this one:

```ts
const useStore = create(
  middleware<State, Presence>(
    (set) => ({
      obj: { a: 0 },
      otherObj: { b: 0 },

      setObj: (newObj) => set({ obj: newObj }),
      setOtherObj: (otherObj) => set({ otherObj: newOtherObj }),
    }),
    {
      client,
      storageMapping: { obj: true, otherObj: true },
    }
  )
);
```

On 0.17, after calling `setObj`, `state.obj` === `newObj`.

If we update the middleware to always synchronize the store with the underlying immutable references, `state.obj` !== `newObj`. It's hard to predict if it will have any impact... it depends on how developers use their store, so I don't think we should take that risk. 

Ideally, we should have a way to replace the immutable cache of a `LiveStructure` without breaking the encapsulation. My first intuition is to add a method on `LiveStructure` that would take an immutable object to update them and replace the immutable cache simultaneously. Basically, moving the code of `patchLiveObject` to `LiveStructure`. What do you think?

nvie: I've codified this concern in an explicit unit test for this case: see f8a4127827b4c8d24b1100fa778f055a9d59d5a2.

PR #406, opened 2022-07-12
Fix auto-detection of external packages in Rollup configs

nvie:
This fixes an issue where, if you had `some-pkg` declared in your dependencies, but then imported from `"some-pkg/sub/module"`, then Rollup would not consider that an external dependency, because it was not an exact match.

The fix made by this PR is to not codify the externals as a static list but to make `external` a callback function that can do a prefix-match.

This fix also allows us to get rid of the custom `@liveblocks/client/internal` hack we had in there to support a similar case.

I am making this fix proactively here (on `main`), because I needed to add `use-sync-external-store` soon on #404 , which made me run into this problem.

### DIFF
--- a/packages/liveblocks-client/rollup.config.js
+++ b/packages/liveblocks-client/rollup.config.js
@@ -11,6 +11,22 @@ import { promises } from "fs";
 const packageJson = require("./package.json");
 const createBabelConfig = require("./babel.config");
 
+function makeExternalFn() {
+  // NOTE: Make sure this list always matches the names of all dependencies and
+  // peerDependencies from package.json
+  const pkgJson = require("./package.json");
+
+  const EXTERNAL_PKGS = [
+    ...Object.keys(pkgJson?.dependencies ?? {}),
+    ...Object.keys(pkgJson?.peerDependencies ?? {}),
+  ];
+
+  return (importPath) =>
+    EXTERNAL_PKGS.some(
+      (ext) => importPath === ext || importPath.startsWith(ext + "/")
+    );
+}
+
 function execute(cmd, wait = true) {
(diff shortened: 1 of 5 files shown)


packages/liveblocks-client/rollup.config.js
```diff
@@ -205,17 +221,7 @@ export default async () => {
...
-
-  // NOTE: Make sure this list always matches the names of all dependencies and
-  // peerDependencies from package.json
-  const pkgJson = require("./package.json");
-  const external = [
-    ...Object.keys(pkgJson?.dependencies ?? {}),
-    ...Object.keys(pkgJson?.peerDependencies ?? {}),
-  ].flatMap((dep) =>
-    // Also include `@liveblocks/.../internal` as an external reference, by convention
-    dep.startsWith("@liveblocks/") ? [dep, dep + "/internal"] : [dep]
```
nvie, line 217: This hack is no longer needed.

packages/liveblocks-node/rollup.config.js
```diff
@@ -192,17 +208,7 @@ export default async () => {
...
-  // peerDependencies from package.json
-  const pkgJson = require("./package.json");
-  const external = [
-    ...Object.keys(pkgJson?.dependencies ?? {}),
-    ...Object.keys(pkgJson?.peerDependencies ?? {}),
-  ].flatMap((dep) =>
-    // Also include `@liveblocks/.../internal` as an external reference, by convention
-    dep.startsWith("@liveblocks/") ? [dep, dep + "/internal"] : [dep]
-  );
+  const external = makeExternalFn();
```
nvie, line 211: Just the same change, repeated 5 times for all of our packages.

PR #413, opened 2022-07-18
Updated sample avatars

stevenfabre:
Replaced the 10 sample avatars with 30 Liveblocks branded avatars on examples that used them.


marcbouchenoire (commented):
Looks good! With a list of 30 I'm wondering if maybe we should follow the same pattern as https://liveblocks.io/loading.svg? Host them on .io and only reference them within the examples. There's nothing wrong to having example-specific resources but here they stay the same between examples so maybe having a single place to maintain/update them could be better?

stevenfabre:
Thanks @marcbouchenoire. I just made your suggested change. Much better!

marcbouchenoire (changes requested):
🥳

It looks like there was a git issue and `examples/nextjs-live-avatars-advanced/pages/api/auth.ts` from https://github.com/liveblocks/liveblocks/pull/365 was added here, we should remove it.

stevenfabre:
My bad! This file got added somehow after switching branch. I just removed it. Thanks @marcbouchenoire

marcbouchenoire (approved):

PR #419, opened 2022-07-23
Make @liveblocks/zustand middleware generic types optional

GuillaumeSalles:
Simplify zustand middleware integration with Typescript. `TPresence`, `TStorage`, `TUserMeta`, and `TRoomEvent` are now optional.

Note that `@liveblocks/zustand` does not work with zustand > v4 because v3 and v4 have completely different type definitions. As soon as zustand v4 is out of the RC phase, we will consider updating our middleware to work with the latest version.

### Example

Let's take a look at our [To-do list](https://github.com/liveblocks/liveblocks/tree/main/examples/zustand-todo-list) example. Without our middleware, the store would look like this:

```ts
import create from "zustand";

type State = {
  draft: string;
  isTyping: boolean;
  todos: Todo[];
  setDraft: (draft: string) => void;
  addTodo: () => void;
  deleteTodo: (index: number) => void;
};

create<State>(/* ... */)
```

With our middleware, you simply need to move the `State` param at the middleware level:

```ts
import create from "zustand";
import { createClient } from "@liveblocks/client";
import { middleware } from "@liveblocks/zustand";

const client = createClient({ /*...*/ });

type State = {
  draft: string;
  isTyping: boolean;
  todos: Todo[];
  setDraft: (draft: string) => void;
  addTodo: () => void;
  deleteTodo: (index: number) => void;
};

create(
  middleware<State>(/* ... */, {
    client,
    presenceMapping: { isTyping: true },
    storageMapping: { todos: true }
  })
);
```

### DIFF
--- a/packages/liveblocks-zustand/types/tsconfig.json
+++ b/packages/liveblocks-zustand/types/tsconfig.json
@@ -3,7 +3,7 @@
     "target": "es2015",
     "module": "es2020",
     "moduleResolution": "node",
-    "lib": ["es2015"],
+    "lib": ["es2015", "DOM"],
     "noImplicitAny": true,
     "noImplicitThis": true,
     "strictFunctionTypes": true,
@@ -12,6 +12,6 @@
     "noEmit": true,
     "forceConsistentCasingInFileNames": true,
 
-    "paths": { "@liveblocks/zustand": [".."] }
+    "paths": { "@liveblocks/zustand": ["../lib"] }
   }
 }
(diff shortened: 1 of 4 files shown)


packages/liveblocks-zustand/types/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "es2015",
     "module": "es2020",
     "moduleResolution": "node",
-    "lib": ["es2015"],
+    "lib": ["es2015", "DOM"],
```
ofoucherot, line 6: Why is it needed?

GuillaumeSalles: This `tsconfig` is only needed for `dtslint` validation. Because we're using fetch in `@liveblocks/client`, this is needed.

PR #428, opened 2022-08-02
Update Examples thumbnails

pierrelevaillant:


marcbouchenoire (changes requested):
👏

There's a few things related to the size I think could be fixed:
- Replace all `width="500"` by `width="536"` to match the new size.
- Improve retina handling → 1072px is the 2x size but the corner radius and border of the mask are 1x, they both should be doubled to look right at 1x and not too small like currently.

marcbouchenoire (approved):

PR #430, opened 2022-08-04
Exploration: canUndo/canRedo

marcbouchenoire:
This PR explores exposing whether `undo` and `redo` will do anything. It's mostly an idea, the API is to be considered as part of the discussion and not set in stone _at all_. ~(e.g. without subscribing, it's currently mostly useless since the main use-case is to update a UI state)~

~I briefly tried to think about how it could be made reactive in `@liveblocks/react` as `useCanUndo` and `useCanRedo` but I'm not sure what to subscribe to (it will only ever change after an undo/redo action) to rerender. I thought `useUndo` and `useRedo` could maybe handle callbacks but that would ignore any change made with `useHistory().undo` or to client directly.~

**Update:** I implemented subscribing as an overload to the existing `subscribe()` (as @GuillaumeSalles suggested), using a single `"history"` event type.

### DIFF
--- a/packages/liveblocks-react/src/factory.tsx
+++ b/packages/liveblocks-react/src/factory.tsx
@@ -791,7 +791,7 @@ Please see https://bit.ly/3Niy5aP for details.`
 
   function useCanUndo(): boolean {
     const room = useRoom();
-    const [canUndo, setCanUndo] = React.useState(false);
+    const [canUndo, setCanUndo] = React.useState(room.history.canUndo);
 
     React.useEffect(() => {
       const unsubscribe = room.subscribe("history", ({ canUndo }) =>
@@ -807,7 +807,7 @@ Please see https://bit.ly/3Niy5aP for details.`
 
   function useCanRedo(): boolean {
     const room = useRoom();
-    const [canRedo, setCanRedo] = React.useState(false);
+    const [canRedo, setCanRedo] = React.useState(room.history.canRedo);
 
     React.useEffect(() => {
       const unsubscribe = room.subscribe("history", ({ canRedo }) =>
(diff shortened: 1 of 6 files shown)


packages/liveblocks-react/src/factory.tsx
```diff
@@ -779,6 +789,38 @@ Please see https://bit.ly/3Niy5aP for details.`
     return useHistory().redo;
   }
 
+  function useCanUndo(): boolean {
+    const room = useRoom();
+    const [canUndo, setCanUndo] = React.useState(false);
```
nvie, line 794:
```suggestion
    const [canUndo, setCanUndo] = React.useState(room.history.canUndo);
```

packages/liveblocks-react/src/factory.tsx
```diff
@@ -779,6 +789,38 @@ Please see https://bit.ly/3Niy5aP for details.`
...
+        unsubscribe();
+      };
+    }, [room]);
+
+    return canUndo;
+  }
+
+  function useCanRedo(): boolean {
+    const room = useRoom();
+    const [canRedo, setCanRedo] = React.useState(false);
```
nvie, line 810:
```suggestion
    const [canRedo, setCanRedo] = React.useState(room.history.canRedo);
```

nvie (approved):
Love this feature and the implementation 😍 ! I left one minor suggestion, namely to initialize it with the correct initial value (instead of assuming `false`). This may avoid "flickering" and requiring two renders to get synced up on the correct value. Thanks for adding this in, @marcbouchenoire !

packages/liveblocks-react/src/factory.tsx
nvie: To be clear, this should be read as `React.useState(() => room.history.canUndo())`, i.e. using a computed initial value, by invoking that `canUndo` function on history.

PR #433, opened 2022-08-15
No need to generate `useCanUndo` wrapper in compat module

nvie:
Small cleanup after #430. This generated layer is only needed for _old_ APIs. New hooks (like `useCanUndo`, etc) shouldn't have an entry in there. I'm enforcing that in this PR by hardcoding the list of APIs to generate for now.

With 0.18, this whole module (and the code generation script) will get removed anyway, since it's no longer needed.

### DIFF
--- a/packages/liveblocks-react/scripts/generate-compat.ts
+++ b/packages/liveblocks-react/scripts/generate-compat.ts
@@ -14,6 +14,29 @@ import {
   isTypeNode,
 } from "typescript";
 
+const GENERATE_FOR_HOOKS = [
+  "RoomProvider",
+  "useBatch",
+  "useBroadcastEvent",
+  "useErrorListener",
+  "useEventListener",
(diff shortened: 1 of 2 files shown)


packages/liveblocks-react/scripts/generate-compat.ts
```diff
@@ -14,6 +14,29 @@ import {
...
+  "useObject",
+  "useObject_deprecated",
+  "useOthers",
+  "useRedo",
+  "useRoom",
+  "useSelf",
+  "useStorage",
+  "useUndo",
+  "useUpdateMyPresence",
+];
```
nvie, line 38: Note: the new APIs `useCanUndo` and `useCanRedo` are _NOT_ in this list.

marcbouchenoire (approved):
Whoops, my bad. I got tricked by the CI. 🫠

PR #435, opened 2022-08-16
Test suite improvement for `@liveblocks/react`

nvie:
This PR reorganizes the test suite for `@liveblocks/react` a bit, so that writing new tests is a lot easier. (I wanted to add a few tests for the upcoming `useSelector()` hook, but found this to be very difficult in the current setup.)

Improvements include:

1. All test code is now moved into `src/__tests__` (relates to #233)
2. The `src/__tests__/_utils.tsx` file wraps the `@testing-library/react` package, to have all tests automatically be wrapped in a `RoomProvider` context (following their own [docs/recommendations](https://testing-library.com/docs/react-testing-library/setup#custom-render))
3. The tests are rewritten to no longer have all these manual testing components that were necessary to test hooks. For example, previously we had to define [an `ObjectComponent`](https://github.com/liveblocks/liveblocks/blob/0d8943d33bb54c90426d4739a7e7a02615272518/packages/liveblocks-react/src/index.test.tsx#L423-L430), which [needed to be wrapped in a `RoomProvider`](https://github.com/liveblocks/liveblocks/blob/0d8943d33bb54c90426d4739a7e7a02615272518/packages/liveblocks-react/src/index.test.tsx#L453-L458), only to be able [to call `useObject()`](https://github.com/liveblocks/liveblocks/blob/0d8943d33bb54c90426d4739a7e7a02615272518/packages/liveblocks-react/src/index.test.tsx#L424), and then we needed [extra constructs to generate test IDs for divs](https://github.com/liveblocks/liveblocks/blob/0d8943d33bb54c90426d4739a7e7a02615272518/packages/liveblocks-react/src/index.test.tsx#L20-L28) for [event handlers](https://github.com/liveblocks/liveblocks/blob/0d8943d33bb54c90426d4739a7e7a02615272518/packages/liveblocks-react/src/index.test.tsx#L250) and [text serialization](https://github.com/liveblocks/liveblocks/blob/0d8943d33bb54c90426d4739a7e7a02615272518/packages/liveblocks-react/src/index.test.tsx#L253), only so that we could `expect()` on the output. I've rewritten all of that using [`renderHook()`](https://github.com/liveblocks/liveblocks/blob/c0eb2c1555594e49606f5599e2a1ce361279313d/packages/liveblocks-react/src/__tests__/index.test.tsx#L199), and [directly `expect()`'ing its results](https://github.com/liveblocks/liveblocks/blob/c0eb2c1555594e49606f5599e2a1ce361279313d/packages/liveblocks-react/src/__tests__/index.test.tsx#L210-L212). Much less noise!
4. Lastly, I've cleaned out the manual mutations done to the mocked websocket to simulate server events, and instead I've abstracted those into a small "Websocket simulator" API, which makes the tests a bit more readable/succinct.

There were no fundamental changes made to the test suite yet. The same stuff is still being tested.

### DIFF
--- a/packages/liveblocks-react/src/__tests__/index.test.tsx
+++ b/packages/liveblocks-react/src/__tests__/index.test.tsx
@@ -5,7 +5,6 @@ import {
 } from "@liveblocks/client/internal";
 import { rest } from "msw";
 import { setupServer } from "msw/node";
-import * as React from "react";
 
 import {
   useMyPresence,
@@ -13,17 +12,7 @@ import {
   useOthers,
   useRoom,
 } from "./_liveblocks.config";
-import { act, fireEvent, render, renderHook, screen, waitFor } from "./_utils"; // Basically re-exports from @testing-library/react
-
-type TestID = "me-x" | "increment" | "othersJson" | "liveObject" | "unmount";
-
-function testId(testId: TestID) {
-  return testId;
-}
-
-function element(testId: TestID) {
-  return screen.getByTestId(testId);
-}
+import { act, renderHook, waitFor } from "./_utils"; // Basically re-exports from @testing-library/react
 
 /**
  * https://github.com/Luka967/websocket-close-codes
--- a/packages/liveblocks-react/src/__tests__/index.test.tsx
+++ b/packages/liveblocks-react/src/__tests__/index.test.tsx
@@ -1,21 +1,14 @@
-import { createClient, LiveObject } from "@liveblocks/client";
 import {
   ClientMsgCode,
   CrdtType,
   ServerMsgCode,
 } from "@liveblocks/client/internal";
-import {
-  act,
-  fireEvent,
-  render,
-  screen,
-  waitFor,
-} from "@testing-library/react";
 import { rest } from "msw";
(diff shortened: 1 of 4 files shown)


packages/liveblocks-react/src/__tests__/index.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+import {
+  useMyPresence,
+  useObject,
+  useOthers,
+  useRoom,
+} from "./_liveblocks.config";
+import { act, renderHook, waitFor } from "./_utils"; // Basically re-exports from @testing-library/react
+
+/**
+ * https://github.com/Luka967/websocket-close-codes
```
ofoucherot, line 20: I would use this reference instead: https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code

ofoucherot (approved):
Small comment but lgtm!

packages/liveblocks-react/src/__tests__/index.test.tsx
nvie: Ah, thx — can update! 👍

Issue #432, opened 2022-08-11
Detect duplicate @liveblocks/client package and mitigate instanceof issues


GuillaumeSalles:
Some [bundlers](https://github.com/liveblocks/liveblocks/pull/98) configurations or mono repo setup duplicate `@liveblocks/client` in the final bundle. It creates bugs that are difficult to identify for our users because we rely on `instanceof` in our code base.

Find a way to mitigate this issue by:
- Detect duplicate `@liveblocks/client` in final bundle (global symbol?) and print an error message that link to specific documentation page
- Remove usage of `instanceof` and use symbol to detect `LiveStructure` (similar to what React is doing I think)

nvie:
I think many library authors struggle with this problem when targeting both CJS and ESM, but I've never seen a reliable detection mechanism for this.

Definitely want to stop relying on `instanceof`, but we still need some kind of oracle to be able to tell whether an incoming object originated from our library one way or the other. Using a unique symbol won't work for the same reason (it's also scoped to the module it's coming from, and thus duplicated). Using a named symbol from the runtime-wide registry (e.g. `Symbol.for()`) may work around the technical issue, but I still think that's only masking the real problem. It would only lead to the "ESM copy" of the bundled code being able to recognize and accept, at runtime, LiveStructure instances that were generated from our "CJS copy" of the bundled code, but that might lead to various wild bugs and broken assumptions elsewhere (deeper down). Stuff like unexpected duplicated memory instances of LiveStructures that should be unique, for example. Or multiple client connections to a room, etc etc.

So I think we should focus on the duplication detection here mostly, and not worry too much about the `instanceof` checks (as it's an unimportant implementation detail).

One hack to detect this situation might be to inject a snippet like this in each of our generated dists:

```ts
if (process.env.NODE_ENV !== "production") {
  const pkg = "@liveblocks/client";
  const format = "ESM";

  const g = globalThis as typeof globalThis & { [key: string]: string };
  if (g && g[pkg] !== format) {
    if (g[pkg] === undefined) {
      g[pkg] = format;
    } else {
      console.warn(
        `${pkg} appears twice in your bundle (as ${g[pkg]} and ${format}). This can lead to hard-to-debug problems. Please see XXX for details.`
      );
    }
  }
}
```

And do a similar thing for all combinations of `@liveblocks/{client,react,redux,zustand,node}` × `{ESM,CJS}`, so emit this code snippet 10 times, with only lines 2 and 3 of this snippet varying.

[...]

PR #437, opened 2022-08-17
Detect accidental duplication in production bundles

nvie:
This adds a code snippet to detect cases where complex bundle setups or certain monorepo setups lead to duplicated inclusion of both the CJS and ESM versions of our library in the final production bundle. The problem with that is that object instances generated with one "runtime copy" of the library end up being sent to APIs from the other "runtime copy", which leads to various bugs.

This is a proposed fix for #432.

Not sure if I actually like this solution. Would love to get everyone's opinion on this? Good idea? Terrible idea? A different approach?

### DIFF
--- a/packages/liveblocks-client/rollup.config.js
+++ b/packages/liveblocks-client/rollup.config.js
@@ -85,6 +85,7 @@ function buildESM(srcFiles, external = []) {
     external,
     plugins: [
       replaceText({
+        __PACKAGE_FORMAT__: "ESM",
         __PACKAGE_VERSION__: packageJson.version,
         preventAssignment: true,
       }),
@@ -109,6 +110,7 @@ function buildCJS(srcFiles, external = []) {
     plugins: [
       resolve({ extensions }),
       replaceText({
+        __PACKAGE_FORMAT__: "CJS",
         __PACKAGE_VERSION__: packageJson.version,
         preventAssignment: true,
       }),
--- a/packages/liveblocks-client/src/index.ts
+++ b/packages/liveblocks-client/src/index.ts
@@ -1,3 +1,21 @@
+if (process.env.NODE_ENV !== "production") {
+  const pkg = "@liveblocks/client";
+  const format =
+    "__PACKAGE_FORMAT__" === "__PACKAGE" + /* don't join */ "_FORMAT__"
+      ? "ESM"
+      : "__PACKAGE_FORMAT__";
+  const g = globalThis as typeof globalThis & { [key: string]: string };
+  if (g && g[pkg] !== format) {
+    if (g[pkg] === undefined) {
+      g[pkg] = format;
+    } else {
+      console.warn(
+        `${pkg} appears twice in your bundle (as ${g[pkg]} and ${format}). This can lead to hard-to-debug problems. Please see XXX for details.`
+      );
+    }
+  }
+}
+
 export { createClient } from "./client";
 export { LiveList } from "./LiveList";
 export { LiveMap } from "./LiveMap";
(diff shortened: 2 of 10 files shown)


ofoucherot:
I don't see any downside to this solution but I don't know if it's the best approach. 

I will do more research tomorrow but could we use the Approach #1 from the Nodejs [doc](https://nodejs.org/api/packages.html#dual-commonjses-module-packages ) on "Writing dual packages while avoiding or minimizing hazards" ?

nvie:
> …could we use the Approach #1 from the Nodejs [doc](https://nodejs.org/api/packages.html#dual-commonjses-module-packages)…

Ah! _There_ it was! I've been trying to find back a blog post of some kind that I read about that trick in the past, but couldn't recollect it. This Node doc page was exactly what I was trying to recall — thanks for surfacing it for me.

I remember studying this once when I wanted to do the same thing for publishing decoders, but found it to be too cumbersome to set up. I think I didn't want to bother writing a custom code generator for those wrappers, and also didn't feel like maintaining the mapping manually. Back then I didn't find a tool that would enable this wrapping automatically for all exported symbols.

Will take a look to see if it's any easier today.

nvie:
Was thinking of adding the following section to the [Troubleshooting](https://liveblocks.io/docs/guides/troubleshooting) section in the docs to link to in the warning message:

> ### Avoiding bundling Liveblocks twice in your production bundle
> 
> We publish Liveblocks packages to NPM in both CJS and ESM formats. When your bundler supports modern code, it will automatically use the ESM sources. This has several benefits, like tree-shaking.
> 
> However, in certain complex bundler setups, we've seen cases where Liveblocks code  accidentally gets included twice—both the CJS and ESM implementations end up in production bundles. If this happens, it self-detects this case and issues the following warning in the console at runtime:
> 
> ```
> @liveblocks/client appears twice in your bundle (as CJS and ESM).
> ```
> 
> This scenario should obviously be avoided. On the one hand, it means your production bundle size is unnecessarily bloated. But worse, it leads to hard-to-debug bugs caused by two separately running Liveblocks client states. When you hit this situation, it's well possible that the Liveblocks package isn't the only package that is getting included twice, so please do double-check your bundlers configuration and dependency tree.

Not sure if we can be more helpful/concrete in the _actual_ way to debug this issue, since so much of the nitty gritty details here depend on the actual bundler and their configuration that there are myriad ways to end up in this scenario…?

Ideas?

ofoucherot:
> ```
> 
> ```

Let's forget the wrapping solution then. It looks like there is no "standard" solution. The warning that contains a link to the documentation is enough I think.

nvie:
I dug a little deeper here, and I think the problem might be bigger than we originally anticipated. This isn't an issue just for complex bundler setups or monorepos. Even our relatively standard Next.js e2e test app setup (basically a standard `npx create-next-app` unless I'm overlooking something here) seems to be affected and is bundling our app twice… 🤔

After some deeper debugging, this seems to happen in the bundle:

```
==> First import of @liveblocks/client CJS in .next/server/pages/storage/list.js:17:75)
==> Dupe import of @liveblocks/client ESM in .next/server/pages/storage/list.js:21:77)
```

So first the CJS version is required from L17, then the ESM version is imported from L21. That seems bad!

That generated Webpack code looks like:

Screen Shot 2022-08-22 at 11 51 37@2x

Here, L17's CJS-style import corresponds to the original source's line: https://github.com/liveblocks/liveblocks/blob/main/e2e/next-sandbox/pages/storage/list.tsx#L1

Next L21's import corresponds to https://github.com/liveblocks/liveblocks/blob/main/e2e/next-sandbox/pages/storage/list.tsx#L4 which in turn does https://github.com/liveblocks/liveblocks/blob/main/e2e/next-sandbox/utils/createClient.ts#L1

It's unclear to me why the first one uses the CJS module, while the second is using an ESM module. The dual package hazard problem seems real.

@ofoucherot @GuillaumeSalles Perhaps it would be better to follow the ESM wrapping technique after all, and sacrifice tree-shakability in return?

nvie:
Closing this in favor of #441.

PR #439, opened 2022-08-18
Reorganize all unit tests into `__tests__` dir

### LINKED ISSUE #233: Move tests to separate folder

nvie:
Fixes #233.

I already did this for `liveblocks-react` as part of #435. This PR just does the same mechanical move for all the other projects.


ofoucherot (approved):
🚀

PR #441, opened 2022-08-22
Publish ESM modules as light wrappers around CJS modules

nvie:
This PR is a new attempt at solving the [Dual Package Hazard™](https://nodejs.org/api/packages.html#dual-commonjses-module-packages) issue #432, and it replaces the initially attempted solution from #437 (now closed).

The initial approach from #437 was to inject an auto-detection code snippet in the ESM and CJS module outputs, and this solution worked. In fact, it worked so well that it became apparent that this issue was more wide-spread than we thought initially. Even in a standard `npx create-next-app`, importing from `@liveblocks/client` from a few modules would lead to a mix of ESM and CJS modules in the generated bundles. For details, see this https://github.com/liveblocks/liveblocks/pull/437#issuecomment-1222184552.

This PR solves this issue by using the recommended solution by the Node documentation: generate the CJS modules, and then use a light ESM wrapper for the public exports. This unfortunately comes at the cost of tree-shakability, but we can try to solve for that later.

Tree-shakability is nice, but in practice, and in the current state of the code base, it's not a huge gain. However, it's much more important that it works out of the box in standard (aka non-complex) bundler setups without issues.

So this PR:

- ~Explicitly uses only `*.cjs` and `*.mjs`, never outputs any `*.js` files in the output anymore, because it's ambiguous which style those will get interpreted as~ Nevermind! Didn't play well with Jest. Other tools out there might assume to find `*.js` files out there, so let's not even try to go down this path.
- Replaces `rollup` with `tsup` while I was at it (= much less config/plugins 🎉 )
- Uses [`gen-esm-wrapper`](https://github.com/addaleax/gen-esm-wrapper) to generate the light ESM wrappings


vinodsantharam (approved):
Amazing! Bravo @nvie!

ofoucherot (approved):
Didn't know about tsup, looks good!

Lgtm!

PR #443, opened 2022-08-24
Presence internals refactoring

nvie:
This is a refactoring of the internals used to maintain our Presence state inside the room's machine. It replaces the `state.me`, `state.others`, and `state.users` fields, and encapsulates them all into a single `state.presence` field, which is a single `Presence` class instance. This way, we can "reason locally" about all updates to presence. More importantly, it enables us to start treating all presence internals as immutable/readonly data, just like we do since #404.

## Why this change?
This is an important change, because it's an enabler for implementing a `usePresenceSelector()` hook (which I'll tackle in a follow-up PR). Basically, this refactoring makes all presence data (`me` +`others`) readonly/immutable, just like we did for our Storage in #404. It means that if you have 3 connected users, and only one is updating their cursor, the other two user's presence data will stay referentially equal until they actually change.

Here's a recent ask for exactly such [from the community](https://discord.com/channels/913109211746009108/913157273046646795/1011337537702338581). We also [want this in 0.18](https://www.notion.so/liveblocks/Release-0-18-997d8dbd04d040598486e4f1c0cf5b1c#4d920ef0b12b4a818c25bb7291e274ca). And it may be the elegant/efficient way to reattempt #140.

## Important details
All Presence state management is now abstracted into the `Presence` class. This class is separately unit tested to provide the object reference guarantees that are important.

## Behavioral change
I've introduced one behavioral change here, compared to our current version (0.17.9). "Others" will only become visible in clients once that client knows both their metadata _and_ their initial presence data. This makes dealing with others data less cumbersome for users, since there will never be a case where metadata or presence data is unknown/undefined for a user. Either we know about the others and we know everything, or we don't know about them yet.

Observable difference in the UI:

### Before

before

### After

after

### DIFF
--- a/packages/liveblocks-client/src/__tests__/room.test.ts
+++ b/packages/liveblocks-client/src/__tests__/room.test.ts
@@ -338,7 +338,7 @@ describe("room", () => {
     );
 
     const users = [];
-    for (const user of machine.selectors.getOthers()) {
+    for (const user of machine.getOthers()) {
       users.push(user);
     }
 
@@ -371,7 +371,7 @@ describe("room", () => {
       })
     );
 
-    expect(machine.selectors.getOthers().toArray()).toEqual([
+    expect(machine.getOthers().toArray()).toEqual([
       { connectionId: 1, presence: { x: 2 } },
     ]);
 
@@ -382,7 +382,7 @@ describe("room", () => {
       })
     );
 
-    expect(machine.selectors.getOthers().toArray()).toEqual([]);
+    expect(machine.getOthers().toArray()).toEqual([]);
   });
 
   describe("broadcast", () => {
@@ -501,12 +501,12 @@ describe("room", () => {
     room.undo();
 
     expect(state.buffer.presence?.data).toEqual({ x: 0 });
-    expect(room.selectors.getPresence()).toEqual({ x: 0 });
+    expect(room.getPresence()).toEqual({ x: 0 });
 
     room.redo();
 
     expect(state.buffer.presence?.data).toEqual({ x: 1 });
-    expect(room.selectors.getPresence()).toEqual({ x: 1 });
+    expect(room.getPresence()).toEqual({ x: 1 });
   });
 
   test("if presence is not added to history during a batch, it should not impact the undo/stack", async () => {
@@ -537,7 +537,7 @@ describe("room", () => {
 
     room.undo();
 
-    expect(room.selectors.getPresence()).toEqual({ x: 1 });
+    expect(room.getPresence()).toEqual({ x: 1 });
     expect(storage.root.toObject()).toEqual({ x: 0 });
 
     room.redo();
@@ -561,7 +561,7 @@ describe("room", () => {
     room.undo();
 
     expect(state.buffer.presence?.data).toEqual({ x: 0 });
-    expect(room.selectors.getPresence()).toEqual({ x: 0 });
+    expect(room.getPresence()).toEqual({ x: 0 });
   });
 
   test("undo redo with presence that do not impact presence", async () => {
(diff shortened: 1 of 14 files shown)


packages/liveblocks-client/src/__tests__/room.test.ts
```diff
@@ -848,6 +868,30 @@ describe("room", () => {
...
+      //   main machine.
+      // - This means the ref machine should first receive either a ROOM_STATE
+      //   or a USER_JOINED message from the server.
+      // - This doesn't seem to be happening. Very possibly this is a detail
+      //   that isn't implemented in the mock implementation of the ref machine.
+      // - I tried calling `refMachine.connect()`, but this didn't make
+      //   a difference.
+      //
+      // - Help?
+      //
```
nvie, line 894: This is a tricky one. Perhaps @ofoucherot you know enough about this test setup to help me solve this issue? I've documented my findings here in this comment. Would love to get some help from you on this one! 🙏

packages/liveblocks-client/src/__tests__/room.test.ts
```diff
@@ -936,11 +980,19 @@ describe("room", () => {
 
       const unsubscribe = machine.subscribe("others", (o) => (others = o));
 
+      machine.onMessage(
+        serverMessage({
+          type: ServerMsgCode.ROOM_STATE,
+          users: { 1: {} },
+        })
+      );
```
nvie, line 988: These tests are updated to reflect the need for each client to receive both `ROOM_STATE`… 1/2

packages/liveblocks-client/src/__tests__/room.test.ts
```diff
@@ -936,11 +980,19 @@ describe("room", () => {
...
+
       machine.onMessage(
         serverMessage({
           type: ServerMsgCode.UPDATE_PRESENCE,
           data: { x: 2 },
           actor: 1,
+          targetActor: 0, // Setting targetActor means this is a full presence update
```
nvie, line 995: …and `UPDATE_PRESENCE` (full) now. 2/2

nvie: I think this litters implementation details in the tests. Perhaps this behavior should be moved into the `setupStateMachine()` or `machine.connect()` somewhere? Not super sure where this belongs. @ofoucherot tips?

packages/liveblocks-client/src/__tests__/room.test.ts
ofoucherot: The util `prepareStorageTest` mocks 2 users connected to a room (one user and one user ref). It doesn’t reproduce the server logic completely, basically, it just forwards storage updates to the other user. We don’t want to replicate the full server logic in the unit tests, if it’s needed we can use the integration tests. If we want to unit test custom flow, we need to send server messages manually.

packages/liveblocks-client/src/types/ServerMsg.ts
```diff
@@ -46,24 +46,46 @@ export type ServerMsg<
...
+  | PartialUpdatePresenceServerMsg<TPresence>;
+
+export type FullUpdatePresenceServerMsg<TPresence extends JsonObject> = {
   type: ServerMsgCode.UPDATE_PRESENCE;
   /**
    * The User whose Presence has changed.
    */
   actor: number;
+  /**
+   * If this message was sent in response to a newly joined user, this field
```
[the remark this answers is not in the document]
nvie: Ah, good point. I already commented this in the _ClientMsg_ type definition, but I forgot to update it here. Thanks for the reminder! Now updated in 05069a5e4343c4cccb600d765e94590713892eb3. The comment for this field now explains that there is no more specific meaning for this field, see https://github.com/liveblocks/liveblocks/blob/05069a5e4343c4cccb600d765e94590713892eb3/packages/liveblocks-client/src/types/ServerMsg.ts#L59-L71

Hope that clarifies 👍

PR #446, opened 2022-08-26
Make tests more stable

nvie:
This tweaks a unit test that is failing more often. Instead of waiting an arbitrary delay (500ms), occasionally this test will fail because the clients haven't established a connection within this timeout. I've expressed this test to wait until a certain text content change happens.

There are likely more places we could make similar updates, but I intend to take a deeper look at the e2e test suite and see if we can speed it up a bit and/or stabilize it a bit more soon. (Once 0.18 is out.)

### DIFF
--- a/e2e/next-sandbox/test/utils.ts
+++ b/e2e/next-sandbox/test/utils.ts
@@ -43,12 +43,20 @@ export async function assertContainText(
   }
 }
 
+export async function getTextContentOrEmpty(
+  page: Page,
+  id: string
+): Promise<string> {
+  const selector = id.startsWith(".") || id.startsWith("#") ? id : `#${id}`;
+  return page.locator(selector).innerText();
+}
+
 export async function getTextContent(page: Page, id: string): Promise<string> {
-  const element = await page.locator(`#${id}`).innerText();
-  if (!element) {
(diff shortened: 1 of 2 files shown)


e2e/next-sandbox/test/utils.ts
```diff
@@ -76,6 +84,51 @@ export function delay(ms: number) {
...
+  } while (Date.now() < timeoutAt);
+
+  const ms = Date.now() - start;
+  attempts.push(`(after ${ms}ms) Timed out`);
+  throw new Error(
+    `Expected text content was never found\n\nid: ${id}\nI tried looking for: ${JSON.stringify(
+      expectedText
+    )}\n\nHere were my attempts:\n${attempts.join("\n")}`
+  );
+}
```
nvie, line 130: The tl;dr here is: try to find a text in a loop. Polling every 60ms, until a timeout of 1500ms is exceeded. Stop as soon as the text is found. If never found, show a log tail of what was attempted and what was the found text (this _hugely_ helps debugging/writing the tests).

FlowFlorent (approved):

ofoucherot (approved):
🚀

PR #447, opened 2022-08-26
Add new selector APIs for presence

nvie:
Recently we added `useSelector(root => ...)` as a selector-based API for accessing our immutable Storage™ tree.

This PR cleans up the definitive names we settled on (so ~useSelector~ now is **useStorage**), and adds similar selector-based hook APIs, but for Presence™ data, by extending the existing `useSelf()` and `useOthers()` hooks with an optional selector argument:

```ts
const root = useStorage();   // immutable root
const me = useSelf();        // immutable current user
const others = useOthers();  // immutable others list

// All of these also have selector overloads now, where you get passed these "root" objects from above
const thing    = useStorage(root => root.some.thing);
const { x, y } = useSelf(me => me.presence.cursor);
const cursors  = useOthers(others => others.map(o => o.presence.cursor), shallow);
```

All of these can potentially return `null`s if not yet connected/storage has been loaded.

The same semantics apply across the board:
- The received argument is immutable
- All data you select from it will remain to be stable references as long as they didn't change
- The component you use this hook in will only re-render if the selected value is different
- The selector will get re-evaluated if either `me` or (something in the) `others` array changes

### DIFF
--- a/packages/liveblocks-client/src/ImmRef.ts
+++ b/packages/liveblocks-client/src/ImmRef.ts
@@ -1,0 +1,37 @@
+import type { EventSource, Observable } from "./EventSource";
+import { makeEventSource } from "./EventSource";
+
+/**
+ * Base class that implements an immutable cache.
+ *
+ * TODO: Document usage.
+ */
+export abstract class ImmRef<T> {
+  /** @internal */
+  private _value: Readonly<T> | undefined;
+
+  /** @internal */
+  private _ev: EventSource<void>;
+
+  constructor() {
+    this._ev = makeEventSource<void>();
+  }
+
+  get didInvalidate(): Observable<void> {
+    return this._ev.observable;
+  }
+
+  /** @internal */
+  protected abstract _toImmutable(): Readonly<T>;
+
+  protected invalidate(): void {
+    if (this._value !== undefined) {
+      this._value = undefined;
+      this._ev.notify();
+    }
+  }
+
+  get current(): Readonly<T> {
+    return this._value ?? (this._value = this._toImmutable());
+  }
+}
--- a/packages/liveblocks-client/src/ImmRef.ts
+++ b/packages/liveblocks-client/src/ImmRef.ts
@@ -2,6 +2,36 @@ import type { EventSource, Observable } from "./EventSource";
 import { makeEventSource } from "./EventSource";
 
 /**
+ * Patches a target object by "merging in" the provided fields. Patch
+ * fields that are explicitly-undefined will delete keys from the target
+ * object. Will return a new object.
+ *
+ * Important guarantee:
+ * If the patch effectively did not mutate the target object because the
+ * patch fields have the same value as the original, then the original
+ * object reference will be returned.
+ */
+export function merge<T>(target: T, patch: Partial<T>): T {
+  let updated = false;
+  const newValue = { ...target };
+
+  Object.keys(patch).forEach((k) => {
+    const key = k as keyof T;
+    const val = patch[key];
+    if (newValue[key] !== val) {
+      if (val === undefined) {
+        delete newValue[key];
+      } else {
+        newValue[key] = val as T[keyof T];
+      }
+      updated = true;
+    }
+  });
+
+  return updated ? newValue : target;
+}
+
+/**
  * Base class that implements an immutable cache.
  *
  * TODO: Document usage.
--- a/packages/liveblocks-client/src/room.ts
+++ b/packages/liveblocks-client/src/room.ts
@@ -210,10 +210,8 @@ type State<
   };
 
   readonly me: MeRef<TPresence>;
-  presence: {
-    readonly __others: OthersRef<TPresence, TUserMeta>;
-    readonly __othersProxy: OthersProxyRef<TPresence, TUserMeta>; // TODO: Deprecate this proxy helper
-  };
+  readonly others: OthersRef<TPresence, TUserMeta>;
+  readonly othersProxy: OthersProxyRef<TPresence, TUserMeta>; // TODO: Deprecate this proxy helper
 
   idFactory: IdFactory | null;
   numberOfRetry: number;
@@ -483,7 +481,7 @@ function makeStateMachine<
     others?: OthersEvent<TPresence, TUserMeta>[];
   }) {
     if (otherEvents.length > 0) {
-      const others = state.presence.__othersProxy.current;
+      const others = state.othersProxy.current;
       for (const event of otherEvents) {
         eventHub.others.notify({ others, event });
       }
@@ -880,10 +878,10 @@ function makeStateMachine<
       // handle it if `targetActor` matches our own connection ID, but we can
       // use the opportunity to effectively reset the known presence as
       // a "keyframe" update, while we have free access to it.
-      const oldUser = state.presence.__others.getUser(message.actor);
-      state.presence.__others.setOther(message.actor, message.data);
+      const oldUser = state.others.getUser(message.actor);
+      state.others.setOther(message.actor, message.data);
 
-      const newUser = state.presence.__others.getUser(message.actor);
+      const newUser = state.others.getUser(message.actor);
       if (oldUser === undefined && newUser !== undefined) {
         // The user just became "visible" due to this update, so fire the
         // "enter" event
@@ -891,10 +889,10 @@ function makeStateMachine<
       }
     } else {
       // The incoming message is a partial presence update
-      state.presence.__others.patchOther(message.actor, message.data), message;
+      state.others.patchOther(message.actor, message.data), message;
     }
 
-    const user = state.presence.__others.getUser(message.actor);
+    const user = state.others.getUser(message.actor);
     if (user) {
       return {
         type: "update",
@@ -909,9 +907,9 @@ function makeStateMachine<
   function onUserLeftMessage(
     message: UserLeftServerMsg
   ): OthersEvent<TPresence, TUserMeta> | null {
-    const user = state.presence.__others.getUser(message.actor);
+    const user = state.others.getUser(message.actor);
     if (user) {
-      state.presence.__others.removeConnection(message.actor);
+      state.others.removeConnection(message.actor);
       return { type: "leave", user };
     }
     return null;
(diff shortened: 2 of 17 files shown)


packages/liveblocks-react/src/factory.tsx
```diff
@@ -585,7 +607,7 @@ export function createRoomContext<
     }
   }
 
-  function useSelector<T>(
+  function useStorageSelector<T>(
```
FlowFlorent, line 610: should this be named `useStorage`? and we would also have `usePresenceSelector` as well? 

Or is the rational to keep `useStorage` for future utils which wouldn't work with `useStorageSelector`?

packages/liveblocks-react/src/factory.tsx
```diff
@@ -630,6 +652,42 @@ export function createRoomContext<
...
+      },
+      [room]
+    );
+
+    const getSnapshot = React.useCallback(
+      (): Snapshot => room.getPresenceSnapshot(),
+      [room]
+    );
+
+    const getServerSnapshot = getSnapshot;
```
FlowFlorent, line 680: I am guessing this is just to fit the function definition's naming in the [doc](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore)?

nvie: Correct! 👍

packages/liveblocks-react/src/factory.tsx
nvie: Yeah, some naming cleanup needs to happen here. I renamed this one to distinguish it from `usePresence()` currently (which also is a selector-based API). The problem is that we cannot rename it to `useStorage()` just yet, because that hook already exists (in 0.17.9!) and some customers are using it, even though it's not documented officially. So we'll have to first deprecate that API to make room for that name… 🤔 

Definitely some polish to do here!

nvie:
@ofoucherot I'd love that shape! Super short due to the 4-letter field names, and the hook itself already makes it clear that this is presence data, so no need to repeat the `.presence` prop in there any longer. I personally love it! It would mean a derivation from our current structure though. But since we're introducing a new API here, it might not be a breaking change. People will have to opt-in to this API anyway, so I like it 😄 

@GuillaumeSalles @stevenfabre What do you think of Olivier's proposal? Worth the derivation, in exchange for brevity?

nvie:
To be clear, the **current** way of returning others is a list of `User`s:

```ts
export type User<
  TPresence extends JsonObject,
  TUserMeta extends BaseUserMeta
> = {
  readonly connectionId: number;
  readonly id: TUserMeta["id"];
  readonly info: TUserMeta["info"];
  readonly presence: TPresence;
};
```

The **proposal** is to return them as a list of `Actor`s instead.

```ts
export type Actor<
  TPresence extends JsonObject,
  TUserMeta extends BaseUserMeta
> = {
  readonly actorId: number;
  readonly meta: TUserMeta;
  readonly data: TPresence;
};
```

Same information richness, but put in a "smaller" object.

GuillaumeSalles:
As discussed during our call, I would approach the selectors API by use cases instead of the symmetry we usually use to describe the product. We often use `Presence` vs `Storage` in our concepts and as "groups" for our APIs, but it's an oversimplification that does not represent the room reality.

See @liveblocks/react API reference navigation structure below. 

Events, errors, and users hooks are not really parts of the "Presence" concept.

Here is how I think about the room's public state:
- **Local state** - read/write only by the client - current user + connection state + undo/redo stack
- **Others state** - read-only
- **Shared state** (Storage) - readable/writeable by everyone - conflicts only happen on this sub-state

Most of the time, granular components read only one sub-state at a time. 

- Avatars are built by reading the "Others state"
- Cursors are built by reading the "Other state"
- An offline indicator is built by reading the "Local state"
- The static version of a document is built by reading the "Shared state"

The only common scenario where we need to derive a value from multiple sub-states is when we implement the concept of "storage selection". It's a link between the "Local state" and "Shared state" or a link between "Others state" and "Shared state"; but not a link between "Local state" and "Others state"

Note that adding the current user picture in the avatar stack can be seen as the derived value from the "current users" and "other users." Still, there are no dependencies between the two sub-states as opposed to the "selection" concept.

Currently, our "read" hooks strictly interact with these sub-states independently.

- Local state => `useSelf`, `useMyPresence`, `useCanUndo`, `useCanRedo`
Other states => `useOthers`
Shared state => `useStorage`, `useList`, `useMap`, `useObject`

With this model in mind, this is what I'm proposing for the selection API.

- Add an optional callback on `useOthers` to derive a value from the "Others state". Useful when building / optimizing components like "LiveCursors" / "LiveAvatars". For example: `const avatars = useOthers(users => users.map(user => user.info.avatar))`
- Add an optional callback on `useSelf` to derive a value from the current user. Useful to optimize `const avatar = useSelf().info.avatar` into `const avatar = useSelf(user => user.info.avatar);` or to optimize presence selection `const [{cursor}] = useMyPresence()` into `const cursor = useSelf(user => user.presence.cursor);`
- Find a way to deprecate the existing `useStorage` and add a optional selector callback to derive storage value.
- General purpose `useSelector` to build "storage selection" concept and other unusual scenarios. E.g `const selectedShape = useSelector(({ myPresence, storage }) => storage.shapes[myPresence.selectedId])`

What I'm not sure about:

- Should we have a dedicated hook for local presence selection instead of `useSelf`?
- How to deprecate `useStorage` smoothly?
- How to name all the fields in the general `useSelector` callback?

LMN, what do you think about this use-case-based approach instead of "Presence" vs "Storage" approach.

nvie:
That line of thinking makes total sense, thanks for describing it so clearly 👍 

It's a good reminder of the subtle difference between the marketing term ("Presence vs Storage Blocks"), and the "presence" field/type/concept we expose for users. I agree that having two things referred to as "Presence" is not a great thing, and I will make sure to stop using "Presence" for these things 👍

So the plan is:
- Ditching the name/hooks `usePresence`
- Also find an alternative name for the recently-introduced internal helper (`class Presence`)
- Make the existing hooks (optional) selector-based APIs instead, according to your suggestions

Will get to work on this today—thanks! 🙏

packages/liveblocks-client/src/ImmRef.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  return updated ? newValue : target;
+}
+
+/**
+ * Base class that implements an immutable cache.
+ *
+ * TODO: Document usage.
+ */
+export abstract class ImmRef<T> {
```
ofoucherot, line 39:
```suggestion
export abstract class ImmutableRef<T> {
```

ofoucherot (approved):
Great work! The new React API is a big upgrade.

packages/liveblocks-react/src/factory.tsx
```diff
@@ -210,23 +242,50 @@ type RoomContextBundle<
   ];
 
   /**
-   * Returns an object that lets you get information about all the the users currently connected in the room.
+   * Returns an object that lets you get information about all the the users
```
GuillaumeSalles, line 245: Just caught an old typo while reviewing

```suggestion
   * Returns an object that lets you get information about all the users
```

packages/liveblocks-react/src/factory.tsx
```diff
@@ -210,23 +242,50 @@ type RoomContextBundle<
...
+   * currently connected in the room.
    *
    * @example
    * const others = useOthers();
    *
    * // Example to map all cursors in JSX
-   * {
+   * return (
```
GuillaumeSalles, line 252: I'm really nitpicking but... should we wrap in a fragment to be valid JSX?

packages/liveblocks-react/src/factory.tsx
```diff
@@ -210,23 +242,50 @@ type RoomContextBundle<
...
    *       return null;
    *     }
    *     return <Cursor key={user.connectionId} cursor={user.presence.cursor} />
    *   })
-   * }
+   * )
    */
   useOthers(): Others<TPresence, TUserMeta>;
 
+  /**
```
GuillaumeSalles, line 263: I love this documentation :chefkiss: Really detailed and clean at the same time.

packages/liveblocks-client/src/room.ts
```diff
@@ -146,7 +148,7 @@ type Machine<
   getSelf(): User<TPresence, TUserMeta> | null;
 
   // Presence
-  getPresence(): TPresence;
+  getPresence(): Readonly<TPresence>;
```
GuillaumeSalles, line 151: 👍

GuillaumeSalles (approved):
Awesome job! I'm so excited to be able to land these selectors hooks in 0.18! I also like how you strengthen the internal typings.

Is there a reason why the optional comparison function takes two `unknown` instead of two `T`? I didn't try it locally so maybe it causes inference issues that I don't see just by reading the code.

```ts
function useStorage<T>(
    selector: (root: ToImmutable<TStorage>) => T,
    isEqual?: (a: unknown, b: unknown) => boolean
  ): T | null;
```

vs

```ts
function useStorage<T>(
    selector: (root: ToImmutable<TStorage>) => T,
    isEqual?: (a: T, b: T) => boolean
  ): T | null;
```

packages/liveblocks-client/src/ImmRef.ts
nvie: Can update! I'm still looking for a slightly better abstraction here. I still feel like this abstract class is a little too clunky/bulky to use and am still looking for a slightly lighter API to enable this, but I can do that in a follow-up PR to clean that up. For now, it works.

Until I found the better abstraction that's still itching, I'll change this name — thanks 👍 

Done in 4607112d8e367a48323835d027064e3c6cadae08.

packages/liveblocks-react/src/factory.tsx
nvie: Returning an array-of-JSX elements is also valid/legal JSX, so I think the example was correct. Nevertheless, I agree wrapping it in a fragment may be nicer to stress the JSX nature of the example 👍 

Done in 9d7fdb25a79357c849daabd30ce2a45d6ae1d48b.

nvie:
> Is there a reason why the optional comparison function takes two `unknown` instead of two `T`?

Great question! I initially went with `T` here, but while implementing them I ran into various type issues that I didn't want to focus on, so I loosened this to `unknown`, then forgot to update them back. I just did that in 3aaf9b0f6d29420e6ca2c9a07b5e1fa9d2e0fb2d 👍 

The typing issues I ran into where that in some selectors, the functions return a `T`, but the hook can also still return `null` (while loading). This means that the comparison functions for those hooks that possibly return `null` should be `T | null`, which I updated in 41745bac1d4ee6f2971b6a236f3f094e7c87fe28.

This was harder to see while implementing, but obvious now that the implementation is simplified so much 🎉

PR #449, opened 2022-08-29
Simplify internal event source implementation

### LINKED ISSUE #411: Clean up Room subscription internals using new `EventSource` abstraction
We currently have a lot of support code inside the Room's state machine to manage several event queue callbacks that can be subscribed to using a single exposed `room.subscribe()` API. This setup means a lot of boilerplate: many local variables exist in the function scope, manual `.push()`'ing and looping-to-notify manually.

This internal complexity could be simplified by using [EventSource](https://github.com/liveblocks/liveblocks/blob/50b709a98f5da870061cf9bdb69aca5938e22990/packages/liveblocks-client/src/EventSource.ts) API I added as part of the `useSelector()` API work.

This would make the internal implementation of the Room simpler and as a result easier to debug or reason about.

Ultimately, the implementation of `room.subscribe()` can be simplified a lot as well if we do this.

nvie:
Fixes #411.

There currently is quite a bit of support code inside the Room's state machine to manage several event queue callbacks that can be subscribed to using a single exposed `room.subscribe()` API. This setup means many local variables exist in the state machine's scope, and there is lots of manual `.push()`'ing and looping-to-notify manually.

This PR simplifies the implementation by reusing the recently-introduced `EventSource` abstraction. This small helper abstracts out a simple event subscription/notification API. This way, those details don't have to be repeated internally everywhere.

The nice thing about the EventSource API is that it offers a standardized way to create both a "writable" (for private use) and a "readable" (for external use) part to these. So internally, we can emit events using `.notify()`, but externally, you can only `.subscribe()` to the events.

This change exposes the various event sources externally, making it possible to change:

```ts
room.subscribe('my-presence', ...);
room.subscribe('others', ...);
room.subscribe('connection', ...);
// etc
```

to

```ts
room.events.me.subscribe(...);
room.events.others.subscribe(...);
room.events.connection.subscribe(...);
// etc
```

Those events are all fully typed, and the type complexity there is a lot less than the complex type signature of `room.subscribe()`. It would be possible to officially deprecate the `room.subscribe('<string>', ...)` APIs now that we have these. (This PR doesn't officially do that yet.)

All other packages use these new APIs though, and only use the `room.subscribe()` to subscribe to Node instance changes.

### DIFF
--- a/packages/liveblocks-client/src/EventSource.ts
+++ b/packages/liveblocks-client/src/EventSource.ts
@@ -1,4 +1,4 @@
-type Callback<T> = (event: T) => void;
+export type Callback<T> = (event: T) => void;
 type UnsubscribeCallback = () => void;
 
 export type Observable<T> = {
@@ -10,6 +10,9 @@ export type EventSource<T> = {
    * Private/controlled notification of events.
    */
   notify(event: T): void;
+  subscribe(callback: Callback<T>): UnsubscribeCallback;
+  clear(): void;
+
   /**
    * Observable instance, which can be used to subscribe to this event source
    * in a readonly fashion. Safe to publicly expose.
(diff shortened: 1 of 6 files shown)


FlowFlorent (approved):

nvie:
This PR is based on #447, but it shouldn't have to be. Since that PR still has some PR feedback that needs to be addressed, I'll change this PR's base branch to `release-0.18` now, and rebase it, so this can be merged in now 👍

PR #452, opened 2022-08-31
Add Suspense support to new selector APIs

nvie:
This implements and exports Suspense-enabled versions of our new selector hooks.

## The purpose
To avoid the `null`s returned by all the selectors.

Normally:

```ts
const title = useStorage((root) => root.title);
if (title === null) {
  // still loading Liveblocks
}
```

When using the Suspense version of this hook:

```ts
const title = useStorage((root) => root.title);
//    ^? string

// Will only get here if Liveblocks has been loaded. No need for null-testing manually
```

Of course, the latter requires the component tree to be wrapped in a `<Suspense>` boundary further up the tree somewhere.

## How to import?
After many attempts and trial and error on real world code, I kind of like [Option 3](https://www.notion.so/liveblocks/Release-0-18-997d8dbd04d040598486e4f1c0cf5b1c#a56d8fd5343d4c3c9198a517a179311b) from our original proposal.

Normally, you would import the hooks like so:

```ts
export const { RoomProvider, useRoom, useStorage, useSelf, useOthers } = createRoomContext(...);
//                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                                 Import "normal" hooks from here

export const {
  RoomProvider, useRoom,

  suspense: { useStorage, useSelf, useOthers },
  //          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  //         Import Suspense-enabled from here
} = createRoomContext(...);
```

This puts the user in the choice which version of the hooks to import and use. They can also import either/both, as long as they re-export them under a different name each.

## Caveat: cannot blindly use this on the server side
One caveat that might not be obvious when using these Suspense-enabled hooks is that they don't work on the server side (when rendered by Node).

To work around this (for example when using Liveblocks in a NextJS app), you can use a client-side only suspense component which will simply use the provided fallback on server-side renderings. Here is an implementation you can simply include in your app for that: [`ClientSideSuspense.tsx`](https://gist.github.com/nvie/b540b1783a6debab836f8155a43e543b)

### DIFF
--- a/packages/liveblocks-react/src/factory.tsx
+++ b/packages/liveblocks-react/src/factory.tsx
@@ -804,8 +804,7 @@ export function createRoomContext<
     // Error early if suspense is used in a server-side context
     if (typeof window === "undefined") {
       throw new Error(
-        // XXX Fill in documentation URL
-        "You cannot use hooks in Suspense mode on the server side. Make sure to only call Suspense APIs on the client side. For tips for structuring your app, see XXX"
+        "You cannot call the Suspense version of this hook on the server side. Make sure to only call them on the client side.\nFor tips for structuring your app, see XXX"
       );
     }
   }


packages/liveblocks-react/src/factory.tsx
```diff
@@ -384,6 +384,26 @@ type RoomContextBundle<
...
+    useSelf<T>(
+      selector: (me: User<TPresence, TUserMeta>) => T,
+      isEqual?: (a: T, b: T) => boolean
+    ): T;
+
+    useOthers(): Others<TPresence, TUserMeta>;
+    useOthers<T>(
+      selector: (others: Others<TPresence, TUserMeta>) => T,
+      isEqual?: (a: T, b: T) => boolean
+    ): T;
```
nvie, line 405: Basically all the same selector functions, but without all the possible `null`s in the return types.

nvie:
After having some initial plays with the new APIs, one frustration came up pretty obviously with real world code. The new selector APIs work best if you use them as deep down in your code as possible, to only trigger rerenders for local component updates. As such, it means there are way more places in the code that now need to perform the `null` checks. You can no longer optimize for that by doing the One-Big-Null-Check-To-Rule-Them-All™ at the top of your app.

Therefore, I think **shipping Suspense support in 0.18 is more important than we initially thought it would be**. By using the Suspense versions of these hooks, all these extra null check annoyances will be avoided.

nvie:
I'm going to get this PR merged in now, so we can roll a first beta release today. Please do still review/comment on this — happy to change anything in hindsight! 🙏

PR #456, opened 2022-09-01
Deprecate (old meaning of) `useStorage` hook in 0.17.x

nvie:
This PR deprecates the `useStorage` hook. This API is currently undocumented, but we know about some customers that are relying on it. We want to free up the name for a new API that's currently planned for 0.18.

By deprecating it now, people that upgrade to the latest version of 0.17.x will be guided to change it to `useStorageRoot()` to retain API compatibility.

### DIFF
--- a/packages/liveblocks-react/src/factory.tsx
+++ b/packages/liveblocks-react/src/factory.tsx
@@ -13,7 +13,7 @@ import type {
 } from "@liveblocks/client";
 import { LiveList, LiveMap, LiveObject } from "@liveblocks/client";
 import type { Resolve, RoomInitializers } from "@liveblocks/client/internal";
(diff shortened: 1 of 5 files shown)


packages/liveblocks-react/src/factory.tsx
```diff
@@ -578,6 +584,13 @@ export function createRoomContext<
     return [root];
   }
 
+  function useStorage(): [root: LiveObject<TStorage> | null] {
+    deprecate(
+      "In the upcoming 0.18 version, the name `useStorage()` is going to be repurposed for a new hook. Please use `useStorageRoot()` instead to keep the current behavior."
+    );
```
nvie, line 590: 🔑 This is the key change for this PR.

GuillaumeSalles (approved):

PR #460, opened 2022-09-05
Make initial presence mandatory

nvie:
This PR forces the `initialPresence` to be provided whenever you enter a room. It's no longer optional. As such, we will always know for sure that there is a `.presence` field on every user, both for `me` and for all the `others`.

### DIFF
--- a/packages/liveblocks-zustand/src/index.ts
+++ b/packages/liveblocks-zustand/src/index.ts
@@ -194,7 +194,7 @@ export function middleware<
         return;
       }
 
-      room = client.enter(roomId);
+      room = client.enter(roomId, { initialPresence: {} as TPresence });
 
       updateZustandLiveblocksState(set, {
         isStorageLoading: true,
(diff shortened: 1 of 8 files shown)


packages/liveblocks-zustand/src/index.ts
```diff
@@ -194,7 +194,7 @@ export function middleware<
         return;
       }
 
-      room = client.enter(roomId);
+      room = client.enter(roomId, { initialPresence: {} as TPresence });
```
nvie, line 197: Zustand and Redux packages seem to get initialized differently, where the initial presence is expected always to be "empty" and immediately followed by a call to broadcastInitialPresence (which emits multiple presence chunk updates—one for every field). Not sure why that happens that way. Need to dig in there, but for now, let's just keep this working the way it currently works.

nvie:
Going to get this PR merged in now, so we can roll a first beta release today. Please do still review/comment on this — happy to change anything in hindsight! 🙏

Issue #463, opened 2022-09-06
Improve `useOthers()` for >90% use cases


nvie:
The goal of this issue is to fix some drawbacks of the recently-implemented selector-based `useOthers()` hook.

It was implemented to be consistent with its current (0.17) implementation, and to be symmetric to the way `useStorage()` and `useSelf()` works.

For example:

```tsx
useStorage(root => root.some.thing)
useSelf(me => me.presence.cursor)
useOthers(others => others.map(...), shallow);
```

Since `useOthers()` is the only one of these hooks that operates on a _list_, the only reasonable operations you can do on those that make sense are:

- Some form of **mapping** - e.g. selecting all cursors
- Some form of **filtering** - e.g. selecting only users with an avatar
- Some form of **folding** - e.g. testing if _some_ or _every_ user matches a predicate

Let's consider a use case for each.

### Use cases for mapping
There are legion of these. In fact, >90% of use cases will be some kind of mapping. For example, extracting cursors, drafts, selections, avatars, etc for each of these users would require a mapping.

```tsx
// 0.18.0-beta0
const cursors = useOthers(others => others.map(other => other.presence.cursor), shallow);
const drafts = useOthers(others => others.map(other => other.presence.pencilDraft), shallow);
const avatars = useOthers(others => others.map(other => other.info.avatar), shallow);
```

It's pretty verbose, right? Also, the `shallow` here is used to compare the outer selection, which in the case of a `.map()` will always be the outer array. Not using a `shallow` comparison makes absolutely no sense with this API, and will likely always be a bug.

While selecting "stable" references means the `shallow` comparison makes sense here, it becomes problematic if we select derived data, for example, when we want to select multiple things:

```tsx
// 0.18.0-beta0
const drafts = useOthers(
  others => others.map(other => ({
    path: other.presence.pencilDraft,
    color: other.presence.penColor,
  })),
  shallow
);
```

Now, the `shallow` won't behave like you expect, because it will be used on the outermost array, but the returned `{ path, color }` objects will be `===`-compared. Because they are always new objects, these will never match. You can only solve this by using a two-level-deep-shallow comparison function 😢  Let's call this the TwoLevelShallow Comparison Problem™.

[...]

PR #467, opened 2022-09-07
Add new APIs: `useOtherIds()` and `useOther(n)` for better others iteration

nvie:
This attempts to fix the annoyances with `useOthers()`, as described in #463.

New APIs introduced here:

```tsx
const ids = useOtherIds();
// [2, 5, 11]

const cursors = useOtherIds(user => user.presence.cursor);
// [
//   { connectionId: 2, data: { x: -18, y: 292 } },
//   { connectionId: 5, data: null },
//   { connectionId: 11, data: { x: 38, y: -1 } },
// ]

const someUser = useOther(2);
// {
//   connectionId: 2,
//   id: 'internal-id-4',
//   info: {
//     avatar: 'https://user-images.githubusercontent.com/83844/188878306-73cece2d-d9b2-46c6-9b11-8bdc02d58d4f.png',
//   },
//   presence: {
//     cursor: { x: -18, y: 292 },
//   },
// }

const { x, y } = useOther(2, user => user.presence.cursor);
// { x: -18, y: 292 }

useOther(99);  // throws! "No such connection ID"
```

### DIFF
--- a/packages/liveblocks-react/src/factory.tsx
+++ b/packages/liveblocks-react/src/factory.tsx
@@ -686,7 +686,7 @@ export function createRoomContext<
           const eq = isEqual ?? Object.is;
           return eq(a, b);
         },
-        [sentinel, isEqual]
+        [isEqual]
       );
 
       const other = _useOthers(wrappedSelector, wrappedIsEqual);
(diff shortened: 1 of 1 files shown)


CTNicholas (approved):
We just had a chat about this, I like this addition 👍 we can keep `useOthers` as it is for basic apps, and make use of these for cases where improved performance is necessary

Issue #416, opened 2022-07-20
Nested batching support


GuillaumeSalles:
It's not currently possible to do this:

```ts
room.batch(() => {
  // This fails 👇 
  room.batch(() => { });
});
```

Batching all operations when the outer batch is over might be intuitive enough.

nvie:
Will be supported starting with 0.18 👍

PR #468, opened 2022-09-07
Convert advanced whiteboard example to 0.18

nvie:
This PR converts `nextjs-whiteboard-advanced` to the new 0.18 APIs, and uses new best practices. I kept the commit history super clean and atomic, so I encourage you to read this PR commit-by-commit, using GitHub's next/prev buttons to navigate. Each change shows one conversion, so you can easily compare "old" vs "new" patterns.

The end result is a massive improvement, not only making the code simpler, but also the app is a lot faster because the amount of rerenders has drastically dropped.

Just look at the rerender counter for the main canvas area to see the difference 😄 

### Before

https://user-images.githubusercontent.com/83844/188910815-9097010d-6094-4e70-8daa-9ed96edecd6e.mp4

### After

https://user-images.githubusercontent.com/83844/188910842-91b1abd8-fd3c-469e-955c-76747e6d5aab.mp4

### DIFF
--- a/examples/nextjs-whiteboard-advanced/src/index.tsx
+++ b/examples/nextjs-whiteboard-advanced/src/index.tsx
@@ -103,7 +103,7 @@ function Canvas() {
   const canUndo = useCanUndo();
   const canRedo = useCanRedo();
 
-  const selectionBounds = useSelectionBounds(liveLayers, selection);
+  const selectionBounds = useSelectionBounds();
 
   useDisableScrollBounce();
 
--- a/examples/nextjs-whiteboard-advanced/src/index.tsx
+++ b/examples/nextjs-whiteboard-advanced/src/index.tsx
@@ -122,8 +122,10 @@ function Canvas() {
   /**
    * Delete all the selected layers.
    */
-  const deleteLayers = useCallback(() => {
-    batch(() => {
+  const deleteLayers = useMutation(
+    ({ root }) => {
+      const liveLayers = root.get("layers");
+      const liveLayerIds = root.get("layerIds");
       for (const id of selection) {
         // Delete the layer from the layers LiveMap
         liveLayers.delete(id);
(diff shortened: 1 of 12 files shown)


FlowFlorent (approved):
:nice:

examples/nextjs-whiteboard-advanced/src/index.tsx
```diff
@@ -438,35 +455,57 @@ function Canvas({
...
+      camera,
+      canvasState,
+      continueDrawing,
+      resizeSelectedLayer,
+      setPresence,
+      startMultiSelection,
+      translateSelectedLayers,
+      updateSelectionNet,
+    ]
+  );
```
nvie, line 508: @GuillaumeSalles Wondering if there is a reason why nesting batches throws an error. Could we not safely combine inner batches with the outermost batch silently (making inner batch wrappings a no-op?), or would that be problematic?

I hit this issue when I tried to convert this callback to a `useMutation()` call. Since it will invoke other mutations, this causes nested batching and thus throws an error. That's why I had to make it a `useCallback()`.

Not a super big deal, but it does mean that the outermost `setPresence()` call on L496 is called _after_ the batch, and not _included with the batch_. Not sure if there is a good reason to actively prevent nesting of batches? Because this seems like a useful pattern to support.

GuillaumeSalles: We throw for nested batches because I never took the time to implement it properly 😬 

It shouldn't be too hard to implement, we just need to be cautious about undo/redo and dispatching. When a batch is finished, we shouldn't send the operations over the wire or add them to the undo/redo stack if there is an outer batch in progress (while being cautious about exceptions coming from the callback).

We can probably squeeze that in 0.18. What do you think?

nvie: Ah cool! Also I just discovered this ticket: #416 👍  Can take a look tomorrow to see if it's easy to implement. If so, would love to squeeze it in.

stevenfabre:
The before/after video to show the performance improvement is :chefkiss: !

PR #474, opened 2022-09-12
Fix bug in whiteboard example

nvie:
While testing some of the 0.18 hooks, I found a tiny bug in the whiteboard example.

To see the bug:
1. Open https://liveblocks.io/examples/collaborative-whiteboard-advanced/nextjs
2. Notice the "Undo" button being disabled
3. Click on the empty canvas once
4. Notice the "Undo" button getting enabled

The cause of the bug is that clicking triggers an unselection, which gets put on the undo stack. This PR fixes that by only unselecting if there is an actual selection.

### DIFF
--- a/examples/nextjs-whiteboard-advanced/src/index.tsx
+++ b/examples/nextjs-whiteboard-advanced/src/index.tsx
@@ -291,9 +291,14 @@ function Canvas() {
     [canvasState, selection]
   );
 
-  const unselectLayers = useMutation(({ setMyPresence }) => {
-    setMyPresence({ selection: [] }, { addToHistory: true });
-  }, []);
+  const unselectLayers = useMutation(
+    ({ setMyPresence }) => {
+      if (selection.length > 0) {
+        setMyPresence({ selection: [] }, { addToHistory: true });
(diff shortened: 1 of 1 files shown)


examples/nextjs-whiteboard-advanced/src/index.tsx
```diff
@@ -291,9 +291,14 @@ function Canvas() {
...
-    setMyPresence({ selection: [] }, { addToHistory: true });
-  }, []);
+  const unselectLayers = useMutation(
+    ({ setMyPresence }) => {
+      if (selection.length > 0) {
+        setMyPresence({ selection: [] }, { addToHistory: true });
+      }
+    },
+    [selection]
+  );
```
nvie, line 301: Unrelated to this bug, but FYI today I'll work on exposing the current presence from within the `useMutation` hook, so we can remove the dependency on `selection` here for this, which will make this mutation a lot simpler to express. (See https://github.com/liveblocks/liveblocks/pull/384#discussion_r967249977)

marcbouchenoire (approved):
Good catch!

PR #475, opened 2022-09-12
Use sync external store hook in more places

nvie:
For consistency in the hooks definitions, there were more places where we can utilize the `useSyncExternalStore` semantics instead of rolling our own. This just makes all of our hooks slightly more similar in terms of their implementation.

### DIFF
--- a/packages/liveblocks-react/src/__tests__/index.test.tsx
+++ b/packages/liveblocks-react/src/__tests__/index.test.tsx
@@ -10,11 +10,15 @@ import { rest } from "msw";
 import { setupServer } from "msw/node";
 
 import {
+  useCanRedo,
+  useCanUndo,
+  useMutation,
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react/src/__tests__/index.test.tsx
```diff
@@ -414,3 +437,56 @@ describe("useStorage", () => {
...
+    expect(canUndo.result.current).toEqual(true);
+    expect(canRedo.result.current).toEqual(true);
+
+    // Undo the last one
+    act(() => undo.result.current());
+
+    expect(canUndo.result.current).toEqual(false);
+    expect(canRedo.result.current).toEqual(true);
+  });
+});
```
nvie, line 492: Also added some unit tests for these (which is how I discovered the bug from #474).

PR #476, opened 2022-09-12
Export all hooks under `suspense` key

nvie:
This PR duplicates all the hooks generated by the factory to also be included underneath the `suspense` subkey. This makes switching all hooks over to Suspense a no-brainer for users. Simply wrap all your exports in a `{ suspense: <exports> }` without having to think which exact hooks to put in there, vs which hooks won't need the Suspense wrapper.

The main downside of this change would be our own maintenance pain. Since JSDocs comments are read off of the type definition _literals_, we cannot pull any magic to DRY up the comments in both version of the hooks. However, since these hooks are never expected to behave differently (except of course for the Suspense part), I'm enforcing parity through a helper script that I added as a part to our CI lint checks.

This means that it will be a CI error if ever we export a new member in the "normal" bundle vs the "suspense" bundle. Also, if they aren't _documented_ the same in the JSdoc strings, that will be a CI issue.

This means that the copy/pasting to keep them in sync still requires manual maintenance, but at least we can be confident that it's impossible to forget about 😄

### DIFF
--- a/packages/liveblocks-react/src/factory.tsx
+++ b/packages/liveblocks-react/src/factory.tsx
@@ -718,50 +718,72 @@ export function createRoomContext<
   }
 
   return {
+    RoomContext,
     RoomProvider,
-    useBatch,
-    useBroadcastEvent,
-    useCanRedo,
-    useCanUndo,
-    useErrorListener,
-    useEventListener,
-    useHistory,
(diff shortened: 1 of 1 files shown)


CTNicholas (commented):
Awesome!

PR #478, opened 2022-09-12
Expose read-only Presence in Mutation contexts

nvie:
The mutation context is the first argument you get passed when using the `useMutation()` hook.

It currently gets passed `{ root, setMyPresence }`, but this PR changes that to `{ root, self, others, setMyPresence }`. This has quite a big practical benefit, namely that if your mutations rely on the current user's (or other user's) presence, you won't have to externally get a reference to those and keep invalidating your callback function.

```ts
const selection = useSelf(user => user.presence.selection);
const myCallback = useMutation(
  ({ root }) => {
    doSomething(root, selection);
  },
  [selection]
);
```

Now:

```ts
const myCallback = useMutation(
  ({ root, self }) => {
    doSomething(root, self.presence.selection);
  },
  []
);
```

This enables you to write much cleaner mutations. See f2623295591a1614437771c9fb2bad1d7ee75986 for an example of that. See also this [feedback](https://github.com/liveblocks/liveblocks/pull/384#discussion_r967249977).

### DIFF
--- a/packages/liveblocks-react/src/factory.tsx
+++ b/packages/liveblocks-react/src/factory.tsx
@@ -618,7 +618,7 @@ export function createRoomContext<
       context: MutationContext<TPresence, TStorage, TUserMeta>,
       ...args: any[]
     ) => any
-  >(callback: F, deps?: unknown[]): OmitFirstArg<F> {
+  >(callback: F, deps: readonly unknown[]): OmitFirstArg<F> {
     const room = useRoom();
     return React.useMemo(
       () => {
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react/src/factory.tsx
```diff
@@ -47,6 +47,49 @@ function getEmptyOthers() {
...
+    get self() {
+      const self = room.getSelf();
+      // NOTE: We could use room.isSelfAware() here to keep the check
+      // consistent with `others`, but we also want to refine the `null` case
+      // away here.
+      if (self === null) {
+        throw new Error(errmsg);
+      }
+      return self;
+    },
```
nvie, line 79: Made these keys lazy, so that if your mutation won't need `self` and doesn't access it, it will not recompute the immutable self value if it would be invalidated at that point.

GuillaumeSalles (approved):
Awesome! Thanks for the quick additions to the mutation context!

PR #483, opened 2022-09-13
Updated to-do list for 0.18

CTNicholas:
This is updated and ready to go out. It is part of the updated React to-do list tutorial, which will be released with 0.18.

This PR previously contained the whiteboard example https://github.com/liveblocks/liveblocks/pull/487, but it's been split as that will release later.

### DIFF
--- a/examples/react-todo-list/package.json
+++ b/examples/react-todo-list/package.json
@@ -9,8 +9,8 @@
     "eject": "react-scripts eject"
   },
   "dependencies": {
-    "@liveblocks/client": "^0.18.0-beta3",
-    "@liveblocks/react": "^0.18.0-beta3",
+    "@liveblocks/client": "^0.18.0-beta4",
+    "@liveblocks/react": "^0.18.0-beta4",
     "@testing-library/jest-dom": "^5.16.2",
     "@testing-library/react": "^12.1.3",
     "@testing-library/user-event": "^13.5.0",
--- a/examples/react-todo-list/package.json
+++ b/examples/react-todo-list/package.json
@@ -9,8 +9,8 @@
     "eject": "react-scripts eject"
   },
   "dependencies": {
-    "@liveblocks/client": "^0.17.11",
-    "@liveblocks/react": "^0.17.11",
+    "@liveblocks/client": "^0.18.0-beta3",
+    "@liveblocks/react": "^0.18.0-beta3",
     "@testing-library/jest-dom": "^5.16.2",
     "@testing-library/react": "^12.1.3",
     "@testing-library/user-event": "^13.5.0",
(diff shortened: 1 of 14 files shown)


examples/react-whiteboard/src/App.js
```diff
@@ -104,31 +108,19 @@ function Canvas({ shapes }) {
...
-              selectionColor={selectionColor}
-              transition={selectedShape !== shapeId}
             />
           );
         })}
       </div>
       <div className="toolbar">
-        <button onClick={insertRectangle}>Rectangle</button>
-        <button onClick={deleteRectangle} disabled={selectedShape == null}>
+        <button onClick={() => insertRectangle()}>Rectangle</button>
```
nvie, line 122: Any reason not to keep it at this? Not that it's super important, but `() => ...()` would undo the benefit of the memorized callback. In the case of the `selectedShape` on the line below, you don't have a choice though. Done for consistency/symmetry with that line?

```suggestion
        <button onClick={insertRectangle}>Rectangle</button>
```

examples/react-whiteboard/src/index.js
```diff
@@ -1,28 +1,12 @@
-import React from "react";
-import ReactDOM from "react-dom/client";
+import { createRoot } from "react-dom/client";
```
nvie, line 1: Thx for upgrading this example to use modern React primitives, at the same time 👍

nvie (approved):
Excellent work, @CTNicholas 👌  I love all of these updates! 😍

examples/react-whiteboard/src/App.js
CTNicholas: Not for symmetry, it's just me being overly cautious! As standard practice I don't pass parameters to functions that don't use any, in case the function changes later. This passes on a click event. If `insertRectangle` were changed to take an optional parameter, this would cause a bug.

You're totally right though, I didn't think about performance! What do you think?

nvie: I see! I didnt even think of it that way because I’d rely so heavily on TypeScript to catch bugs like that. It doesnt have enough of a performance gain to be important here, so I’m perfectly okay with your version. Was mostly commenting out of curiosity!

examples/react-todo-list/package.json
```diff
@@ -9,8 +9,8 @@
     "eject": "react-scripts eject"
   },
   "dependencies": {
-    "@liveblocks/client": "^0.17.11",
-    "@liveblocks/react": "^0.17.11",
+    "@liveblocks/client": "^0.18.0-beta3",
+    "@liveblocks/react": "^0.18.0-beta3",
```
nvie, line 13: I just published 0.18.0-beta4, which hopefully is the final version that we can launch as 0.18.0 next week. Could you upgrade once more? Sorry 'bout the hassle!

CTNicholas:
Updated to the newest version and using regular React Suspense now.

PR #485, opened 2022-09-13
react-native todo-list upgrade client and react packages 0.18

vinodsantharam:
Upgrade `@liveblocks/react` and `@liveblocks/client` to 0.18-beta-3. 
Really easy upgrades in the code. 
Also added `initialePresence` to `RoomProvider`. 

Tested with two iPhone simulators. "Someone is typing" and the "others.count" work as expected.

### DIFF
--- a/examples/react-native-todo-list/App.tsx
+++ b/examples/react-native-todo-list/App.tsx
@@ -9,19 +9,32 @@ import {
   KeyboardAvoidingView,
   Platform,
 } from "react-native";
-import { useOthers, useList, useUpdateMyPresence } from "./liveblocks.config";
+import {
+  useOthers,
+  useUpdateMyPresence,
+  useStorage,
+  useMutation,
+} from "./liveblocks.config";
 import Header from "./src/components/Header";
 import Todo from "./src/components/Todo";
 import TextInputWithButton from "./src/components/TextInputWithButton";
 import NoMoreTodos from "./src/components/NoMoreTodos";
 
 const App = () => {
-  const todos = useList("todos");
-  const others = useOthers();
+  const todos = useStorage((root) => root.todos);
+  const userCount = useOthers((others) => others.length);
   const updateMyPresence = useUpdateMyPresence();
-  const isSomeoneTyping = others
-    .toArray()
-    .some((user) => user.presence?.isTyping);
+  const isSomeoneTyping = useOthers((others) =>
+    others.some((user) => user.presence?.isTyping)
+  );
+
+  const addTodo = useMutation(({ root }, text) => {
+    root.get("todos").push({ text });
+  }, []);
+
+  const deleteTodo = useMutation(({ root }, index) => {
+    root.get("todos").delete(index);
+  }, []);
 
   if (todos === null) {
     return (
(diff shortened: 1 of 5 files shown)


examples/react-native-todo-list/App.tsx
```diff
@@ -9,19 +9,32 @@ import {
...
-  const todos = useList("todos");
-  const others = useOthers();
+  const todos = useStorage((root) => root.todos);
+  const userCount = useOthers((others) => others.length);
   const updateMyPresence = useUpdateMyPresence();
-  const isSomeoneTyping = others
-    .toArray()
-    .some((user) => user.presence?.isTyping);
+  const isSomeoneTyping = useOthers((others) =>
+    others.some((user) => user.presence?.isTyping)
```
nvie, line 28: Since it's now mandatory to set `initialPresence` in the RoomProvider, we will always know a user's presence value, so this uncertainty is a thing of the past.

```suggestion
    others.some((user) => user.presence.isTyping)
```

examples/react-native-todo-list/App.tsx
```diff
@@ -57,9 +70,7 @@ const App = () => {
               updateTypingStatus={(isTyping: boolean) =>
                 updateMyPresence({ isTyping: isTyping })
               }
-              handleOnSubmitEditing={(todo: string) =>
-                todos.push({ text: todo })
-              }
+              handleOnSubmitEditing={(todo: string) => addTodo(todo)}
```
nvie, line 73: Optional, but I think you can improve this example a little bit further. If you move the `isSomeoneTyping` variable and this `updateTypingStatus` callback out of the props of the `TextInputWithButton` component, and use our hooks there directly. This not only reduces necessary prop drilling, it's also better for rendering efficiency, since only the `TextInputWithButton` will have to be rerendered if someone starts/stops typing, instead of the entire `App` component. In general, I think it's helpful to push the use of our hooks down into the component tree as deep as possible as a best practice.

nvie (approved):
Nice work, @vinodsantharam — thanks for lending a hand getting these examples converted! 🙏

examples/react-native-todo-list/package.json
```diff
@@ -9,8 +9,8 @@
     "test": "jest"
   },
   "dependencies": {
-    "@liveblocks/client": "^0.17.11",
-    "@liveblocks/react": "^0.17.11",
+    "@liveblocks/client": "^0.18.0-beta3",
+    "@liveblocks/react": "^0.18.0-beta3",
```
nvie, line 13: I just published 0.18.0-beta4, which hopefully is the final version that we can launch as 0.18.0 next week. Could you upgrade once more? Sorry 'bout the hassle!

vinodsantharam: Done! No worries 👍

examples/react-native-todo-list/App.tsx
vinodsantharam: gotcha! fixed!

examples/react-native-todo-list/App.tsx
vinodsantharam: Thanks! you are 100% right! I was dealing in my head with the tradeoff performance vs code clarity. 
I decided to centralise all code related to Liveblocks at root component and let all child components  to be self responsible about their behaviours and UI.

nvie (approved):
Thanks, Vinod! 🙌

PR #495, opened 2022-09-15
feat(nextjs-live-cursors): update to 0.18

FlowFlorent:
- No code changes

### DIFF
--- a/examples/nextjs-live-cursors/package-lock.json
+++ b/examples/nextjs-live-cursors/package-lock.json
@@ -7,8 +7,8 @@
     "": {
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


nvie (approved):
Great! No changes are the best changes! I think there is one tiny change you could still add here and that is this line:

https://github.com/liveblocks/liveblocks/blob/abb3e57b1c3b0be9653ce21c0e219f504145c88a/examples/nextjs-live-cursors/pages/index.tsx#L66

The null check for the `.presence` field here is no longer necessary, so this could become:

```tsx
          if (presence.cursor == null) {
```

But that's very minor.

Thanks for helping to convert this example! 🙌

PR #499, opened 2022-09-16
Convert `nextjs-form` example to 0.18

nvie:


examples/nextjs-form/pages/index.tsx
```diff
@@ -31,13 +33,22 @@ export default function Example() {
...
-  if (!data) {
+  const updateName = useMutation(({ storage }, name: string) => {
+    storage.get("logo").set("name", name);
+  }, []);
+
+  const updateTheme = useMutation(({ storage }, theme: "light" | "dark") => {
+    storage.get("logo").set("theme", theme);
+  }, []);
+
+  if (!logo) {
```
FlowFlorent, line 51: I was also on it yesterday but forgot to communicate it, thanks for doing it! I was wondering if suspense could have been a good call here 🤔

nvie: Yeah, I considered it, but since this is the only null check in the entire example, I just decided to keep it as-is. I think it's fine.

FlowFlorent: Glad we have a similar reaction, proves 0.18 is straightforward to use!

PR #500, opened 2022-09-16
📖 Read Only Websockets: add `isReadOnly`

FlowFlorent:
### Description

Adds `isReadonly` from token scopes for self and gets `isReadonly` for other users

Blocked by: https://github.com/liveblocks/liveblocks-cloudflare/pull/216

Coming in the next PR: optimizing calls when connection `isReadonly` (all update storage calls should be noops)

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

#### Related issue(s)

Closes https://github.com/liveblocks/liveblocks.io/issues/378
Closes https://github.com/liveblocks/liveblocks.io/issues/438
Fixes https://github.com/liveblocks/liveblocks.io/issues/443
Fixes https://github.com/liveblocks/liveblocks.io/issues/445


packages/liveblocks-client/src/room.ts
```diff
@@ -892,7 +893,13 @@ function makeStateMachine<
     }
   }
 
+  function isReadonly(scopes: string[]) {
+    return scopes.includes("room:read");
+  }
+
   function authenticationSuccess(token: RoomAuthToken, socket: WebSocket) {
+    // Can separate read only and read/write tokens here
+    // Update addEventListener with onReadOnlyMessage
```
FlowFlorent, line 902: coming in a follow-up PR

packages/liveblocks-react/src/factory.tsx
```diff
@@ -445,32 +445,36 @@ export function createRoomContext<
     return [useMutableStorageRoot()];
   }
 
+  // Returns errors / warnings on read only
```
FlowFlorent, line 448: Not sure about this, perhaps just let it be a no-op

packages/liveblocks-client/src/OthersRef.ts
```diff
@@ -110,12 +111,14 @@ export class OthersRef<
   setConnection(
     connectionId: number,
     metaUserId: TUserMeta["id"],
-    metaUserInfo: TUserMeta["info"]
+    metaUserInfo: TUserMeta["info"],
+    metaIsReadonly: User<TPresence, TUserMeta>["isReadonly"]
```
nvie, line 115:
```suggestion
    metaIsReadonly: boolean
```

Since this isnt a complicated type, for readability it may help to see that it’s just a boolean.

packages/liveblocks-client/src/room.ts
```diff
@@ -982,7 +990,13 @@ function makeStateMachine<
     for (const key in message.users) {
       const user = message.users[key];
       const connectionId = Number(key);
-      state.others.setConnection(connectionId, user.id, user.info);
+      // TODO: How to set isReadonly here?
```
nvie, line 993: Looks like this TODO has been addressed? Think you can clean up the comment!

packages/liveblocks-client/src/room.ts
```diff
@@ -892,7 +893,13 @@ function makeStateMachine<
     }
   }
 
+  function isReadonly(scopes: string[]) {
+    return scopes.includes("room:read");
```
nvie, line 897: Here is a more philosophical question: should we consider a connection read _only_ only if they also don’t have write permissions?

If I have scopes `['room:write', 'room:read']` then I would not think of that as a readonly connection?

nvie: Perhaps?

```suggestion
    return scopes.includes("room:read") && !scopes.includes("room:write");
```

packages/liveblocks-client/src/room.ts
```diff
@@ -1003,15 +1017,19 @@ function makeStateMachine<
...
-      type: ClientMsgCode.UPDATE_PRESENCE,
-      data: state.me.current,
-      targetActor: message.actor,
-    });
+    state.others.setConnection(
+      message.actor,
+      message.id,
+      message.info,
+      isReadonly(message.scopes)
+    ),
```
nvie, line 1025: Typo?

```suggestion
    );
```

packages/liveblocks-client/src/types/ServerMsg.ts
```diff
@@ -114,6 +114,11 @@ export type UserJoinServerMsg<TUserMeta extends BaseUserMeta> = {
    * endpoint.
    */
   info: TUserMeta["info"];
+
+  /**
+   * Permissions that the user has in the Room.
+   */
+  scopes: string[];
```
nvie, line 121: Another philosophical question: should we expose the exact scopes to other clients? Or should the server translate this to “readonly” before sending it over the wire? I worry that exposing this could be a potential attack vector.

packages/liveblocks-client/src/types/index.ts
```diff
@@ -260,6 +260,11 @@ export type User<
    * The user presence.
    */
   readonly presence: TPresence;
+
+  /**
+   * True if the user can modify the room storage, false otherwise.
```
nvie, line 265:
```suggestion
   * False if the user can modify the room storage, true otherwise.
```

nvie (commented):
Left some minor and some more philosophical comments, but I think overall this is shaping up nicely! 👍

packages/liveblocks-client/src/types/ServerMsg.ts
FlowFlorent: We always verify from the token and the token cannot be forged so it's not a security risk to have a them public

Also, it'd be easy to infer anyone's scopes even if they were obfuscated since `isReadonly` matches `room:read` + `room:presence:write` in the documentation

packages/liveblocks-client/src/types/index.ts
FlowFlorent: 🤦 thank you

packages/liveblocks-client/src/types/ServerMsg.ts
ofoucherot: It's already public in the REST API, we are using scopes for managing a room permissions.
I think it's ok if our REST API + WS protocol is more low level/granular and that we hide this complexity in the client package. 
When we will release a SDK for the REST API we will do the same, hide the complexity.

nvie: So @ofoucherot, are you suggesting to switch to `isReadonly: boolean` here, instead of `scopes: string[]`?

ofoucherot: No, I think it's ok to keep scopes for the Websocket messages. There is no risk exposing that since we are already exposing scopes in the REST API.

I think it's ok if we only simplify/add abstraction in the client.

nvie: 👍

packages/liveblocks-client/src/room.ts
```diff
@@ -1064,38 +1085,47 @@ function makeStateMachine<
     batchUpdates(() => {
       for (const message of messages) {
         switch (message.type) {
+          // Read event
```
ofoucherot, line 1088: Do we really have a notion of "read" or "write" for messages received from the server?

ofoucherot (approved):
Small comment but lgtm! I'm waiting for the next PRs :).

nvie (approved):
I think this change is looking good, but I realized I may not have the full final picture of how scopes are intended to match to permissions _exactly_, see the comment I left. Are there implicit scopes?

@FlowFlorent Approved the PR, but would appreciate the clarification/confirmation 🙏 

Thanks! 🙌

packages/liveblocks-client/src/__tests__/AuthToken.test.ts
```diff
@@ -36,24 +36,20 @@ describe("isTokenExpired", () => {
...
-        "websocket:storage",
-        "room:read",
-        "room:write",
-      ],
+      actor: 87,
+      appId: "605a4fd31a36d5ea7a2e08f1",
+      exp: 1664570010,
+      iat: 1664566410,
+      roomId: "KXa6UclvraeGZNdXVz67Z",
+      scopes: ["room:write"],
```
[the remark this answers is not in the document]
nvie: I guess what I'm lacking is an understanding of which scopes are implicit. If you have `["room:write"]` scope, what permissions do you have for presence? Are `"room:read"` and `"room:presence:read"` implied? Or does `["room:write"]` explicitly disallow read permissions everywhere, because those scopes aren't in there?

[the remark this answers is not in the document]
FlowFlorent: Good point, if you're asking the question we need clarification somewhere.


- `room:write` = `room:storage:write` + `room:presence:write`
    and the same logic would apply for `room:read`

- `room:write` = `room:write` + `room:read`
    this is an implicit simplification so that users are not required to add `read` whenever something is writable since writing is a stronger permission than reading
    
The simplifications are made so that less permissions are required to be set by our users

Let me know if that makes sense 🙏

packages/liveblocks-client/src/room.ts
```diff
@@ -892,7 +893,17 @@ function makeStateMachine<
...
+    return (
+      scopes.includes("room:read") &&
+      scopes.includes("room:presence:write") &&
+      !scopes.includes("room:write")
+    );
+  }
+
   function authenticationSuccess(token: RoomAuthToken, socket: WebSocket) {
+    // Can separate read only and read/write tokens here
+    // Update addEventListener with onReadOnlyMessage
```
FlowFlorent, line 906:
```suggestion
```

packages/liveblocks-client/src/__tests__/AuthToken.test.ts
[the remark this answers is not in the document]
nvie: Aha! That's good to know.

So, are the following conclusions correct?

- `["room:storage:read"]` expands to `["room:storage:read"]`
- `["room:storage:write"]` expands to `["room:storage:read", "room:storage:write"]`
- `["room:presence:read"]` expands to `["room:presence:read"]`
- `["room:presence:write"]` expands to `["room:presence:read", "room:presence:write"]`
- `["room:write"]` expands to `["room:storage:read", "room:storage:write", "room:presence:read", "room:presence:write"]`
- `["room:read"]` expands to `["room:storage:read", "room:presence:read"]` (but no write permissions)
- `["room:write", "room:storage:read"]` expands to `["room:storage:read", "room:storage:write", "room:presence:read", "room:presence:write"]` (the explicit read perms are unnecessary?)

If so, what does "nothing" expand to?

- `[]` expands to `[]`? So no access whatsoever?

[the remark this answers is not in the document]
FlowFlorent: Exact! 

Except that we'll allow at the API level only two options for now: `room:write` or `room:read` + `room:presence:write`

`[]` will not be possible as basically if the user has no access, then it gets a 403 when trying to obtain a token.

But yes, `[]` would expand to `[]` if we were to emit a token (which we do not)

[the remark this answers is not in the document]
nvie: Amazing! That is great, and makes so much sense! Thanks for clarifying.

PR #508, opened 2022-09-22
Update all non-React examples to 0.18

nvie:
Since the changes in 0.18 are mostly focused on the React package, I think we can safely update these. To non-React examples, this upgrade should almost be a no-op. The only observable difference for these is that initial presence is now mandatory when entering the room, so I updated those.


marcbouchenoire (approved):
LGTM 🧹

PR #510, opened 2022-09-23
Deprecate websocket:read/write permissions

FlowFlorent:
### Description

Deprecates permissions:
- `websocket:presence`
- `websocket:storage`
- `rooms:write`
- `rooms:read`

### Related PR

**BLOCKED BY**: https://github.com/liveblocks/liveblocks-cloudflare/pull/207

### Ticket
Related to https://github.com/liveblocks/liveblocks.io/issues/435

### DIFF
--- a/packages/liveblocks-client/src/AuthToken.ts
+++ b/packages/liveblocks-client/src/AuthToken.ts
@@ -1,14 +1,7 @@
 import type { Json, JsonObject } from "./types";
 import { b64decode, isPlainObject, tryParseJson } from "./utils";
 
-export const SCOPES = [
-  "websocket:presence",
-  "websocket:storage",
-  "room:read",
-  "room:write",
-  "rooms:read",
-  "rooms:write",
-] as const;
+export const SCOPES = ["room:read", "room:write"] as const;
 
 export type Scope = typeof SCOPES[number];
 
--- a/packages/liveblocks-client/src/AuthToken.ts
+++ b/packages/liveblocks-client/src/AuthToken.ts
@@ -1,10 +1,6 @@
 import type { Json, JsonObject } from "./types";
 import { b64decode, isPlainObject, tryParseJson } from "./utils";
 
-export const SCOPES = ["room:read", "room:write"] as const;
-
-export type Scope = typeof SCOPES[number];
-
 export type AppOnlyAuthToken = {
   appId: string;
   roomId?: never; // Discriminating field for AuthToken type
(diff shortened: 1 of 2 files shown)


packages/liveblocks-client/src/AuthToken.ts
```diff
@@ -1,14 +1,7 @@
...
 
-export const SCOPES = [
-  "websocket:presence",
-  "websocket:storage",
-  "room:read",
-  "room:write",
-  "rooms:read",
-  "rooms:write",
-] as const;
+export const SCOPES = ["room:read", "room:write"] as const;
```
nvie, line 4: One caveat that may not be obvious now that I think if it, is that this constant isn't at all used by any of our client packages, but is instead consumed to [implement the scope decoder](https://github.com/liveblocks/liveblocks-cloudflare/blob/main/src/decoders/AuthToken.ts#L17-L20) **in the cloudflare repo** (!).

If this PR is merged, published to NPM, and the cloudflare repo upgraded to the latest version of our client, it will start throwing errors, aka rejecting incoming JWT tokens that include any of these removed scopes. That seems not the intent here.

This disconnect existed for historical/DRY reasons, but now that I see this, I think it's super unobvious. A better approach may be to cut this `SCOPES` constant and in the `isScope` helper function from the client, and just move it into the cloudflare repo directly, so this effect will be more obvious. The client packages don't need this constant anyway—it only adds to the weight of the published bundle at this point.

nvie, line 4: One caveat that may not be obvious now that I think if it, is that this constant isn't at all used by any of our client packages, but is instead consumed to [implement the scope decoder](https://github.com/liveblocks/liveblocks-cloudflare/blob/main/src/decoders/AuthToken.ts#L17-L20) **in the cloudflare repo** (!).

If this PR is merged, published to NPM, and the cloudflare repo upgraded to the latest version of our client, it will start throwing errors, aka rejecting incoming JWT tokens that include any of these removed scopes. That seems not the intent here.

This disconnect existed for historical/DRY reasons, but now that I see this, I think it's super unobvious. A better approach may be to cut this `SCOPES` constant and in the `isScope` helper function from the client, and just move it into the cloudflare repo directly, so this effect will be more obvious. The client packages don't need this constant anyway—it only adds to the weight of the published bundle at this point.

nvie: Scratch that. It's even better. This `SCOPES` constant, the `isScope` helper, and the `scope` decoder in the cloudflare repo all seems to be dead code, so let's just remove it all?

FlowFlorent: No code is best code, nice catch! Thanks

FlowFlorent: Associated PR :) https://github.com/liveblocks/liveblocks-cloudflare/pull/214

nvie (approved):
Thanks for cleaning this up! 🙌

PR #512, opened 2022-09-27
Fix confusing runtime error message in Zustand

nvie:
This fixes a confusing error message in Zustand (this came up in a [help request in Discord](https://discord.com/channels/913109211746009108/913158542565994606/1023968984569217024)).

### DIFF
--- a/packages/liveblocks-zustand/src/index.ts
+++ b/packages/liveblocks-zustand/src/index.ts
@@ -372,7 +372,7 @@ function updatePresence<
 ) {
   for (const key in presenceMapping) {
     if (typeof newState[key] === "function") {
-      throw mappingToFunctionIsNotAllowed("value");
+      throw mappingToFunctionIsNotAllowed(key);
     }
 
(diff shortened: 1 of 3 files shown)


packages/liveblocks-zustand/src/index.ts
```diff
@@ -372,7 +372,7 @@ function updatePresence<
 ) {
   for (const key in presenceMapping) {
     if (typeof newState[key] === "function") {
-      throw mappingToFunctionIsNotAllowed("value");
+      throw mappingToFunctionIsNotAllowed(key);
```
nvie, line 375: @GuillaumeSalles You authored this 7 months ago. Not sure if there is a subtle reason for this that I'm overlooking, but if I had to guess this was a TDD'ed version that wasn't not generalized after getting to green?

nvie: Going to assume this is the case, and including this with 0.18.3. Lmk if I happen to have overlooked something after all.

PR #513, opened 2022-09-27
Structural fix for the zombie child problem

### LINKED ISSUE #511: Investigate potential zombie child problem
Last week, we published a bugfix release 0.18.2 (see https://github.com/liveblocks/liveblocks/pull/506). The bug targeted by this issue might be an instance of the well-known zombie child problem in React. We should investigate a bit deeper and see if there is a more generic fix for this.

nvie:
This PR reverts the fix I made last week (#506), and instead fixes the zombie child problem structurally—not just for the `useOther` hook. In the `useOther` hook, the problem was most apparent because that function explicitly throws, but the problem was more broad than this.

Fixes #511.

As mentioned in https://github.com/liveblocks/liveblocks/issues/511#issuecomment-1258500696, this zombie child problem is only a problem in React<18, and not in React>=18. See that comment for an explanation of why this is a problem in React 17 and lower.

In this more structural solution, the RoomProvider can pass a reference to `unstable_batchedUpdates` (from ReactDOM or ReactNative), down to the client, which will then wrap all notification callbacks in a single "render batch". It looks like this:

```tsx
import { unstable_batchedUpdates } from "react-dom";  // or…
// import { unstable_batchedUpdates } from "react-native";

<RoomProvider
  id="my-room"
  unstable_batchedUpdates={unstable_batchedUpdates}  // 👈
/>
```

Including this line fixes the zombie child problem and is generally a good idea to use. In React>=18, this is not needed, but if you include it, it still works, as it's a no-op there.

## Downsides
While this solution may be the Proper/Robust/Structural Fix™, it has a few downsides that I don't like much 🤔

First of all, ideally we'd hide this implementation detail entirely. But we cannot "automatically" fix this under the hood, because `unstable_batchedUpdates` is a renderer-specific implementation detail. It either must come from ReactDOM or ReactNative (or whatever render you’re using). The `@liveblocks/react` package only directly depends on `react`, not on renderers like `react-dom` or `react-native`.

Secondly, this is an optional field, but ideally it should be set in all cases where users are on React 17 or lower. Ideally it would be mandatory to set in those cases. If we don't make it mandatory to configure, then users may use the non-batching version for a long time in production before stumbling on the stable props / zombie child issue, and will have a hard time debugging it. After that pain, they may learn about the existence of this fix, but it's a frustrating journey that can be prevented if this would be mandatory.

### DIFF
--- a/packages/liveblocks-client/src/client.ts
+++ b/packages/liveblocks-client/src/client.ts
@@ -21,6 +21,13 @@ type EnterOptions<
   // Enter options are just room initializers, plus an internal option
   RoomInitializers<TPresence, TStorage> & {
     /**
+     * HACK: This option should be passed in from React v17 or earlier if you
+     * want to avoid the zombie child problem. Not necessary in React v18 or
+     * later.
+     */
+    unstable_batchedUpdates?: (cb: () => void) => void;
+
+    /**
      * INTERNAL OPTION: Only used in a SSR context when you want an empty room
      * to make sure your react tree is rendered properly without connecting to
      * websocket
@@ -118,6 +125,8 @@ export function createClient(options: ClientOptions): Client {
         polyfills: clientOptions.polyfills,
         WebSocketPolyfill: clientOptions.WebSocketPolyfill, // Backward-compatible API for setting polyfills
         fetchPolyfill: clientOptions.fetchPolyfill, // Backward-compatible API for setting polyfills
+
+        unstable_batchedUpdates: options?.unstable_batchedUpdates,
 
         liveblocksServer:
           // TODO Patch this using public but marked internal fields?
--- a/packages/liveblocks-client/src/client.ts
+++ b/packages/liveblocks-client/src/client.ts
@@ -21,9 +21,13 @@ type EnterOptions<
   // Enter options are just room initializers, plus an internal option
   RoomInitializers<TPresence, TStorage> & {
     /**
-     * HACK: This option should be passed in from React v17 or earlier if you
-     * want to avoid the zombie child problem. Not necessary in React v18 or
-     * later.
+     * A custom callback that will get wrapped around all emitted
+     * notifications. Typically used in the context of a React 17 (or lower)
+     * application to pass in a reference to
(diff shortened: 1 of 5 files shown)


nvie:
Had a chat about this with @GuillaumeSalles, and here's my proposal to address the downsides above:

1. We keep the optional `unstable_batchedUpdates` field on the RoomProvider.
2. The RoomProvider will check `React.version` at runtime. If it's 17 or lower, then `unstable_batchedUpdates` should be set or it will throw an error in development mode. If 18 or higher, then `unstable_batchedUpdates` should not be set, and it will show a warning that it can be removed.

In the error message, we say something like:

> _"We noticed you're using React 17. There's a well-known problem with React 17 that is known as the "stable props / zombie child" problem, which can be avoided if you specify `unstable_batchedUpdates` at the RoomProvider level. [Link to more detailed docs]"_

packages/liveblocks-client/src/client.ts
```diff
@@ -20,6 +20,17 @@ type EnterOptions<
 > = Resolve<
   // Enter options are just room initializers, plus an internal option
   RoomInitializers<TPresence, TStorage> & {
+    /**
+     * A custom callback that will get wrapped around all emitted
+     * notifications. Typically used in the context of a React 17 (or lower)
+     * application to pass in a reference to
+     * a `ReactDOM.unstable_batchedUpdates` or
+     * `ReactNative.unstable_batchedUpdates` here, to announce to the React
+     * renderer that a series of updates can follow that should lead to
```
ofoucherot, line 29: I'm not sure I understand this sentence 🤔 .

nvie: Ugh, me neither. Let me rewrite it… 😂

nvie: @ofoucherot I rewrote the comment! Is this better?

ofoucherot: Yes 🙏

ofoucherot (approved):
🚀

PR #515, opened 2022-09-28
Fancy console badges

nvie:
Adds fancy/branded badges to our console logs:

Screen Shot 2022-09-28 at 22 11 02@2x

These can be used like normal console methods, simply by importing this module as a global:

```tsx
import * as console from "./fancy-console";

console.log('foo');
console.warn('foo');
console.error('foo');
```

---

In addition, there also exists a version that allows you to log out a title (printed in bold):

```tsx
console.errorBold('Some title', 'Lorem ipsum dolor sit amet');
```

Screen Shot 2022-09-28 at 22 16 46@2x

### DIFF
--- a/packages/liveblocks-client/.eslintrc.js
+++ b/packages/liveblocks-client/.eslintrc.js
@@ -1,3 +1,6 @@
+const rulesDirPlugin = require("eslint-plugin-rulesdir");
+rulesDirPlugin.RULES_DIR = "./rules";
+
 module.exports = {
   root: true,
   parser: "@typescript-eslint/parser",
@@ -6,6 +9,7 @@ module.exports = {
     "@typescript-eslint",
     "eslint-plugin-import",
     "eslint-plugin-simple-import-sort",
+    "eslint-plugin-rulesdir",
   ],
   extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
   rules: {
@@ -54,6 +58,7 @@ module.exports = {
     // -------------------------------
     // Custom syntax we want to forbid
     // -------------------------------
+    "rulesdir/console-must-be-fancy": "error",
     "no-restricted-syntax": [
       "error",
       {
--- a/packages/liveblocks-client/rules/console-must-be-fancy.js
+++ b/packages/liveblocks-client/rules/console-must-be-fancy.js
@@ -1,0 +1,26 @@
+module.exports = {
+  meta: {
+    messages: {
+      addMissingImport:
+        "Don't use `console` directly. To fix: `import * as console from './fancy-console'` in this module.",
+    },
+  },
+
+  create(context) {
+    let seenFancyImport = false;
+
+    return {
+      ImportDeclaration(node) {
+        if (node.source.value.includes("fancy-console")) {
+          seenFancyImport = true;
+        }
+      },
+
+      CallExpression(node) {
+        if (node.callee.object?.name === "console" && !seenFancyImport) {
+          context.report({ node: node.callee, messageId: "addMissingImport" });
+        }
+      },
+    };
+  },
+};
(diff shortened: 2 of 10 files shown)


packages/liveblocks-client/src/fancy-console.ts
```diff
@@ -0,0 +1,35 @@
...
+          badge,
+          bold,
+          message,
+          ...args
+        );
+}
+
+// export const logBold = wrapBold("log");
+// export const warnBold = wrapBold("warn");
+export const errorBold = wrapBold("error");
```
nvie, line 35: Not sure about this name. Should we perhaps use `errorWithTitle`?

FlowFlorent: I prefer `errorWithTitle` indeed!

marcbouchenoire: +1 on `***WithTitle` 👍

packages/liveblocks-client/src/fancy-console.ts
```diff
@@ -0,0 +1,35 @@
+const badge =
+  "background:radial-gradient(106.94% 108.33% at -10% -5%,#ff1aa3 0,#ff881a 100%);border-radius:6px;color:#fff;padding:2px 6px";
+const bold = "font-weight:600";
```
nvie, line 3: @pierrelevaillant If we want to change the style of this branded badge, we can simply change the CSS here in one place.

packages/liveblocks-client/src/fancy-console.ts
nvie: Thx! Then `xxxWithTItle` it is — f8a98e5493727c62a7a32639f46af2bc65cccbf9!

packages/liveblocks-client/.eslintrc.js
```diff
@@ -6,6 +9,7 @@ module.exports = {
     "@typescript-eslint",
     "eslint-plugin-import",
     "eslint-plugin-simple-import-sort",
+    "eslint-plugin-rulesdir",
```
nvie, line 12: The "rulesdir" ESLint plugin allows us to add custom ESLint rules with syntax rules we want to apply to this code base.

packages/liveblocks-client/rules/console-must-be-fancy.js
```diff
@@ -0,0 +1,26 @@
...
+      },
+
+      CallExpression(node) {
+        if (node.callee.object?.name === "console" && !seenFancyImport) {
+          context.report({ node: node.callee, messageId: "addMissingImport" });
+        }
+      },
+    };
+  },
+};
```
nvie, line 26: Here is a custom ESLint rule I wrote to ensure that we don't accidentally use `console.log()`. Every module that uses a `console.xxx` statement will be reminded about doing that fancy import automatically.

nvie: The reason this needed to be a full custom plugin is that it's a little more complicated rule than just disallowing a certain syntax. We cannot just disallow all console statements. "It depends" on whether the fancy console module has been imported or not. If it were simpler, I wouldn't have used a custom plugin, but instead disallowed the syntax in the .eslintrc file directly, using the `"no-restricted-syntax"` config.

packages/liveblocks-client/src/fancy-console.ts
CTNicholas: I quite like this styling:
```suggestion
const badge =
  "background:radial-gradient(106.94% 108.33% at -10% -5%,#ff1aa3 0,#ff881a 100%);border-radius:3px;color:#fff;padding:2px 5px;font-family:sans-serif;font-weight:600";
const bold = "font-weight:600";
```

Try it out by pasting this in your console:
```
console.log("%cLiveblocks%c Connection to websocket...", "background:radial-gradient(106.94% 108.33% at -10% -5%,#ff1aa3 0,#ff881a 100%);border-radius:3px;color:#fff;padding:2px 5px;font-family:sans-serif;font-weight:600", "")
```

nvie: Nice! Will use that 👍

PR #516, opened 2022-09-29
Upgrade examples to 0.18.3

nvie:


examples/react-native-todo-list/index.js
```diff
@@ -13,7 +13,16 @@ const initialStorage = () => ({
...
+    <RoomProvider
+      id={roomId}
+      initialStorage={initialStorage}
+      initialPresence={{ isTyping: false }}
+      unstable_batchedUpdates={
+        // NOTE: Only needed while this project is on React 17. After upgrading
+        // to React 18, this can be removed.
+        unstable_batchedUpdates
+      }
+    >
```
nvie, line 25: This is our only example that's still on React 17, it seems like. @vinodsantharam Could we upgrade this to React 18? I don't know enough about React Native or how to test this to do it myself.

PR #517, opened 2022-10-01
Expose option to connect or not to Liveblocks servers when entering a room

GuillaumeSalles:
## Why?

https://github.com/liveblocks/liveblocks/discussions/477

## @liveblocks/client

- Rename and expose `withoutConnecting` option on `client.enter` (previously internal)

## @liveblocks/react

- Add new option on `RoomProvider` to not connect to Liveblocks servers. By default, equals to `typeof window === "undefined"` to make sure the `RoomProvider` not connect by default on SSR.

## Tasks

- [x] Changelog
- [x] @nvie curious to have your opinion on the options naming
- [ ] Documentation - Will do it in @marcbouchenoire branch when we release 0.18.4 

## Out of scope

Expose methods on the room to be able to connect/disconnect after the RoomProvider has been rendered with `withoutInitiallyConnecting={true}`. Will be done in a subsequent releases because it requires more tests around storage initialization.


packages/liveblocks-react/src/types.ts
```diff
@@ -27,6 +27,15 @@ export type RoomProviderProps<
     id: string;
     children: React.ReactNode;
 
+    /**
+     * Weather or not the room should connect to Liveblocks servers
```
nvie, line 31: Small typo:

```suggestion
     * Whether or not the room should connect to Liveblocks servers
```

packages/liveblocks-react/src/types.ts
```diff
@@ -27,6 +27,15 @@ export type RoomProviderProps<
...
+     * when the RoomProvider is rendered.
+     *
+     * By default, the RoomProvider tries to connect to Liveblocks servers
+     * only on the client side.
+     */
+    withoutInitiallyConnecting?: boolean;
```
nvie, line 37: Two things about the naming:

1. I'd prefer to use positive phrasings for boolean fields. Rather than saying "without", could we flip the meaning by making it a positive expression? That'd mean that `initiallyConnect={true}` would be the default, and in order to _not_ connect, you'd have to use `initiallyConnect={false}` explicitly. We can handle that default easily using the `??` operator, i.e. `props.initiallyConnect ?? true`.
2. Could we ditch the "initial" part here, and make this a toggle'able property? I leave like this would make it possible to use this property control entering/leaving a room from a simple React prop for people. Do you see any downsides to that? And if we do, could we rename this to `enabled`?

Example:

```tsx
// Flip toggle to leave/enter room as you wish
<RoomProvider id="my-room" enabled={toggle}>
</RoomProvider>
```

packages/liveblocks-react/src/types.ts
GuillaumeSalles: 🙈 Ahah nice catch 🌤️

packages/liveblocks-react/src/types.ts
GuillaumeSalles: > I'd prefer to use positive phrasings for boolean fields. Rather than saying "without", could we flip the meaning by making it a positive expression? That'd mean that initiallyConnect={true} would be the default, and in order to not connect, you'd have to use initiallyConnect={false} explicitly. We can handle that default easily using the ?? operator, i.e. props.initiallyConnect ?? true.

Agree! Positive phrasing is better. What about `shouldInitiallyConnect`?

> Could we ditch the "initial" part here, and make this a toggle'able property? I feel like this would make it possible to use this property control entering/leaving a room from a simple React prop for people. Do you see any downsides to that? And if we do, could we rename this to enabled?

I think having a prop that looks like a state and triggers a connection or disconnection might introduce confusion with the connection state. It would be possible to have `enabled={true}` while the connection is closed and `enabled={false}` while the connection is opened (calling `useRoom().connect` directly).

nvie: > I think having a prop that looks like a state and triggers a connection or disconnection might introduce confusion with the connection state. It would be possible to have `enabled={true}` while the connection is closed and `enabled={false}` while the connection is opened (calling `useRoom().connect` directly).

Ah yes, that's a good point. Let's stick to initial behavior, then! 👍 

> `shouldInitiallyConnect`

I like it! 👍

PR #484, opened 2022-09-13
Added withoutConnecting flag to RoomProvider

kachar:
Added option `withoutConnecting` to skip the initial connection on mount of `<RoomProvider />` component.

```tsx
<RoomProvider id={roomId} withoutConnecting>
  <App />
</RoomProvider>
```

We're using the internal option to skip the connection in SSR mode:

https://github.com/liveblocks/liveblocks/blob/528ff8d3d0d589d0078dddf340c9b22fc655f4f2/packages/liveblocks-react/src/factory.tsx#L396-L402

https://github.com/liveblocks/liveblocks/blob/c70fed4818f243dc476fbb6a9d7286568f6e6486/packages/liveblocks-client/src/client.ts#L23-L28

Related to https://github.com/liveblocks/liveblocks/discussions/477

PR #519, opened 2022-10-07
TypeScript config cleanup pass

nvie:
Fixes #484.


ofoucherot (approved):
🚀 . I like that we now have one common ts config for all packages!

PR #529, opened 2022-10-13
Fix type of `BaseUserMeta`

nvie:
Removes `scopes` from `BaseUserMeta` and adds the property to the definition of `RoomStateServerMsg`.

Scopes should not be defined by clients as they are internal to Liveblocks.

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -907,7 +907,7 @@ function makeStateMachine<
     }
   }
 
-  function isStorageReadOnly(scopes: string[] = [RoomScope.Write]) {
+  function isStorageReadOnly(scopes: string[]) {
     return (
       scopes.includes(RoomScope.Read) &&
       scopes.includes(RoomScope.PresenceWrite) &&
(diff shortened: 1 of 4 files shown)


FlowFlorent (approved):

packages/liveblocks-core/src/room.ts
```diff
@@ -927,7 +927,7 @@ function makeStateMachine<
         id: token.actor,
         userInfo: token.info,
         userId: token.id,
-        isReadOnly: isReadOnly(token.scopes),
+        isReadOnly: isStorageReadOnly(token.scopes),
```
nvie, line 930: Rather than adding a default value, would this behavior be clearer if we avoid even calling this in the case of an `undefined` value?

```suggestion
        isReadOnly: token.scopes !== undefined ? isStorageReadOnly(token.scopes) : false,
```

nvie: The idea being that `user.scopes` never is going to be undefined, so we isolate that edge case, and only call the helper function in the expected case. We can later ensure that TS understands that it can never be undefined.

GuillaumeSalles (approved):

nvie:
Ohh, I like this fix!

PR #533, opened 2022-10-14
Simplify instructions for local linking

nvie:
I've removed a step from the instructions to link the examples locally. This seems to be enough already. I think I got confused here originally ¯\\\_(ツ)\_/¯!


marcbouchenoire (approved):
🧹💨

PR #534, opened 2022-10-14
Code organization cleanups in base packages

nvie:
Most importantly, this moves client-related code back into @liveblocks/client.


nvie:
Going to try this in a different way later.

PR #535, opened 2022-10-14
Increase speed of subsequent `tsc` builds

nvie:
Enable [`incremental`](https://www.typescriptlang.org/tsconfig#incremental) tsc builds for all of our projects. Makes running `tsc` slightly faster locally.


nvie:
Hmm 🤔 

Seems incompatible with `tsup`. If you want to use this, we'll just need to run `tsc --incremental` manually on the command line.

PR #536, opened 2022-10-20
Update Get started with Vue.js

guswillemann:
This PR updates the guide to integrate Liveblock with Vue.js

changes:

- fix `client.leave` call:
it was being called with 2 arguments instead of only with the `roomId`;

- add Vue3 Options Api section:
the `Vue.extend` method and the `destroyed` lifecycle hook are deprecated in Vue3;


CTNicholas:
Hey @guswillemann, thank you for the catching the mistakes, and updating us to Vue 3. I've updated the Options API example to your version, and added a Composition API example too. 

Thanks for contributing!

PR #544, opened 2022-10-27
Zustand internals prefactorings

nvie:
Small cleanups that were part of the Zustand V4 branch, but that be merged as-is, beforehand. Simplifies the actual PR.

### DIFF
--- a/e2e/next-sandbox/pages/zustand/index.tsx
+++ b/e2e/next-sandbox/pages/zustand/index.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect } from "react";
 import create from "zustand";
 
-import { middleware } from "@liveblocks/zustand";
+import { liveblocks } from "@liveblocks/zustand";
 import createLiveblocksClient from "../../utils/createClient";
 
 const client = createLiveblocksClient();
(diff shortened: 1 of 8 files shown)


packages/liveblocks-zustand/src/index.ts
```diff
@@ -368,16 +379,22 @@ function updatePresence<
...
-      const val = newState[key] as unknown as Json | undefined;
-      room.updatePresence({ [key]: val } as any);
+      const val = newState[key];
+      patch[key] = val;
     }
   }
+
+  if (Object.keys(patch).length > 0) {
+    room.updatePresence(patch);
+  }
```
nvie, line 397: This change will only call `room.updatePresence()` once per `updatePresence()` call, not once _for every key_.

e2e/next-sandbox/pages/zustand/index.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useEffect } from "react";
 import create from "zustand";
 
-import { middleware } from "@liveblocks/zustand";
+import { liveblocks } from "@liveblocks/zustand";
```
nvie, line 4: Moving to the Zustand ecosystem convention of naming the middlewares after your service/product/library. "Middleware" is too generic and every middle could be named that way. The `middleware` constant is still available as an export, but I've deprecated it in favor of `liveblocks`.

nvie:
Going to close this PR after all. This change shouldn't land on `main` because it would not be safe to release as 0.18.x. Will move these changes to the main Zustand v4 PR.

PR #547, opened 2022-11-01
Internal module structure cleanup

nvie:
## What does this PR do?
This PR only shuffles modules around, it contains no functional changes.

All the changes are for the `liveblocks-core` project. It removes a few (simple) cyclical dependencies, but not all of them (yet). The more complex part of that will require changes to the `AbstractCrdt` class hierarchy, so I'm saving that for a later PR.

The rationale for this change is threefold:
1. Making the code more navigable from a 10-foot view
2. Making further refactoring to the code easier
3. Removing complexity coming from cyclical dependencies

Just to give one example of why we want to achieve (2) is that it's been impossible to isolate and move the `client.ts` module outside of the `liveblocks-core` project, and into `liveblocks-client` as of yet due to the current complex dependencies. This PR brings that isolation a step closer.

## Nature of changes
All of these changes are made by doing basic mechanical transformations to make the module dependency graph more straightforward:
- Splitting apart modules that were doing multiple things
- Avoiding "re-export all" constructs in `index.ts` files (which makes direct dependencies invisible)
- Grouping together related modules into directories
- Defining import "rules"

The directories that are introduced here are intended to help understand the structure of the code a bit more at a glance, but by no means do we have to keep them as-is. If a better way of structuring exists, we can keep changing these.

## Dependency cruiser
I've been using a tool called [dependency cruiser](https://github.com/sverweij/dependency-cruiser) to help visualize and untangle the dependencies, detect cyclical dependencies, and help refactor the internals based on rules that we can define ourselves and programmatically enforce.

dependency-graph

I haven't made that tool an official dev dependency yet, but we could consider it, because it could enforce the following rules for us automatically in CI. By defining import rules and adhering to them, you effectively create a sort of "swimlanes" and the idea is that each module can import from modules in the same swimlane, or any swimlane to the right of it.

- The rightmost swimlane is `lib/`: only the most generic modules live here. These modules are characterized by:
  - They don't have any dependencies themselves
  - They are not Liveblocks-specific
  - They could, theoretically, be packages of their own
- Everything can import from `lib/`, but inside `lib/` you cannot import anything else (except other local `lib/` modules)
- Same goes for `protocol/`. Inside `protocol/`, only "local" imports or imports from `lib/` are allowed.
- Same for `types/`, same for `refs/`, etc.
- Finally, everything else can import from anything else.

In the image above you can see the high-level import graph where I collapsed all the modules into their directories. You can see all arrows pointing "to the right", except the cyclical import area that is `crdts/`, which we'll address later.

Compare this to the graph I started from, which includes a lot more arrows "pointing left":

dependency-graph


GuillaumeSalles (approved):
:chef-kiss:

The structure is so much better and the commit-by-commit approach made the review delightful!

I would be down to have dependency cruiser as a dev dep to enforce this structure at build time!

nvie:
> I would be down to have dependency cruiser as a dev dep to enforce this structure at build time!

I will add dependency cruiser in a follow-up PR, and make sure these rules are enforced by CI 👍

PR #548, opened 2022-11-02
Redux cleanups

nvie:
Small subset of #546, that is safe to be merged in after all. No observable changes from the outside.

### DIFF
--- a/packages/liveblocks-redux/src/index.ts
+++ b/packages/liveblocks-redux/src/index.ts
@@ -78,10 +78,10 @@ export type WithLiveblocks<
   TUserMeta extends BaseUserMeta
 > = TState & { readonly liveblocks: LiveblocksContext<TPresence, TUserMeta> };
 
-const internalEnhancer = <T>(options: {
+const internalEnhancer = <TState>(options: {
   client: Client;
-  storageMapping?: Mapping<T>;
-  presenceMapping?: Mapping<T>;
+  storageMapping?: Mapping<TState>;
+  presenceMapping?: Mapping<TState>;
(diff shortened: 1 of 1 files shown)


packages/liveblocks-redux/src/index.ts
```diff
@@ -289,13 +289,13 @@ const internalEnhancer = <T>(options: {
...
       }
 
-      function newDispatch(action: any, state: any) {
+      function newDispatch(action: any) {
         if (action.type === ACTION_TYPES.ENTER) {
           enterRoom(action.roomId);
         } else if (action.type === ACTION_TYPES.LEAVE) {
           leaveRoom(action.roomId);
         } else {
-          store.dispatch(action, state);
```
nvie, line 298: This seems to be a remnant of a past iteration? Going from the Redux type definitions, it looks like dispatch() only takes a single argument, not two. This may just have been passing implicit-undefined around in previous versions.

packages/liveblocks-redux/src/index.ts
```diff
@@ -183,7 +186,7 @@ const internalEnhancer = <T>(options: {
         }
       };
 
-      const store = createStore(newReducer, initialState, enhancer);
```
nvie, line 186: Same goes for the `createStore()` function, which only seems to be taking 2 arguments, not three?

nvie: Actually, I stand corrected. While the _type definitions_ don't mention it, the documentation does. I'm not going to touch this and revert this change.

PR #549, opened 2022-11-02
Dependency cruiser

nvie:
This PR adds dependency cruiser, a tool to enforce implicit import rules and make them explicit. It can also be used to generate a dependency graph for the project.

This PR adds to the `@liveblocks/core` package:

- The `.dependency-cruiser.js` config file, which contains all the rules to enforce
- Two new commands to visualize the dependency graph: `npm run showdeps` (detailed) and `npm run showdeps:high-level`
- A new `npm run test:deps` command, and uses it to enforce the rules in CI

The config file is large, but is, for the most part, the standard/generated config file. It contains generic rules that are good to enforce, so I kept those in. In addition, I added [our own custom rules](https://github.com/liveblocks/liveblocks/blob/fa844409925efd8229e57457fdb69463b8dffd7c/packages/liveblocks-core/.dependency-cruiser.js#L138-L186), which is the relevant section to review here.

To enable the fourth swimlane for the `crdts/` directory, I had to move the `liveblocks-helpers.ts` file into that `crdts/` directory. This means that the only remaining "cyclical imports" are now isolated in that directory.

Output of `npm run showdeps:high-level`:

dependency-graph

<details>
<summary>Output of <code>npm run showdeps</code> for comparison. (You can see the dependency cycle inside `crdts/` here.)</summary>

dependency-graph

</details>

As you can see, the module hierachy is very clear from the "high level" output. If you look at the dependencies in full detail, it's a bit more chaotic, but what's important is that you cannot "just import" stuff from anywhere: the hierarchy is and will remain maintained. It means that if you're studying code inside `crdts/`, you can know for sure that those modules don't know about the concept of a "room" or a "client".

Similarly, if you're studying code inside the `protocol/` directory, you can trust that none of that code is intertwined with other concepts, and the messages in there are just data definitions.

Hopefully this helps to bring a little bit more structure to navigating the code base, and to enforce that this setup will remain to be maintained over time, as the code base and the number of contributors grows.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -51,6 +51,9 @@ jobs:
       - name: Run type tests
         run: npm run test:types -- --filter ${{ matrix.pkg }}
 
+      - name: Run dependency tests
+        run: npm run test:deps -- --filter ${{ matrix.pkg }}
+
       - name: Run lint checks
         run: npm run lint -- --filter ${{ matrix.pkg }}
 
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -123,5 +123,5 @@ jobs:
         run: npm install
 
       - name: Run e2e client specs tests
-        run: npm run test-e2e
+        run: npm run test:e2e
         working-directory: packages/liveblocks-core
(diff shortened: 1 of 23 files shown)


vinodsantharam (approved):
@nvie this so much next level to me haha. I went through the documentation and tried to understand how it's applied in our case.
I can only say: Bravo!

PR #551, opened 2022-11-04
Improve code coverage

nvie:
This PR adds missing unit tests. This is an ongoing effort to increase code coverage in the core library. I'm starting with the `src/lib` directory here now, because those are the smallest building blocks.

This PR:

- Changes the jest test suite for `@liveblocks/core` to output a coverage report
- Locally, you can see the results when running the test script
- Also, it outputs a directory with an HTML report, in `coverage/` (see screenshots below)
- Adds missing unit tests for all `src/lib` modules, brining it to 100%

Other parts are still uncovered, and that is TBD later, incrementally.

### Before

Screen Shot 2022-11-04 at 14 15 03@2x

### After

Screen Shot 2022-11-04 at 14 15 22@2x

### DIFF
--- a/packages/liveblocks-core/jest.config.js
+++ b/packages/liveblocks-core/jest.config.js
@@ -1,1 +1,12 @@
-module.exports = require("@liveblocks/jest-config");
+/** @type {import('jest').Config} */
+
+const commonJestConfig = require("@liveblocks/jest-config");
+
+module.exports = {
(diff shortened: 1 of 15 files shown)


ofoucherot (approved):
So satisfying to see the code coverage goes all green! Thank you for that improvement.
Adding a check that verifies the % coverage change for each new PR will be nice.

FlowFlorent:
This is neat! Nice @nvie !

PR #555, opened 2022-11-05
[docs] Minor code fixes in Whiteboard React + Zustand tutorial

rayoverweij:
Hi, loved getting to know liveblocks tonight!

A couple of minor fixes in the Whiteboard React + Zustand tutorial code blocks:
* In the first version of `deleteShape()`, moved the null check for `selectedShape` one line up as it's necessary for the destructuring assignment to work
* In the second version of `deleteShape()`, added this same null check as it seems to have been mistakenly left out
* In both times the Delete Shape toolbar button is mentioned, changed the disabled check from a double-equals to triple-equals sign

Looking forward to experimenting a bit more!


nvie (approved):
Amazing 😍 ! Thanks for your contribution, @rayoverweij — we appreciate it! 🙏

PR #559, opened 2022-11-09
Better info in package.json files

CTNicholas:
This should help more people find us on NPM.


stevenfabre (approved):
Nice one, this will definitely help with people searching for those keywords.

@GuillaumeSalles cool if we get this merged?

nvie (approved):
Lovely! I'm about to roll 0.19.1 with a fix, so I can include these in that release! 🙌

PR #565, opened 2022-11-21
Internal refactoring of the `Pos` lib

nvie:
This PR refactors the internals of the `lib/position.ts` module (without observable change on the outside). This isn't related to an epic, but part of ongoing maintenance. I actually wanted to make a change to the internals of the LiveList module, but had to go down this rabbit hole first.

The result of this PR is:
- The `lib/position.ts` module is now documented (it's a beautiful data structure, so better if more people understand it)
- The "position" datastructure is now a **branded type**
- I've added a lot more **property-based tests** to get more confidence in/during the refactoring
- The internals have been simplified and made faster
- I've updated the use of this structure in LiveList, so that `parentPos` is now available next to `parentKey`

## Branded type
I've made `Pos` a "branded type". Every position is actually a simple string value, so you can pass it to any API that needs a string. But the opposite is not true. Some APIs require a valid Pos value, so they can make assumptions about this string, like the string will only contain legal chars, won't be empty, won't have trailing zeroes, etc. without having to double-check those at runtime over and over again. To achieve this, when we receive a string from the outside, you call `asPos(str)` on it, and it will guarantee the string is indeed a valid `Pos`.

Screen Shot 2023-04-20 at 09 15 31@2x

Branding is a TypeScript trick. It doesn't have a runtime presence.

## Property-based tests
Previously we had unit tests for statically hardcoded values. These remain the same on every test run. This PR adds property-based testing in addition to those tests, to keep enforcing certain properties always hold. These tests generate random input data, and you double check that assumptions hold. The more often these tests get run, the more confident you can be in the code.

For example, that `after(pos) > pos` is always true. Or `before(pos) < pos`, for _any_ value of `Pos` that you ever give it. Take a look at those tests (here's [just one example](https://github.com/liveblocks/liveblocks/blob/aa2f13e93759856a2da5be72cf6a2f99b5d7e7c7/packages/liveblocks-core/src/lib/__tests__/position.test.ts#L383), but there are many more).

## Simplifications
Currently, internally, the helpers are converting input strings to array-of-numers and back again. Using the property based tests as guard rails, I've refactored these internals to produce the same outputs, but without ever having to convert-to-numbers-then-back, and instead produce the results at once, using just direct string manipulations.

Some notes on perf:
- `makePosition()` runs (slightly) faster
- `comparePosition()` is now trivial: just a string comparison, so faster too
- `asPos()` is new and checks whether a string is valid using a regex. I've run a couple benchmarks, and all of them where (slightly) faster with the regex than calling `.charCodeAt()` in a loop like we did before.

Minor thing: In the process, I found a "bug", where "`before(.567)`" would produce "`.4`" before, instead of "`.5`". Not really a bug, but I took the opportunity to "fix" it anyway, while I was at it. (This makes it consistent like how `after(.567)` generates `.6`, not `.7`.)

## Future work
Now that this module _guarantees_ that Positions are alphabetically ordered, I can make the change that I originally wanted to make: to make insertions in LiveLists a faster (_O_(log _n_)) operation. PR coming soon after this is approved/merged.


nvie:
@jrowny This PR wasn't stale, it just never got reviewed in the last 5 months, and I forgot about it. I still think this is a good refactoring, and an overall simplification with better documentation of the purpose. Would love to get a review on this when you feel up for it. It's also a nice introduction to the fractional indexing to how we define element positions in LiveLists! 🙏

packages/liveblocks-core/src/lib/utils.ts
```diff
@@ -1,5 +1,8 @@
 import type { Json } from "./Json";
 
+declare const brand: unique symbol;
+export type Brand<T, TBrand extends string> = T & { [brand]: TBrand };
```
jrowny, line 4: Any reason not to use the ts-brand lib?

packages/liveblocks-core/src/lib/position.ts
```diff
@@ -1,133 +1,340 @@
-export const min = 32;
-export const max = 126;
+/**
+ * Positions, aka the Pos type, are efficient encodings of "positions" in
```
jrowny, line 2: this is excellent documentation! thank you!

packages/liveblocks-core/src/lib/position.ts
```diff
@@ -1,133 +1,340 @@
...
+ * A valid/verified "position" string. These values are used as "parentKey"s by
+ * LiveList children, and define their relative ordering.
+ */
+export type Pos = Brand<string, "Pos">;
 
-  // Insert at the end
-  else if (before !== undefined) {
-    return getNextPosition(before);
-  }
+const minCode = 32;
```
jrowny, line 55: sorry to nit pick, but this const won't change, so maybe rename to `MIN_CODE`, also maybe add a comment that this references an ASCII code. `MAX_CODE` as well.

packages/liveblocks-core/src/lib/position.ts
```diff
@@ -1,133 +1,340 @@
...
-  // No children
-  return pos([min + 1]);
+/**
+ * Given two positions, returns the position value that lies in the middle.
+ * When given only a high bound, computes the canonical position "before" it.
+ * When given only a low bound, computes the canonical position "after" it.
+ * When given no bounds at all, returns the "first" canonical position.
+ */
+function makePosition(x?: Pos, y?: Pos): Pos {
+  return x !== undefined && y !== undefined
```
jrowny, line 93: I personally don't mind nested ternaries, but I was once given the advice to never nest more than one because they become very hard to read.

jrowny (changes requested):
I love the documentation you added to the code, the refactor makes a lot more sense to me. My only requested changes are very minor.

packages/liveblocks-core/src/crdts/AbstractCrdt.ts
```diff
@@ -157,23 +188,6 @@ export abstract class AbstractCrdt {
     return this._parent;
   }
 
-  /** @internal */
-  get _parentNode(): LiveNode | null {
```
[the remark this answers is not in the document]
nvie: Seems like it. I just double-checked, and this seems to be dead code indeed.

packages/liveblocks-core/src/crdts/AbstractCrdt.ts
```diff
@@ -138,6 +152,23 @@ export abstract class AbstractCrdt {
     }
   }
 
+  /** @internal */
+  get parentPos(): Pos {
```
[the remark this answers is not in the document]
nvie: Ah yeah, probably better to do that. I don't think we are 100% consistent on this, but in most cases we definitely do. Fixed in 9dc8379b3153d0745c4d585c72bd3028d6bb1249 – thanks!

packages/liveblocks-core/src/lib/utils.ts
nvie: My main reason is that I didn't know of its existence. Now that I do know about it, not sure if this is worth the dependency, since it's only a one-liner.

packages/liveblocks-core/src/lib/position.ts
nvie: Agreed! Fixed in 3b9d843f1247aaf39e41384cea14bab22b541ddf.

packages/liveblocks-core/src/lib/position.ts
nvie: Completely agree, there is no reason to use this here. Fixed in a6b93bf3723128aa8865b375e7dc1a609cd13ec1.

nvie:
@jrowny Thanks for the review, I've addressed all the things you brought up, and I think this is good to go now. Can you give it another look?

jrowny (approved):

packages/liveblocks-core/src/lib/position.ts
```diff
@@ -1,133 +1,342 @@
...
+ *   - The string must not be the empty string
+ *
+ * The first range [\x20-\x7E] checks whether all characters are from the
+ * alphabet. Here, \x20 = 32 (min), and \x7E = 126 (max).
+ *
+ * In the second range, we check [\x21-\x7E], which excludes the " " char
+ * (\x20). This ensures there are no trailing zeroes, _and_ ensures the string
+ * is non-empty.
+ *
+ * Using this regex is faster than iterating over the string
```
GuillaumeSalles, line 285: Curious to see a benchmark, especially on positions with 1 or 2 characters 👀

packages/liveblocks-core/src/lib/position.ts
GuillaumeSalles: Awesome documentation!

packages/liveblocks-core/src/lib/position.ts
GuillaumeSalles: I like my constants when they SCREAM at me

GuillaumeSalles (approved):
Amazing work @nvie!

Can you confirm that all the previous tests are still passing against the new implementation? It's not obvious at a glance after the tests refactoring.

packages/liveblocks-core/src/lib/position.ts
nvie: You're right, I think the benchmark I used for this when I wrote this was not representative. Since most Pos strings will be super short, the regex indeed does not beat it, and the perf benefit only starts to kick in with strings from length 5 and up. I've replaced this with a manual check again in 80cbc4a2aaa98f2588765584830cd054115b65e6.

nvie:
> Can you confirm that all the previous tests are still passing against the new implementation? It's not obvious at a glance after the tests refactoring.

Here is the [proof](https://github.com/liveblocks/liveblocks/compare/pos...pos-proof-of-unit-tests-passing), constructed by reverting the test suite to exactly what it was on `main` (and patch a few small necessary things). These commits are of course not intended to be committed, but does that help?

Screen Shot 2023-04-25 at 10 08 57@2x

PR #567, opened 2022-11-23
Make all message/op types readonly

nvie:
In some places (in this repo, and in other private repos), we're mutating the fields of Ops and other messages in-place, which makes reasoning about the data and refactoring of those bits unnecessarily hard. This PR changes the data types to readonly, so we can ensure that in-place mutation won't happen.

Targeted for release **0.19.2**.

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -614,8 +614,8 @@ function isConnectionSelfAware(
 type HistoryOp<TPresence extends JsonObject> =
   | Op
   | {
-      type: "presence";
-      data: TPresence;
+      readonly type: "presence";
+      readonly data: TPresence;
     };
 
 type IdFactory = () => string;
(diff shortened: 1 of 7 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -1048,26 +1048,36 @@ function makeStateMachine<
...
   ): {
-    reverse: HistoryOp<TPresence>[];
+    // Input Ops can get opIds assigned during application.
+    ops: O[];
+    reverse: O[];
     updates: {
       storageUpdates: Map<string, StorageUpdate>;
       presence: boolean;
     };
   } {
```
nvie, line 1062: Apply will now output the (potentially) modified ops array, and won't mutate its input in-place.

Ops originating from undo/redo stacks won't have an `opId` assigned yet.

Before this PR:

```ts
apply(ops, …)
for (const op of ops) {
  // op _will_ have opId set here!
}
```

This led to a few surprises when refactoring this code. This crucial side effect was a bit hidden here and I didn't realize this was happening at first.

After this PR:

```ts
result = applyOps(ops, …)
for (const op of ops) {
  // op.opId might be undefined here! Input array won't be mutated!
}

// Instead, use the _result_ of applyOps()
for (const op of result.ops) {
  // op.opId is set here!
}
```

GuillaumeSalles (approved):

PR #577, opened 2022-12-05
room.hasPendingStorageModifications

GuillaumeSalles:
Until we persist offline operations to IndexedDB, we want to let the developer know if any storage modifications have not been acknowledged by the server.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -1276,6 +1276,45 @@ describe("room", () => {
         },
       ]);
     });
+
+    test("hasPendingStorageModifications", async () => {
+      const { storage, assert, machine, refStorage, reconnect, ws } =
+        await prepareStorageTest<{ x: number }>(
+          [createSerializedObject("0:0", { x: 0 })],
+          1
+        );
+
+      assert({ x: 0 });
+
+      expect(machine.hasPendingStorageModifications()).toBe(false);
+
+      ws.closeFromBackend(
+        new CloseEvent("close", {
+          code: WebsocketCloseCodes.CLOSE_ABNORMAL,
+          wasClean: false,
+        })
+      );
+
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -480,6 +480,11 @@ export type Room<
    */
   batch<T>(fn: () => T): T;
 
+  /**
+   * Whether or not the room has storage modifications that have not been acknowledged by the server.
+   */
+  hasPendingStorageModifications(): boolean;
```
nvie, line 486: Do we want to document these in our public documentation, too?

packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -1276,6 +1276,45 @@ describe("room", () => {
...
+      const newInitStorage: IdTuple<SerializedCrdt>[] = [
+        createSerializedObject("0:0", { x: 0 }),
+      ];
+
+      reconnect(2, newInitStorage);
+
+      assert({
+        x: 1,
+      });
+      expect(machine.hasPendingStorageModifications()).toBe(false);
```
nvie, line 1316: A little surprised that `reconnect()` is synchronous so that this even works. I would expect we'd have to `await` it to be able to send the offlineOps?

nvie (approved):
Approved if this passes, but one thing was a little surprising to me. Can you detail?

packages/liveblocks-core/src/room.ts
GuillaumeSalles: 👌

packages/liveblocks-core/src/__tests__/room.test.ts
GuillaumeSalles: In this context, `reconnect` is a synchronous test utility. We send the initial storage after reconnecting, and the client sends the diff synchronously.

PR #586, opened 2022-12-08
Use CredentialsProvider instead of custom cookie based auth

GuillaumeSalles:
Instead of building our own cookie based auth, we can use [CredentialsProvider](https://next-auth.js.org/configuration/providers/credentials) to simulate the existing behavior. The benefits are:

- Remove `cookies-next` dependency
- All the demo authentication is in 2 files so easier to remove
- No more wrapping of `next-auth`

TODO:
- [x] Documentation
- [x] Comments

@CTNicholas: curious to know if you considered this approach. Do you see any downsides to doing the demo auth like this?


starter-kits/nextjs-starter-kit/layouts/Authentication/Authentication.tsx
```diff
@@ -1,15 +1,14 @@
...
-import { signIn } from "../../lib/client";
-import { AUTHENTICATION_DEMO_MODE } from "../../liveblocks.config";
 import { Button } from "../../primitives/Button";
 import { Select } from "../../primitives/Select";
 import { getProviders } from "next-auth/react";
 import styles from "./Authentication.module.css";
 
 interface Props extends ComponentProps<"div"> {
-  providers?: ReturnType<typeof getProviders>;
+  providers?: Awaited<ReturnType<typeof getProviders>>;
```
GuillaumeSalles, line 11: This type was not accurate, so I fixed it while refactoring this code.

CTNicholas:
Hey thanks for the PR! I didn't add `CredentialsProvider` because it wasn't working for me (this branch doesn't work for me either), and I didn't have too much time to debug before the deadline. I've just spent a while looking into it now, and I think it's an issue with a package NextAuth uses to generate a temporary secret in dev mode, it might not be compatible with Windows...

...but if I add a `secret` to `authOptions` in `[...nextauth].ts` it starts working! 👍 We'll switch to `CredentialsProvider` now. I'll make the installer generate a random secret for you, and add it as an env var.

CTNicholas:
Updated the guide 👍

PR #587, opened 2022-12-09
update RoomAuthToken type

ofoucherot:
Add `groupIds` and `anonymousId` to type `RoomAuthToken`.

### DIFF
--- a/packages/liveblocks-core/src/protocol/AuthToken.ts
+++ b/packages/liveblocks-core/src/protocol/AuthToken.ts
@@ -21,11 +21,9 @@ export type RoomAuthToken = {
   maxConnectionsPerRoom?: number;
 
   // Extra payload as defined by the customer's own authorization
-  id?: string;
-  anonymousId?: string;
   info?: Json;
   groupIds?: string[];
-};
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/protocol/AuthToken.ts
```diff
@@ -22,7 +22,9 @@ export type RoomAuthToken = {
 
   // Extra payload as defined by the customer's own authorization
   id?: string;
+  anonymousId?: string;
```
nvie, line 25: Will either  `id` or  `anonymousId` always be set, and the other always be empty (like an XOR)? If so, it'd be helpful to codify that knowledge right in the type definition with something like:

```ts
export type RoomAuthToken = {
  ...
} & (
  | { id: string; anonymousId?: never }
  | { id?: never; anonymousId: string }
);
```

That way, TypeScript can be aware of the relation between these fields, and for example help us understand that if one is missing, the other field will be present.

nvie (approved):

packages/liveblocks-core/src/protocol/AuthToken.ts
ofoucherot: Yes it's a XOR, I updated the type. thank you!

PR #588, opened 2022-12-09
feat(auth): ensure public auth sets cookie

FlowFlorent:
### Description

- Adds `credentials: "include"` to ensure cookie is set from API call

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -2444,6 +2444,7 @@ async function fetchAuthEndpoint(
     headers: {
       "Content-Type": "application/json",
     },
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -2444,6 +2444,7 @@ async function fetchAuthEndpoint(
     headers: {
       "Content-Type": "application/json",
     },
+    credentials: "include",
```
FlowFlorent, line 2447:
```suggestion
    // necessary for cookies
    credentials: "include",
```

ofoucherot (approved):

nvie (approved):

PR #592, opened 2022-12-18
0.19.3-beta

GuillaumeSalles:
## room.getStorageStatus()

Get the storage status.
- `not-loaded`: Initial state when entering the room.
- `loading`: Once the storage has been requested via room.getStorage().
- `synchronizing`: When some local updates have not been acknowledged by Liveblocks servers.
- `synchronized`: Storage is in sync with Liveblocks servers.

## room.subscribe("storage-status", status => { })

Subscribe to storage status changes.

Returns an unsubscribe function.

```typescript
room.subscribe("storage-status", (status) => {
  switch(status) {
     case "not-loaded":
       break;
     case "loading":
       break;
     case "synchronizing":
       break;
     case "synchronized":
       break;
     default:
       break;
  }
});
```

## Room.reconnect

Close the room connection and try to reconnect.

## Internal

- Remove precommit


packages/liveblocks-core/src/room.ts
```diff
@@ -1837,9 +1873,15 @@ function makeStateMachine<
...
 
     if (storageOps.length > 0) {
+      const hasAlreadyPendingModifications = hasPendingStorageModifications();
       storageOps.forEach((op) => {
         state.offlineOperations.set(nn(op.opId), op);
       });
+      if (!hasAlreadyPendingModifications) {
+        eventHub.pendingStorageModifications.notify({
+          hasPendingStorageModifications: true,
+        });
```
nvie, line 1883: @GuillaumeSalles EventSources don't require a payload for an event. If in this case the payload isn't useful (i.e. always true), you may as well just not send a payload at all and just make it a `void` event source, and call `.notify()` without arguments. Nothing wrong with that. Or would it be useful to send an event whenever this boolean value "flips", i.e. also send an event when it changes to `false`?

packages/liveblocks-core/src/room.ts
```diff
@@ -353,9 +358,23 @@ export type Room<
...
+     *
+     * @returns Unsubscribe function.
+     *
+     * @example
+     * room.subscribe("pending-storage-modifications", ({ hasPendingStorageModifications }) => {
+     *   // Do something
+     * });
+     */
+    (
+      type: "pending-storage-modifications",
```
nvie, line 375: We call these changes "storage updates" everywhere else. Should we stick with that nomenclature, i.e. rename this to `pending-storage-updates`? Also, because the meaning of this event is captured in a boolean value (has or has-not), should we reflect that in the name? For example: add a "has" or "is" to the name, like `has-pending-storage-updates`?

Going a step further even, could we rename this event to `is-synchronized` instead? To me, it's about avoiding double-negation in concepts, compare the "happy paths" `has-pending-storage-updates? false!` vs `is-synchronized? true!`. We're mostly interested in acting on the event when the system is in-sync again, after all.

nvie: Just adding my $0.02 here now, because you only get one free shot when picking new names in an API.

packages/liveblocks-core/src/room.ts
```diff
@@ -510,7 +529,8 @@ export function isRoomEventName(value: string): value is RoomEventName {
     value === "event" ||
     value === "error" ||
     value === "connection" ||
-    value === "history"
+    value === "history" ||
+    value === "pending-storage-modifications"
```
nvie, line 533: Btw, when I introduced the `EventSource` and event "hub" abstraction (as part of the 0.18 release), I definitely intended those to be public APIs that developers could use directly. Because I didn't want to call too much attention to it as part of 0.18, I kept the `Room.subscribe()` API intact in a backward-compatible way, but internally it's just a wrapper around those new (typed) event hub APIs.

I had in mind that the event hub might one day become the primary/stable API for subscribing to room events, and do it in a way that's somewhat easier to maintain (and to type with TypeScript, too). The various overloads of `Room.subscribe()` have historically been a complicated part of the client code base.

Now that we're extending the system with a new event, do we want to consider only adding this new event to the event hub only, and not "backport" it to the `Room.subscribe()` API, i.e. stop maintaining it and eventually at one point deprecate it?

The main rationale for the change was to avoid the various overloads of `room.subscribe()` in favor of a more direct subscription model. As a user, you can subscribe via `room.events.XXX.subscribe(...)` — in this case `room.events.pendingStorageModifications.subscribe(...)`.

Would you mind using that API directly, starting now?

packages/liveblocks-core/src/room.ts
GuillaumeSalles: I like `is-synchronized`! It's way better. Do you think we should add the word storage to make it clear we're talking about the storage only? `is-storage-synchronized`?

Or maybe, we should go a step further and have a non boolean value? `storage-status` so we can support 'synchronized' / `synchronizing` / `offline`? 🤔

packages/liveblocks-core/src/room.ts
GuillaumeSalles: The events hub is a way better API than `subscribe` for the reasons you're mentioning and will become the public API at some points. However, I want to do it properly and not have events that are supported only using a single pattern until we explain how to migrate and why we're recommending this API and we adapt our examples. I don't want to avoid explaining to users why this event is only supported using this pattern until we're ready to support the events hub officially.

packages/liveblocks-core/src/room.ts
nvie: Yeah, `is-storage-synchronized` might be better! Just to make it explicit _what_ is synced, agreed.

I have another alternative that I just thought of… if you look at the event sources we currently have defined, here:

https://github.com/liveblocks/liveblocks/blob/75419a17e5c667d61be041dd8c5901fb44998c6c/packages/liveblocks-core/src/room.ts#L599-L612

Look at the `storageDidLoad` event. This is the event without payload that only fires once during the lifetime of a room. What would you think of following this naming convention and change this setup to `storageDidSync` instead? When we use that name, we can remove the payload from the event, and just interpret each event as an indicator that storage just emptied its pending Ops buffer. Of course, there would not be a way to tell (event-wise) when it gets out of sync again, but I don't think we're interested in that event anyway?

GuillaumeSalles: Updated the implementation. Far from ideal but the external API is as we discussed. Let's rework our internal a bit later.

packages/liveblocks-core/src/room.ts
```diff
@@ -2160,6 +2193,36 @@ function makeStateMachine<
...
+   * Currently okay because we only have 4 callers and shielded by tests.
+   */
+  let _lastStorageStatus = getStorageStatus();
+  function notifyStorageStatus() {
+    const storageStatus = getStorageStatus();
+    if (_lastStorageStatus !== storageStatus) {
+      _lastStorageStatus = storageStatus;
+      eventHub.storageStatus.notify(storageStatus);
+    }
+  }
```
nvie, line 2224: The comment is 👌 I think this is fine!

packages/liveblocks-core/src/room.ts
```diff
@@ -2160,6 +2193,36 @@ function makeStateMachine<
...
+        ? "synchronized"
+        : "synchronizing";
+    }
+
+    if (_getInitialStatePromise !== null) {
+      return "loading";
+    }
+
+    return "not-loaded";
+  }
```
nvie, line 2208: Small readability suggestion: eat all the "edge cases" first at the top, and move the happy path down:

```suggestion
  function getStorageStatus(): StorageStatus {
    if (_getInitialStatePromise === null) {
      return "not-loaded";
    }
    
    if (state.root === undefined) {
      return "loading";
    }
    
    return state.offlineOperations.size === 0
      ? "synchronized"
      : "synchronizing";
  }
```

Another nice side benefit is that it subconciously reflects the order of the state transitions that way, too.

nvie (approved):
Nice! Love this new event 👍

packages/liveblocks-core/src/room.ts
GuillaumeSalles: Yep! Way cleaner!

CHANGELOG.md
```diff
@@ -1,3 +1,42 @@
...
+## Room.subscribe("storage-status", status => { })
+
+Subscribe to storage status changes.
+
+Returns an unsubscribe function.
+
+```typescript
+room.subscribe("storage-status", (status) => {
+  switch(status) {
+     case "not-loaded":
```
taylorpwhipp, line 22: just a question about best practices- should this be "not-loaded" or empty?

nvie (approved):

CHANGELOG.md
nvie: We went with `not-loaded` here to make it clear that the initial load for Storage hasn't even begun yet. Client that don't use Storage, for example, will always remain in this state. Once you successfully loaded the initial Storage state, it could be empty. That's why we wanted to avoid the term "empty" here. A loaded, but empty, state is still perfectly valid and would be "synchronized".

PR #620, opened 2023-01-12
chore(core): add wip failing test for nested storage updates

BitPhinix:


BitPhinix:
@GuillaumeSalles, @nvie I'm not entirely sure but I think I stumbled upon another potential bug while fixing the storage update related one. This PR just contains the minimal reproduction case for now, will investigate a bit more to see if I can figure out what if happening here.

BitPhinix:
It turns out it was just an issue with the way the `prepareStorageUpdateTest` setup the clients (they both had a connectionId of 0, so they clashed with the existing items).

PR #622, opened 2023-01-12
Fix prepareStorageUpdateTest helper room setup

BitPhinix:
Fixes the underlying issue which caused my confusion in https://github.com/liveblocks/liveblocks/pull/620.

We initialize both clients with an actor id of `0` + the room storage with events with an actor id of `0`, so inserting new LiveNodes inside tests using the helper will cause unexpected issues. Seems like it was just luck this didn't happen until now.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/_utils.ts
+++ b/packages/liveblocks-core/src/__tests__/_utils.ts
@@ -451,14 +451,14 @@ export function prepareStorageUpdateTest<
 ): () => Promise<void> {
   return async () => {
     const { storage: refStorage, machine: refMachine } =
-      await prepareRoomWithStorage(items, 1);
+      await prepareRoomWithStorage(items, -1);
 
     const { storage, machine } = await prepareRoomWithStorage<
       TPresence,
       TStorage,
       TUserMeta,
       TRoomEvent
-    >(items, 0, (messages) => {
+    >(items, 1, (messages) => {
       for (const message of messages) {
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/__tests__/_utils.ts
```diff
@@ -451,14 +451,14 @@ export function prepareStorageUpdateTest<
...
-      await prepareRoomWithStorage(items, 1);
+      await prepareRoomWithStorage(items, -1);
 
     const { storage, machine } = await prepareRoomWithStorage<
       TPresence,
       TStorage,
       TUserMeta,
       TRoomEvent
-    >(items, 0, (messages) => {
+    >(items, 1, (messages) => {
```
nvie, line 461: Perhaps use `-2` here, so that even if we write tests with more than one actor, we'll not run into the same problem in the future?

nvie (approved):
Amazing debugging + fix, @BitPhinix! 🙌

Btw, I'd love to include some minimal version of your unit test from #620 in this PR, so the bug is demonstrated & fixed!

BitPhinix:
> Btw, I'd love to include some minimal version of your unit test from https://github.com/liveblocks/liveblocks/pull/620 in this PR, so the bug is demonstrated & fixed!

Not sure if it's worth it since that would be a "duplicate" test to ensure the test helper is working, but I don't have that strong of an opinion about it, so if you prefer it, I can do it.

PR #623, opened 2023-01-12
Fix/nested update storage update handling

BitPhinix:
Based upon https://github.com/liveblocks/liveblocks/pull/622 (needed that helper to work).

Seems like there are two small error in the way we handle nested changes inside `applyOps`:
1. Because we are adding `applyOpResult.modified.node._id` to the `createdNodeIds` we are storing the parent of the inserted node, not the inserted node itself. => we are ignoring any further changes to that parent.
2. We should not be checking if `applyOpResult.modified.node.parent.type` was inserted because that is the parent of the node we are trying to modify. We should check if the node we are trying to modify is a node that was inserted instead (so `applyOpResult.modified.node.id`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+
+# v0.19.7
+
+Fix nested storage event handling issue.
+
 # v0.19.6
 
 Support authentication with cookies.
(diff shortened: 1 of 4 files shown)


nvie (approved):
Again, amazing find & fix, @BitPhinix! 😍

Would love to get @GuillaumeSalles's approval just to be sure there is no caveat that I'm overlooking, but this fix absolutely make sense.

Feel free to also update the `CHANGELOG` file, so we can publish this as a fix once it's merged! 🙌

GuillaumeSalles (approved):
Nice catch @BitPhinix! So impressed that you went this deep on your first contribution to the core!

I added an extra check to your test to make sure an immutable state (used with zustand / redux) would be able to be patched with the generated updates. I was mostly to wrap my head around it.

PR #624, opened 2023-01-12
Update brand, assets, and docs

pierrelevaillant:
This refreshes Liveblocks README assets (_header, wordmark, examples thumbnails..._).


stevenfabre (commented):
Looks great overall!

Small feedback:
- Why have you moved the logo up in `header-wordmark-dark` and `header-wordmark-light`? It looks slightly off-center vertically, but I think we shouldn't even align it in the center vertically. The reason why it's close to the bottom is so that it feels close to the badges. I'd suggest keeping that.

marcbouchenoire (changes requested):
I'd suggest renaming the branch and PR (to `Update brand` maybe?) since it includes docs fixes too, but it's good to have only one PR here!

I have the same feedback as Steven, all animated headers should keep the same alignment (you can even see the cursors passing above the content when it's in the center like that).

marcbouchenoire (changes requested):
The logo in `header-light.svg` and `header-dark.svg` also needs to move down in the same way.

marcbouchenoire (approved):
👌

PR #627, opened 2023-01-15
New pricing and limits

stevenfabre:
Documentation guides associated with the new Liveblocks pricing and limits.
Changes include:
- Updated limits page
- New Fair Use Policy page that will be linked from the pricing page
- Update navigation to include the Fair Use Policy under "Limits"

You'll notice that I'm not using the `<Limits />` components available in `liveblocks.io` anymore. Should we update that component there to include all the limits we need? That way we can make sure the limits come from a single source of truth. @ofoucherot @GuillaumeSalles 

Thanks! ✌️

### DIFF
--- a/docs/pages/guides/limits.mdx
+++ b/docs/pages/guides/limits.mdx
@@ -8,14 +8,13 @@ meta:
 
 ## General limits
 
-|                                   | Starter     | Pro           | Startup          | Enterprise |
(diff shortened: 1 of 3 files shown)


FlowFlorent:
Should we have a section for legacy limits? That we'll deprecate in x months

ofoucherot (approved):

PR #633, opened 2023-01-19
Upgrade nextjs-spreadsheet-advanced example to 0.19.7

nvie:
This upgrades our spreadsheet example to technically be on 0.19.7. It's still not using the better new APIs introduced with 0.18 unfortunately, but we can fix that later. At least the app is now inspectable with devtools.


marcbouchenoire (approved):
I couldn't reproduce the issue I had before.

Issue #628, opened 2023-01-15
[BUG] Undo after a LiveList `set` results in delete/insert events


plamzd:
**Describe the bug**

We are trying to build multiplayer undo/redo into a collaborative text editor. The editor is, in essence, a sequence of paragraphs (only text) which we store in a LiveList. Whenever a user types some letter into a paragraph the LiveList `set` method is called for the respective element. The issue we are experiencing happens when we invoke `room.history.undo()` too soon after `set` for the same client (e.g. the user clicks ctrl-Z very quickly after typing) - in such cases the local client's room subscription receives 3 updates - a `set` update (which is expected, this is the reverse of our `set`) and then a `delete` and an `insert` for the respective live list element (FWIW, the remote client subscriptions correctly receive only the `set`).

For testing purposes I've added a 1 second delay after the `set` and the `undo` and this only results in a single 'set' event (which is what we'd expect).

**To Reproduce**

Have a single participant in a room. Subscribe for storage updates on the room. Call `undo` immediately after a LiveList `set`. Observe that 3 updates are receives, a `set`, a `delete` and an `insert`.

**Expected behavior**

The undo should reliably produce only a single `set` event that reverses the initial `set`.

**Environment (please complete the following information):**

"@liveblocks/client": "^0.19.4"
"@liveblocks/node": "^0.19.4"
"@liveblocks/react": "^0.19.4
Chrome v108.0.5359

GuillaumeSalles:
Hi @plamzd, 

It definitely look like an issue on our end. We're going to look into it.

nvie:
Hi @plamzd, thanks again for filing this bug report. We have recently fixed two bugs[^1][^2] in this area that might have been related. We haven't been able to reproduce this issue. Would you mind trying to upgrade to the latest version (1.0) to see if these fixes have solved this issue for you? If it's still an issue, please let us know! Thanks! 🙏 

[^1]: https://github.com/liveblocks/liveblocks/pull/636
[^2]: https://github.com/liveblocks/liveblocks/pull/637

weiningning:
I also encountered a similar problem. When I invoke `blocks.insert(child, i)`, and then invoke `blocks.set(i, child)` too soon, in such cases the local client's room subscription receives 2 updates - an insert for the respective live list element and a delete.
![image](https://user-images.githubusercontent.com/20291747/220347216-ef47c8bf-3e7b-4751-8cd8-864f0abd1af8.png)
[...]

PR #636, opened 2023-01-20
Fix LiveObject object key set undo with pending changes

BitPhinix:
While looking into https://github.com/liveblocks/liveblocks/issues/628 I discovered a bug in the LiveObject `_applyDeleteObjectKey` function that causes local history changes not to be applied. Probably not the cause of that issue, but something worth fixing regardless.

### DIFF
--- a/packages/liveblocks-core/src/crdts/__tests__/LiveObject.test.ts
+++ b/packages/liveblocks-core/src/crdts/__tests__/LiveObject.test.ts
@@ -1,6 +1,8 @@
+import { objectUpdate } from "../../__tests__/_updatesUtils";
 import {
   createSerializedList,
   createSerializedObject,
+  prepareDisconnectedStorageUpdateTest,
   prepareIsolatedStorageTest,
   prepareStorageTest,
   reconnect,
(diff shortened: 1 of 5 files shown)


packages/liveblocks-core/src/crdts/__tests__/LiveObject.test.ts
```diff
@@ -405,7 +407,56 @@ describe("LiveObject", () => {
...
+      prepareDisconnectedStorageUpdateTest<{
+        items: LiveObject<{ b?: string; a?: string }>;
+      }>(
+        [
+          createSerializedObject("0:0", {}),
+          createSerializedObject("0:1", { a: "B" }, "0:0", "items"),
+        ],
+        async ({ assert, batch, root, machine }) => {
+          const items = root.get("items");
+          machine.pauseHistory();
```
GuillaumeSalles, line 421: Here,`pauseHistory` and `resumeHistory` should have no impact because batch already group the two `set`.

GuillaumeSalles (approved):
Awesome catch! Super impressed that found this bug and made a fix!

nvie (approved):
The new unit test makes total sense, and I see that it indeed fails on main, but not on your branch — so thanks for the great fix 🙏 ! In full honesty, I don't think I fully follow how the fix works yet, but that's mostly my own lack of understanding of what `_propToLastUpdate` and `isLocal` mean/do precisely.

Feel free to also update the CHANGELOG with an entry for a new patch version. Perhaps provide a small example of what bug is fixed by this — I bet users will want to know about this one! 👍

PR #637, opened 2023-01-20
Fix undo batch with modify before set

BitPhinix:
Fixes another small issue I've discovered while digging into history: We use `activeBatch.reverseOps.push` instead of `activeBatch.reverseOps.unshift` as we do for non-batched ops (see [addToUndoStack](https://github.com/liveblocks/liveblocks/blob/fa7126aa07deb393b0d83bd89f07f014b87c5648/packages/liveblocks-core/src/room.ts#L1084)) resulting in incorrect ordering of the reverse ops for changes done in `batch`

I based this upon https://github.com/liveblocks/liveblocks/pull/636 because I used the new `prepareDisconnectedStorageUpdateTest` helper introduced in that pr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # v0.19.8
 
-Fix LiveObject key set undo if key has pending local changes.
+Fixes two history bugs:
+- The history didn't reliably undo LiveObject key set changes if any pending local changes existed on that key.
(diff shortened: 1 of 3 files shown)


GuillaumeSalles (approved):
Another great catch! I can't believe we didn't have a test that caught this earlier 😬

nvie (approved):
Great unit test + fix, @BitPhinix! 😍

Again, feel free to mention this in the CHANGELOG and perhaps include a small example to illustrate to developers what this fixes.

PR #642, opened 2023-01-25
Small update to DevTools installation instructions

nvie:


marcbouchenoire (changes requested):
Good idea to add it here. I've personally just disabled it (which is available in all browsers) to make switching between the two easier, and I also feel like `official` could be replaced by `production`?

> Disable the production Liveblocks extension, if you already have it installed (it will conflict with the development version)

nvie:
Great feedback, just updated it!

PR #644, opened 2023-01-26
Add new server message to communicate Op rejections

nvie:
This adds a new protocol message that the backend will start sending in response to a storage Op that is incompatible with the desired room schema.

### DIFF
--- a/packages/liveblocks-core/src/protocol/ServerMsg.ts
+++ b/packages/liveblocks-core/src/protocol/ServerMsg.ts
@@ -14,6 +14,7 @@ export enum ServerMsgCode {
   // For Storage
   INITIAL_STORAGE_STATE = 200,
   UPDATE_STORAGE = 201,
+  REJECT_STORAGE_OP = 299,
 }
 
 /**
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/protocol/ServerMsg.ts
```diff
@@ -180,3 +182,14 @@ export type UpdateStorageServerMsg = {
...
 };
+
+/**
+ * Sent by the WebSocket server to the client to indicate that certain opIds
+ * have been received but were rejected because they caused mutations that are
+ * incompatible with the Room's schema.
+ */
+export type RejectedStorageOpServerMsg = {
+  readonly type: ServerMsgCode.REJECT_STORAGE_OP;
+  readonly opIds: string[];
```
nvie, line 193: Since these rejection responses will necessarily trickle in asynchronously, there is no natural way to link back the rejection message to the original call site that is useful context for the developer.

For example, if you write:

```ts
root.get('foo').set('x', 'haha');
```

And it will get rejected by the server because `x` must be an int, then we may get a rejection message like "root.foo.x must be an Int", but it may be unclear _where_ in the app the code is that caused the failure.

Therefore, I suggest that, whenever we create the outgoing Op, we take a snapshot of the stack trace and remember it, just like we keep an administration of unacknowledged Ops. Then, when the Ops are reported back, we can use this `opIds` list to dig up the stack trace and use it when throwing the error message, so it provides the original context, along with the rejection reason.

nvie: Related issue: https://github.com/liveblocks/liveblocks.io/issues/716

BitPhinix (approved):

PR #645, opened 2023-01-26
0.19.10

GuillaumeSalles:
- Allow throttle to go as low as 16ms


nvie (approved):
LGTM, but curious about the rationale for this. Is it to support game-like updates with 60fps?

GuillaumeSalles:
16ms threshold increases the pressure on our servers, but it allows smoother cursors. It might only work with a few people in the room. Ideally, the throttling mechanism should be smarter and adapt based on how fast the server acknowledges an operation.

ofoucherot (approved):

stevenfabre (approved):
Oh yeah! 💪

PR #646, opened 2023-01-26
feat(webhooks): add WebhookHandler

FlowFlorent:
### Description

Adds a `WebhookHandler` class.
- `WebhookHandler(secret)` throws if secret is invalid or missing

`WebhookHandler.verifyRequest`:
- asserts that the event timestamp is with a 5 minute bound of `Date.now()` or throws
- asserts that the signature is correct


packages/liveblocks-node/src/webhooks.ts
```diff
@@ -0,0 +1,192 @@
...
+      .filter(isNotUndefined);
+
+    if (expectedSignatures.includes(signature) === false)
+      throw new Error(
+        `Invalid signature, expected one of ${expectedSignatures}, got ${signature}`
+      );
+
+    const event = JSON.parse(request.rawBody);
+
+    return event;
```
nvie, line 51: You're effectively assigning an `any`-typed value to a `WebhookEvent` return value here implicitly. In this case that's safe because we just verified that this value is coming from a Liveblocks origin, but it may be good to make this explicit using:

```suggestion
    const event: unknown = JSON.parse(request.rawBody);

    return event as WebhookEvent;
```

nvie: Thinking about forward-compatibility, one downside of this return value is that, if we ever start emitting a new, fourth, webhook event type, this would effectively be a breaking change because this current return type promises that there will only ever be max three possible event types.

This means that, strictly speaking, if we ever start emitting a new event type in the future, and we "just return" it, it will not match the TypeScript type we're promising here.

I think the best way to handle this is to inspect the `type` field here before returning it, and if it's a new unknown event type, throw an error saying "unknown event type, please upgrade to a higher version of @liveblocks/node".

That way, the TypeScript is fully correct and the implementation future-proof. We recently ran into an issue just like this in @liveblocks/core, that made it impossible for us to let the server send a new OpCode, because it would crash existing old clients if we did. We had to come up with a dirty hack as a workaround. I don't wish you this future 😉

nvie (approved):
Looking great, @FlowFlorent ! I left one suggestion to save your future-self from headaches, but otherwise this is awesome! Love this API!

packages/liveblocks-node/src/webhooks.ts
FlowFlorent: Oh indeed, that's a great catch, I'll ensure that this is done!

We do expect a couple more events to arrive such as `RoomCreatedEvent` and `RoomDeletedEvent`

Thank you for the review, saved a future headache! 😂

packages/liveblocks-node/src/webhooks.ts
```diff
@@ -0,0 +1,192 @@
...
+  type: "userEntered";
+  data: {
+    appId: string;
+    roomId: string;
+    connectionId: number;
+    userId: string | null;
+    userInfo: Record<string, unknown> | null;
+    /**
+     * ISO 8601 datestring
+     * @example "2021-03-01T12:00:00.000Z"
```
FlowFlorent, line 161:
```suggestion
     * @example "2021-03-01T12:00:00.000Z"
     * @description The time when the user entered the room.
```

ofoucherot (approved):
🚀

PR #650, opened 2023-01-30
add a why webhooks section

taylorpwhipp:
Add a "Why Webhooks" Section to guide

### DIFF
--- a/docs/pages/guides/webhooks.mdx
+++ b/docs/pages/guides/webhooks.mdx
@@ -10,24 +10,23 @@ you can react to Liveblocks events as they happen. Now that our platform is
 observable, you can use webhooks to reduce development time and the need for
 polling.
 
+## Why Webhooks
+Webhooks connect two applications and enable sending data asynchronously through event subscription and HTTP requests.  Using webhooks with Liveblocks decreases the complexity of managing and creating cron jobs which reach out to our API endpoints and poll. Using webhooks make a platform observable because updates between applications are immediately available.
+
 ## What is a webhook?
 
-Webhooks make a platform observable. They let users subscribe to events on the
-platform for them to react to. In this sense, webhooks streamline the collection
-of data between two connected services.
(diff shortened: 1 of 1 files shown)


docs/pages/guides/webhooks.mdx
```diff
@@ -10,24 +10,23 @@ you can react to Liveblocks events as they happen. Now that our platform is
 observable, you can use webhooks to reduce development time and the need for
 polling.
 
+## Why Webhooks
+Webhooks connect two applications and enable sending data asynchronously through event subscription and HTTP requests.  Using webhooks with Liveblocks decreases the complexity of managing and creating cron jobs which reach out to our API endpoints and poll. Using webhooks make a platform observable because updates between applications are immediately available.
```
adigau, line 14:
```suggestion
Webhooks connect two applications and enable sending data asynchronously through event subscription and HTTP requests. Using webhooks with Liveblocks decreases the complexity of managing and creating cron jobs which reach out to our API endpoints and poll. Using webhooks make a platform observable because updates between applications are immediately available.
```

adigau: Why don’t we give a few examples to inspire a less technical person?
Here's an example:
> For instance, you could imagine a scenario where you must save your room storage to your database once all users have left it. Our webhooks mechanism will notify your system every time someone leaves the room.

taylorpwhipp: sure that makes sense, will add example (or 2)

FlowFlorent, line 14: Sweet! That's gonna be a great section for those that are unfamiliar to webhooks!

I took the liberty of writing some suggestions, let me know what you think:

```suggestion
Webhooks connect two applications by letting one of them subscribe to events created by the other. The listening application (yours) subscribes to events and the event emitting application (Liveblocks) submits events via HTTP requests to the listening application's URL. Thus, using webhooks with Liveblocks reduces the need to poll Liveblocks data as you are now alerted when the data changes. This essentially makes Liveblocks observable and greatly reduces the need for cron jobs to poll Liveblocks data. As a result, your application (the listening application) will greatly reduce in complexity.
```

adigau (approved):

FlowFlorent (approved):
LOVE IT! 🚀

PR #651, opened 2023-01-31
Fix publish script

nvie:
This fixes the bug in the publish script where it errors with "turbo not found". It was a thing I missed, because I'm using [direnv](https://direnv.net/) to automatically manage my env vars on a per-directory basis, and "turbo" was always on my path.

Also, I think we're not enforcing specific `npm` and `node` versions yet here, which might be another issue. @FlowFlorent Lmk if this fix doesn't resolve the issues you're facing locally with this.


BitPhinix (approved):
:shipit:

PR #647, opened 2023-01-27
Schema validation guide

taylorpwhipp:
This PR encompasses the majority of work for the Schema Validation Guide but is in need of a few Assets:
- [ ] Animated Diagram for how Schema Validation Works between clients
- [ ] Design System Call to Action component/ implementation
- [ ] Merge PR for Error Message in liveblocks/cloudflare and  PR for api/v2/schema to include error code

Aslo as an aside, should I be referring to Schema Validation with a TM each time?

PR #652, opened 2023-01-31
Add handling for RejectedStorageOpServerMsg

BitPhinix:
Adds rudimentary client-side handling for RejectedStorageOp server messages by

- In development builds: log the error to the console and also throw it, so it's in the developer's face
- In production builds: only log the error to the console

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -1752,6 +1752,8 @@ function makeStateMachine<
                 `Storage ops rejected by server: ${message.reason}`
               );
             }
+
+            break;
           }
         }
       }
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -1743,13 +1743,13 @@ function makeStateMachine<
           // as a developer-owned bug. In production, these errors are not expected to happen.
           case ServerMsgCode.REJECT_STORAGE_OP: {
             console.errorWithTitle(
-              "Storage op rejection error",
+              "Storage mutation rejection error",
               message.reason
             );
 
             if (process.env.NODE_ENV !== "production") {
               throw new Error(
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -1735,6 +1735,24 @@ function makeStateMachine<
...
+              "Storage op rejection error",
+              message.reason
+            );
+
+            if (process.env.NODE_ENV !== "production") {
+              throw new Error(
+                `Storage ops rejected by server: ${message.reason}`
+              );
+            }
+          }
```
nvie, line 1755: Not strictly necessary, but let's end this `case` block with a `break`, despite it being the last one here.

nvie (approved):
Yesss! 🙌

packages/liveblocks-core/src/room.ts
```diff
@@ -1735,6 +1735,26 @@ function makeStateMachine<
...
+          // as a developer-owned bug. In production, these errors are not expected to happen.
+          case ServerMsgCode.REJECT_STORAGE_OP: {
+            console.errorWithTitle(
+              "Storage op rejection error",
+              message.reason
+            );
+
+            if (process.env.NODE_ENV !== "production") {
+              throw new Error(
+                `Storage ops rejected by server: ${message.reason}`
```
nvie, line 1752: While working on the [error message screenshots](https://github.com/liveblocks/liveblocks/pull/647#discussion_r1095647226), I noticed the use of the word "op" here, which is a concept that most Liveblocks users will never see. Let's avoid it in public messages:

```suggestion
            console.errorWithTitle(
              "Storage mutation rejection error",
              message.reason
            );

            if (process.env.NODE_ENV !== "production") {
              throw new Error(
                `Storage mutation rejected by server: ${message.reason}`
```

PR #654, opened 2023-02-01
Log traces of calls that caused rejected storage ops in non-production builds

BitPhinix:
Not sure about the DX of this one, but I want to try it out with the actual implementation.

Sadly because we can't use `console.trace` etc., the stack traces aren't correctly highlighted


nvie (approved):
Excellent work! 🤩

PR #659, opened 2023-02-02
Add references to DT in the documentation

adigau:
Adding references to the DT in the documentation.

To note:
- I used the <Banner> component everywhere I could
- I replaced the "storage in beta" banner where I needed to put the new banner, duplicates felt weird and we're not in beta anymore
- I moved some paragraphs to make them more coherent between elements of a series
- I added a blurb in the openapi file but couldn't style it


marcbouchenoire (approved):
✨

PR #660, opened 2023-02-03
Auth endpoint: add section on userId for MAUs

ofoucherot:
Update the authorize endpoint description:

image


FlowFlorent (approved):

adigau:
@ofoucherot I don't understand that phrase:
> The body for this POST request lets you declare what user ID the token is for, by means of the `userId` property.
What is it you want to say? I'd like to simplify it

ofoucherot:
> @ofoucherot I don't understand that phrase:
> 
> > The body for this POST request lets you declare what user ID the token is for, by means of the `userId` property.
> > What is it you want to say? I'd like to simplify it

@adigau This part is already there. I only added the second paragraph.
I agree that it's unclear; what do you think of that: "You can pass the property `userId` in the request's body. This can be whatever internal ...".

docs/references/v2.openapi.json
```diff
@@ -754,7 +754,7 @@
...
...tend) can use to enter a Liveblocks room. You use this endpoint to implement your own application’s custom authentication endpoint. When making this request, you’ll have to use your secret key.\n\nThe body for this POST request lets you declare what user ID the token is for, by means of the `userId` property. This can be whatever internal identifier you use for your user accounts, as long as it uniquely identifies an account. Setting the `userId` is optional if you want to use public rooms, but it is required to enter a private room (because permissions are assigned to specific user IDs). You can also declare which `groupIds` this user belongs to, in case you want to use the group permission system.\n\nThe property userId is used by Liveblocks to calculate your account’s Monthly Active Users. One unique userId corresponds to one MAU. If you don’t pass a userId, we will create for you a new anonymous userId on each connection, but your MAUs will be higher.\n\nAdditionally, you can set custom metadata to the token, which will be publicly accessible by other clients, through the `user.info` property. This is useful to store static data like avatar images, or the user's display name.",
```
adigau, line 757:
```suggestion
        "description": "This endpoint lets your application server (= your backend) obtain a token that one of its clients (= your frontend) can use to enter a Liveblocks room. You use this endpoint to implement your own application’s custom authentication endpoint. When making this request, you’ll have to use your secret key.\n\nYou can pass the property `userId` in the request's body. This can be whatever internal identifier you use for your user accounts as long as it uniquely identifies an account. Setting the `userId` is optional if you want to use public rooms, but it is required to enter a private room (because permissions are assigned to specific user IDs). In case you want to use the group permission system, you can also declare which `groupIds` this user belongs to.\n\nThe property userId is used by Liveblocks to calculate your account’s Monthly Active Users. One unique userId corresponds to one MAU. If you don’t pass a userId, we will create for you a new anonymous userId on each connection, but your MAUs will be higher.\n\nAdditionally, you can set custom metadata to the token, which will be publicly accessible by other clients through the `user.info` property. This is useful for storing static data like avatar images or the user’s display name.",
```

adigau: @ofoucherot made a few edits to the copy, and added your new phrase. Looks good to you?

ofoucherot: yes! i will merge this today

taylorpwhipp (commented):
I can't remember how long an anonymous user is valid for-- was it 24h? 30 days? we should probably mention this.

ofoucherot:
> I can't remember how long an anonymous user is valid for-- was it 24h? 30 days? we should probably mention this.

@taylorpwhipp With the secret api key endpoint, a new anonymous userId is generated at each connection, we don't reuse the ids.

With the public api endpoint, it's valid up until the client deletes their cookies.

PR #671, opened 2023-02-07
Fix create-liveblocks-app ssh repo url

BitPhinix:
Fixes one of two issues encountered by https://www.youtube.com/watch?v=8afaNZVzLVc. We are constructing the ssh repo URL incorrectly


CTNicholas (approved):
Thanks for fixing!

PR #681, opened 2023-02-10
Schema validation docs polish

nvie:
Hi! This morning, while checking the schema docs, I got an idea for slightly restructuring them, and to stop referring to the concept of a "schema language" or "language specification", because I think it may sound a bit too daunting for what it really is. What I would want to avoid is interested people thinking "Oh god, do I need to learn a new language for this? Nevermind."

Here is a small proposal to rename the doc structure. What do you think?

# Before vs after (main guide)

Screen Shot 2023-02-10 at 12 22 05@2x

Screen Shot 2023-02-10 at 12 22 22@2x

# Before vs after (syntax)

Screen Shot 2023-02-10 at 12 23 37@2x

Screen Shot 2023-02-10 at 12 24 04@2x

### DIFF
--- a/docs/pages/guides/schema-validation.mdx
+++ b/docs/pages/guides/schema-validation.mdx
@@ -64,11 +64,10 @@ of the data you want to store in your rooms.
   </video>
 </Figure>
 
-<Banner title="Schema language specification">
+<Banner title="Schema syntax">
 
-Take a look at our
-[language specification](/docs/guides/schema-validation/language) to learn more
-about what kind of validations are currently supported.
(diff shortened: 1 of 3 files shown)


taylorpwhipp (approved):

taylorpwhipp:
This makes more sense 👍  Nice :)

GuillaumeSalles:
I like it. Less intimidating. Thank you @nvie for the attention to details

GuillaumeSalles (approved):

stevenfabre (approved):
I like that better too. Nice one @nvie!

PR #683, opened 2023-02-10
Update example READMEs with `create-liveblocks-app` instructions

CTNicholas:
New format for the READMEs containing the installer command. The READMEs vary slightly, depending on whether environment variables are used or not.

---

**How it looks on GitHub** 

chrome_EA9E86FNPE

---

**How it looks on examples pages**

image (2)

---

**TODO**
- [x] Finish all READMEs
- [ ] Merge this: https://github.com/liveblocks/liveblocks.io/pull/909

### DIFF
--- a/docs/pages/guides/nextjs-starter-kit.mdx
+++ b/docs/pages/guides/nextjs-starter-kit.mdx
@@ -42,7 +42,7 @@ The Next.js Starter Kit includes the following
 
 You can get started by running the following command:
 
-```shell
+```bash
 npx create-liveblocks-app@latest --next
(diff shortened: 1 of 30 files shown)


marcbouchenoire (changes requested):
Much better, especially combined with `details`.

Given that `shell` isn't in the Prism language list, we've started using `bash` everywhere. In addition to adding highlighting, the code blocks also do some other things only for `bash` blocks like adding a `$` if it's a single line.

marcbouchenoire (approved):

PR #684, opened 2023-02-12
Starter kit: Avoid throwing error when user input causes "user not found" 

CTNicholas:


GuillaumeSalles (approved):

GuillaumeSalles:
Thanks for the quick fix @CTNicholas

PR #693, opened 2023-02-16
Zustand/Next.js flowcharts example

CTNicholas:
Under popular demand!

**TODO**
- [x] Add image
- [ ] Complete https://github.com/liveblocks/liveblocks.io/pull/924

image (4)


examples/nextjs-flowchart/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "nextjs-flow-chart",
```
marcbouchenoire, line 2: This would need to follow the folder so `nextjs-flowchart`.

examples/nextjs-flowchart/pages/_app.tsx
```diff
@@ -0,0 +1,8 @@
+import "@/styles/globals.css";
```
marcbouchenoire, line 1: We don't use TS paths in many places (anywhere even maybe?) so it'd be in favor of not using them here, especially if it's just used here and is synonymous to `./styles/globals.css`.

marcbouchenoire (changes requested):
Nice! I think the README should mention React Flow and more importantly Zustand, I first thought it was a regular Next.js/React example instead of a Zustand one.

Even the example name might be problematic, should it be `zustand-flowchart`? To me knowing that it uses `@liveblocks/zustand` instead of `@liveblocks/react` is more important than knowing that it uses Next.js.

examples/nextjs-flowchart/pages/_app.tsx
CTNicholas: Whoops! Didn't notice this, must be the new Next.js default

CTNicholas:
> Nice! I think the README should mention React Flow and more importantly Zustand, I first thought it was a regular Next.js/React example instead of a Zustand one.
> 
> Even the example name might be problematic, should it be `zustand-flowchart`? To me knowing that it uses `@liveblocks/zustand` instead of `@liveblocks/react` is more important than knowing that it uses Next.js.

Great points, I've updated it with your suggestions 👍 also added a few comments

examples/zustand-flowchart/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+<p>
+  <a href="https://liveblocks.io/examples/flowchart/nextjs/preview">
+    <img src="https://img.shields.io/badge/live%20preview-message?style=flat&logo=data:image/svg+xml;base64,PHN2ZyB2aWV3Qm94PSIwIDAgMjQgMjQiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+PHBhdGggZD0iTTE2Ljg0OSA0Ljc1SDBsNC44NDggNS4wNzV2Ny4wMDhsMTItMTIuMDgzWk03LjE1IDE5LjI1SDI0bC00Ljg0OS01LjA3NVY3LjE2N2wtMTIgMTIuMDgzWiIgZmlsbD0iI2ZmZiIvPjwvc3ZnPg==&color=333" alt="Live Preview" />
+  </a>
+  <a href="https://codesandbox.io/s/github/liveblocks/liveblocks/tree/main/examples/nextjs-flowchart">
+    <img src="https://img.shields.io/badge/open%20in%20codesandbox-message?style=flat&logo=codesandbox&color=333&logoColor=fff" alt="Open in CodeSandbox" />
+  </a>
+  <img src="https://img.shields.io/badge/react-message?style=flat&logo=react&color=0bd&logoColor=fff" alt="React" />
+  <img src="https://img.shields.io/badge/next.js-message?style=flat&logo=next.js&color=07f&logoColor=fff" alt="Next.js" />
```
marcbouchenoire, line 20: Last nitpick: we should add the Zustand badge here like other Zustand examples.

examples/zustand-flowchart/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "module": "esnext",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "jsx": "preserve",
+    "incremental": true,
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["./*"]
+    }
```
marcbouchenoire, line 20: Since we're not using them we can remove this from the config.

marcbouchenoire (changes requested):

marcbouchenoire (changes requested):
I thought it was part of the template but it looks like the README image is missing the usual border.

CTNicholas:
> I thought it was part of the template but it looks like the README image is missing the usual border.
> 
> 

Good spot! Fixed

marcbouchenoire (approved):
🙌

PR #697, opened 2023-02-17
feat(publish): publish from CI

FlowFlorent:
## What does this PR do?

This PR enables publishing of packages from the CI using `workflow_dispatch`

**Check out a successful run [here](https://github.com/liveblocks/liveblocks/actions/runs/4266706391/jobs/7427540895)!**

## Where should the reviewer start?

The process happens in 2 places.

### 1. Running a release
When merged on `main`, it will be done through the github UI with a `tag` input:
actions-manually-run-workflow

Though for now it must be done through the `gh` CLI:
`gh workflow run "Publish packages" --ref feat/automatic-deploy -f tag=<GITHUB_TAG>`

### 2. Workflow takes over
1. Install all dependencies
2. Setup the release `release.sh`
3. Build packages
4. Publish the release (`publish.sh`) and pushes to the branch
Screen Shot 2023-02-24 at 5 39 41 PM
5. Creates & pushes the release tag and prompts for release notes
Screen Shot 2023-02-24 at 5 35 16 PM

## How should this be manually tested?

Run `gh workflow run "Publish packages" --ref feat/automatic-deploy -f tag=<GITHUB_TAG>`
Check that action runs
Check that packages are properly published
Check that `beta` npm tag won't work, ex: `gh workflow run "Publish packages" --ref feat/automatic-deploy -f tag=1.0.2-beta0` since its a topic branch
Check that release without a tag won't work, ex: `gh workflow run "Publish packages" --ref feat/automatic-deploy -f tag=1.0.2` since its not `main` branch

## What are the relevant tickets?

https://github.com/liveblocks/liveblocks.io/issues/891

## Further discussion
Next step will be to use it for schema packages, it will require setting up some modularity, ex: repo url, package paths


scripts/publish.sh
```diff
@@ -368,22 +105,4 @@ echo ""
 echo "==> Pushing changes to GitHub"
 BRANCH="$(git current-branch)"
```
FlowFlorent, line 106: will fail as requires git

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,89 @@
...
+          GIT_TAG: ${{ steps.git_tag.outputs.value }}
+        run: |
+          get_tag_if_exists () {
+            if grep -q "-" <<< "$1"; then
+                echo "${1##*-}" | sed 's/[0-9]//g'
+            else 
+                echo "latest"
+            fi
+          }
+          echo value=$(get_tag_if_exists "$GIT_TAG") >> $GITHUB_OUTPUT
```
nvie, line 54: Nice! This is the part where `v0.1.2-beta6` will "infer" `beta` and use it as an NPM tag 👌

I happen to be able to read this, but it's quite hidden/obscure, and I think this convention is really important to document somewhere — did you document this process somewhere already?

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,89 @@
...
+        env:
+          GIT_TAG: ${{ steps.git_tag.outputs.value }}
+        run: echo "value=${GIT_TAG#v}" >> $GITHUB_OUTPUT
+
+      - name: Set npm tag
+        id: npm_tag
+        env:
+          GIT_TAG: ${{ steps.git_tag.outputs.value }}
+        run: |
+          get_tag_if_exists () {
```
nvie, line 47: Naming suggestion: `get_npm_tag_if_exists`. I would stress this is the _npm_ tag, and not the _git_ tag.

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,89 @@
...
+            if grep -q "-" <<< "$1"; then
```
nvie, line 48: TIL about `<<<` 🤯

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v3.6.0
+        with:
+          node-version: 16.17.1
+
+      - name: Preparing release
+        env:
+          VERSION: ${{steps.version.outputs.value}}
+        run: ./scripts/release.sh -V "$VERSION"
```
nvie, line 75: Could we flip these sections, so that the "setup git" and "prepare release" are closer together?

```suggestion
      - name: Setup Node.js environment
        uses: actions/setup-node@v3.6.0
        with:
          node-version: 16.17.1

      - name: Setup git config
        run: |
          git config user.name "GitHub Actions Bot"
          git config user.email "<>"

      - name: Preparing release
        env:
          VERSION: ${{steps.version.outputs.value}}
        run: ./scripts/release.sh -V "$VERSION"
```

Here, I'm assuming that the "setup git" part is required because the "preparing release" section will create a new git commit, right?

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Build all packages
+        run: npm run build
+
+      # Add approve step here?
```
nvie, line 80: What would this look like in your view, since this is a non-interactive script?

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Publish packages
+        # setup-node created .npmrc file to publish to npm
+        # https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          VERSION: ${{ steps.version.outputs.value }}
+          TAG: ${{ steps.npm_tag.outputs.value }}
+        run: ./scripts/publish.sh -V "$VERSION" -t "$TAG"
```
nvie, line 89: I'd love to move the `release.sh` and `publish.sh` scripts into the workflows directory, to keep them local here and to avoid them from being run by users locally.

nvie (commented):
Super nice work, @FlowFlorent 😍 ! Can't wait to have this. I left a couple thoughts below, lmk what you think!

scripts/test-running-script.sh
```diff
@@ -0,0 +1,3 @@
+echo "Running test scripts..."
+
+git current-branch
```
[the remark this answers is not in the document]
FlowFlorent: Yes, I was using for tinkering, good catch 👁️

scripts/create-release-tag.sh
```diff
@@ -0,0 +1,43 @@
...
+    if ! [[ "$1" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[a-z0-9]+)?$ ]]; then
+        err "Invalid tag: $1"
+        err "Tag must be in the form of vX.Y.Z or vX.Y.Z-<tag>"
+        exit 2
+    fi
+}
+
+# Check branch is main or beta, only allow tags without -<tag> on main
+check_current_branch () {
+  git branch --show-current | grep -E "^(main|beta)$" || {
```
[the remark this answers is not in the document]
FlowFlorent: What I predict is that people will `git push -f` their experimental branch to `beta` and then publish from beta, meaning that it won't solve the issue.

It seems that beta tags are used by the community, whether it's expected or not.

I would propose the following, as another step / process to setup (out of scope for this PR as it does not currently downgrade the experience)

`beta` & `latest` tags are officially supported (or maybe just `latest`), we use `alpha`, `experimental`, `test` or some kid of other tag for our own testing purposes, we make this very clear in our github documentation: `latest`, `beta` are officially supported, anything else is not, use at your own risk. 

If we want greater protection, we could also ensure that these `alpha`-like versions are privately published, meaning that we don't risk community users accidentally falling on a version which we might not want used.

`beta` (beta branch) is ahead of `latest` (main branch) and `beta` cannot be overridden with a `git push -f`, only incremented, the same way that we deal with `main`. This will ensure that versions don't have unexpectedly missing features.

This process:
- allows freedom to safely publish experimental versions
- communicates what we officially support
- ensures stability for officially supported versions

What do you think?

scripts/create-release-tag.sh
```diff
@@ -0,0 +1,43 @@
...
+      exit 2
+  }
+  echo "Tag $1 does not exist yet!"
+}
+
+create_and_push_tag () {
+  echo "Creating tag $1"
+  git tag "$1" -m "Release $1" && git push origin "$1"
+  echo "Tag $1 created and pushed"
+}
```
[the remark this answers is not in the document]
FlowFlorent: Yes! So there's zero install on the user's machine!

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,89 @@
...
+        id: set-up-homebrew
+        uses: Homebrew/actions/setup-homebrew@master
+
+      - name: Install moreutils
+        run: brew install moreutils
+
+      - name: Install git-toolbelt
+        run:
+          brew install fzf && brew tap nvie/tap && brew install
+          nvie/tap/git-toolbelt
```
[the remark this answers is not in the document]
FlowFlorent: You are absolutely correct on this one, I'd love to reduce this to a few seconds. Currently using a bazooka to swap a fly. 😅 

| To "install" git-toolbelt, all you'd have to do is unzip the repo somewhere, and put it on the $PATH.

Would love to have pointers to achieve this 🙏

.github/workflows/publish-packages.yml
FlowFlorent: Good call! I changed it to `get_npm_tag` because it defaults to latest

.github/workflows/publish-packages.yml
FlowFlorent: github copilot ftw!

.github/workflows/publish-packages.yml
FlowFlorent: absolutely!

scripts/create-release-tag.sh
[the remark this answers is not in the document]
FlowFlorent: Created a ticket for this: https://github.com/liveblocks/liveblocks.io/issues/937

.github/workflows/publish-packages.yml
FlowFlorent: That's indeed the case, the created tag within the workflow did not trigger another github action 🎉 
Screen Shot 2023-02-23 at 2 03 25 PM
Screen Shot 2023-02-23 at 2 03 19 PM

scripts/create-release-tag.sh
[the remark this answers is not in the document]
nvie: Agreed, happy to consider both `main` and `beta` as special branches, so that:
- we can only publish `beta` tags from a `beta` branch (and they _must_ be named vX.Y.Z-betaN)
- we can only publish `latest` tags from the `main` branch (and they _must_ be named vX.Y.Z)

But other than that, it's fine to publish other NPM tags freely from any branch, no need for extra conventions or limits. We'll probably only ever need `test` for that, but hey, any tag name is fine AFAIAC. This way, we avoid multiple people from using `test` at the same time, too.

If I've used these custom tags in the past, I've always (manually) removed them after I was done testing things in NPM anyway. Just to keep our view of https://www.npmjs.com/package/@liveblocks/client?activeTab=versions clean. (Ironically, I just noticed there still is a `test` version out there now—I'll clean that up right away 😉)

Sounds good?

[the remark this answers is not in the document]
nvie: > I've always (manually) removed them after I was done testing things in NPM anyway.

Done so by running the following, btw:

```
npm dist-tag rm "@liveblocks/node" <tagname>
npm dist-tag rm "@liveblocks/zustand" <tagname>
npm dist-tag rm "@liveblocks/redux" <tagname>
npm dist-tag rm "@liveblocks/react" <tagname>
npm dist-tag rm "@liveblocks/client" <tagname>
npm dist-tag rm "@liveblocks/core" <tagname>
```

.github/workflows/publish-packages.yml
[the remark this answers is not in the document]
nvie: I haven't tested this, but here is one way that I think should work, as suggested by https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-adding-a-system-path:

    - name: Install git-toolbelt
      run: |
        wget https://github.com/nvie/git-toolbelt/archive/v1.8.0.tar.gz
        tar -xzf v1.8.0.tar.gz
        mkdir -p "$HOME/.local/bin"
        cp git-toolbelt-1.8.0/git-* "$HOME/.local/bin"
        rm -rf v1.8.0.tar.gz git-toolbelt-1.8.0
        echo "$HOME/.local/bin" >> $GITHUB_PATH

Could you try this out and lmk if this works?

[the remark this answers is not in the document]
FlowFlorent: Absolutely, checking now, thank you!

.github/workflows/publish-packages.yml
nvie: What happens to your local tag in that case? Will this conflict be resolved when you do `git pull` next?

.github/workflows/publish-packages.yml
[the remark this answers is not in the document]
nvie: Btw, not sure if you've set up any form of caching for this job, but if so, you may download the `.tar.gz` file to the cached location, and only `wget` it if it doesn't exist on disk locally (and of course not `rm -rf` it in that case). Saves a download on every run.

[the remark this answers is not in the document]
FlowFlorent: Yeah absolutely, will look into this once the whole workflow runs properly

.github/workflows/publish-packages.yml
FlowFlorent: It will conflict on the git pull, with vscode you get a popup that solves the conflict for you
Screen Shot 2023-02-23 at 3 21 53 PM

FlowFlorent: Perhaps some kind of `temp-v1.0.1-test1` tag could be created with `npm publish v1.0.1-test1` so we wouldn't have to delete it, and CI creates a proper one. 
That would mean that we'd get double the tags though, not sure we necessarily want that.

scripts/create-release-tag.sh
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+check_tag_does_not_exist () {
+  git tag -l "$1" | grep -E "^$1$" && {
+      err "Tag $1 already exists"
+      exit 2
+  }
+  echo "Tag $1 does not exist yet!"
+}
```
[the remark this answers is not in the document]
FlowFlorent: true!

.github/workflows/publish-packages.yml
nvie: Alternatively, you can change the local script to delete the tag directly after pushing it out?

```
git tag "$tag" && git push origin "$tag" && git tag -d "$tag"
```

scripts/create-release-tag.sh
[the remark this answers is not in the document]
FlowFlorent: Done, added the checks in the `npm run publish` script

.github/workflows/publish-packages.yml
FlowFlorent: Well, there we go simple and effective, thank you!

.github/workflows/publish-packages.yml
FlowFlorent: I did not document it but plan on adding it to the `README.MD` or somewhere obvious for us

scripts/create-release-tag.sh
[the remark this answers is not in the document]
GuillaumeSalles: Great! I like your propositions. Let's put that in place in a subsequent PR

nvie:
Btw, @FlowFlorent! I just noticed that @GuillaumeSalles linked to [this article](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) in the original ticket. Did you explore that option too? To me, it looks like this is a pretty nice way to trigger these actions, rather than invoking a custom script locally.

Think:

Screen Shot 2023-02-24 at 09 15 38@2x

The workflow script would do the sanity checks itself directly. (I.e. check value entered is in the expected version string format, and validate against branch limit rules.)

Sounds like an even better option than having the `publish` script locally.

nvie:
Also asking because for the https://github.com/liveblocks/liveblocks-schema repo, we forked the monorepo setup and our trusty old `publish.sh` script is also used there. Would be nice to have the same setup in both repos once we nail this one.

FlowFlorent:
@nvie That is true! I'll take a crack at it

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,106 @@
...
+                echo "latest"
+            fi
+          }
+          echo value=$(get_npm_tag "$GIT_TAG") >> $GITHUB_OUTPUT
+
+      - name: Check values
+        run: |
+          echo ${{ inputs.tag }}
+          echo ${{ steps.version.outputs.value }}
+          echo ${{ steps.npm_tag.outputs.value }}
```
nvie, line 62: Always quote shell arguments that contain variables:

```suggestion
          echo "${{ inputs.tag }}"
          echo "${{ steps.version.outputs.value }}"
          echo "${{ steps.npm_tag.outputs.value }}"
```

package.json
```diff
@@ -13,7 +13,8 @@
     "test:types": "turbo run test:types",
     "test:deps": "turbo run test:deps",
     "lint": "turbo run lint",
-    "dev": "turbo run dev"
+    "dev": "turbo run dev",
+    "publish": "./scripts/create-release-tag.sh"
```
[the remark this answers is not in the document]
FlowFlorent: You're right, forgot to remove it! I can link the `CONTRIBUTING.MD` file there or some kind of link towards an instruction

PR #700, opened 2023-02-20
fix: Add `!` where it's missing

CTNicholas:


nvie (approved):
Thanks for the quick fix! 🚀

PR #709, opened 2023-02-22
Fix: server message ROOM_STATE should clear removed users

ofoucherot:
When a client connects to a room, the server sends the message `ROOM_STATE` containing the list of users present in the room.

If a client has an inactive user in its "others" list, and the client **re**connects to the server and receives a new `ROOM_STATE` message, the client should remove the inactive user (which was already removed in the server).


nvie (approved):
Thank you for pin-pointing the issue. The fix is super plausible and fortunately simple. Great work, @ofoucherot ✨ !

nvie:
@ofoucherot I merged this with `main`, [because of this](https://liveblocks.slack.com/archives/C02PZL7QAAW/p1677151489703639). CI should now pass again here.

ofoucherot:
> @ofoucherot I merged `main` back in here, [because of this](https://liveblocks.slack.com/archives/C02PZL7QAAW/p1677151489703639). CI should now pass again.

@nvie thank you!

PR #712, opened 2023-02-24
Core: `PlainLson` types

marcbouchenoire:
This PR adds and exposes `PlainLson` types from `@liveblocks/core` so that they can be shared and used across multiple Liveblocks projects.

@nvie I wasn't exactly sure about the difference between living under `protocol` and `types` in this case (the types are generic but not too much).


nvie (approved):
Thanks for doing this!

Putting this in `types/` is fine. The import rules are [enforced by dependency cruiser](https://github.com/liveblocks/liveblocks/blob/core/plain-lson-types/packages/liveblocks-core/.dependency-cruiser.js#L148-L166). As you can see from those descriptions, you can pick either of the two locations.

Btw, you can get CI to pass by running `npm run format`.

PR #713, opened 2023-02-25
fix(docs): renamed to `liveblocksEnhancer`

daikiojm:
Updated docs according to the following changes.

```ts
/**
 * Redux store enhancer that will make the `liveblocks` key available on your
 * Redux store.
 */
declare const liveblocksEnhancer: <TState>(options: {
    client: Client;
    storageMapping?: Mapping<TState> | undefined;
    presenceMapping?: Mapping<TState> | undefined;
}) => StoreEnhancer;
/**
 * @deprecated Renamed to `liveblocksEnhancer`.
 */
declare const enhancer: <TState>(options: {
    client: Client;
    storageMapping?: Mapping<TState> | undefined;
    presenceMapping?: Mapping<TState> | undefined;
}) => StoreEnhancer;
```

### DIFF
--- a/docs/pages/api-reference/liveblocks-redux.mdx
+++ b/docs/pages/api-reference/liveblocks-redux.mdx
@@ -12,13 +12,13 @@ Enhancer that lets you connect a Redux state to Liveblocks Presence and Storage
 features.
 
 ```js
-import { enhancer } from "@liveblocks/redux";
+import { liveblocksEnhancer } from "@liveblocks/redux";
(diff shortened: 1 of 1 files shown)


taylorpwhipp (commented):
@daikiojm Thanks for making this contribution/ correction - it is much appreciated! This looks good to me, but will have another teammate look over as well.

nvie (approved):
Thank you!

PR #719, opened 2023-03-01
Fix some typos in README instructions

nvie:
I noticed these while following the instructions. Not sure if the Starter Kit ones use different commands to run them locally, but I don't think so?

Depending on which project type, there is variation between `dev`, `serve`, `start`, and `run` 😬


CTNicholas:
I redid every README recently, must have missed this! Will check over this and the other examples later, thank you 🙌

CTNicholas (approved):
Found the same mistake in other places, thanks for highlighting this 👍

PR #722, opened 2023-03-02
Add installer info to DevTools guide

CTNicholas:
All of the examples on our website can be installed with the following command:

```
npx create-liveblocks-app@latest --example [GITHUB EXAMPLE DIRECTORY NAME]
```

Updated the guide to add this info!


CTNicholas:
Hey, I've made another quick change—undo it if you're think I'm wrong!

I haven't seen the `⌥` symbol before, my keyboard says "alt". So I changed it to say "alt/option key" instead of "`⌥` key (option on mac)".

taylorpwhipp (approved):
Thanks for adding this Chris!

PR #735, opened 2023-03-14
Avoid re-rendering twice when mounting RoomProvider

### LINKED ISSUE #730: [FEATURE_REQUEST] Children under the `RoomProvider` component should render only once initially
I am not too familiar with the codebase, so I may be incorrect in a lot of my assumptions, so please feel free to correct me if anything seems incorrectly construed.

**Is your feature request related to a problem? Please describe.**

Yes. I noticed that all children components placed inside the `RoomProvider` component render twice initially (technically four times, but we can discount two due to React's Strict Mode behaviour). This may not be a major issue, but I wanted to start a discussion on why that may be and how we can solve it.

From what I understand after taking a look at the code, the `Room` object is created using the `useState` hook, and the code that deals with creating this `Room` object and updating it looks like the following:

https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react/src/factory.tsx#L172-L196

```tsx
  const [room, setRoom] = React.useState<
    Room<TPresence, TStorage, TUserMeta, TRoomEvent>
  >(() =>
    client.enter(roomId, {
      unstable_batchedUpdates: frozen.unstable_batchedUpdates,
      shouldInitiallyConnect: frozen.shouldInitiallyConnect,
    } as RoomInitializers<TPresence, TStorage>)
  );

  React.useEffect(() => {
    setRoom(
      client.enter(roomId, {
        unstable_batchedUpdates: frozen.unstable_batchedUpdates,
        withoutConnecting: frozen.shouldInitiallyConnect,
      } as RoomInitializers<TPresence, TStorage>)
    );

    return () => {
      client.leave(roomId);
    };
  }, [roomId, frozen]);
```

This may be inefficient because `RoomProvider` and all its children first render with the initially set value of `room`, and then immediately render again due to the `room` object being updated inside the `useEffect` hook. I understand the need of `entering` and `leaving` the room as the component mounts and un-mounts respectively but I think this can be done without triggering multiple renders.

Another potential issue with re-initializing `room` through the `useEffect` hook every time `roomId` changes could be that internal state of children of `RoomProvider` won't be reset when `room` changes. For example, in the sample code below, when the `roomId` value is updated, the value of `count` (if updated) in `MyComponent` doesn't reset, but the value of `shapes` does. This may be okay depending on what we believe the expected result to look like. I would, however, assume everything to be reset when the `roomId` is updated. This can still be achieved in the code block below by explicitly resetting `count` when `room` updates, but that is inefficient and we will have another double render issue.

```tsx
  function App() {
    const [roomId, setRoomId] = useState(1);
    
    return (
        <RoomProvider roomId={roomId} initialPresence={...} initialStorage={...}>
            <MyComponent />
            <button onClick={() => setRoomId(room + 1)}>Room {roomId}</button>
        </RoomProvider>
    )
}

function MyComponent() {
    const [count, setCount] = useState(0);
    const shapes = useStorage((root) => root.shapes);

    return <button onClick={() => setCount(count + 1)}>Increment</button>
}
```

**Describe the solution you'd like**

This is my very naive attempt at the solution, and please feel free to point out anything may doesn't make sense or is blatantly incorrect. And, please note that some of the functions that are called/used in the code aren't currently exposed/available.

First of all, instead of creating `room` using the `useState` hook, we could instantiate it as a mutable ref using the `useRef` hook. This will also require updating the code inside `useEffect` to connect the room and add it to client on mount, and disconnect the room and remove it from the client on unmount. 

And, in order to reset all internal states of the children of `RoomProvider`, we pass the `roomId` as the `key` property. This way, we are asking React to treat two `RoomProvider` component with different `roomIds` (as keys) as two different components that should not share any internal state. 

```tsx
function RoomProviderInner({ roomId, initialPresence, initialStorage } : { roomId: string, initialPresence: ..., initialStorage: ... }) {
    const roomRef = useRef<Room<Presence, Storage, BaseUserMeta, never>>();
  
    if (!roomRef.current) {
      roomRef.current = createInternalRoom(roomId, {
        initialPresence: initialPresence,
        initialStorage: initialStorage,
      });
    }
  
    useEffect(() => {
      const room = roomRef.current;
  
      if (!room) return;
  
      room.connect();
      client.add(room); // method not exposed/available 
  
      return () => {
        room.disconnect();
        client.remove(room); // method not exposed/available 
      };
    }, []);

    return (
      <RoomContext.Provider value={roomRef.current}>
        {children}
      </RoomContext.Provider>
    );
}

function RoomProvider({ roomId, initialPresence, initialStorage } : { roomId: string, initialPresence: ..., initialStorage: ... }) {
    return <RoomProviderInner key={roomId} roomId={roomId} initialPresence={initialPresence} .... />
}

function useRoom() {
    const room = useContext(RoomContext);
    
    if (!room) {
      throw new Error("RoomProvider not found");
    }
  
    return room;
}
```

nvie:
Fixes #730.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# v1.0.2
+
+### `@liveblocks/react`
+
+- Fixes a bug in React where the children of `RoomProvider` would get
+  re-rendered twice on mount
+
 # v1.0.1
 
 Log stack traces of function calls that resulted in rejected storage mutations
@@ -12,7 +19,8 @@ to the console in non-production builds to ease debugging.
 
 # v1.0.0
 
-This major release marks the maturity of Liveblocks. For upgrade instructions, see the [1.0 upgrade guide](https://liveblocks.io/docs/guides/upgrading/1.0).
+This major release marks the maturity of Liveblocks. For upgrade instructions,
+see the [1.0 upgrade guide](https://liveblocks.io/docs/guides/upgrading/1.0).
 
 ## `@liveblocks/node`
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -15894,10 +15894,10 @@
     },
     "packages/liveblocks-client": {
       "name": "@liveblocks/client",
-      "version": "1.0.1",
+      "version": "1.0.2-test5",
       "license": "Apache-2.0",
       "dependencies": {
-        "@liveblocks/core": "1.0.1"
+        "@liveblocks/core": "1.0.2-test5"
       },
       "devDependencies": {
         "@liveblocks/eslint-config": "*",
@@ -15911,7 +15911,7 @@
     },
     "packages/liveblocks-core": {
       "name": "@liveblocks/core",
-      "version": "1.0.1",
+      "version": "1.0.2-test5",
       "license": "Apache-2.0",
       "devDependencies": {
         "@liveblocks/eslint-config": "*",
@@ -16077,7 +16077,7 @@
     },
     "packages/liveblocks-node": {
       "name": "@liveblocks/node",
-      "version": "1.0.1",
+      "version": "1.0.2-test5",
       "license": "Apache-2.0",
       "dependencies": {
         "node-fetch": "^2.6.1"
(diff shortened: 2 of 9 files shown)


nvie:
Closing because this turned out not to be a real issue.

PR #750, opened 2023-03-22
DRAFT- Ts migration nextjs whiteboard

taylorpwhipp:
This is a Draft PR for migrating the Simple whiteboard example

**TODO** 
- [ ] Add to [`/src/constants/examples.ts`](https://github.com/liveblocks/liveblocks.io/blob/main/src/constants/examples.ts), check info is correct @taylorpwhipp 
  - https://github.com/liveblocks/liveblocks.io/pull/1058
- [ ] ~Improve design @pierrelevaillant~ We'll do this later when editing the example for the interactive tutorial
  - https://github.com/liveblocks/liveblocks.io/issues/917

Info on improved design:
>**Styles**
> - We want the example to work inside the two windows in the tutorial, so the UI should be fairly minimal and not too large. 
> - By default the example should have light mode styling.
> - When liveblocks.io is on dark mode the interactive tutorial will add `data-theme="dark"` to `<html>`. We can use this to add dark mode styles too. 
> ``` [data-theme="dark"] body { ... } ```
>
>**Notes**
> - I'm applying default styling to EVERY preview window in `/src/tutorial/config/defaultStyles.ts`. Feel free to change this too, but this should be very minimal. 
> - Remember that the example must work by itself, but also with the interactive tutorial system.
>


CTNicholas:
Heya, I've done a quick review, thank you for working on this! Most of my changes are just tidying up, and making the example match our others 👍

I made one change to the types, which was removing all the `any`s, as using `any` disables TypeScript. For a moment I was confused why the types weren't working, but it was just being caused by `type Shape = { ... }` instead of `type Shape = LiveObject<{ ... }>`

Make sure to add this to liveblocks.io when releasing: https://github.com/liveblocks/liveblocks.io/blob/main/src/constants/examples.ts

Thanks!

CTNicholas (approved):

taylorpwhipp:
@GuillaumeSalles should we 
1. replace the existing example, 
2. merge this and keep both
2. update the existing tutorial (guessing no here) until interactive is ready ?

PR #754, opened 2023-03-27
Update syntax doc to match latest state for schema validation

nvie:
This updates the syntax doc to match the latest implementation.

### DIFF
--- a/docs/pages/guides/schema-validation/syntax.mdx
+++ b/docs/pages/guides/schema-validation/syntax.mdx
@@ -18,9 +18,10 @@ This document describes the language rules for writing your own Liveblocks
 schemas. It is an exhaustive account of all feature that will be implemented and
 supported before the public beta.
 
-Currently in the private beta phase we support: scalars, optionals, objects, and
-`LiveObject`. We’re sharing our plans for other syntaxes so you can give us
-early feedback [here](https://github.com/liveblocks/liveblocks/discussions/674).
+Currently in the private beta phase we support: scalars, arrays, objects,
+optionals, `LiveObject`, `LiveList`, `LiveMap`, and most unions. We’re sharing
+our plans for other syntaxes so you can give us early feedback
+[here](https://github.com/liveblocks/liveblocks/discussions/674).
 
 ## Storage root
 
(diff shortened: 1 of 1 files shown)


docs/pages/guides/schema-validation/syntax.mdx
```diff
@@ -64,7 +65,6 @@ root.set("hasSiblings", true);
 
 // ❌ Invalid storage updates
 root.set("name", true);
-root.set("age", 66.5); // Invalid because `age` should be a whole number
```
nvie, line 67: Noticed this was still in here, but this is not true since we removed the `Int` type for now. We'll reintroduce it later again as a custom data validator type (when we introduce regex, numberic ranges, etc).

docs/pages/guides/schema-validation/syntax.mdx
```diff
@@ -18,9 +18,10 @@ This document describes the language rules for writing your own Liveblocks
 schemas. It is an exhaustive account of all feature that will be implemented and
 supported before the public beta.
 
-Currently in the private beta phase we support: scalars, optionals, objects, and
-`LiveObject`. We’re sharing our plans for other syntaxes so you can give us
-early feedback [here](https://github.com/liveblocks/liveblocks/discussions/674).
+Currently in the private beta phase we support: scalars, arrays, objects,
+optionals, `LiveObject`, `LiveList`, `LiveMap`, and most unions. We’re sharing
+our plans for other syntaxes so you can give us early feedback
```
nvie, line 23: Updated to match the current feature set!

docs/pages/guides/schema-validation/syntax.mdx
```diff
@@ -243,6 +243,20 @@ root.set(
...
+from TypeScript. Here are some examples:
+
+```ts
+type Storage {
+  ids: (string | number)[]
+  selectedIds: string | null
+  person: LiveObject<Person> | null
+  people: LiveList<LiveObject<Person>> | null
+}
+```
```
nvie, line 258: @taylorpwhipp What do you think? Examples make sense? Too complicated? Feel free to come up with simpler examples! 🙏

taylorpwhipp: makes sense to me!

taylorpwhipp (approved):
Thanks for making these changes!

PR #756, opened 2023-03-29
Update `starter-kits` Liveblocks versions from script

CTNicholas:
Hey @nvie, please check this is correct! We want to automatically update Liveblocks in projects in the `starter-kits` directory.


nvie (approved):
Correct!

PR #757, opened 2023-03-30
Extend todo list demo app with ability to cross off items

nvie:
Here is a slightly modified version of our Todo list example app, with the ability to cross items off. I've done this to facilitate a simpler demo for schema validation to record for Taylor, and this app can be used as a foundation to show how to make the `checked` property required.

The app is improved in such a way that I think it's a better demo app anyway, even without schema validation in mind.

Here is the new behavior, see:

https://user-images.githubusercontent.com/83844/228759143-f62e34c7-f371-4bc0-ab05-73b7c32ed604.mp4

The updates:

1. Makes every todo list item a `LiveObject`, rather than a JSON object (more idiomatic Liveblocks)
2. Adds an optional `checked` boolean on there that gets toggled on click

I've also fixed a bug where submitting the empty text was possible. (You could hold down Enter in the previous version to create thousands of items per second 😬)

### DIFF
--- a/examples/nextjs-todo-list/pages/index.tsx
+++ b/examples/nextjs-todo-list/pages/index.tsx
@@ -7,7 +7,7 @@ import {
   useMutation,
 } from "../liveblocks.config";
 import "@liveblocks/react";
-import { LiveList } from "@liveblocks/client";
+import { LiveList, LiveObject } from "@liveblocks/client";
 import { useRouter } from "next/router";
 import { ClientSideSuspense } from "@liveblocks/react";
 
@@ -37,7 +37,7 @@ function Example() {
   const todos = useStorage((root) => root.todos);
 
   const addTodo = useMutation(({ storage }, text) => {
-    storage.get("todos").push({ text });
+    storage.get("todos").push(new LiveObject({ text }));
   }, []);
 
   const deleteTodo = useMutation(({ storage }, index) => {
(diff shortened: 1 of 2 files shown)


examples/nextjs-todo-list/pages/index.tsx
```diff
@@ -88,7 +102,7 @@ function Loading() {
 }
 
 export default function Page() {
-  const roomId = useOverrideRoomId("nextjs-todo-list");
+  const roomId = useOverrideRoomId("nextjs-todo-list-v2");
```
nvie, line 105: Technically, the switch to `LiveObject`s here can possibly break when clicking on items that have previously been created as JSON todo items. The easiest fix for this was to just bump to a different room ID, but it's the first time we've done that in our examples, I think? Any advice? Worth it?

marcbouchenoire: We do [clean up unused-since-at-least-24h example rooms daily](https://github.com/liveblocks/liveblocks.io/blob/main/scripts/cron/clean-up-examples.ts), but that still leaves plenty of room for error so yeah I'd say it's worth it to do versioning.

marcbouchenoire (approved):
Nice 🙂

taylorpwhipp (approved):

CTNicholas (approved):

PR #798, opened 2023-04-05
Fix bug where `shouldInitiallyConnect` config option would not always work

nvie:
We forgot to update the old `withoutConnecting` name in the useEffect callback. In this case, the bug would only get surfaced in React Strict Mode, where an application using `<RoomProvider id="abc" shouldInitiallyConnect={false}>...</RoomProvider>` would _still_ connect eventually (on the second render, as triggered by React Strict Mode).

This has been a search & replace overlook, and TypeScript didn't catch it for us because we were annotating the type manually instead of using inference 🙈

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -2489,7 +2489,10 @@ export function createRoom<
   TUserMeta extends BaseUserMeta,
   TRoomEvent extends Json
 >(
-  options: RoomInitializers<TPresence, TStorage>,
+  options: Omit<
+    RoomInitializers<TPresence, TStorage>,
+    "shouldInitiallyConnect"
+  >,
   config: Config
 ): InternalRoom<TPresence, TStorage, TUserMeta, TRoomEvent> {
   const { initialPresence, initialStorage } = options;
--- a/packages/liveblocks-react/src/factory.tsx
+++ b/packages/liveblocks-react/src/factory.tsx
@@ -186,7 +186,7 @@ export function createRoomContext<
           initialPresence: frozen.initialPresence,
           initialStorage: frozen.initialStorage,
           unstable_batchedUpdates: frozen.unstable_batchedUpdates,
-          withoutConnecting: frozen.shouldInitiallyConnect,
+          shouldInitiallyConnect: frozen.shouldInitiallyConnect,
         } as RoomInitializers<TPresence, TStorage>)
       );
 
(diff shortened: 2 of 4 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -2489,7 +2489,10 @@ export function createRoom<
   TUserMeta extends BaseUserMeta,
   TRoomEvent extends Json
 >(
-  options: RoomInitializers<TPresence, TStorage>,
+  options: Omit<
+    RoomInitializers<TPresence, TStorage>,
+    "shouldInitiallyConnect"
+  >,
```
nvie, line 2495: Unnecessary for the change, but strictens the type param a bit furhter. This function will ignore the `shouldInitiallyConnect` flag if you pass it.

packages/liveblocks-react/src/factory.tsx
```diff
@@ -175,19 +175,19 @@ export function createRoomContext<
       client.enter(roomId, {
         initialPresence: frozen.initialPresence,
         initialStorage: frozen.initialStorage,
-        unstable_batchedUpdates: frozen.unstable_batchedUpdates,
         shouldInitiallyConnect: frozen.shouldInitiallyConnect,
-      } as RoomInitializers<TPresence, TStorage>)
+        unstable_batchedUpdates: frozen.unstable_batchedUpdates,
```
nvie, line 179: There was no need to "hide" this API, it's already a public API, and _needs_ to be public for all people on React 17 or lower. By making this public now, we no longer have to overwrite the TypeScript inference with a forced `as`-annotation. Had we done this before, this bug would have never existed.

packages/liveblocks-react/src/factory.tsx
```diff
@@ -175,19 +175,19 @@ export function createRoomContext<
...
     React.useEffect(() => {
       setRoom(
         client.enter(roomId, {
           initialPresence: frozen.initialPresence,
           initialStorage: frozen.initialStorage,
+          shouldInitiallyConnect: frozen.shouldInitiallyConnect,
           unstable_batchedUpdates: frozen.unstable_batchedUpdates,
-          withoutConnecting: frozen.shouldInitiallyConnect,
-        } as RoomInitializers<TPresence, TStorage>)
+        })
```
nvie, line 190: Ignore the change to the type annotation and the whitespace here. The core of the fix here is the change of the key in this object: `withoutConnecting` (old name) → `shouldInitiallyConnect` (new name).

taylorpwhipp: ✅

PR #802, opened 2023-04-07
Mention in the API docs that calling these endpoints will disconnect all connected users

nvie:
Belongs to https://github.com/liveblocks/liveblocks-cloudflare/pull/405.

Screen Shot 2023-04-07 at 10 42 11@2x

Screen Shot 2023-04-07 at 10 42 57@2x


adigau (approved):
👌

PR #804, opened 2023-04-11
Open source the Liveblocks schema language

nvie:
Parser and type checker for the [Liveblocks schema language](https://liveblocks.io/docs/guides/schema-validation/syntax). This merges the private repo https://github.com/liveblocks/liveblocks-schema into this repo, making all future development public. The other repo will be archived once this merge is completed.

> **Warning** **This is not an API that users of Liveblocks would use directly.** The intended consumers for this library are the Liveblocks server backend implementation, and tools (editors, linters, prettifiers, code generators). Development of this library is early and the return value of `parse()` is unstable, which means it can receive breaking changes any time without upfront acknowledgement.

## Example

Example of the outer API that gets exposed by this:

```ts
import { parse } from "@liveblocks/schema";

const example = `
  type Storage {
    theme: "light" | "dark"
    layers: LiveList<Circle>
  }

  type Circle {
    cx: number
    cy: number
    r: number
    fill?: string
    stroke?: string
  }
`;

const doc = parse(example);
//    ^^^ CheckedDocument instance
```


FlowFlorent:
Nice! 🎉 

Do we also want to add the automatic package deployment with this PR? 

If so, the path for the package should be added here:
https://github.com/liveblocks/liveblocks/blob/main/.github/scripts/release.sh#L5
https://github.com/liveblocks/liveblocks/blob/main/.github/scripts/publish.sh#L5

nvie:
> Do we also want to add the automatic package deployment with this PR?
> 
> If so, the path for the package should be added here: https://github.com/liveblocks/liveblocks/blob/main/.github/scripts/release.sh#L5 https://github.com/liveblocks/liveblocks/blob/main/.github/scripts/publish.sh#L5

Nope! The language has its own release cadance, independent of the main packages. This is the main reason I put these new workspaces into their own new top-level directory `schema-lang/`, next to `packages/`. Each "group" are their own "version siblings" if that is a thing. With `schema-lang/`, we are still very early, and can make breaking changes a lot more carelessly. (Which is fine because we're the only user at this point still.)

I would love to make the publishing work similarly with a GitHub action, but I'll do that later.

GuillaumeSalles (approved):

PR #805, opened 2023-04-11
Upgrade Turborepo to 1.8 (and some other smaller maintenance upgrades)

nvie:
In preparation of the open sourcing of the language repo, I've upgraded a couple of utilities:

- `prettier` (trivial)
- `tsup` (trivial)
- `turbo` (1.7 -> 1.8)

Turborepo got a lot better with 1.8. The two big improvements:

1. [Workspace configurations](https://turbo.build/blog/turbo-1-8-0#workspace-configurations) - I've been able to simplify our setup a bit, by moving the specific Turbo config for our DevTools into the DevTools project itself. This was previously not possible in 1.7.
2. [Automatic Workspace Scoping](https://turbo.build/blog/turbo-1-8-0#automatic-workspace-scoping) - This means that if you cd into one of our projects, and run `turbo build`, it will automatically run `turbo --scope=<current package name>` from the project root 🥹 (I've been using a custom shell script for this, which is no longer needed 😍 )


turbo.json
```diff
@@ -10,13 +10,6 @@
       "dependsOn": ["^build"],
       "outputs": ["dist/**"]
     },
-    "@liveblocks/devtools#dev:firefox": {
-      "dependsOn": ["^build"],
-      "cache": false
-    },
-    "@liveblocks/devtools#build:firefox": {
-      "dependsOn": ["^build"]
-    },
```
nvie, line 19: @marcbouchenoire These no longer have to pollute the global `turbo.json`…

packages/liveblocks-devtools/turbo.json
```diff
@@ -0,0 +1,13 @@
...
+  "pipeline": {
+    "dev:firefox": {
+      "dependsOn": ["^build"],
+      "cache": false
+    },
+    "build:firefox": {
+      "dependsOn": ["^build"]
+    }
+  }
+}
```
nvie, line 13: …instead, you can "extend" the global version, and make these local, so this config lives in the workspace where these commands make sense.

PR #814, opened 2023-04-17
Naming cleanup of test helpers

nvie:
Renames a few of our exposed `assert` methods, to be more clear in call sites. What "assert" means depends on which prep function was used.

This PR is just a naming cleanup. I want to make more fundamental changes to our test setup, but that's not the point of this PR just yet.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/immutable.test.ts
+++ b/packages/liveblocks-core/src/__tests__/immutable.test.ts
@@ -87,7 +87,7 @@ export async function prepareStorageImmutableTest<
     itemsCount?: number,
     storageOpsCount?: number
   ) {
-    assertStorage(data);
+    expectStorageInBothClients(data);
 
     if (itemsCount !== undefined) {
       expect(machine.getItemsCount()).toBe(itemsCount);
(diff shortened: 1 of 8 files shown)


packages/liveblocks-core/src/__tests__/immutable.test.ts
```diff
@@ -82,25 +82,25 @@ export async function prepareStorageImmutableTest<
...
+  function expectStorageAndStateInBothClients(
     data: ToJson<TStorage>,
     itemsCount?: number,
     storageOpsCount?: number
   ) {
-    assertStorage(data);
+    expectStorageInBothClients(data);
 
-    if (itemsCount) {
+    if (itemsCount !== undefined) {
```
jrowny, line 92: This is nitpicky but in general I prefer to use typeof var !== "undefined" as it works more consistently, although in this case it makes no difference.

jrowny (approved):

packages/liveblocks-core/src/__tests__/immutable.test.ts
nvie: You mean to protect against cases where someone would define a global variable called `undefined`?

nvie: Haha, yes. I don’t know how we did things without TypeScript 😬

PR #815, opened 2023-04-17
Flip callback-based helper signatures inside-out…

nvie:
...so they look more like our other promise-based helper APIs.

Honestly not sure if there was a reason why these test helpers were structured like this, but using a simple return-value based structures makes these a lot easier to work with, and look more consistent with the other test helpers we have.

Best reviewed using `?w=1`, since most lines are indentation changes only.


nvie:
For clarity, what I mean by "best reviewed with `?w=1`" is this "hide whitespace-only lines from the diff" setting:

Screen Shot 2023-04-19 at 11 57 46@2x

It stems from the old days of GitHub, from before they added this UI for it. Thanks @taylorpwhipp for asking, I should have explained that.

jrowny (approved):
much cleaner!

Issue #731, opened 2023-03-09
[BUG] LiveList insert and set the same block events quickly


weiningning:
I encountered a problem when developing collaborative editors with Liveblocks and slatejs in our Web App[[Allset](https://www.allset.day/)]

## Describe the bug

Locally, LiveList inserts blockA + sets blockA quickly, and the local subscribe will receive two updates.

These two updates have passed, and the business layer will use these two ops, leading to bugs.

https://user-images.githubusercontent.com/20291747/223987967-e4b75e1a-b9cf-4ab7-9ce9-364a0ade86c6.mov

The  editor example of Liveblocks also has the same problem：

https://user-images.githubusercontent.com/20291747/223988106-0dd9ac26-cc54-476e-b0ef-63f417401094.mov

## Expected behavior

Locally does not accept any updates changed by itself.

[Currently, when inserting blockA and setting blockA, there is an interval between them, and the local will not receive any updates, which meets the requirements.]

## P**roblem a**nalysis

Looking at the Liveblocks source code, my analysis is:

### Operating block

1. Insert blockA "op.id:4248:0":
    - Add block "op.id:4248:0" to this._pool.
2. Set blockA "op.id:4248:1":
    - Detach blockA "op.id:4248:0" from this._pool.
    - Attach "op.id:4248:1" to this._pool.

At this time, this._pool only records the current "op.id:4248:1" block.

### onMessage  callback

Then, onMessage listens to messages from ws:

Receive two updates at a time, one is the previous insert blockA "op.id:4248:0" and the other is set blockA "op.id:4248:1".

applyOp:

- Op is insert blockA "op.id:4248:0":
    - When _applyInsertAck:
    Because this.__pool deleted blockA "op.id:4248:0" when blockA was assigned "op.id:4248:1"), this__pool now lacks blockA "op.id:4248:0", so a blockA "op.id:4248:0" is created and an update is generated. The logic goes to the red box below.
![Untitled](https://user-images.githubusercontent.com/20291747/223988627-2f2676a9-b72b-4a84-bfeb-2c2a9e765da6.png)

*I think this op is not found in  this._pool, and there is already a block at the same position, so this op can be considered as a past action, and it can be discarded directly. I don't know if this will cause other problems.*

- Op is set blockA "op.id:4248:1":
    - When _applySetAck,
[...]

PR #819, opened 2023-04-18
Fix "ghosting bug"

nvie:
There is a bug where a specific succession of quickly sent events can cause too many storage update events to get fired. While the updates ultimately lead to the correct end result, the in-between states that happen cause the storage state to temporarily be incorrect in the client. This bug manifests in an application as something described as "ghosting". The application seems to "flicker" or jump to incorrect state and quickly recover afterwards.

It addresses the symptoms described by Fermat on the customer call we had with them last week. Bug reports #628 and #731 seem to be incarnations of the same bug.

Commit eefca593aa8d3fd56784a811755291110272e17c adds a unit test, to try to capture the issue in a failing unit test. This is mostly the work of Eric, but I adjusted it a bit to be a little more explicit, and to capture another case.

Next, commit 70ccf3df317834aa8c10930c8cd25ddcb5c5d56c is a potential fix that I tried to make the tests pass. However, I'm not sure if this is the correct fix. ~While the unit tests all pass,~ Hmm, nevermind 🤔 The E2E suite definitely fails, fortunately! (I hadn't been running those locally, because they're so slow.)

Definitely shows I haven't found the fix yet.

---

## How should this be manually tested?

To see the weird behavior yourself, do this:

1. Check out this PR
2. `cd examples/nextjs-block-text-editor-advanced`
3. Run `link-example-locally.sh` (this will create a temporary local commit, please remove this local commit when you're done testing)
4. Run `turbo dev` to run the example locally (built against the version of `@liveblocks/client` from this PR)

### DIFF
--- a/packages/liveblocks-core/src/crdts/LiveList.ts
+++ b/packages/liveblocks-core/src/crdts/LiveList.ts
@@ -1,4 +1,4 @@
-import { nn } from "../lib/assert";
+import { assertNever, nn } from "../lib/assert";
 import { nanoid } from "../lib/nanoid";
 import { comparePosition, makePosition } from "../lib/position";
 import type { CreateChildOp, CreateListOp, CreateOp, Op } from "../protocol/Op";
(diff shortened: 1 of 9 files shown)


packages/liveblocks-core/src/crdts/LiveList.ts
```diff
@@ -585,31 +585,44 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
...
+  _dispatchByIntent(op: CreateChildOp, source: OpSource): ApplyResult {
     if (op.intent === "set") {
       if (source === OpSource.REMOTE) {
-        result = this._applySetRemote(op);
-      } else if (source === OpSource.ACK) {
-        result = this._applySetAck(op);
+        return this._applySetRemote(op);
+      } else if (
+        source === OpSource.PURE_ACK ||
+        source === OpSource.PATCHED_ACK
```
nvie, line 602: I've changed all call sites that used `ACK` to use _either_ of these two, but I think what we'll want to do is to go over each use case, and decide which of the two cases we are intending to handle here. My assumption is that we're not interested in `PURE_ACK`, and only want to handle `PATCHED_ACK`, but let's talk about this in depth.

nvie:
Fixed with a different approach, in #1177.

PR #821, opened 2023-04-20
Update Vercel "should deploy" script

nvie:
This updates the Vercel "should deploy" script to fail with different exit codes for different errors, so we can see which one is failing. All we know is that it's [failing with exit code 2](https://vercel.com/liveblocks/examples-nextjs-whiteboard/3xViCNVWjC9HDneZdVV7HDsXXisz) now, which is not that useful.

### DIFF
--- a/scripts/should-example-rebuild-on-vercel.sh
+++ b/scripts/should-example-rebuild-on-vercel.sh
@@ -48,12 +48,12 @@ shift $(($OPTIND - 1))
 
 if [ "$#" -lt 1 ]; then
     usage
-    exit 2
+    exit 3
 elif [ "$#" -gt 1 ]; then
     shift 1
     err "Superfluous arguments: $@"
     usage
-    exit 2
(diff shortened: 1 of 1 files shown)


marcbouchenoire (approved):
(I also fixed the whiteboard example that still had the script as `react-whiteboard` instead of `nextjs-whiteboard` in Vercel)

nvie:
LOL, I just figured that out!

Screen Shot 2023-04-20 at 09 44 26@2x

nvie:
@marcbouchenoire Thanks for the fix!

nvie:
@marcbouchenoire Should we also use this for the starter kit deployment, btw? It's currently the only one that always gets deployed, but I don't think it relies on the local workspaces, does it? Would be nice to save the extra carbon emissions and free up our Vercel queues :)

Screen Shot 2023-04-20 at 10 16 56@2x

Issue #823, opened 2023-04-20
Formalize the connection state machine


nvie:
The client currently uses a home-grown state machine implementation, but it isn't formalized, so it's hard to reason about. I think we would benefit from such formalization, and be very explicit about each state transition and its effects, to prevent connectivity issues and to be able to reason about possible client states. It will also help us explain the behavior of the client better.

nvie:
I've added the ideal formalized state machine as this [Stately.ai model](https://stately.ai/registry/editor/e6c62b06-19f2-44fe-9f9a-d69472d262fd?mode=Design&machineId=cee0b455-b631-4be9-a614-07789f711b01). There is no implementation for this one yet, but I think it offers some substantial benefits and simplifications over the current implementation.

PR #828, opened 2023-04-21
Prepare for formalization of state machine

nvie:
In preparation of #823, I did a few "cleanup" passes over the current state machine implementation. **I haven't changed any functionality here**—all changes here are pure refactorings, like renames, or moving code around that is internal. I've started adopting some of the XState terminology, like using the term `context` for the "extended state" being tracked in the Finite State Machine. I'll continue with this work next week, and then I'll use more terminology from XState, like "invoked actors", "transitions", etc in an effort to further formalize the machine.

The reason I'm opening this PR, is to remove a lot of noise from that PR. This PR should be able to be merged as-is, without any observable changes to the public APIs and/or behavior. Having this PR merged will make reviewing the real changes next week a lot easier / less noisy.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -21,6 +21,7 @@ import type { _private_Effects as Effects } from "../room";
 import {
   _private_makeStateMachine as makeStateMachine,
   createRoom,
+  makeClassicSubscribeFn,
 } from "../room";
 import type { Others } from "../types/Others";
 import { WebsocketCloseCodes } from "../types/WebsocketCloseCodes";
@@ -98,7 +99,14 @@ function setupStateMachine<
     initialPresence,
     undefined // no initialStorage
   );
-  return { machine, state: machine.state, effects };
+  return {
+    machine: {
+      ...machine,
+      subscribe: makeClassicSubscribeFn(machine),
+    },
+    state: machine.state,
+    effects,
+  };
 }
 
 describe("room / auth", () => {
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -18,7 +18,7 @@ import type { IdTuple, SerializedCrdt } from "../protocol/SerializedCrdt";
 import { CrdtType } from "../protocol/SerializedCrdt";
 import { ServerMsgCode } from "../protocol/ServerMsg";
 import {
-  _private_defaultState as defaultState,
+  _private_defaultMachineContext as defaultMachineContext,
   _private_makeStateMachine as makeStateMachine,
   createRoom,
 } from "../room";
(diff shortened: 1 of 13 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -67,31 +67,31 @@ type AuthCallback = (room: string) => Promise<{ token: string }>;
...
     }
   /* Connection lost unexpectedly, considered a temporary hiccup, will retry */
-  | { state: "unavailable" }
+  | { status: "unavailable" }
   /* Connection failed due to known reason (e.g. rejected). Will throw error, then immediately jump to "unavailable" state, to attempt to reconnect */
-  | { state: "failed" };
+  | { status: "failed" };
 
-export type ConnectionState = Connection["state"];
+export type ConnectionStatus = Connection["status"];
```
nvie, line 94: While this type is renamed here, there is no impact on the public types we expose, so this isn't a breaking change. `Connection` and `ConnectionStatus` are not publicly exported types.

As you can see, these exist in the bundle as _declared_, but not _exported_ types:

Screen Shot 2023-04-21 at 17 35 46@2x

The _contents_ of `ConnectionStatus` is `"closed" | "authenticating" | ...`, so where we use those types, we not breaking any contracts. Therefore, this name change is fine.

packages/liveblocks-core/src/room.ts
```diff
@@ -2483,7 +2471,7 @@ export type InternalRoom<
   onVisibilityChange: (visibilityState: DocumentVisibilityState) => void;
 };
 
-export function createRoom<
+export function createRoomMachine<
```
jrowny, line 2474: I'm on board with most of this renaming, but personally I don't think we should rename the room to roomMachine. I think it's fairly obvious now what is is through types, especially if you return a `RoomMachine` type.

nvie: Sure, can change it back. There used to be a difference between "internal room" and "public room" with different APIs, think:

```tsx
const internalRoom = createRoom();

// The public Room API, as documented on
// https://liveblocks.io/docs/api-reference/liveblocks-client#Room
const room = internalRoom.room;
```

With this function name change, I wanted to stress that this function creates the _internal_ room type, aka the "RoomMachine".

But the more I'm refactoring this, though, the less of a difference there seems to be needed here. I think we may be able to ditch the idea of an "internal room" altogether and just return the (public) room API directly from this function.

I'll refactor this a bit more today, and see if I can remove the idea of the "internal room" (aka the "room machine") and instead just make the public room API. That will help to simplify the implementation a bit more internally.

packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -190,20 +213,20 @@ describe("room / auth", () => {
   });
 
   test("private authentication that does not returns json should throw", async () => {
-    const room = createRoom(
+    const controlledRoom = createTestableRoom(
```
nvie, line 216: Renamed `room` to `controlledRoom` here, to make it more obvious that this isn't a `Room` instance.

packages/liveblocks-core/src/client.ts
```diff
@@ -124,7 +124,7 @@ export function createClient(options: ClientOptions): Client {
 
   const rooms = new Map<
     string,
-    InternalRoom<JsonObject, LsonObject, BaseUserMeta, Json>
+    Room<JsonObject, LsonObject, BaseUserMeta, Json>
```
nvie, line 127: This structure now directly holds and returns `Room` instances. Private methods (like `.connect()` and `.disconnect()`) that are needed for the internal implementation are no longer sent in an enclosing wrapper type (e.g. `InternalRoom`), but now just live on `room.__internal.privateMethod()`, like some other internal methods that already lived there. These methods are not supposed to be called from the outside, and they will be invisible to TypeScript users.

jrowny (approved):
looks good, the types are a lot simpler now without internalRoom

PR #830, opened 2023-04-24
Prepare for formalization of state machine (continued)

nvie:
More of the same stuff that #828 is made of. Again, this only contains pure refactorings. This is a continuation of cleaning up the room's internals in order to simplify them and document them, so that making the real refactoring will soon be much easier. This is likely not the last one. I'll keep doing this until the switch to a more formal state machine is less daunting to make, and we can do so with more confidence. (In the spirit of Kent Beck: _"First make the change easy, then make the easy change"_.)

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -687,8 +687,16 @@ type MachineContext<
   } | null;
   lastConnectionId: number | null; // TODO: Move into Connection type members?
   socket: WebSocket | null;
-  lastFlushTime: number;
+
+  /**
+   * All pending changes that yet need to be synced.
+   */
   buffer: {
+    // When the last flush happened. Together with config.throttleDelay, this
+    // will control whether the next flush will be sent out immediately, or if
+    // a flush will get scheduled for a few milliseconds into the future.
+    readonly lastFlushedAt: number;
+
     // Queued-up "my presence" updates to be flushed at the earliest convenience
     me:
       | { type: "partial"; data: Partial<TPresence> }
@@ -844,13 +852,14 @@ function makeStateMachine<
     lastConnectionId: null,
     socket: null,
     numberOfRetry: 0,
-    lastFlushTime: 0,
     timeoutHandles: {
       flush: undefined,
       reconnect: undefined,
       pongTimeout: undefined,
     },
+
     buffer: {
+      lastFlushedAt: 0,
       me:
         // Queue up the initial presence message as a Full Presence™ update
         {
@@ -860,6 +869,7 @@ function makeStateMachine<
       messages: [],
       storageOperations: [],
     },
+
     intervalHandles: {
       heartbeat: undefined,
     },
@@ -1881,10 +1891,9 @@ function makeStateMachine<
     }
 
     const now = Date.now();
+    const elapsedMillis = now - context.buffer.lastFlushedAt;
 
-    const elapsedTime = now - context.lastFlushTime;
-
-    if (elapsedTime > config.throttleDelay) {
+    if (elapsedMillis > config.throttleDelay) {
       const messages = flushDataToMessages(context);
 
       if (messages.length === 0) {
@@ -1892,18 +1901,18 @@ function makeStateMachine<
       }
       effects.send(messages);
       context.buffer = {
+        lastFlushedAt: now,
         messages: [],
         storageOperations: [],
         me: null,
       };
-      context.lastFlushTime = now;
     } else {
       if (context.timeoutHandles.flush !== null) {
         clearTimeout(context.timeoutHandles.flush);
       }
 
       context.timeoutHandles.flush = effects.delayFlush(
-        config.throttleDelay - (now - context.lastFlushTime)
+        config.throttleDelay - elapsedMillis
       );
     }
   }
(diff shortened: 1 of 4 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -1730,12 +1723,10 @@ function makeStateMachine<
...
 
-    clearTimeout(context.timeoutHandles.pongTimeout);
-    clearInterval(context.intervalHandles.heartbeat);
-    if (context.timeoutHandles.flush) {
-      clearTimeout(context.timeoutHandles.flush);
-    }
-    clearTimeout(context.timeoutHandles.reconnect);
+    clearTimeout(context.timers.flush);
+    clearTimeout(context.timers.reconnect);
+    clearInterval(context.timers.heartbeat);
```
jrowny, line 1728: in the original code it was more clear that heartbeat was an interval, and thus needed to be cleared via `clearInterval`. Maybe instead of moving it to timers, put it in a `context.intervals`.

packages/liveblocks-core/src/room.ts
```diff
@@ -1730,12 +1723,10 @@ function makeStateMachine<
...
     context.socket = null;
 
-    clearTimeout(context.timeoutHandles.pongTimeout);
-    clearInterval(context.intervalHandles.heartbeat);
-    if (context.timeoutHandles.flush) {
-      clearTimeout(context.timeoutHandles.flush);
-    }
-    clearTimeout(context.timeoutHandles.reconnect);
+    clearTimeout(context.timers.flush);
+    clearTimeout(context.timers.reconnect);
```
jrowny, line 1727: if it's a rule that all timers must are cleared on close, then maybe just

`Object.values(context.timers).forEach(clearTimeout)`; 

This way if we add more timers in the future no further code changes are needed.

packages/liveblocks-core/src/room.ts
```diff
@@ -1603,9 +1573,33 @@ function makeStateMachine<
...
+
+    notify(result.updates, batchedUpdatesWrapper);
+
+    effects.send(messages);
+  }
+
   function onMessage(event: MessageEvent<string>) {
     if (event.data === "pong") {
-      clearTimeout(context.timeoutHandles.pongTimeout);
+      clearTimeout(context.timers.pongTimeout);
```
jrowny, line 1602: If I am understanding this correctly, we don't reschedule here because any activity will mean we can wait for the heartbeat to run again? Shouldn't we also restart the heartbeat interval on pong? Otherwise the heartbeat might happen like 1 second after this and then we've got a ping/pong happening when we don't really need it.

packages/liveblocks-core/src/room.ts
```diff
@@ -1868,47 +1862,50 @@ function makeStateMachine<
...
+    if (context.socket) {
+      context.socket.removeEventListener("open", onOpen);
+      context.socket.removeEventListener("message", onMessage);
+      context.socket.removeEventListener("close", onClose);
+      context.socket.removeEventListener("error", onError);
+      context.socket.close();
+      context.socket = null;
+    }
 
-    messages.push({
```
jrowny, line 1895: this part of the diff got really whacky, I think that this is remnants of `applyAndSendOps` which is further up in the file, right?

packages/liveblocks-core/src/room.ts
```diff
@@ -1868,47 +1862,50 @@ function makeStateMachine<
...
-      type: ClientMsgCode.UPDATE_STORAGE,
-      ops: result.ops,
-    });
+    clearTimeout(context.timers.flush);
+    clearTimeout(context.timers.reconnect);
+    clearInterval(context.timers.heartbeat);
+    clearTimeout(context.timers.pongTimeout);
```
jrowny, line 1896: same code as in onClose right? Maybe we need a cleanup function or something to abstract that away so they don't diverge.

jrowny (approved):
Looks good, My comments don't necessarily need to be addressed in this PR and can be part of further refactoring efforts as they're not newly introduced things. Nice work!

packages/liveblocks-core/src/room.ts
nvie: Can do, but I think what this will evolve into is some kind of grouping of the heartbeat + pong timers. These two things have a relationship, and the the reconnect timer has a different trigger. In the eventual state machine this will evolve into, I'd love to make the heartbeat and pong handler part of the enter/exit criteria for the "connected" phase. The reconnect timer is part of a different state of the state machine, where the reconnect backoff strategy gets implemented.

So I think no matter how we organize them here now, I think they will eventually get separated again by function more than by the fact that they are timers.

packages/liveblocks-core/src/room.ts
nvie: See previous comment. I think the fact they they are timers is a lousy grouping mechanism, and they should become part of different states in the eventual state machine instead.

packages/liveblocks-core/src/room.ts
nvie: I think that's correct, and it would be nice to do that indeed 👍

packages/liveblocks-core/src/room.ts
nvie: Oh wow yeah. I moved some functions to colocate them, just to get a better overview during my changes. There is no new behavior introduced here. For a better view, it may be more useful to look at each commit individually instead of looking at the final diff.

PR #836, opened 2023-04-25
Avoid another `any` in the public code

nvie:
Removes the last use of `any` in `src/room.ts`. This was probably typed this way because our unit test's `MockWebSocket` somehow is not assignable to this type. Need to dig in to deeper understand why not, but that's no reason to ship a public API with an `any` in there.

This will be the impact on our published types:

Screen Shot 2023-04-25 at 15 03 48@2x

Technically speaking, that's a breaking change, but one that will only be a breaking change to people that are using this API (small %), and only to those who are using it incorrectly (even smaller %). Given this very narrow use case, I think we can publish this as a patch-level release?


jrowny (approved):
I'm not sure on releasing this as patch level, but I do think it's a reasonable change.

nvie:
> I'm not sure on releasing this as patch level

Perhaps that's right. I'll wait with merging this until we have a bit more substance and we can release it as 1.1, and combined it with a few more changes.

GuillaumeSalles:
This change makes sense. Can you double-check to see if `WebSocket` from the node module `ws` is assignable to the web `WebSocket` implementation? Just to see if we need to change this doc https://liveblocks.io/docs/api-reference/liveblocks-client#createClientNode

Releasing as 1.1.0 makes sense even if not strictly semantic

GuillaumeSalles (approved):

nvie:
@jrowny @GuillaumeSalles Re-requesting another review here, since the changes I made since your approval are non-trivial.

packages/liveblocks-core/src/types/IWebSocket.ts
```diff
@@ -0,0 +1,60 @@
...
+  readonly readyState: number;
+
+  addEventListener(type: "close", listener: (this: IWebSocketInstance, ev: IWebSocketCloseEvent) => unknown): void; // prettier-ignore
+  addEventListener(type: "message", listener: (this: IWebSocketInstance, ev: IWebSocketMessageEvent) => unknown): void; // prettier-ignore
+  addEventListener(type: string, listener: (this: IWebSocketInstance, ev: IWebSocketEvent) => unknown): void; // prettier-ignore
+  //                     ^^^^^^ Includes "open" and "error"
+
+  removeEventListener(type: "close", listener: (this: IWebSocketInstance, ev: IWebSocketCloseEvent) => unknown): void; // prettier-ignore
+  removeEventListener(type: "message", listener: (this: IWebSocketInstance, ev: IWebSocketMessageEvent) => unknown): void; // prettier-ignore
+  removeEventListener(type: string, listener: (this: IWebSocketInstance, ev: IWebSocketEvent) => unknown): void; // prettier-ignore
```
jrowny, line 30: instead of string, would it be better to do `"open" | "error"`?

shared/eslint-config/restricted-syntax.js
```diff
@@ -10,6 +10,12 @@ module.exports = [
     message: "Don't refer to AbstractCrdt as a type. Use LiveNode instead.",
   },
 
+  {
+    selector: "TSTypeReference[typeName.name='WebSocket']",
```
jrowny, line 14: since this is a shared config, is there any chance this may leak out into a repo where we do want to be able to use the original websockets?

jrowny (approved):
Looks good to me, had a few questions but no blockers.

packages/liveblocks-core/src/types/IWebSocket.ts
nvie: Good point! I initially tried doing this, but thought it didn't work, but it must have been something else. I changed it in 96f6d21ae0040bdaa2a1f400656bd6b46e933b20, and it still works. Thanks! 🙏

PR #837, opened 2023-04-25
Only send messages on OPEN sockets

nvie:
Matches the same guard we have in place for sending `ping` socket messages. A socket can be in either of these 4 states during its lifetime:

- `CONNECTING` - ⚠️ Calling `.send()` here can throw ← **This PR fixes that!**
- `OPEN`
- `CLOSING` - Calling `.send()` here won't throw (silently discards)
- `CLOSED` - Calling `.send()` here won't throw (silently discards)

We'll only want to call `.send()` when we have an `OPEN` connection. During the `CONNECTING` phase, calling `socket.send()` can throw an exception.

There is no risk for a race condition here, because once the socket reaches `CLOSING` or `CLOSED` phases, calling `socket.send()` will no longer throw. Instead, those messages won't be received, and we can expect those messages to be re-sent after reconnection, later in the future.

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -1016,7 +1016,9 @@ function makeStateMachine<
       if (context.socket === null) {
         throw new Error("Can't send message if socket is null");
       }
-      context.socket.send(JSON.stringify(messageOrMessages));
+      if (context.socket.readyState === context.socket.OPEN) {
+        context.socket.send(JSON.stringify(messageOrMessages));
+      }
     },
     delayFlush(delay: number) {
       return setTimeout(tryFlushing, delay);


packages/liveblocks-core/src/room.ts
```diff
@@ -1016,7 +1016,9 @@ function makeStateMachine<
       if (context.socket === null) {
         throw new Error("Can't send message if socket is null");
       }
-      context.socket.send(JSON.stringify(messageOrMessages));
+      if (context.socket.readyState === context.socket.OPEN) {
```
nvie, line 1019: What we're _really_ checking here is how far along its lifecycle this WebSocket is. In a way, we're checking

```suggestion
      if (context.socket.readyState >= context.socket.OPEN) {
```

here. Since the state diagram for WebSockets looks like this:

```mermaid
stateDiagram-v2
    direction LR
    [*] --> CONNECTING
    CONNECTING --> OPEN
    OPEN --> CLOSING
    CLOSING --> CLOSED
    CLOSED --> [*]
```

There is no way to "go back" a phase.

packages/liveblocks-core/src/room.ts
```diff
@@ -1016,7 +1016,9 @@ function makeStateMachine<
...
+        context.socket.send(JSON.stringify(messageOrMessages));
```
jrowny, line 1020: should we log or queue this message? I understand in the current code this might throw, but maybe we should also handle that? I'm afraid this might mask an undesirable behavior rather than solving it.

jrowny (changes requested):

packages/liveblocks-core/src/room.ts
nvie: Yes, I think we do want to do that. I didn't want to go on a tangent there for this PR, but this is something I think we need to do better as part of the stability improvement. Still not sure what the final result for that will look like, but I'll put this on my todo list.

jrowny (approved):

PR #843, opened 2023-04-26
Make unit tests tests fewer implementation details

nvie:
Some minor refactorings, mostly of our test helpers. I have a bunch more local changes that aren't ready to be committed just yet, but these are the ones that should be able to land already. This changes the testing setup a bit, so it's more obvious where we're testing implementation details. Most of the machine's `context` is now hidden away already, the goal is to make it entirely opaque to unit tests and only use publicly observable properties and behavior in those tests.


nvie:
Closing this in favor of an upcoming PR that contains broader refactorings of the internals.

PR #845, opened 2023-04-27
Tidy & standardise all code snippets in API docs

CTNicholas:
Our code snippets could be better.

### DIFF
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -52,6 +52,7 @@ const client = createClient({
       },
       body: JSON.stringify({ room }),
     });
+
     return await response.json();
   },
 });
@@ -77,7 +78,7 @@ import { createClient } from "@liveblocks/client";
 import { decode } from "base-64";
 
 const client = createClient({
-  /* ... other options ... */
+  // Other options
   polyfills: {
     atob: decode,
   },
@@ -104,7 +105,7 @@ import fetch from "node-fetch";
 import WebSocket from "ws";
 
 const client = createClient({
-  /* ... other options ... */
+  // Other options
   polyfills: {
     fetch,
     WebSocket,
@@ -128,7 +129,7 @@ It is possible to override that configuration with the `throttle` option.
 import { createClient } from "@liveblocks/client";
 
 const client = createClient({
-  /* ... other options ... */
+  // Other options
   throttle: 16,
 });
 ```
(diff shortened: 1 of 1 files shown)


docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -77,7 +78,7 @@ import { createClient } from "@liveblocks/client";
 import { decode } from "base-64";
 
 const client = createClient({
-  /* ... other options ... */
+  // Other options
```
nvie, line 81: I do feel the `/* ... */` style block works better here, because it's a placeholder for where code needs to be written. The `// Other options` line comment is something I would read as describing the lines below it, which is not the case here.

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -104,7 +105,7 @@ import fetch from "node-fetch";
 import WebSocket from "ws";
 
 const client = createClient({
-  /* ... other options ... */
+  // Other options
```
nvie, line 108: Ditto.

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -128,7 +129,7 @@ It is possible to override that configuration with the `throttle` option.
 import { createClient } from "@liveblocks/client";
 
 const client = createClient({
-  /* ... other options ... */
+  // Other options
```
nvie, line 132: Ditto.

nvie (approved):
I like most of these updates, except the one that I left a comment on. Thanks for cleaning up these snippets! ✨ 🧽 ✨

CTNicholas:
> I like most of these updates, except the one that I left a comment on. Thanks for cleaning up these snippets! ✨ 🧽 ✨

Thank you @nvie! When I have a chance I'll go through and update every other page too.

100% agree on your recommendation, thanks for pointing it out! Might change it to this, so we keep the same comment style throughout (`//`):

```ts
const client = createClient({
  throttle: 16,  

  // Other options
  // ...
});
```

nvie:
> Thank you @nvie! When I have a chance I'll go through and update every other page too.

Yes, consistency FTW! 👍 

> 100% agree on your recommendation, thanks for pointing it out! Might change it to this, so we keep the same comment style throughout (`//`)

Love it, that works for me! 👍

CTNicholas:
Implemented here, much later:
- https://github.com/liveblocks/liveblocks/pull/1519

PR #852, opened 2023-05-01
Fix exported `Pos` APIs

nvie:
This PR changes the following exported private APIs, which should have been part of #565.

- Removes the `comparePosition` API
- Exposes the `asPos` API

Needed by our backend.


packages/liveblocks-core/src/index.ts
```diff
@@ -41,7 +41,7 @@ export { freeze } from "./lib/freeze";
 export type { Json, JsonArray, JsonObject, JsonScalar } from "./lib/Json";
 export { isJsonArray, isJsonObject, isJsonScalar } from "./lib/Json";
 export { asArrayWithLegacyMethods } from "./lib/LegacyArray";
-export { comparePosition, makePosition } from "./lib/position";
+export { asPos, makePosition } from "./lib/position";
```
nvie, line 44: @marcbouchenoire This is a change I needed to make to be able to properly fix the issue you ran into with https://github.com/liveblocks/liveblocks-cloudflare/pull/414. I'll need to publish this as a release first, though.

marcbouchenoire (approved):

PR #857, opened 2023-05-02
Internal prefactoring: combine `Machine` with and `Room` types

nvie:
This PR prefactors the internals of the (internal) `Machine` and (public) `Room` types to further DRY things up and to only have a single interface, the Room. **This PR introduces no observable changes, it's a pure refactoring.**

The previous internal API of the Room was (mostly) facilitated by the `Machine` type, but soon I want to extract out a formal real state machine, so I wanted to remove the "Machine" name first to be ahead of that confusion. Also, there was a lot of duplication between `Machine` and `Room`, which made it really hard to understand how it all fit together. Roughly, what previously was the `Machine` type, can now be found on the `room.__internal` object. This value should only be accessed by unit tests, and will make testing implementation details more explicit. My goal is to eventually shrink the fields on this `__internal` structure, and decrease the amount of implementation detail testing we do, but this PR does not do that yet.

The outer API has not been changed, so the exported types are still the same, compared to what's on `main`, with these small differences:

- I've introduced the `SubscribeFn` type to capture the type of the generic `room.subscribe()` function, just to get rid of the giant nested type inside the Room type definition. Makes the Room type more readable. Since this is just giving a name to the existing type definition, this is a non-breaking change. No overloads have been changed.
- I've marked all `events` fields as readonly. I consider this a bugfix. Since the `events` API is not publicly documented anywhere yet, I think we can consider this a patch-level release still.

It could be released as **v1.0.8** as-is.


packages/liveblocks-core/src/room.ts
```diff
@@ -234,170 +234,186 @@ export type BroadcastOptions = {
   shouldQueueEventIfNotReady: boolean;
 };
 
-export type Room<
+type SubscribeFn<
   TPresence extends JsonObject,
-  TStorage extends LsonObject,
+  _TStorage extends LsonObject,
```
jrowny, line 245: I'm curious why this type received an underscore

packages/liveblocks-core/src/room.ts
```diff
@@ -881,7 +848,7 @@ function makeStateMachine<
       storageOperations: [],
     },
 
-    connection: new ValueRef<Connection>({ status: "closed" }),
+    connection: new ValueRef({ status: "closed" }),
```
jrowny, line 852: I know this isn't introduced in this PR, but should we use an enum/const here?

packages/liveblocks-core/src/room.ts
```diff
@@ -2248,28 +2237,20 @@ function makeStateMachine<
 
     // Storage
     batch,
-    undo,
-    redo,
-    canUndo,
-    canRedo,
-    pauseHistory,
-    resumeHistory,
+    history: {
```
jrowny, line 2245: does this not change the public interface, or is this re-exported somewhere?

jrowny (commented):
It looks good to me, I just have a few questions (because I'm new 😂 ) and then happy to approve.

packages/liveblocks-core/src/__tests__/_utils.ts
```diff
@@ -80,7 +74,7 @@ export class MockWebSocket implements WebSocket {
     public url: string,
     private onSend: (message: string) => void = () => {}
   ) {
-    this.readyState = this.CLOSED;
+    this.readyState = this.CONNECTING;
```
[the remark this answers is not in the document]
nvie: This is our testing helper to mock an actual WebSocket instance, and to allow us to control it during unit testing. I'm planning to rewrite some of it to make that easier to do soon. While I was taking a look here and investigating, I noticed that it would initialize a new WebSocket connection in `CLOSED` state by default (aka when doing `new WebSocket()`). We would typically not notice, because most tests will call `ws.open()` immediately, putting it in `OPEN` state.

This behavior is incorrect however, as new WebSockets in real life will never start in `CLOSED` state, but in `CONNECTING` state instead. This little adjustment will make it behave more closely to a real WebSocket implementation.

(There are many more small issues with this mock that I would like to fix. I just changed this one here and now as it didn't have an effect on our unit tests. It's unrelated to the rest of these refactorings.)

packages/liveblocks-core/src/__tests__/_utils.ts
```diff
@@ -184,30 +178,32 @@ export async function prepareRoomWithStorage<
...
 
-  machine.connect();
-  machine.authenticationSuccess(makeRoomToken(actor, scopes), ws as any);
+  room.__internal.connect();
+  room.__internal.authenticationSuccess(makeRoomToken(actor, scopes), ws);
   ws.open();
 
   // Start getting the storage, but don't await the promise just yet!
-  const getStoragePromise = machine.getStorage();
+  const getStoragePromise = room.getStorage();
```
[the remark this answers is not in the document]
nvie: I think this is done to fire the promise, and then immediately call `room.__internal.onMessage()` on the next line, and only _then_ await the promise. Likely just an optimization? Not sure why it's written like this, this would be my guess.

packages/liveblocks-core/src/room.ts
nvie: The `TStorage` type isn't used in this definition, and this `_`-prefix avoids a linting error about it. I added the type param despite it not being used, to maintain the same param order as all of our other APIs. This seemed important for future compatibility, because even though we don't explicitly export the `SubscribeFn` type, this type is somewhat part of our public API[^1], and adding in the `TStorage` type param later might result in a breaking change if we did.

[^1]: For example, if you access `Room<P, S, U, E>["subscribe"]`.

packages/liveblocks-core/src/room.ts
nvie: It does not, in fact this structure mimicks our external/public Room API. Internally it was structured differently, probably just for legacy reasons.

packages/liveblocks-core/src/room.ts
nvie: Would probably be a good idea indeed. I may tackle this in an upcoming PR 👍

nvie:
@jrowny Thanks for the quick review. I've answered all your questions — lmk if anything is still unclear!

jrowny (approved):

PR #859, opened 2023-05-03
Tighten ESLint checks

nvie:
This PR tightens our ESLint checks, by enabling the following preset:

    @typescript-eslint/recommended-requiring-type-checking

This complements the `@typescript-eslint/recommended` preset we were already using, but this preset will use static type information from TypeScript to provide even better linting errors. These checks make it much easier to catch issues like implicit `any`s that you can shoot yourself in the foot with, or dangling promises that you aren't awaiting, or functions that are async that don't have to be, or `x as Y` annotations where `x` already is `Y`, so the `as`-clause can be safely removed, etc.

In general: more code hygiene. I've fixed a handful of the issues that popped up, and also ignored some (mostly in our unit tests). For now, at least.

I decided to tighten these checks first, before continuing with my refactoring efforts, which primarily heavily rely on our TypeScript types being accurate and reliable.

### DIFF
--- a/packages/liveblocks-core/src/immutable.ts
+++ b/packages/liveblocks-core/src/immutable.ts
@@ -204,7 +204,9 @@ export function patchLiveObjectKey<
     const nonSerializableValue = findNonSerializableValue(next);
     if (nonSerializableValue) {
       console.error(
-        `New state path: '${nonSerializableValue.path}' value: '${nonSerializableValue.value}' is not serializable.\nOnly serializable value can be synced with Liveblocks.`
+        `New state path: '${nonSerializableValue.path}' value: '${String(
+          nonSerializableValue.value
+        )}' is not serializable.\nOnly serializable value can be synced with Liveblocks.`
       );
       return;
     }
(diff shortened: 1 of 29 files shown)


packages/liveblocks-core/src/immutable.ts
```diff
@@ -304,17 +308,13 @@ function legacy_patchImmutableNode<S extends Json>(
...
-          Array.isArray(state)
-        ) {
+        if (!isJsonObject(state)) {
           throw new Error(
             "Internal: received update on LiveObject but state was not an object"
           );
         }
 
-        const newState = Object.assign({}, state as JsonObject);
+        const newState: JsonObject = Object.assign({}, state);
```
nvie, line 317: FYI: Using `as` is an unsafe construct in TypeScript, telling TypeScript you know it better which is potentially unsafe, especially the value you're casting changes in the future. Here, I'm switching to using an explicit variable annotation to cause an intermediate explicit type check, which has to always match TypeScript's infered types. If this later causes a mismatch, it will be reported. This is safer in the light of future changes than using an `as` construct.

packages/liveblocks-core/src/room.ts
```diff
@@ -970,7 +970,7 @@ export function createRoom<
         authenticationSuccess(prevToken.parsed, socket);
         return undefined;
       } else {
-        return auth(config.roomId)
+        void auth(config.roomId)
```
nvie, line 973: Chosing to _not_ await a promise should never be done carelessly/accidentally/implicitly. Here, we're deliberately ignoring the returned Promise, which we can express using the `void` keyword. This will not have a runtime presence.

The ESLint rule will force you to declare for each Promise to either `await` or `void` them, so we're explicit about how we intend to handle them.

ofoucherot (approved):
🧹

PR #861, opened 2023-05-03
Improve WebSocket mock

nvie:
This PR simplifies our `MockWebSocket` class in our unit tests. It only implements the bare minimum that's needed for the now required `IWebSocket` interface, and ensures that it's always assignable to it. I've removed all the cruft from the mock that wasn't used or needed, and tweaked a few things to make it behave more closely to a real `WebSocket` implementation.

I've also renamed the `.open()` method to `.simulateOpen()` to make it more obvious that this isn't a method that is part of the WebSocket API.

One bug I found is that you could call `ws.simulateOpen()` multiple times for a connection, which should never happen in practice, so I'm now [forbidding that explicitly](https://github.com/liveblocks/liveblocks/commit/b04fa2671dd7e062cd2c00e3aa671dc15f0311c4). Doing so [revealed a bug](https://github.com/liveblocks/liveblocks/actions/runs/4872579293/jobs/8690953670#step:7:85) in our unit test, where we were asserting a property on the wrong instance of a WebSocket. I've [fixed this unit test](https://github.com/liveblocks/liveblocks/commit/c30894fdf335b821b62dc83789e8806604d01558) to more accurately test what should be tested: that the _second instance_ of the WebSocket will move to "open" state correctly. (The unit test only passed because the old instance remained in "open" state previously.)

### DIFF
--- a/packages/liveblocks-core/src/__tests__/_utils.ts
+++ b/packages/liveblocks-core/src/__tests__/_utils.ts
@@ -24,6 +24,12 @@ import type { ServerMsg } from "../protocol/ServerMsg";
 import { ServerMsgCode } from "../protocol/ServerMsg";
 import type { _private_Effects as Effects, Room } from "../room";
 import { createRoom } from "../room";
+import type {
+  IWebSocket,
+  IWebSocketCloseEvent,
+  IWebSocketEvent,
+  IWebSocketMessageEvent,
+} from "../types/IWebSocket";
 import type { JsonStorageUpdate } from "./_updatesUtils";
 import { serializeUpdateToJson } from "./_updatesUtils";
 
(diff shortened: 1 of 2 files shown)


packages/liveblocks-core/src/__tests__/_utils.ts
```diff
@@ -50,94 +56,114 @@ function deepClone<T extends Json>(items: T): T {
...
-  removeEventListener<T extends "open" | "close" | "error" | "message">(
-    event: T,
-    callback: (event: WebSocketEventMap[T]) => void
-  ) {
-    remove(this.callbacks[event], callback as any);
+  removeEventListener(event: "message", listener: MessageListener): void; // prettier-ignore
+  removeEventListener(event: "close", listener: CloseListener): void; // prettier-ignore
+  removeEventListener(event: "open" | "error", listener: Listener): void; // prettier-ignore
+  // prettier-ignore
+  removeEventListener(event: "open" | "close" | "message" | "error", listener: Listener | MessageListener | CloseListener): void {
```
jrowny, line 112: maybe consider making an event type rather than a list of strings.

packages/liveblocks-core/src/__tests__/_utils.ts
```diff
@@ -50,94 +56,114 @@ function deepClone<T extends Json>(items: T): T {
...
-    callback: (event: WebSocketEventMap[T]) => void
-  ) {
-    this.callbacks[event].push(callback as any);
+  addEventListener(event: "message", listener: MessageListener): void; // prettier-ignore
+  addEventListener(event: "close", listener: CloseListener): void; // prettier-ignore
+  addEventListener(event: "open" | "error", listener: Listener): void; // prettier-ignore
+  // prettier-ignore
+  addEventListener(event: "open" | "close" | "message" | "error", listener: Listener | MessageListener | CloseListener): void {
+    if (event === "open") {
+      this.#openListeners.push(listener as Listener);
```
jrowny, line 98: Is there a clean way to avoid the "as" ?

I can think of one way, but it's a lot of code for little benefit.

```function isCloseListener(event: string; listener: SomeListener | AnotherListener):  is CloseListener {
  return event === "close"
}```

and then your if statement is ```if (isCloseListener(event, listener)) {
...
}```

jrowny (approved):
A few minor suggestions, but looks good!

packages/liveblocks-core/src/__tests__/_utils.ts
nvie: Unfortunately, type guards [are just as unsafe](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABDAzgZSgJxmA5gCgDcBDAGxAFMAuRcAazDgHcwBKGk8i5FRFLHLkQBvAFCJEmClBCYkUAJ4AHCnGCJOlRAF5diAORgQAWwBGFTPoDcEgPS3EgXg3AEHuiAvqNEQE-RAA8aekYWHUQAFgAmK1EYdXxUDGw8fD9WVhFxCX8AOn5iTCgUAHUYKAALfH1iUwh9dMR7RBLSUkRgYhhW4ihJcFhjCndRIA) as an `as` would be here. TypeScript happily assume you know what you're doing in both approaches, and it's on us to keep these claims accurate 😬

I dislike `as`-expressions and always try to avoid them, but at the very least, an `as`-expression will fail if you're stating something that would be clearly nonsense. This is one of the reasons why I dislike complicated function overloads.

Here, the `as`-expressions are only used to select which union member applies to each `event` param, like the three overloads dictate should already be the case.

If typeguards would be actually typesafe to use, I would agree with your comment!

packages/liveblocks-core/src/__tests__/_utils.ts
nvie: I had that first, but went back this way with string literals to more clearly see from the outside which event names exist without having to look up those names. TypeScript will protect from typos in these string literals, so the duplication is fine here I think.

PR #862, opened 2023-05-03
Refactor the room's internals a bit more

nvie:
This PR continues the refactoring of our Room's internals, which is a prereq to be able to implement the new room connectivity state machine soon. This PR still **contains no behavioral changes** — this PR is a pure refactoring. I suspect this will be the last (or at least one of the last) big refactorings, before I start making actual potentially behavioral changes, by creating the actual state machine (and making it more robust).

Put differently, merging this PR should be okay to be published as **1.0.8** (patch level release). Whereas the next upcoming PR (likely to land next week) will contain a real formalized state machine, and will have to be released as **1.1.0-beta** (and be robust against all kinds of connectivity issues).


packages/liveblocks-core/src/room.ts
```diff
@@ -2205,31 +2231,80 @@ export function createRoom<
...
+  //
+  // NOTE:
+  // The Finite State Machine's .transition() function will eventually be
+  // the central gate keeper of sanity. This function will likely be abstracted
+  // away in a proper FSM abstraction (like a separate class).
+  //
+  // The idea is that, eventually, the handler functions can only be called
+  // from here, and never from anywhere else, so this function can perform its
+  // central gatekeeping.
+  //
```
nvie, line 2243: If you squint a bit, you can see the state machine structure start to take shape here. There still is quite some work to be done before it's a real one, but this is the closest to that intended structure I can take the current implementation without starting to introduce changes that aren't solely pure refactorings, so out of scope for this PR.

packages/liveblocks-core/src/room.ts
```diff
@@ -1574,12 +1591,23 @@ export function createRoom<
...
+   */
+  function handleRawSocketMessage(event: IWebSocketMessageEvent) {
     if (event.data === "pong") {
-      clearTimeout(context.timers.pongTimeout);
-      return;
+      transition({ type: "RECEIVE_PONG" });
+    } else {
+      handleServerMessage(event);
     }
+  }
```
nvie, line 1604: Notice that this is presorting on the future state where the handling of the "pong" and handling any other incoming messages are separate parts of the room's client. Handling "pong"s is directly related to maintaining the socket connectivity, whereas handling incoming server messages can always happen, and doesn't need a state transition. Hence the split introduced here. **No behavioral change here yet, though.**

packages/liveblocks-core/src/client.ts
```diff
@@ -223,15 +223,17 @@ export function createClient(options: ClientOptions): Client {
...
+        room.__internal.simulate.navigatorOnline();
       }
     });
   }
 
   if (typeof document !== "undefined") {
     document.addEventListener("visibilitychange", () => {
-      for (const [, room] of rooms) {
-        room.__internal.onVisibilityChange(document.visibilityState);
+      if (document.visibilityState === "visible") {
```
nvie, line 233: Another subtle change: I've renamed the `onVisibilityChange` to `windowGotFocus` here, meaning it should only get called when the window is refocused, and no longer when the window loses focus. This check used to live in the handler, but now it lives as a guard here, where the event will be fired. This will match more closely with the events from our [target state diagram](https://stately.ai/registry/editor/e6c62b06-19f2-44fe-9f9a-d69472d262fd?machineId=cee0b455-b631-4be9-a614-07789f711b01&mode=Simulate).

packages/liveblocks-core/src/client.ts
```diff
@@ -198,7 +198,7 @@ export function createClient(options: ClientOptions): Client {
         global.atob = clientOptions.polyfills.atob;
       }
 
-      newRoom.__internal.connect();
+      newRoom.__internal.simulate.connect();
```
jrowny, line 201: Did you mean to call `handleConnect` here and not simulate? This is the actual client right?

packages/liveblocks-core/src/client.ts
```diff
@@ -210,7 +210,7 @@ export function createClient(options: ClientOptions): Client {
 
     const room = rooms.get(roomId);
     if (room !== undefined) {
-      room.__internal.disconnect();
+      room.__internal.simulate.disconnect();
```
jrowny, line 213: also, handleDisconnect?

packages/liveblocks-core/src/client.ts
```diff
@@ -223,15 +223,17 @@ export function createClient(options: ClientOptions): Client {
     // TODO: Expose a way to clear these
     window.addEventListener("online", () => {
       for (const [, room] of rooms) {
-        room.__internal.onNavigatorOnline();
+        room.__internal.simulate.navigatorOnline();
```
jrowny, line 226: maybe I'm just not understanding the naming convention

jrowny (changes requested):
I am not following this change exactly, maybe I'm just tripped up on the naming.

packages/liveblocks-core/src/client.ts
nvie: Agreed, this name change was made mainly from a unit test perspective. The pattern here is "sending messages to the room's internals", so it can act on it. Of course, from where we use this in the actual client, this isn't _simulating_, but _actually doing_. So I agree the naming is bad here.

Perhaps I can name this group `send`? So it reads as:

```
newRoom.__internal.send.connect()
```

That way, it may be less confusing?

Either way, what's important here is the grouping itself (independent of the name), which I've done to more clearly see which parts are for "writing to" vs "reading from" the room's internals. I'll soon replace this code with a much cleaner API anyway! (Working on that still.)

nvie: Rebased + fixed the name in f3c71caf63870221b5fc169521554dd89661cbbe. Hope this is more clear?

jrowny: yes, thank you!

jrowny (approved):

PR #864, opened 2023-05-08
Add generic state machine abstraction

nvie:
This PR adds a generic finite state machine abstraction, called `FSM`. **It does not use it yet.** In my next PR, I will _use_ an FSM instance to implement our new connection manager with this. But we can reuse this generic abstraction beyond this one direct use case. I've developed this API based around our specific needs, and optimized it for our own ergonomics, which will make it really neat to implement the new client's internals soon based on the state diagram.

Think of this abstraction as something super generic, not Liveblocks-specific. Inspired by other state machine libraries out there, but not as heavyweight, yet still as powerful. Building it ourselves avoids an external dependency, which is undeseriable for `@liveblocks/client`.

Most state management libraries define the machine using a big JSON object, but in my experience the larger the state machine gets (and they can get large in practice), the less clear those big config objects become to read.

## Basic example
```ts
const machine = new FSM({ /* context */ })
  // Define all states
  .addState('red')
  .addState('green')

  // Define all transitions
  .addTransition('red', { GO: 'green' })
  .addTransition('green', { HOLD: 'red' })
  .start();  // After calling start you can no longer configure the machine

fsm.currentState;  // 'red'
fsm.send({ type: "GO" });
fsm.currentState;  // 'green'
fsm.send({ type: "GO" });  // Throws: invalid
fsm.send({ type: "HOLD" });
fsm.currentState;  // 'red'
```

For the best overview of _all_ features, see the documentation of each method, and/or the unit test suite.

## Basic API
The machine can be configured before starting it. Once started, you cannot mutate it anymore. The class has the following methods to configure/define it:

- `addState()`
- `addTransition()`
- `addTimedTransition()` (see below)
- `onEnter()` (see below)
- `onEnterAsync()` (see below)

### Wildcard patterns for state groups
To be practical for our needs, it supports state grouping, based on wildcard patterns using dotted state name syntax:

```ts
const machine = new FSM({ /* context */ })
  // Define all states
  .addState('aaa')
  .addState('bbb.one')
  .addState('bbb.two')
  .addState('ccc')

  // Define all transitions
  .addTransition('aaa', { GO: 'bbb.one' })
  .addTransition('bbb.*', { ERROR: 'ccc' })
  .addTransition('bbb.one', { GO: 'bbb.two' })
  .addTransition('bbb.two', { GO: 'bbb.one' })

  ...
```

Here, the `"bbb.*"` wildcard pattern will allow `ERROR` transitions from _any_ substate.

### onEnter / onExit

You can assign `onEnter` and `onExit` handlers for each state, or state group! Nesting these is possible. For example:

```ts
const machine = new FSM(/* context */)
  /* Define all states… */

  .onEnter('foo.bar', () => {
    console.log('State foo.bar entered')
  })

  // onEnter functions can return a cleanup function, which will be installed as the onExit handler
  .onEnter('foo.bar', () => {
    console.log('Entered foo.bar');
    return () => { console.log('Left foo.bar'); }
  })

  // onEnter functions can be nested
  .onEnter('foo.*', () => {
    console.log('Entered the foo.* substate');
    return () => { console.log('Left the foo.* substate'); }
  })

  .start();

fsm.send(/* suppose we jump to foo.bar */);
// => Entered the foo.* substate
// => Entered foo.bar
fsm.send(/* suppose we jump to foo.qux */);
// => Left foo.bar
fsm.send(/* suppose we jump to baz */);
// => Left the foo.* substate
```

This also happens upon `fsm.stop()`, as you may expect.

### Timed transitions

You can define a timed transition for a state.

The simplest case is this:

```ts
fsm
  .addTimedTransition('green', 10000, 'red')
```

This will jump to state `red` 10 seconds after entering state `green`, _unless_ another transition out of green has happened in the mean time. This is all automatically managed with timers and cleaned up. (Internally, this nice API is just a special case of `.onEnter()`.)

### Promise-based states

The last nice feature is a "promise based state". This will trigger a promise when entering a state, and lets you define what to do when the promise resolves and when it is rejected. What's nice is that if any transition is made out of the state that triggered the promise, and then the promise resolves much later, it will be ignored automatically.

```ts
  .onEnterAsync(
    'connecting',
    () => fetch('/auth', { method: "POST" }),

    // When OK
    (result) => 'good',

    // When failed
    (err) => err.status === 401 ? 'failed' : 'backoff',
  )
```


packages/liveblocks-core/src/fsm.ts
```diff
@@ -0,0 +1,493 @@
...
+      };
+    });
+  }
+
+  private getMatches(nameOrPattern: TState | Wildcard<TState>): TState[] {
+    const matches: TState[] = [];
+
+    // We're trying to match a group pattern here, i.e. `foo.*` (which might
+    // match `foo.bar` and `foo.qux` states)
+    if (nameOrPattern.endsWith("*")) {
```
jrowny, line 272: can you do `foo.bar*`? If not, maybe split on `.` and take the last index to make sure it's entirely "*".

packages/liveblocks-core/src/fsm.ts
```diff
@@ -0,0 +1,493 @@
...
+   * Call this before changing the current state to the next state.
+   *
+   * @param levels Defines how many "levels" of nesting will be exited. For
+   * example, if you transition from `foo.bar.qux` to `foo.bar.baz`, then
+   * the level is 1. But if you transition from `foo.bar.qux` to `bla.bla`,
+   * then the level is 3.
+   */
+  private exit(levels: number | null) {
+    for (let i = 0; i < (levels ?? this.cleanupStack.length + 1); i++) {
+      this.cleanupStack.pop()?.();
```
jrowny, line 395: I don't this will work when levels is not provided because cleanupStack.length gets evaluated during each iteration while `i` keeps increasing and you will only end up emptying half the array. Usually I do `while(myarray.length) array.pop()` but in your case I think you can just set length before the loop and it will work.

nvie:
> _One argument for the JSON configuration approach is that it's easier to add types. I'm personally in favor of a large up-front setup and be able to call methods with auto-completion and guaranteed correctness. Maybe there's another way we can do add that typing but it'd probably result in a typeset as large as the config right?_

I realize I haven’t made something clear enough in the PR description and in these examples. The way I have been playing with this in my local playground (basically trying to use this API to implement the new client) is by explicitly providing type annotations for these machines in the initial constructor.

It’s a bit of a compromise between full type _inheritance_ (which you get from one giant config) and this method based API for configuration. The small price you pay is that you have to duplicate the state names and allowed events (and their payloads), but beyond that, everything is fully type-safe. Even the group wildcard patterns will be inferred correctly, and autocomplete works everywhere.

A more complete usage example showing the types too would be:

```
type State =
  | "idle.bla"
  | "idle.foo"
  | "xyz"

type Event =
  | { type: "KICK" }
  | { type: "PUNCH" }
  | { type: "SING”, song: string }
  
const fsm = new FSM<State, Event, Context>(…)
  .addTransition("
//                ^ Will auto-complete with "idle.bla", "idle.foo", "xyz", "idle.*", "*”
```

Having these explicit `State` and `Event` types at the top of your state machine implementation makes it more readable to see what the machine contains at a glance.

jrowny:
> > _One argument for the JSON configuration approach is that it's easier to add types. I'm personally in favor of a large up-front setup and be able to call methods with auto-completion and guaranteed correctness. Maybe there's another way we can do add that typing but it'd probably result in a typeset as large as the config right?_
> 
> I realize I haven’t made something clear enough in the PR description and in these examples. The way I have been playing with this in my local playground (basically trying to use this API to implement the new client) is by explicitly providing type annotations for these machines in the initial constructor.
> 
> It’s a bit of a compromise between full type _inheritance_ (which you get from one giant config) and this method based API for configuration. The small price you pay is that you have to duplicate the state names and allowed events (and their payloads), but beyond that, everything is fully type-safe. Even the group wildcard patterns will be inferred correctly, and autocomplete works everywhere.
> 
> A more complete usage example showing the types too would be:
> 
> ```
> type State =
>   | "idle.bla"
>   | "idle.foo"
>   | "xyz"
> 
> type Event =
>   | { type: "KICK" }
>   | { type: "PUNCH" }
>   | { type: "SING”, song: string }
>   
> const fsm = new FSM<State, Event, Context>(…)
>   .addTransition("
> //                ^ Will auto-complete with "idle.bla", "idle.foo", "xyz", "idle.*", "*”
> ```
> 
> Having these explicit `State` and `Event` types at the top of your state machine implementation makes it more readable to see what the machine contains at a glance.

Cool! As long as we have that I'm on board.

packages/liveblocks-core/src/fsm.ts
nvie: Ah yes, good point! I pulled it out of the conditional and before the for loop in ec2df9212d9f05ed2f51dea3df5b12588c6a5b3b. Also noticed that the `+ 1` was a remnant of a previous version, but it wasn't needed. It never broke because `[].pop()` happens to be `undefined`. Thanks! 🙏

packages/liveblocks-core/src/fsm.ts
nvie: Fixed that by including the `.` in the check in 0fd123e383a13c1a263f075ba5db6aefd4578c93. Does that look okay to you? I wanted to avoid the unnecessary string split if we don't use the parts here.

jrowny (approved):

nvie:
Added two changes here, based on practical needs:

1. 696ad4e As expected, it's better to have `.send()` ignore events that cannot be handled by the current state, than to throw. So basically I've removed `.sendIfPossible()`, and those semantics are now the default for `.send()`. Sending a completely unknown event still throws, because that would be a configuration error.
2. 3126f9c I've added event sources, so a user of this API can observe the changes within the machine from the outside, e.g. to produce a debug trace. This is going to be a major power tool for finding bugs!

PR #866, opened 2023-05-09
Improve config generation + match interactive tutorial

CTNicholas:
1. Added flags for creating config files e.g.: you can now use all these flags and skip the UI.
```
npx create-liveblocks-app@latest --init --framework react --suspense --typescript --comments
```
2. Show a friendly error if an incorrect flag is used
3. Slightly changed the config file so it works better with the interactive tutorial.


stevenfabre (approved):
Looks good!

Issue #582, opened 2022-12-06
Export types for LiveMapUpdates, LiveObjectUpdates, and LiveListUpdates


SamyPesse:
Currently, only `StorageUpdates` is exported (https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-core/src/crdts/StorageUpdates.ts#L13), but `LiveMapUpdates`, `LiveObjectUpdates`, and `LiveListUpdates` are not.
It makes it hard to implement advanced logic based on `room.subscribe(root, fn)`

**Describe the solution you'd like**

Expose the types from `@liveblock/client`.

PR #867, opened 2023-05-09
Add more update exports

jrowny:
### Description

- Exports LiveObjectUpdate, LiveMapUpdate, and LiveListUpdate types.
- Fixes a typing issue I saw
- Fixes incorrect documentation

#### How to test it

The attached docs example should show you the new update types inside of the switch statement:

Screenshot 2023-05-09 at 2 21 43 PM

#### Related issue(s)
https://github.com/liveblocks/liveblocks/issues/582


packages/liveblocks-core/src/room.ts
```diff
@@ -126,6 +126,7 @@ type RoomEventCallbackMap<
   error: Callback<Error>;
   connection: Callback<ConnectionStatus>;
   history: Callback<HistoryEvent>;
+  storage: Callback<StorageUpdate[]>;
```
jrowny, line 129: I had a typescript error because this was missing, but it's not related to my changes I don't think.

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -281,29 +281,33 @@ const unsubscribe = room.subscribe(root, (root) => {
...
+list of updates instead. Each such update is a `{ type, node, updates }` object.
 
 ```ts
 const { root } = await room.getStorage();
 const unsubscribe = room.subscribe(
   root,
-  (updates) => {
-    for (const update of updates) {
+  (storageUpdates) => {
+    for (const update of storageUpdates) {
```
stevenfabre, line 291: This is much clearer to read. Nice one JR!

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -281,29 +281,33 @@ const unsubscribe = room.subscribe(root, (root) => {
...
       const {
-        type, // "update" or "delete"
+        type, // "LiveObject", "LiveList", or "LiveMap"
         node,
+        updates,
```
stevenfabre, line 295: Reading this, I was first slightly confused as to what `updates` was. But since it's different for each type, it became clear when reading the comments for each switch case below.

packages/liveblocks-core/src/room.ts
nvie: For full context, the `.subscribe(...)` API is complicated because it has so many overloads. It has been historically really hard to type out in TypeScript.

Last year, I refactored things internally to have individual event sources for each event, so subscribing to specific events is much easier, because each event source has a specific payload type. Then, I reimplemented `.subscribe()` on top of those, as a backward-compatible layer with the API as it has always existed (aka today's API with many overloads).

These APIs also exist, and are recommended for internal use:

```ts
room.events.connection.subscribe(callback);  // .subscribe('connection', callback) calls this internally
room.events.me.subscribe(callback);          // .subscribe('my-presence') calls this internally
room.events.storage.subscribe(callback);     // .subscribe(root, callback) calls this internally
...
```

We have not advertised this API publicly (yet), but we could if we wanted to. It's stable, and here to stay, and I think slightly nicer to use. We just didn't want to cause confusion by having multiple subscription APIs. The single kitchen-sink style `room.subscribe()` API is just a wrapper around those more type-safe and cleaner APIs, and should only be used externally (as a Liveblocks user).

nvie (changes requested):
Lovely — thanks for fixing the bug in the example and making it much cleaner!

CTNicholas (approved):
Thanks for updating the docs!

packages/liveblocks-core/src/room.ts
jrowny: removed the dead code!

nvie (approved):
Nice, @jrowny — thank you! 🙏 
If we're exposing new types, let's mention this in the `CHANGELOG.md` file (you can create a new section for `1.0.9` there) as part of this change, so the next person publishing this will include that in the release notes.

PR #869, opened 2023-05-11
Updated 128 kB limits docs

stevenfabre:
The limits page on the documentation was misleading.
Only LiveObjects and JSON cannot exceed 128 kB.

### DIFF
--- a/docs/pages/guides/limits.mdx
+++ b/docs/pages/guides/limits.mdx
@@ -37,5 +37,4 @@ characters once serialized to JSON.
 
 ### Storage
 
-The value of a `LiveObject`, `LiveList`, and `LiveMap` storage item cannot
-exceed 128 kB. This limit can be increased on the Enterprise plan.
+The `LiveObject` data structure and JSON values in storage cannot exceed 128 kB.
--- a/docs/pages/guides/limits.mdx
+++ b/docs/pages/guides/limits.mdx
@@ -37,4 +37,17 @@ characters once serialized to JSON.
 
 ### Storage
 
-The `LiveObject` and JSON values in storage cannot exceed 128 kB.
+Our conflict-free data structures have a limit.
+
+- A `LiveObject` cannot exceed 128 kB when totalling the size of the keys and
+  values.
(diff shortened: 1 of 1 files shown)


docs/pages/guides/limits.mdx
```diff
@@ -37,5 +37,4 @@ characters once serialized to JSON.
 
 ### Storage
 
-The value of a `LiveObject`, `LiveList`, and `LiveMap` storage item cannot
-exceed 128 kB. This limit can be increased on the Enterprise plan.
+The `LiveObject` and JSON values in storage cannot exceed 128 kB.
```
nvie, line 40:
```suggestion
Each `LiveObject` and JSON value in storage cannot exceed 128 kB.
```

nvie (approved):
Looks good! Left a small tweak that hopefully makes this less ambiguous!

docs/pages/guides/limits.mdx
CTNicholas, line 40: I think we need more info, otherwise people won't fully understand. If @nvie confirms this is correct, I'll give it the ✅.

~~~suggestion
- A `LiveObject` cannot exceed 128 kB when totalling the size of the keys and values. 
- A `LiveMap` can be any size, so long as each individual value does not exceed 128kB.
- A `LiveList` can be any size, so long as each individual value does not exceed 128kB.

Note that when one real-time data structure is nested inside another, it does not count towards the limit. Only the JSON leaves of your data structure count towards the limit. 

For example, given this `LiveObject`:
```ts
const person = new LiveObject({
  name: "Marie",
  traits: new LiveObject({ kind: true }),
  friends: new LiveList(["Albert"]),
  age: 30,
})
```

The data limit for `person` is counted from the following:
```ts
{
  "name": "Marie",
  "age": 30,
}
```

This also applies to values inside `LiveMap` and `LiveList` data structures.
~~~

CTNicholas (changes requested):

CTNicholas (approved):

Issue #625, opened 2023-01-13
Expose helper functions to serialize Lson structure into JSON for using liveblocks api endpoints


gcrespi:
**Is your feature request related to a problem? Please describe.**

For some endpoints in liveblocks API such as [this](https://liveblocks.io/docs/api-reference/rest-api-endpoints#post-rooms-roomId-storage) you need to provide as the request body a serializable version of your Lson object. 

**Describe the solution you'd like**

It would be nice if `@liveblocks/client` would provide a function to transform Lson objects (LiveObject, LiveMap, LiveList, etc) in the needed structure, something like:

```
function convertLsonToJSON(value: Lson): JSONValue {
....
}
```

nvie:
Hi @gcrespi, thanks for opening this issue. We agree this different data format is a bit cumbersome to deal with manually. We don't have a helper in the public API because it's we may soon offer a better API to populate storage, which will require less of this juggling. In the mean time, here is a code snippet you can use to convert an LSON tree to the needed data format: https://gist.github.com/nvie/09a34a3e3b219c442a418f033ce73e3a

jrowny:
This has been completed

PR #871, opened 2023-05-12
Add toPlainLson util

jrowny:
### Description

- [x] add toPlainLson function to assist in 
- [x] documentation
- [x] tests

#### Related issue(s)
https://github.com/liveblocks/liveblocks/issues/625

### DIFF
--- a/packages/liveblocks-core/src/lib/utils.ts
+++ b/packages/liveblocks-core/src/lib/utils.ts
@@ -1,8 +1,3 @@
-import { LiveList } from "../crdts/LiveList";
-import { LiveMap } from "../crdts/LiveMap";
-import { LiveObject } from "../crdts/LiveObject";
-import type { Lson } from "../crdts/Lson";
-import type { PlainLson } from "../types/PlainLson";
 import type { Json } from "./Json";
 
 declare const brand: unique symbol;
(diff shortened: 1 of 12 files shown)


nvie (approved):
Looking good so far! 👍

packages/liveblocks-core/src/lib/utils.ts
nvie, line 1: Btw, the failure you're seeing in CI here is coming from dependency-cruiser, which is a tool to enforce some "high level" rules for which modules can import from which other modules. (We set this up, because we had some serious circular dependencies in our code base before.)

The error thrown here is called `illegal-import-from-libs`, and you can see its definition and rationale here:

https://github.com/liveblocks/liveblocks/blob/03fb96b613dc85ffd3d27d28ed9a92400f4da1b1/packages/liveblocks-core/.dependency-cruiser.js#L138-L146

This rule ensures that we don't accidentally introduce complicated dependency cycles in our packages unknowingly over time, and CI will remind us when we do accidentally, like in this case.

The solution is to put this utility function in another module, not under `src/lib/`. In this particular case, I would recommend renaming the file `src/crdts/ToImmutable.ts` to `src/crdts/utils.ts` and move this function there instead.

nvie (approved):
LGTM!

PR #872, opened 2023-05-12
Refactor internal auth response structures slightly

nvie:
This PR is a pure prefactoring. Its purpose is to make the internals of the current code base more easy to plug in to the new connection manager that is in the making.

**No behavioral changes.**

It specifically restructures the response of the `auth` + `createWebSocket` delegate pair to their input/output 100% symmetric. Until now, we've been returning `{ token: string }` in the `authenticate` promise, and having take the `createWebSocket` delegate take a `string` directly. This worked, because the client would be controlling things, eating off the `token` value and passing it to the creation function. In fact, it would also use this `{ token: string }` and turn it into a "richer" token, which it would keep itself in memory (for metadata).

With the upcoming separation of responsibilities, the connection manager (the state machine) is going to be a black box, and its requirement is that you provide it this delegate pair that have matching input/output (so they can "just" be chained without the manager itself having to know/care/understand that payload). It's just a value that is needed for authentication.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/_utils.ts
+++ b/packages/liveblocks-core/src/__tests__/_utils.ts
@@ -634,7 +634,7 @@ export function mockEffects<
   TRoomEvent extends Json
 >(): Effects<TPresence, TRoomEvent> {
   return {
-    authenticate: jest.fn(),
+    authenticateAndConnect: jest.fn(),
     send: jest.fn(),
     scheduleFlush: jest.fn(),
     scheduleReconnect: jest.fn(),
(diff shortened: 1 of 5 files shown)


packages/liveblocks-core/src/__tests__/_utils.ts
```diff
@@ -634,7 +634,7 @@ export function mockEffects<
   TRoomEvent extends Json
 >(): Effects<TPresence, TRoomEvent> {
   return {
-    authenticate: jest.fn(),
+    authenticateAndConnect: jest.fn(),
```
nvie, line 637: Changed name to better reflect what this effect was actually doing.

packages/liveblocks-core/src/room.ts
```diff
@@ -635,16 +633,14 @@ type HistoryOp<TPresence extends JsonObject> =
 
 type IdFactory = () => string;
 
-type MachineContext<
+type RoomState<
```
nvie, line 636: Still found a reference to "machine" here, which I've now removed.

nvie:
Found some more references to "machine" in our test suite. Changed these too in 12bf55cf542e33686e23f6c5add3961de31c8c60 and 07b5ab6c81d8bae48129aa8159d83d18c8a7dba3, to avoid potential confusion.

jrowny (approved):

PR #875, opened 2023-05-16
Fix HTTP status code for when `@liveblocks/node` cannot connect to our infra

nvie:
This PR changes the status code that our `authorize()` helper from `@liveblocks/node` will return when implementing your own auth backend. This is about the case when the user's backend uses this wrapper to try to relay the request to our infra, but somehow this is not possible (network down, infra down, etc). In those cases, we should return a 5xx response back to the clients, not a 4xx response, and certainly no _403_ response, which will soon indicate to the client to stop retrying. We definitely want the client to keep retrying in this case!

### DIFF
--- a/packages/liveblocks-node/src/authorize.ts
+++ b/packages/liveblocks-node/src/authorize.ts
@@ -101,7 +101,7 @@ export async function authorize(
     };
   } catch (er) {
     return {
-      status: 403,
+      status: 503,
       body: 'Call to "https://api.liveblocks.io/v2/rooms/:roomId/authorize" failed. See "error" for more information.',
       error: er as Error | undefined,
     };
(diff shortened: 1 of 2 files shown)


packages/liveblocks-node/src/authorize.ts
```diff
@@ -90,18 +91,18 @@ export async function authorize(
...
 
     return {
-      status: 200,
+      status: 200 /* OK */,
       body: await result.text(),
     };
   } catch (er) {
     return {
-      status: 403,
+      status: 503 /* Service Unavailable */,
```
nvie, line 105: The only real (and important) change here is this change of status code. That's a fixed bug. The rest is just cleanup while I was in this module.

marcbouchenoire (approved):

jrowny:
My only concern is just to make sure the server returns the right codes as well, as we just saw with mongo returning 5xx errors for the country being blocked. I'm pretty sure our backend turns that back into a 403, but I'm not sure about this exact path, I'd want to check that.

nvie:
Fixed in bd5cfc5a5b04b616113304b902648ce90e80ade6!

jrowny (approved):

PR #883, opened 2023-05-19
Remove the `scheduleFlush` effect

nvie:
Removes the `scheduleFlush` effect that is configured in the current testing setup. The goal here is to get rid of the effects and rely less on testing implementation details, and try to treat the room under test more as a black box. All the other effects (except `send()`) will be removed in the upcoming connection manager PR which I'll open soon, since those effects are all responsibility of the connection manager instead of the room soon.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -277,26 +277,40 @@ describe("room", () => {
     ]);
   });
 
-  test("initial presence followed by updatePresence should delay sending the second presence event", () => {
+  test("initial presence followed by updatePresence should delay sending the second presence event", async () => {
+    jest.useFakeTimers();
+
     const { room, effects } = createTestableRoom({ x: 0 });
 
     const ws = new MockWebSocket();
     room.__internal.send.connect();
     room.__internal.send.authSuccess(defaultRoomToken, ws);
 
-    const now = new Date(2021, 1, 1, 0, 0, 0, 0).getTime();
+    const now = Date.now();
 
-    withDateNow(now, () => ws.simulateOpen());
-
-    withDateNow(now + 30, () => room.updatePresence({ x: 1 }));
-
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -277,26 +277,40 @@ describe("room", () => {
...
+  test("initial presence followed by updatePresence should delay sending the second presence event", async () => {
+    jest.useFakeTimers();
+
     const { room, effects } = createTestableRoom({ x: 0 });
 
     const ws = new MockWebSocket();
     room.__internal.send.connect();
     room.__internal.send.authSuccess(defaultRoomToken, ws);
 
-    const now = new Date(2021, 1, 1, 0, 0, 0, 0).getTime();
```
jrowny, line 287: Makes sense, now was once 2021, but not presently.

packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -277,26 +277,40 @@ describe("room", () => {
...
-    expect(effects.scheduleFlush).toBeCalledWith(
-      makeRoomConfig().throttleDelay - 30
-    );
-    expect(effects.send).toHaveBeenCalledWith([
-      { type: ClientMsgCode.UPDATE_PRESENCE, targetActor: -1, data: { x: 0 } },
+    expect(effects.send).toHaveBeenCalledTimes(1);
+    expect(room.__internal.buffer.me?.data).toEqual({ x: 2 });
+
+    // Forwarding time by the flush threshold will trigger the future flush
+    await jest.advanceTimersByTimeAsync(makeRoomConfig().throttleDelay);
```
jrowny, line 308: this is really cool, I never really used these jest functions before, super useful! The new test code and comments are a lot easier to follow than the scheduleFlush ones.

jrowny (approved):

PR #885, opened 2023-05-19
Add `ManagedSocket` implementation

nvie:
This PR adds the implementation of the `ManagedSocket` class, aka our new Connection Manager™, aka the State Machine™.

In the next PR, I'll _use_ this class to replace the internals of the Room with it. This `ManagedSocket` implementation is the result of a lot of manual/practical testing. I consider it pretty much complete, bugs aside if they are found.

---

> **Note**
> Before reading the code, please look at the visualized state diagram [here](https://stately.ai/registry/editor/e6c62b06-19f2-44fe-9f9a-d69472d262fd?machineId=cf1e6328-1cd0-4c41-a967-24f478bf4b1c&mode=Design). You can also try the simulator to play with it. I've kept the implementation and the diagram in sync as much as possible, even though the implementation does not depend on XState. **If you don't understand a part of the implementation, refer to the visual—it hugely helps!**

State diagram visualization

---

When reviewing this work, keep in mind the context where this will be used. Each `Room` instance will have _exactly one_ `ManagedSocket` instance, which it uses internally to maintain its WebSocket connection reliably. Schematically:

The sole purpose of the connection manager is to be responsible for keeping a socket alive and refresh it by replacing it with a new `WebSocket` instances in the case of an error happening. To the user of this `ManagedSocket` (i.e. each Room instance), it's very much like a normal WebSocket, except... way more reliable.


packages/liveblocks-core/src/connection.ts
```diff
@@ -0,0 +1,756 @@
...
+
+type State =
+  | "@idle.initial"
+  | "@idle.failed"
+  | "@auth.busy"
+  | "@auth.backoff"
+  | "@connecting.busy"
+  | "@connecting.backoff"
+  | "@ok.connected"
+  | "@ok.awaiting-pong";
```
nvie, line 80: FYI, there is no meaning to these `@`'s in the strings. I just found the convention useful to instantly tell state names apart from `EVENT_NAMES` and other string literals.

packages/liveblocks-core/src/connection.ts
```diff
@@ -0,0 +1,756 @@
...
+   * The current active WebSocket connection to the room. If this is not null
+   * on the context, then the socket has successfully been opened.
+   */
+  socket: IWebSocketInstance | null;
+
+  /**
+   * The current retry delay when automatically retrying.
+   */
+  backoffDelay: number;
+  // numRetries: number; // Keep track of the number of retry attempts
```
jrowny, line 107: seems like we're using the backoff delays array, so you can just delete numRetries

packages/liveblocks-core/src/connection.ts
```diff
@@ -0,0 +1,756 @@
...
+  didDisconnect: Observable<void>;
+} {
+  // Emitted whenever a new WebSocket connection attempt suceeds
+  const statusDidChange = makeEventSource<PublicConnectionStatus>();
+  const didConnect = makeEventSource<void>();
+  const didDisconnect = makeEventSource<void>();
+
+  let oldPublicStatus: PublicConnectionStatus | null = null;
+
+  fsm.events.didEnterState.subscribe((newState) => {
```
jrowny, line 240: should we keep track of this to unsubscribe later?

packages/liveblocks-core/src/connection.ts
jrowny: I don't have a great reason for it, but I find myself wanting this to be an enum. Then you wouldn't really need to worry about differentiating with the `@` because you'd just have like `State.Idle.Initial` or something and you never really worry about the underlying string.

packages/liveblocks-core/src/connection.ts
```diff
@@ -0,0 +1,756 @@
...
+      assign: increaseBackoffDelay,
+    },
+
+    DISCONNECT: "@idle.initial",
+  });
+
+  //
+  // Configure the @idle.* states
+  //
+  fsm.addTransitions("@idle.*", {
```
jrowny, line 307: oh right, I forgot about the wildcard part of the names, I guess that makes using enums a bit harder.

packages/liveblocks-core/src/connection.ts
```diff
@@ -0,0 +1,756 @@
...
+    socket: null,
+
+    // Bumped to the next "tier" every time a connection attempt fails (no matter
+    // whether this is for the authentication server or the websocket server).
+    // Reset every time a connection succeeded.
+    backoffDelay: LOW_DELAY,
+    // numRetries: 0,
+  };
+
+  const fsm = new FSM<Context, Event, State>(initialContext)
```
jrowny, line 279: could we find a more descriptive name for the state machine? maybe connectionState or something?

packages/liveblocks-core/src/connection.ts
```diff
@@ -0,0 +1,756 @@
...
+      }),
+
+      // Auth failed
+      (failedEvent) =>
+        failedEvent.reason instanceof UnauthorizedError
+          ? {
+              target: "@idle.failed",
+              effect: () =>
+                console.error(
+                  `Unauthorized, will stop retrying: ${
```
jrowny, line 352: I don't know that we have to explicitly say we'll stop retrying here.

packages/liveblocks-core/src/connection.ts
```diff
@@ -0,0 +1,756 @@
...
+
+      () =>
+        Promise.race([delegates.authenticate(), timeoutAfter(AUTH_TIMEOUT)]),
+
+      // On successful authentication
+      (okEvent) => ({
+        target: "@connecting.busy",
+        assign: {
+          token: okEvent.data as BaseAuthResult,
+          backoffDelay: LOW_DELAY,
```
jrowny, line 341: this is just resetting it back to the quicker backoff in case of future failures right? We're not actually doing any delays or anything here on success?

packages/liveblocks-core/src/connection.ts
```diff
@@ -0,0 +1,756 @@
...
+      // no longer valid.
+      return () => {
+        teardownSocket(ctx.socket);
+        (ctx as any).socket = null;
+        //   ^^^^^^
+        //   XXX Really, I should add an official API to the FSM to support
+        //   XXX assigning to context when entering/leaving a state, since this
+        //   XXX is a well-defined moment in a state transition.
+        //   XXX I've only temporarily pulled this hack here, for which I ask
+        //   XXX forgiveness.
```
jrowny, line 523: I'm not sure if I have the authority to forgive

packages/liveblocks-core/src/connection.ts
```diff
@@ -0,0 +1,756 @@
...
+
+/**
+ * The ManagedSocket will set up a WebSocket connection to a room, and maintain
+ * that connection over time.
+ *
+ * It's a light wrapper around the actual FSM that implements the logic,
+ * exposing just a few safe actions and events that can be called or observed
+ * from the outside.
+ */
+export class ManagedSocket<T extends BaseAuthResult> {
```
jrowny, line 647: As a dev, I feel like maybe these should be in separate files. One file is my state machine definition, the other is the simple wrapper class.

jrowny (changes requested):
Overall this looks good but I'd like to see a few more tests in connection.test, maybe around the kinds of delays that are being set and transitioning from success states.

packages/liveblocks-core/src/connection.ts
nvie: Yeah, like you said in the other comment below, not sure how to combine that with the wildcard patterns. Those patterns are not just syntactic luxury, but an important construct that is needed for the machine.

packages/liveblocks-core/src/connection.ts
nvie: Agreed, fixed in e8445a96b149ea4f5bd71a3e0c7d8b12e67ac048 👍

packages/liveblocks-core/src/connection.ts
nvie: Yeah, good point. I've renamed it from `fsm` → `machine` now in 5d1955272c4353ce72bd10f85b0c44255bd69404, and documented its purpose where it's first introduced.

I tried `connectionState` and some other longer/more verbose names, but it makes the code less readable IMHO, mostly because the machine instance isn't that important. The attention should go to all the transitions being configured there.

(A bit weird maybe, but it feels a bit like how you wouldn't name a `req`/`res` arg in a request handler `createRoomReq`/`createRoomRes` either—longer names would make it less readable because it draws attention from what's really important.)

Either way, I agree `fsm` wasn't a good variable name—hope this helps!

packages/liveblocks-core/src/connection.ts
nvie: Correct, it's a reset indeed, setting it back to the lowest delay (it's never 0). In a prior iteration, I named that `LOW_DELAY` variable `RESET_DELAY`, which may make this easier. I'll change that back, since it would have been clearer. Fixed in 7117ea4066ee366487336dd735dcd171986515b7.

packages/liveblocks-core/src/connection.ts
nvie: Fixed in c44a469f6f1045b615398780765a1c15ea9e8de1 👍

packages/liveblocks-core/src/connection.ts
nvie: This started as a single class, that defined the state machine inline, but I pulled out the helper function because it was getting so large and the indentation in the class made things super hard to read. Basically, the entire [`createConnectionStateMachine()`](https://github.com/liveblocks/liveblocks/blob/c44a469f6f1045b615398780765a1c15ea9e8de1/packages/liveblocks-core/src/connection.ts#L259-L639) helper function is supposed to ["just" be the constructor](https://github.com/liveblocks/liveblocks/blob/c44a469f6f1045b615398780765a1c15ea9e8de1/packages/liveblocks-core/src/connection.ts#L677-L681)—they're not really different things. For that reason, I'd prefer to keep them in the same file, as they are just one conceptual thing and need to always change together.

nvie:
> Overall this looks good but I'd like to see a few more tests in connection.test, maybe around the kinds of delays that are being set and transitioning from success states.

Fair enough! The thing here is that those tests actually overlap with out current `room.test.ts` suite almost entirely, so my plan is to basically rewrite the existing unit tests for room so they work with these new internals. Perhaps I'm able to split the tests that are testing things around connectivity from the ones that test other room internals (like buffer flushes, or undo history, etc) and move the connectivity parts to `connection.test.ts` eventually.

Working on that! Will update this PR and mark it finished when I'm there.

nvie:
@jrowny I merged the changes from #889 into this PR, and addressed all your PR feedback as well. This is ready for a final review.

jrowny (approved):

Issue #679, opened 2023-02-09
[BUG] Client still in presence room even after laptop goes to sleep


talksik:
**Describe the bug**

I have been trying to figure out why my connections are inconsistent with turning off when a client's computer goes to sleep. Is there some delay in pinging clients to check connection?

I see over and over again where client A saw that client B was connected for hours even when client B computer was in sleep mode.

Device: macbook pro silicon chips | both of them

**To Reproduce**

1. connect client A to a room
2. connect client B to a room
3. Put client b laptop to sleep by closing lid
4. Client B is still shows as connected for client A for hours

**Expected behavior**
I should see that the client B is disconnected and a message sent for presence to the rest of the chat that the client has left.

**Environment (please complete the following information):**
Electron app on both clients. Same app same version.
liveblocks javascript vanilla client sdk

talksik:
can I please get help on this? looking to do a prod release soon :)

talksik:
@taylorpwhipp any ideas here? sorry for bugging you once again

nvie:
Hi @talksik, thanks for filing this report. We'd love to help you out to get this issue addressed.

To confirm and fully understand your bug report, is your problem that there is an inconsistency between two clients' behaviors? For example, if you close the lid of client A and observe with client B: does it work then, but not the other way around?

Or is your bug report that _any_ client that closes their lid never gets cleared from the room?

Do you have a minimal replication (perhaps an online running instance) that we could use to investigate this bug further?

talksik:
Thanks for getting back to me! 

Yep I have tried both ways. both laptops switched.

I also tried the react sdk in addition to js sdk to make sure it wasn't some client side bug. Same issue.

hmmm I don't really have anything other than our beta which you could try theoretically. It's hard to code sandbox this as it's an electron app. Feel free to download and try it with two accounts: https://flowy.live

What I have been seeing clearly is that issue of others firing inaccurate data for others as soon as a client disconnects. Maybe that will help find the bug in your servers. I believe it's what is fired when emitting a disconnect event to other participants on socket disconnect

talksik:
to provide more nuance detail I just noticed. this inaccurate firing of others a split second after the other client disconnects is likely not incorrect state in your servers. If client b refreshes their app, then the `others` event correctly returns empty array []. So:

- client a connects
- client b connects
- both see each other
- client a goes to sleep
- for 200 ms or so, client b gets an event that `others` is empty array []
[...]

Issue #414, opened 2022-07-19
Allow authentication to fail


m1guelpf:
Today, if the call to the authentication endpoint fails, it just keeps getting called forever. However, there are certain use-cases where you might want to handle the exception instead (such as when a user is trying to connect to a room they shouldn't have access to).

In order to make this customizable, we could add an extra key to the client config (`onAuthError`), which gets called with the exception thrown (or the response, if the error is a malformed response), and can return `false` to avoid retries (or anything else to retry anyway).

Digging through the codebase, this seems to be the function I'd add the logic to:

https://github.com/liveblocks/liveblocks/blob/6e5f2189c717426214577b658c838d01543e1001/packages/liveblocks-client/src/room.ts#L855-L863

ofoucherot:
Hello,
You make a good point; we should add a way to customize the logic retry. Your solution makes sense; another solution would be to add a list of error codes that should not retry. We will also have to think about if we want to add customization for the retry delay.
I added a task to our backlog, we will update you when we have a clearer idea on the implementation.

npapagna:
Hi! Just bumped into this issue and I'm checking in to see if there was any progress on this one.

TerminalFi:
Ran into this today. Seems like a much needed capability.

basicNew:
Hi folks! Are there any updates on this one? We would benefit a lot from having this. Thanks!

nvie:
Hello, we have this on our radar. I'm currently thinking about making some changes to holistically improve our client's stability and this would be a natural part of it. It may still take a bit of time before we can publish an updated version, but we're actively thinking about and working on this, so there is progress. We'll report back when we have more information to share — thanks for your patience!

Issue #105, opened 2022-03-31
Client: re-work authentication error messages


GuillaumeSalles:
There are lot of point of failures during the authentication process. 

Here is a non exhaustive list:

Public authentication:
- Unknown public key
- Public key is deactivated
- Private key used instead of public key
- Network error while calling `/api/authorize/public`

Private authentication:
- Invalid secret key
- Public key used instead of private key
- Invalid endpoint
- Network error between client and auth endpoint
- Network error in @liveblocks/node
- Invalid payload returned by authentication endpoint

Custom authentication (callback)
- Function throw
- Invalid return

Providing a descriptive error message would probably improve our activation rate.

This task can be split in smaller one to ship a few error message at a time.

PR #886, opened 2023-05-19
Reimplement connection management in the Room

### LINKED ISSUE #823: Formalize the connection state machine

nvie:
This PR replaces the Room internals related to connection management with the new `ManagedSocket` (introduced in #885), with the purpose of alleviating all known (and unknown) connectivity issues, basically making the implementation robust and reliable under all circumstances (or in cases where it's not: easier to debug at least).

Fixes #823, #105, #414, #679 and https://github.com/liveblocks/liveblocks.io/issues/1083.

## Todo

- [x] Unit tests
  - [x] Make all unit tests pass in `@liveblocks/core`
  - [x] Make all unit tests pass in `@liveblocks/client`
  - [x] Make all unit tests pass in `@liveblocks/react`
  - [x] Make all unit tests pass in `@liveblocks/redux`
  - [x] Make all unit tests pass in `@liveblocks/zustand`
  - [x] Make all unit tests pass in `@liveblocks/node`
  - [x] Make all E2E tests pass
- [x] Link all issues here that are fixed by this work
- [x] Disable console tracing by default
- [x] Publish this as `1.1.0-test1` (eventually going to be `1.1.0-beta1`)


packages/liveblocks-core/src/room.ts
```diff
@@ -667,18 +678,13 @@ type RoomState<
     storageOperations: Op[];
   };
 
-  /**
-   * Number of reconnect() retries.
-   */
-  numRetries: number;
   readonly timers: {
+    // XXX Should we move this into the `buffer` structure, if _that_ is what
+    // is being flushed, close to the `lastFlushedAt` value?
```
jrowny, line 683: probably, especially since this is now the loneliest timer in the timers object.

packages/liveblocks-core/src/room.ts
```diff
@@ -671,18 +679,7 @@ type RoomState<
...
-  numRetries: number;
-  readonly timers: {
-    flush: TimeoutID | undefined;
-    reconnect: TimeoutID | undefined;
-    heartbeat: IntervalID | undefined;
-    pongTimeout: TimeoutID | undefined;
-  };
-
-  readonly connection: ValueRef<Connection>;
+  readonly connection: ValueRef<Connection>; // TODO Make this a derived property?
```
jrowny, line 682: is this TODO something you meant to be part of this merge or further down the road?

packages/liveblocks-core/src/room.ts
```diff
@@ -899,6 +881,84 @@ export function createRoom<
...
+    batchUpdates(() => {
+      eventHub.connection.notify(newStatus);
+    });
+  }
+
+  function onDidConnect() {
+    const conn = context.connection.current;
+    if (conn.status !== "open") {
+      // Totally unexpected by now
+      throw new Error("Unexpected non-open state here");
```
jrowny, line 907:
```suggestion
      throw new Error("Unexpected not-open state");
```

packages/liveblocks-core/src/room.ts
```diff
@@ -2652,15 +2353,3 @@ async function fetchAuthEndpoint(
   const { token } = data;
   return { token };
 }
```
jrowny, line 2355: really nice to see this file shed 300 lines of code!

jrowny (approved):
amazing work!

packages/liveblocks-core/src/room.ts
nvie: It's something I wanted to initially make part of this work before the scope grew. At this point, I'm happy to leave it and tackle it in a future change.

PR #887, opened 2023-05-20
use http as a fallback for large messages

jrowny:
### Adding a feature

- for messages over 1MB, use HTTP send endpoint

- [ ] add tests
- [ ] add documentation?

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -608,6 +608,8 @@ const BACKOFF_RETRY_DELAYS_SLOW = [2000, 30000, 60000, 300000];
 const HEARTBEAT_INTERVAL = 30000;
 const PONG_TIMEOUT = 2000;
 
+const MAX_MESSAGE_SIZE = 1 * 1024 * 1024 - 128;
+
 function makeIdFactory(connectionId: number): IdFactory {
   let count = 0;
   return () => `${connectionId}:${count++}`;
@@ -768,6 +770,8 @@ type RoomConfig<TPresence extends JsonObject, TRoomEvent extends Json> = {
   throttleDelay: number;
   authentication: Authentication;
   liveblocksServer: string;
+  httpSendEndpoint: string;
+  unstable_fallbackToHTTP: boolean;
 
   polyfills?: Polyfills;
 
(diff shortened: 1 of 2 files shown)


nvie (commented):
Looking good! Seems like you're done with it once the CORS issue is resolved. Hope my comments help and were the source of your issues! 🤞

packages/liveblocks-core/src/room.ts
```diff
@@ -608,6 +608,8 @@ const BACKOFF_RETRY_DELAYS_SLOW = [2000, 30000, 60000, 300000];
 const HEARTBEAT_INTERVAL = 30000;
 const PONG_TIMEOUT = 2000;
 
+const MAX_MESSAGE_SIZE = 1 * 1024 * 1024 - 128;
```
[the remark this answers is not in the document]
jrowny: yea, figured it was better to have it not right on the line.

packages/liveblocks-core/src/room.ts
```diff
@@ -1011,7 +1015,33 @@ export function createRoom<
...
+          // if our message contains UTF-8, we can't simply use length. See: https://stackoverflow.com/questions/23318037/size-of-json-object-in-kbs-mbs
+          // if this turns out to be expensive, we could just guess with a lower value.
+          const size = new TextEncoder().encode(message).length;
+          if (
+            size > MAX_MESSAGE_SIZE &&
+            context.token?.raw &&
+            config.httpSendEndpoint
+          ) {
+            if (isTokenExpired(context.token.parsed)) {
+              return reconnect();
```
jrowny, line 1029: @nvie if we hit this path, the messages here will be lost, I couldn't see a clean way to put the message back in the queue for processing after reconnect and I didn't see a way to await here. Any ideas on a clean way to do that?

nvie: I don't think you have to worry about this case. All outgoing ops will be stored in an unacknowledged ops administration, which will not be cleared until the ops actually get acknowledged by the server. After a reconnect happens, the unacknowledged ops are resent automatically, in response to the `INITIAL_STORAGE_STATE` server message, see this part:

https://github.com/liveblocks/liveblocks/blob/cdb3cf41b6a5342b07df6f3b3a7520179d05ffae/packages/liveblocks-core/src/room.ts#L1692-L1697

So reconnecting here if the token is expired should already work, I think? Did it not?

nvie (approved):

packages/liveblocks-core/src/room.ts
```diff
@@ -1010,7 +1016,33 @@ export function createRoom<
...
+            }
+            // check for expiration?
```
nvie, line 1032: You dealt with this already.

```suggestion
```

PR #889, opened 2023-05-22
Better API for patching contexts in the state machine

nvie:
This PR improves the API for patching the context as a side-effect, as part of a transition or in an onEnter/onExit handler. It unifies the previously-separate `assign` and `effect` APIs into a single `effect`. An `assign` is now considered just a special case of a side-effect, namely one that happens to mutate the `context`. Mutating the context is perfectly fine during a transition or during onEnter/onExit (but not at any other time!).

This change allows us to unify the API for safely mutating the context from an onEnter/onExit handler as well. In both situations, the `context` that is received allows for patching the context, like so:

```ts
// Only a side-effect
.addTransitions('one', {
  GO: {
    target: 'two',
    effect: () => console.log('Wheee'),
}})

// Only a context assignment
.addTransitions('one', {
  GO: {
    target: 'two',
    effect: (ctx) => ctx.patch({ x: 1 }),
    //                   ^^^^^ ✨ The new API ✨
}})

// Both
.addTransitions('one', {
  GO: {
    target: 'two',
    effect: [
      () => console.log('Wheee'),
      (ctx) => ctx.patch({ x: 1 })
    ],
}})
```

Because of this unified API, we can now also call `.patch()` in onEnter/onExit handlers, too:

```ts
// Only a side-effect
.onEnter('two', (ctx) => {
  ctx.patch({ x: 1 });
  return () => { /* cleanup */ };
});
```

These `.patch()` handlers are guarded, so they can only be used when called from a "safe window"—they will throw if you try to hold on to them and call them later, like so:

```ts
// Only a side-effect
.onEnter('two', (ctx) => {
  setTimeout(() => ctx.patch({ x: 1 }), 1000);  // Won't work!
  return () => { /* cleanup */ };
});
```

### DIFF
--- a/packages/liveblocks-core/src/connection.ts
+++ b/packages/liveblocks-core/src/connection.ts
@@ -174,6 +174,9 @@ enum LogLevel {
   ERROR,
 }
 
+/**
+ * Generic "log" effect. Use it in `effect` handlers of state transitions.
+ */
 function log(level: LogLevel, message: string) {
   const logger =
     level === LogLevel.ERROR
--- a/packages/liveblocks-core/src/connection.ts
+++ b/packages/liveblocks-core/src/connection.ts
@@ -3,7 +3,7 @@ import type { Observable } from "./lib/EventSource";
 import { makeEventSource } from "./lib/EventSource";
 import * as console from "./lib/fancy-console";
 import type { BuiltinEvent, Target } from "./lib/fsm";
-import { FSM } from "./lib/fsm";
+import { FSM, Patchable } from "./lib/fsm";
 import type {
   IWebSocketCloseEvent,
   IWebSocketEvent,
@@ -158,14 +158,14 @@ function nextBackoffDelay(
   );
 }
 
-function increaseBackoffDelay(context: Context) {
-  return { backoffDelay: nextBackoffDelay(context.backoffDelay) };
+function increaseBackoffDelay(context: Patchable<Context>) {
+  context.patch({ backoffDelay: nextBackoffDelay(context.backoffDelay) });
 }
 
-function increaseBackoffDelayAggressively(context: Context) {
-  return {
+function increaseBackoffDelayAggressively(context: Patchable<Context>) {
+  context.patch({
     backoffDelay: nextBackoffDelay(context.backoffDelay, BACKOFF_DELAYS_SLOW),
-  };
+  });
 }
 
 /**
@@ -206,10 +206,6 @@ function enableTracing(machine: FSM<Context, Event, State>) {
     }),
     machine.events.willTransition.subscribe(({ from, to }) => {
       log("Transitioning", from, "→", to);
-    }),
-    machine.events.didPatchContext.subscribe((patch) => {
-      log(`Patched: ${JSON.stringify(patch)}`);
-      // log(`New context: ${JSON.stringify(machine.context)}`);
     }),
     machine.events.didIgnoreEvent.subscribe((e) => {
       log("Ignored event", e, "(current state won't handle it)");
@@ -256,6 +252,12 @@ function defineConnectivityEvents(machine: FSM<Context, Event, State>) {
   };
 }
 
+const assign = (patch: Partial<Context>) => {
+  return (ctx: Patchable<Context>) => {
+    ctx.patch(patch);
+  };
+};
+
 function createConnectionStateMachine<T extends BaseAuthResult>(
   delegates: Delegates<T>
 ) {
@@ -294,7 +296,7 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
   machine.addTransitions("*", {
     RECONNECT: {
       target: "@auth.backoff",
-      assign: increaseBackoffDelay,
+      effect: increaseBackoffDelay,
     },
 
     DISCONNECT: "@idle.initial",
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/connection.ts
```diff
@@ -207,10 +207,6 @@ function enableTracing(machine: FSM<Context, Event, State>) {
     machine.events.willTransition.subscribe(({ from, to }) => {
       log("Transitioning", from, "→", to);
     }),
-    machine.events.didPatchContext.subscribe((patch) => {
-      log(`Patched: ${JSON.stringify(patch)}`);
-      // log(`New context: ${JSON.stringify(machine.context)}`);
-    }),
```
nvie, line 213: This event isn't really needed, it mostly adds noise, because the only moments the context _can_ actually be patched is during transitions or onEnter/onExit handlers, which we already fire events for.

packages/liveblocks-core/src/connection.ts
```diff
@@ -506,20 +509,14 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
...
+      return (ctx) => {
         teardownSocket(ctx.socket);
-        (ctx as any).socket = null;
-        //   ^^^^^^
-        //   XXX Really, I should add an official API to the FSM to support
-        //   XXX assigning to context when entering/leaving a state, since this
-        //   XXX is a well-defined moment in a state transition.
-        //   XXX I've only temporarily pulled this hack here, for which I ask
-        //   XXX forgiveness.
+        ctx.patch({ socket: null });
```
nvie, line 519: Here, in the new setup, if we'd accidentally use `ctx` from the onEnter callback, that would throw an error and catch that bug for you.

packages/liveblocks-core/src/lib/fsm.ts
```diff
@@ -28,51 +28,48 @@ export type AsyncErrorEvent = {
...
-   * state is entered.
-   */
-  assign?: Assigner<TContext, TEvent>;
-
   /**
    * Emit a side effect (other than assigning to the context) when this
    * transition is taken.
    */
-  effect?: Effect<TContext, TEvent>;
+  effect: Effect<TContext, TEvent> | Effect<TContext, TEvent>[];
```
jrowny, line 65: does this mean we require an effect now? Maybe an empty array at least?

packages/liveblocks-core/src/lib/fsm.ts
```diff
@@ -128,6 +125,58 @@ function patterns<TState extends string>(
...
+          self.curr = Object.assign({}, self.curr, patch);
+
+          // Also patch the temporary mutable context helper itself, in case
+          // there are multiple calls in a succession that need
+          for (const pair of Object.entries(patch)) {
+            const [key, value] = pair as [
+              keyof TContext,
+              TContext[keyof TContext]
+            ];
+            if (key !== "patch") {
```
jrowny, line 162: I'm not sure about this limitation, what if someone actually does want to patch a prop named patch?

packages/liveblocks-core/src/lib/fsm.ts
```diff
@@ -538,8 +588,12 @@ export class FSM<
       nextState = nextTarget;
     } else {
       nextState = nextTarget.target;
-      assign = nextTarget.assign;
-      effect = nextTarget.effect;
+      effects =
+        nextTarget.effect !== undefined
```
jrowny, line 592: I love concise code but setting something to undefined if it's undefined because of nested ternaries may be a sign to just not use a double ternary 😂 

```
if (nextTarget.effect !== undefined) {
 effects = Array.isArray(nextTarget.effect)
            ? nextTarget.effect
            : [nextTarget.effect];
}
```

Would be clearer here I think.

jrowny (approved):
a few small nits but overall I like it!

packages/liveblocks-core/src/connection.ts
```diff
@@ -464,14 +465,16 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
...
-            console.error(
-              `Connection to WebSocket could not be established, reason: ${String(
-                failedEvent.reason
-              )}`
-            );
-          },
+          effect: [
+            increaseBackoffDelay,
+            () => {
+              console.error(
```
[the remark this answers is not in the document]
nvie: Oh yes, I like it!

packages/liveblocks-core/src/lib/fsm.ts
nvie: Yes, but only if you use the "expanded form", i.e. `{ target: "@my-state", effect: <something> }`. If you wouldn't want to use an effect, you'd typically use the shorter string form, i.e. `"@my-state"`.

packages/liveblocks-core/src/lib/fsm.ts
nvie: It would not be possible. I thought about it, but chose pragmatism over purity, because otherwise we'd have to send it as two values into each callback, requiring unpacking, i.e. `(ctx, event, patch) => patch({ … })` or `({ ctx, patch }, event) => patch({ … })`, both of which feels overly complicated.

If ever the need for a `context.patch` property arises, I'd be happy to reconsider this API.

packages/liveblocks-core/src/lib/fsm.ts
nvie: Better even, because `effect` can no longer be undefined, we could remove it entirely 👍 Fixed in 84d6d2e229b0e542d06a91db59efad8bce20f4a8.

packages/liveblocks-core/src/connection.ts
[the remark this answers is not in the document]
nvie: How about this? c1acc07b24ea7f20ce7d299a54f76f5f37672372

PR #890, opened 2023-05-23
Add generic async cache

marcbouchenoire:
This PR adds a generic async cache system called `AsyncCache` which handles things like caching, deduplication, invalidation, and even revalidation with optimistic data. It is not used anywhere yet.

This system's raison d'être is that we have various use cases (present and future) in our packages with very similar shared requirements (a form of cached state and request deduplication). It is meant as a private utility that we can use as a foundation while adding use case-specific logic around it (e.g. a token system will invalidate based on expiration and will never return stale data, a comments system will use optimistic data when revalidating after submitting a new comment, etc).

In a future PR, I'll make a React wrapper for it (that's the main reason it has subscription methods) so that we can also use it within `@liveblocks/react`, possibly even while sharing the same caches between `@liveblocks/client` and `@liveblocks/react` (e.g. `client.resolveUser(userId)` and `useUser(userId)` would both share the same `AsyncCache` instance).

### DIFF
--- a/packages/liveblocks-core/src/lib/AsyncCache.ts
+++ b/packages/liveblocks-core/src/lib/AsyncCache.ts
@@ -46,6 +46,9 @@ export type AsyncCacheItem<TData = any, TError = any> = Observable<
   get(): Promise<AsyncResolvedState<TData, TError>>;
   getState(): AsyncState<TData, TError>;
   invalidate(options?: InvalidateOptions): void;
+  revalidate(
+    options?: InvalidateOptions
+  ): Promise<AsyncResolvedState<TData, TError>>;
 };
 
 export type AsyncCache<TData = any, TError = any> = {
@@ -53,6 +56,10 @@ export type AsyncCache<TData = any, TError = any> = {
   get(key: string): Promise<AsyncResolvedState<TData, TError>>;
   getState(key: string): AsyncState<TData, TError> | undefined;
   invalidate(key: string, options?: InvalidateOptions): void;
+  revalidate(
+    key: string,
+    options?: InvalidateOptions
+  ): Promise<AsyncResolvedState<TData, TError>>;
   subscribe(
     key: string,
     callback: Callback<AsyncState<TData, TError>>
@@ -168,6 +175,12 @@ function createCacheItem<TData = any, TError = any>(
     return getState() as AsyncResolvedState<TData, TError>;
   }
 
+  function revalidate(options?: InvalidateOptions) {
+    invalidate(options);
+
+    return get();
+  }
+
   function getState() {
     return {
       isLoading: context.isLoading,
@@ -181,6 +194,7 @@ function createCacheItem<TData = any, TError = any>(
     get,
     getState,
     invalidate,
+    revalidate,
   };
 }
 
@@ -207,7 +221,7 @@ export function createAsyncCache<TData = any, TError = any>(
     return cacheItem;
   }
 
-  async function get(key: string) {
+  function get(key: string) {
     return create(key).get();
   }
 
@@ -217,6 +231,10 @@ export function createAsyncCache<TData = any, TError = any>(
 
   function invalidate(key: string, options?: InvalidateOptions) {
     cache.get(key)?.invalidate(options);
+  }
+
+  function revalidate(key: string, options?: InvalidateOptions) {
+    return create(key).revalidate(options);
   }
 
   function subscribe(
@@ -239,6 +257,7 @@ export function createAsyncCache<TData = any, TError = any>(
     get,
     getState,
     invalidate,
+    revalidate,
     subscribe,
     has,
     clear,
--- a/packages/liveblocks-core/src/lib/AsyncCache.ts
+++ b/packages/liveblocks-core/src/lib/AsyncCache.ts
@@ -2,7 +2,6 @@ import type { Callback, Observable, UnsubscribeCallback } from "./EventSource";
 import { makeEventSource } from "./EventSource";
 
 const DEDUPLICATION_INTERVAL = 2000;
-const KEEP_PREVIOUS_DATA = false;
 
 type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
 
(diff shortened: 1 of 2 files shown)


packages/liveblocks-core/src/lib/AsyncCache.ts
```diff
@@ -0,0 +1,345 @@
...
+  deduplicationInterval?: number;
+};
+
+type AsyncCacheItemOptions = WithRequired<
+  AsyncCacheOptions,
+  "deduplicationInterval"
+>;
+
+type InvalidateOptions<TData = any> =
+  | { clearData?: false; setOptimisticData?: never }
```
ofoucherot, line 20: When do you think the option `clearData` will be used?

marcbouchenoire: For use-cases that should work in a stale-while-revalidate fashion instead of the default which is to never return stale data.

The use-case's own `get()` function would check if the key needs to be invalidated (e.g. based on time), then do `invalidate({ clearData: false })`, call the cache's `get()` function but without awaiting it (just to schedule it) and immediately return the cached/stale data.

packages/liveblocks-core/src/lib/AsyncCache.ts
```diff
@@ -0,0 +1,345 @@
...
+  | {
+      clearData?: never;
+      setOptimisticData: (data: TData | undefined) => TData | undefined;
+    };
+
+export type AsyncState<TData = any, TError = any> = {
```
nvie, line 26: Can we model this as a three-legged ADT?

```
type AsyncState<T, E> =
  | AsyncStateLoading
  | AsyncStateResolved<T>
  | AsyncStateErorr<E>

type AsyncStateLoading = {
  readonly loading: true;
  readonly data?: never;
  readonly error?: never;
};

type AsyncStateResolve<T> = {
  readonly loading: false;
  readonly data: T;
  readonly error?: never;
};

type AsyncStateError<E> = {
  readonly loading: false;
  readonly data?: never;
  readonly error?: E;
};
```

This will make any impossible state impossible to express at the type level. The object `{ isLoading: false, data: undefined, error: undefined }` or `{ isLoading: true, data, error }` are currently possible, but should be invalid, right?

nvie: Also, I would advise to never use default type arguments in TypeScript, especially when defaulting to `any`. It makes it way too easy to end up with unintended `any` types values. Why not remove them? Anywhere you use the `AsyncState` type, you should specify what its expected "ok" and "error" type payloads are. Like any other container, think a Map or Array, it would also be weird to use them without their payload types, right?

packages/liveblocks-core/src/lib/AsyncCache.ts
```diff
@@ -0,0 +1,345 @@
...
+  lastInvokedAt?: number;
+  previousState: AsyncState<TData, TError>;
+  previousNonOptimisticData?: TData;
+};
+
+export type AsyncCacheItem<TData = any, TError = any> = Observable<
+  AsyncState<TData, TError>
+> & {
+  get(): Promise<AsyncResolvedState<TData, TError>>;
+  getState(): AsyncState<TData, TError>;
```
nvie, line 53: Since React introduced the `useSyncExternalStore` API, they talk about `.getSnapshot()` for things like this. Here, it looks like `.getState()` is exactly that: a synchronous snapshot of the current state of this cache item. Shall we rename this method?

packages/liveblocks-core/src/lib/AsyncCache.ts
nvie: Lastly, when writing _generic_ data structures, I wouldn't mind using single-letter type variables like `T` and `E`. Just like how `Array<T>` and `Map<K, V>` are used in the TypeScript standard library. They may make the rest of the code in this module more concice.

packages/liveblocks-core/src/lib/AsyncCache.ts
```diff
@@ -0,0 +1,345 @@
...
+      eventSource.notify(state);
+    }
+  }
+
+  function invoke() {
+    context.lastInvokedAt = Date.now();
+    context.error = undefined;
+    context.isLoading = true;
+    context.isInvalid = true;
+    context.promise = asyncFunction(key);
```
nvie, line 162: Rather than "mutating properties" on the context one by one, the benefit of using the 3-legged ADT and having all of its members use _readonly_ properties, is that you cannot do that. The only way to change from one state to another (e.g. from "loading" to "resolved") is to create a new object whenever you change to a new state.

Writing your code that way massively reduces the opportunity for bugs to lurk in there, and makes reasoning about the possible states easier, because there are literally only 3 states to consider.

Right now, for example, I see this `invoke()` will change 5 properties on `context`, but… I don't know (without look at all the other code), if there are perhaps other properties that must also be reset but perhaps have been forgotten, or if there are properties that deliberately remain the same.

Writing this as a state change using the ADT method, you'd write it as either:

```ts
// Option A: these are literally the only 5 fields on there! Easy to see. All the rest is wiped!
context = {
  lastInvokedAt: Date.now(),
  error: undefined,
  isLoading: true,
  isInvalid: true,
  promise: asyncFunction(key),
}
```

or

```ts
// Option B: these are literally the only 5 fields on there that are changed. All the rest remains the same. Also easy to see.
context = {
  lastInvokedAt: Date.now(),
  error: undefined,
  isLoading: true,
  isInvalid: true,
  promise: asyncFunction(key),
  ...context,
}
```

Another benefit is that if you (or someone else) later must introduce another property on the `context` object, then TypeScript will help you remember to consider/update the impact of that new property on this method. Is it relevant here in `invoke`, or not?

packages/liveblocks-core/src/lib/AsyncCache.ts
marcbouchenoire: It's precisely here for the `useSyncExternalStore`-based React implementation that's coming soon but since it can be used for other synchronous use-cases (e.g. advanced/custom revalidation logic) and the `isLoading`/`data`/`error` type is called `AsyncState` I thought it would make more sense to map to that instead of just the React use-case.

packages/liveblocks-core/src/lib/AsyncCache.ts
marcbouchenoire: I've refactored things to use a discriminated union but because of the API's genericness (to support vanilla and React use-cases) and options like `clearData` and `optimisticData`, there are 4 possible states with some overlap: `AsyncStateInvalid`, `AsyncStateLoading`, `AsyncStateSuccess`, and `AsyncStateError`.

packages/liveblocks-core/src/lib/AsyncCache.ts
marcbouchenoire: I separated the state from the context and moved to always create a new state when modifying it, but the context is more of a collection of variables so I'm not sure if it would make sense to separate it into a discriminated union of possibilities.

PR #893, opened 2023-05-23
Improve WebSocket mocking in unit tests

nvie:
This PR changes some testing internals around the MockWebSocket helper, to help with making it easier to express unit tests more precisely, and to help observe outside behavior, instead of implementation details.

The main conceptual change made here is that no `MockWebSocket` instance stands on its own. Each such socket instance really is the client end to a _pipe_ with sockets on both ends: the client side and the server side.

Using this API, we can pass the client end to the delegate. It basically acts like a normal WebSocket polyfill. But… we can now control the server's bahavior precisely, for example:

```ts
// ❌ Old way
const ws = new MockWebSocket();

// ✅ New way
const wss = new MockWebSocketServer();
//                           ^^^^^^ ⚠️
const ws1 = wss.newSocket();
const ws2 = wss.newSocket();
```

The benefits are: very precise control over the server behavior in unit tests, and being able to observe the effects of using the client, by inspecting the received server messages (rather than observing which effects the client emits).

```ts
ws1.server.accept();
ws2.server.close(/* close event */);
```

Later, we can use this to implement a better fake Liveblocks server (a lite/fake implementation of our backend), which _relays_ messages between clients in a correct way.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/client.node.test.ts
+++ b/packages/liveblocks-core/src/__tests__/client.node.test.ts
@@ -31,7 +31,7 @@ function createClientAndEnter(options: ClientOptions) {
   const client = createClient(options);
   client.enter("room", { initialPresence: {} });
 
-  // Entering starts asynchronous jobs in the background (times, promises,
+  // Entering starts asynchronous jobs in the background (timers, promises,
   // etc). Not leaving the room would leave those open handles dangling which
   // doesn't make Jest happy.
   client.leave("room");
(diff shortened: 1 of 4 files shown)


packages/liveblocks-core/src/__tests__/client.node.test.ts
```diff
@@ -30,6 +30,11 @@ function atobPolyfillMock(data: string): string {
 function createClientAndEnter(options: ClientOptions) {
   const client = createClient(options);
   client.enter("room", { initialPresence: {} });
+
+  // Entering starts asynchronous jobs in the background (times, promises,
+  // etc). Not leaving the room would leave those open handles dangling which
+  // doesn't make Jest happy.
+  client.leave("room");
```
jrowny, line 37: so should this function be renamed to createClientAndEnterAndThenLeave? 😄

jrowny (approved):

PR #894, opened 2023-05-24
Fix token expiry bug

nvie:
This fixes the bug in token expiry. The original intent for this was to allow near-future tokens to be considered valid, in case a user's system clock would not be in sync with the server clock.

This PR adds an explicit unit test for the edge case where this bug happened (❌), and fixes the bug (✅). I also rewrote the code using DeMorgan to make the condition a positive one, because they're much easier to read mentally.

## Future Work
This isn't really the proper way to fix this. What we should do instead, is as soon as we receive the new token from the server, convert the `exp` values to the "local" clock (using `localExp = (Date.now() / 1000) + (exp - iat)`), and use that value to decide on the client side to proactively know when the server will reject the token. That way it works, even if the user's system clock is way off. (Not tackled in this PR.)


jrowny (approved):
Looks good to me

PR #896, opened 2023-05-24
Add `$ref` to 401 response for storage initialization openapi documentation

nimeshnayaju:
Fixes https://github.com/liveblocks/liveblocks.io/issues/1143#issue-1699752443

> Clicking on Status 401 on the Initialize Storage section on the Rest API documentation causes the corresponding responses section to disappear until page is refreshed. Here's a screen capture and steps to reproduce:
> 1. Go to https://liveblocks.io/docs/api-reference/rest-api-endpoints#post-rooms-roomId-storage
> 2. On the Responses section, click on the Status dropdown and select 401 .

**Problem:** The code that deals with displaying responses requires a response to have either a `$ref` or `content` property - the `Initialize Storage` OpenAPI documentation only contained a `description` property, hence the disappearing section. 

**Fix:** This PR adds the `$ref` field and removes the `description` field. Doing so means the response section for status code `401` shows the `WRONG_KEY_USED` or `MISSING_SECRET_KEY` error message, which aligns with the result from our API backend too.

### DIFF
--- a/docs/references/v2.openapi.json
+++ b/docs/references/v2.openapi.json
@@ -550,7 +550,7 @@
             }
           },
           "401": {
-            "description": "Unauthorized"
+            "$ref": "#/components/responses/401"
           },
           "403": {
             "$ref": "#/components/responses/403"


nimeshnayaju:
The response value for `401`:

https://github.com/liveblocks/liveblocks/blob/bbefd7a2ecaa7743c36aad4e16f3e505ba630a5d/docs/references/v2.openapi.json#L1434-L1461

nimeshnayaju:
Curious what should be done to prevent issues like this the future? We seem to be currently handwriting all the information about each endpoint, maybe generate these information automatically using a script that ensures proper validation? 

Or if that is too big of an undertaking for now, maybe ensuring the JSON file follows a certain config or type-safety?

taylorpwhipp (approved):

PR #899, opened 2023-05-25
Update `create-liveblocks-app` non-React init export

CTNicholas:
Adding `export type TypedRoom = Room<...>` for non-React TypeScript projects.


taylorpwhipp (approved):
This looks good to me.  The only thing I was unsure about was the testing procees-  I am assuming that `create-liveblocks-app` needs to be released privately in order to test, yes?

CTNicholas:
> This looks good to me. The only thing I was unsure about was the testing procees- I am assuming that `create-liveblocks-app` needs to be released privately in order to test, yes?

To test it, you can run `npm run demo` then check inside the `.demo` repo!

PR #901, opened 2023-05-25
Disable code coverage by default on local test runs

nvie:
Still runs code coverage on CI, but locally, don't run it by default. You can still run it using the `--coverage` flag, but don't run them by default. It's pretty annoying to have to silence it when you're actively working on getting tests to pass.


marcbouchenoire (approved):
Agreed 👍

Issue #900, opened 2023-05-25
Undo/redo in a loop causes exponential blowup of history


slon:
**Describe the bug**

Repeatedly doing undo/redo of the same update causes slowdown and room disconnect.

**To Reproduce**

1. Open whiteboard demo. https://liveblocks.io/examples/collaborative-whiteboard/nextjs
2. Create rectangle
3. Move rectangle
4. Press undo followed by redo multiple times. (5 times seems enough)
5. One instance of whiteboard reconnects in a loop.

**Expected behavior**

Rectangle moves between two positions.

**Environment (please complete the following information):**

- @liveblocks/react@1.0.10, @liveblocks/client@1.0.10

**Additional context**

It seems that history entries are duplicated during undo/redo. I can see in network debugger that websocket messages are growing exponentially in size and contain duplicate entries.
![image](https://github.com/liveblocks/liveblocks/assets/120082/1458cae4-d17e-49f2-8bb8-622f6ed6f9f1)
![image](https://github.com/liveblocks/liveblocks/assets/120082/0c696905-54a9-4f7f-8b7c-b19d6175c5e0)

Upon brief inspection code in [_applyUpdate](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-core/src/crdts/LiveObject.ts#L369) seems suspicious. `reverseUpdate` is first added to `reverse` array on line [378](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-core/src/crdts/LiveObject.ts#L378) and later added to the same array on line [426](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-core/src/crdts/LiveObject.ts#L426)

jrowny:
@slon Your intuition about the reverse was correct. We've fixed this in #902, added a test,  and published v1.0.11 to resolve this issue. Thank you for the report!

PR #902, opened 2023-05-25
fix exponential update size

jrowny:
### Description

An inadvertent double push to the reverse op array caused the history op size to double. Once this size reaches 1mb the client will be stuck in a reconnect loop. 

#### How to test it

- Check websocket message size when undoing and redoing in the next whiteboard advanced example.

#### Related issue(s)

https://github.com/liveblocks/liveblocks/issues/900

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # vNEXT (not released yet)
 
+### `@liveblocks/core`
+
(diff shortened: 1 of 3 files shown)


nvie (approved):
Amazing work! Thanks for the quick investigation and fix, @jrowny! 💯

PR #904, opened 2023-05-25
Some `MockWebSocketServer` tweaks

nvie:
Betters internal names, better public names for APIs (matching other WebSocket Server implementations), better comments, and… a bug fix.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/_utils.ts
+++ b/packages/liveblocks-core/src/__tests__/_utils.ts
@@ -323,12 +323,18 @@ export class MockWebSocket {
     }
   }
 
+  /**
+   * Send a message from the client to the WebSocket server.
+   */
   public send(message: string) {
     if (this.readyState === this.OPEN) {
       this.server.receivedMessages.push(message);
     }
   }
 
+  /**
+   * Close the socket from the client side.
+   */
   public close(_code?: number, _reason?: string): void {
     this.#readyState = this.CLOSED;
   }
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/__tests__/_utils.ts
```diff
@@ -141,58 +146,46 @@ export class MockWebSocketServer {
...
     // Run the callback in the next tick. This is important, because we first
     // need to return the socket.
     setTimeout(() => {
-      this.newSocketCallback?.(socket);
+      // Call the provided callback once, for this connection only.
+      initFn?.(clientSocket);
+
+      // ...then proceed to call the rest of the callbacks, which will be
+      // executed on every new connection
+      this.newConnectionCallbacks.notify({
```
jrowny, line 178: how does this get populated? It looks like the previous code wrapped the callback everytime it was set.

jrowny (approved):
looks good

packages/liveblocks-core/src/__tests__/_utils.ts
nvie: Correct, and that was a bug. It gets populated by calling `wss.onConnection()`. In a unit test, you can use this to program how the server will behave once a connection gets made. For example:

Screen Shot 2023-05-26 at 01 15 23@2x

nvie: The point of stubbing out this backend like in this example is to let us interact with and observe effects from the outside. Compare this rewritten test above to the same test as it currently exists on `main`, which fakes the behavior to test here by not using a socket connection at all, but instead making equivalent calls to room's internals.

https://github.com/liveblocks/liveblocks/blob/e0bd5a582370d2bec2b1e68e4873fa00a7332cd5/packages/liveblocks-core/src/__tests__/room.test.ts#L641-L661

PR #907, opened 2023-05-26
Fix some bugs I found in the `ManagedSocket` implementation

nvie:
This PR fixes a couple of bugs in the `ManagedSocket` (aka "connection manager") implementation. This code is still unused, so this PR should be safe to merge as-is.

### DIFF
--- a/packages/liveblocks-core/src/connection.ts
+++ b/packages/liveblocks-core/src/connection.ts
@@ -417,32 +417,24 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
     .onEnterAsync(
       "@connecting.busy",
 
+      //
+      // Use the "createSocket" delegate function (provided to the
+      // ManagedSocket) to create the actual WebSocket connection instance.
+      // Then, set up all the necessary event listeners, and wait for the
+      // "open" event to occur.
+      //
+      // When the "open" event happens, we're ready to transition to the
+      // OK state. This is done by resolving the Promise.
+      //
       async (ctx) => {
         let capturedPrematureEvent: IWebSocketEvent | null = null;
 
-        //
-        // Use the "connect" delegate to create the WebSocket connection (which
-        // will initiate the connection), and set up all the necessary event
-        // listeners, then wait until the 'open' event has fired.
-        //
         const promise = new Promise<[IWebSocketInstance, () => void]>(
-          (res, rej) => {
+          (resolve, rej) => {
             if (ctx.token === null) {
               throw new Error("No auth token"); // This should never happen
             }
 
-            /**
-             * Instantiate the WebSocket, and set up event listeners.
-             *
-             * The `error` and `close` event handlers marked (*) are installed
-             * here only temporarily, just to handle this promise-based state.
-             * When those get triggered, we reject this promise.
-             *
-             * When an "open" event happens, we're OK and ready to let the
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/connection.ts
```diff
@@ -412,54 +418,103 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
...
+      // Then, set up all the necessary event listeners, and wait for the
+      // "open" event to occur.
+      //
+      // When the "open" event happens, we're ready to transition to the
+      // OK state. This is done by resolving the Promise.
+      //
+      async (ctx) => {
+        let capturedPrematureEvent: IWebSocketEvent | null = null;
+
+        const promise = new Promise<[IWebSocketInstance, () => void]>(
```
jrowny, line 433: I know that 90% of my critiques are about naming things, but can we call this something other than promise?

packages/liveblocks-core/src/connection.ts
```diff
@@ -412,54 +418,103 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
...
+                socket.removeEventListener("close", reject); // Remove (*)
+              };
+
+              // Resolve the promise, which will take us to the @ok.* state
+              resolve([socket, unsub]);
+            });
           }
+        );
+
+        return withTimeout(promise, SOCKET_CONNECT_TIMEOUT).then(
```
jrowny, line 494: does this need a catch somewhere?

nvie: It does not! The whole point here is that the promise _should_ fail in all cases but the happy path. When that happens, the machine will [take the failure transition](https://github.com/liveblocks/liveblocks/blob/208f9c0531b3e9e69e76d33ddfc19d414970ae48/packages/liveblocks-core/src/connection.ts#L529-L564) out of that state, and either retries-with-backoff, or (purposefully) stops retrying.

nvie: That said, I kinda hate how long this `.onEnterAsync()` definition has become. Makes the structure of that call really hard to follow now. The structure is:

```
    .onEnterAsync(
      "@connecting.busy",
      tryConnectingTheSocket,  // a Promise
      ok => { /* transition like this */ },
      failure => { /* transition like this */ },
    )
```

jrowny: in the failure case, what happens to the socket? Is it cleaned up? Are those reject listeners removed?

nvie: Great question! If the promise succeeds, we retain a permanent reference to the socket, via [this assignment](https://github.com/liveblocks/liveblocks/blob/208f9c0531b3e9e69e76d33ddfc19d414970ae48/packages/liveblocks-core/src/connection.ts#LL524-L524). Otherwise, the socket instance will get garbage collected, so my understanding is that there is no need to remove the listeners.

packages/liveblocks-core/src/connection.ts
nvie: Fixed in e4060b6f8c63247cfbe3986e00511d6a74f85b62. I've renamed it to `connect$`, using the convention of naming promises (values that still need to be `await`'ed) with a `$`-suffix, which we also do elsewhere in code to make that more explicit.

PR #911, opened 2023-05-30
New `EventSource` APIs

nvie:
This PR adds two new EventSource APIs, that turned out to be really useful and generic:

### `EventSource.waitUntil()`
While working on rewriting some of the unit tests, I found that the pattern of "wait until some event fires, then observe some property" was really useful to express. Setting up a one-time event listener, then tearing it down again became quite verbose to do often, until I realized that it would be much more elegantly expressed as a promise, and I realized every event source could easily turn itself into such promise. This isn't recommended for use in production apps, as you risk waiting forever, but use in unit tests it's great.

### `EventSource.pause()` / `EventSource.unpause()`
This allows pausing/unpausing event delivery for an event source. I have this particular situation in the connection manager where events can be delivered before listeners are ready to receive them. By programmatically pausing and unpausing event delivery, I can better control when events get emitted. If an event source is unpaused (= the default), then there is no overhead, and calling `.notify()` will instantly emit the events, like before. But if the event source is paused, those events are buffered in-memory instead, until `.unpause()` is called, at which time it will instantly emit all the buffered events, and free up that memory. Of course, because of this increased memory pressure, you should only use pausing temporarily.

### `EventSource.count()`
As requested by @marcbouchenoire in his PR comment below, we've exposed a method to query the number of current subscribers for this event source, so we can let logic depend on it (like stopping a poller after the last subscriber unsubscribes).

### DIFF
--- a/packages/liveblocks-core/src/lib/__tests__/EventSource.test.ts
+++ b/packages/liveblocks-core/src/lib/__tests__/EventSource.test.ts
@@ -30,7 +30,7 @@ describe("EventSource", () => {
           hub.notify(payload);
           hub.notify(payload);
 
-          expect(callback.mock.calls.length).toBe(3);
+          expect(callback).toHaveBeenCalledTimes(3);
           for (const [arg] of callback.mock.calls) {
             expect(arg).toBe(payload);
           }
@@ -56,12 +56,12 @@ describe("EventSource", () => {
           hub.notify(payload);
           hub.notify(payload);
 
-          expect(callback1.mock.calls.length).toBe(3);
+          expect(callback1).toHaveBeenCalledTimes(3);
           for (const [arg] of callback1.mock.calls) {
             expect(arg).toBe(payload);
           }
 
-          expect(callback2.mock.calls.length).toBe(2);
+          expect(callback2).toHaveBeenCalledTimes(2);
           for (const [arg] of callback2.mock.calls) {
             expect(arg).toBe(payload);
           }
@@ -123,7 +123,7 @@ describe("EventSource", () => {
           hub.notify(payload);
           hub.notify(payload);
 
-          expect(callback.mock.calls.length).toBe(1); // Called only once, not three times
+          expect(callback).toHaveBeenCalledTimes(1); // Called only once, not three times
           for (const [arg] of callback.mock.calls) {
             expect(arg).toBe(payload);
           }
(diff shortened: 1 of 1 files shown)


marcbouchenoire:
On the topic of new APIs, I've recently had the need to know more about the number of subscribers (e.g. to implement a clean-up-if-last-to-unsubscribe system). Could we formalize that? I'm not entirely sure what would be the way/granularity to expose that especially because of the two lists of subscribers (infinite and one-time). Here's what I had for my use case:

```ts
function hasObservers() {
  return _onetimeObservers.size + _observers.size > 0;
}
```

nvie:
I suppose formalizing the total subscriber count wouldn't hurt? How about returning this as a number instead of a boolean?

```ts
/**
 * Returns the number of active subscribers for this event source.
 */
function count() { return _onetimeObservers.size + _observers.size }
```

nvie:
@marcbouchenoire Implemented your request in 7e7ecd3266aad0a96820b7bb617fa6b10336c3fa, and added JSDoc comments for all methods in 51ee33f284e8a8723ec4d94719d6abd52ad6d3ae.

packages/liveblocks-core/src/lib/__tests__/EventSource.test.ts
```diff
@@ -154,4 +193,58 @@ describe("EventSource", () => {
...
+
+          hub.pause();
+          hub.notify(payload);
+          hub.notify(payload);
+          hub.notify(payload);
+
+          expect(callback.mock.calls.length).toBe(0); // No events get delivered until unpaused
+
+          hub.unpause();
+          expect(callback.mock.calls.length).toBe(3); // Buffered events get delivered
```
marcbouchenoire, line 242: Very minor nitpick (or question in case you explicitly want `mock.calls.length`)
 
```suggestion
          expect(callback).not.toHaveBeenCalled(); // No events get delivered until unpaused

          hub.unpause();
          expect(callback).toHaveBeenCalledTimes(3); // Buffered events get delivered
```

marcbouchenoire (approved):

packages/liveblocks-core/src/lib/__tests__/EventSource.test.ts
marcbouchenoire: Oh the rest of the test uses `mock.calls.length`, never mind ignore this 😅

nvie: Fixed in b0961257b12e9ecdb28d689d69c5cdee3d4e5f6e 👍

PR #916, opened 2023-06-02
Upgrade Turborepo to 1.10.1

nvie:


turbo.json
```diff
@@ -25,6 +25,18 @@
...
+      "inputs": [
+        "jest.*",
+        "e2e/**/*.tsx",
+        "e2e/**/*.ts",
+        "src/**/*.tsx",
+        "src/**/*.ts",
+        "test/**/*.ts",
+        "test/**/*.tsx"
+      ]
+    },
```
nvie, line 39: I've also added this missing definition to the `turbo.json` file.

packages/liveblocks-core/turbo.json
```diff
@@ -0,0 +1,19 @@
...
+      "inputs": [
+        "jest.*",
+        "e2e/**/*.tsx",
+        "e2e/**/*.ts",
+        "src/**/*.tsx",
+        "src/**/*.ts",
+        "test/**/*.ts",
+        "test/**/*.tsx"
+      ]
+    }
```
nvie, line 17: I've also added this missing definition to the `turbo.json` file. I've used a local `turbo.json` file here, since this command only makes sense inside the `liveblocks-core` package.

PR #919, opened 2023-06-03
fix: use PATCH instead of DELETE when passing body

antoinewg:
<!-- We appreciate your efforts in opening a PR! Your contribution means a lot.
In order to ensure a seamless handling of your PR, we kindly ask you to refer to the checklist sections provided below.
Please select the appropriate checklist that corresponds to the changes you are making:

### Fixing a bug

Hello, when trying to remove a group access from a document, I have this error:

```
- error Error [SyntaxError]: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at DELETE (webpack-internal:///(api)/./pages/api/liveblocks/documents/[documentId]/teams.ts:45:31)
    at groups (webpack-internal:///(api)/./pages/api/liveblocks/documents/[documentId]/teams.ts:64:26)
    at ...
}
  41 |   const documentId = req.query.documentId as string;
  42 |   console.log("DELETE ", documentId);
> 43 |   const { groupId }: RemoveGroupRequest = JSON.parse(req.body);
     |                                               ^
  44 | 
  45 |   const { data, error } = await removeGroupAccess(req, res, {
  46 |     documentId,
```

I think it's because we're passing a body with a method `DELETE`, which doesn't support.
Using `PATCH` instead fixes it.

### DIFF
--- a/starter-kits/nextjs-starter-kit/lib/client/documents/removeGroupAccess.ts
+++ b/starter-kits/nextjs-starter-kit/lib/client/documents/removeGroupAccess.ts
@@ -26,7 +26,7 @@ export async function removeGroupAccess({
   };
 
   return fetchApiEndpoint<DocumentGroup[]>(url, {
-    method: "DELETE",
+    method: "PATCH",
     body: JSON.stringify(request),
   });
 }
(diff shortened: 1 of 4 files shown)


CTNicholas (approved):
Ah, nice spot! Thanks for updating this 👍

PR #925, opened 2023-06-06
[draft] New authorize method

ofoucherot:
In progress

### DIFF
--- a/packages/liveblocks-node/src/authorize.ts
+++ b/packages/liveblocks-node/src/authorize.ts
@@ -7,12 +7,10 @@ enum LiveblocksScope {
   MyAccess = "my:access",
   CommentWrite = "comment:write",
   CommentRead = "comment:read",
+  Events = "events",
 }
 
-type LiveblocksPermission = {
-  resource: string;
-  scopes: LiveblocksScope[];
-};
+type LiveblocksPermissions = Record<string, LiveblocksScope[]>;
 
 type AuthorizeOptions = {
   /**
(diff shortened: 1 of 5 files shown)


packages/liveblocks-node/src/authorize.ts
```diff
@@ -1,27 +1,52 @@
...
   secret: string;
+
   /**
-   * The room ID for which to authorize the user. This will authorize the user
-   * to enter the Liveblocks room.
+   * The permissions to grant to the user. This will authorize the user to access
+   * the different resources with the specified scopes.
    */
-  room: string;
+  permissions: LiveblocksPermissions;
```
nvie, line 26: @ofoucherot Thinking about your comment about the breaking change, but not sure if we really have to make it a breaking change. Given that in the old method there will always be a mandatory `room` field, and in the new method there will always be a mandatory `permissions` field, I think we can clearly distinguish these cases and add a function overload to make this a non-breaking change?

This was different in the previous approach, where we wanted to make `room` optional (and behaviorally different as such), but given the new `permissions` configuration, I think we can nicely avoid it. Or am I missing something?

nvie: Additionally, I think it may make sense to start versioning our auth tokens explictly? This way, rather than looking at presence/absence of certain properties in order to understand what token we're dealing with, we could ask the backend to produce an `authv1` token (= our current token), or an `authv2` token (= our new token format). We should make it part of the JWT payload, and also make it part of the auth request. (If it's not explicitly requested, we assume it's an old/v1 token.)

Issue #924, opened 2023-06-06
When a client reconnects, it can temporarily retrigger a Suspense boundary


nvie:
**Describe the bug**

When a client reconnects, it can momentarily trigger a Suspense boundary. The Suspense boundary should only be triggered for the initial load, never beyond that.

**To Reproduce**

1. Open a React example that uses a Suspense boundary, for example https://examples-nextjs-todo-list-liveblocks.vercel.app
2. Simulate network outage by "working offline"
3. Continue using the disconnected app
4. Notice how the Suspense boundary can get triggered incorrectly

**Expected behavior**

The suspense boundary should never be triggered again after the initial load.

**Illustrations**

https://github.com/liveblocks/liveblocks/assets/83844/7109d1da-a4c9-4710-8a72-811d39305d09

nvie:
Fixed in [1.1.0](https://liveblocks.io/blog/whats-new-in-v1-1).

Issue #878, opened 2023-05-18
Improved status subscription APIs


CTNicholas:
Currently if you use `room.subscribe("connection")` a number of different statuses can return: `authenticating`, `connecting`, `open`, `failed`, `closed` or `unavailable`. These are the statuses the users sees when connecting to Liveblocks: `closed -> authenticating -> connecting -> open` . 

However, when building your UI/app, most of these statuses are _not helpful_, as they're too low-level. @nvie and I came up with the idea of two new subscriptions, which we're calling `room.subscribe("status")` and `room.subscribe("lost-connection")`, which can hugely simplify things for the user by providing a higher-level abstraction.

## `room.subscribe("status")` 

This would **replace** `room.subscribe("connection")` and would return much more user friendly connection statuses. 

```ts
type Status =
  | "initial"
  | "connecting"
  | "connected"
  | "reconnecting"
  | "disconnected"

room.subscribe("status", (status: Status) => {
  // "initial" | "connecting" | "connected" | "reconnecting" | "disconnected"
  console.log(status)
})
```

- The state when loading the page would be `initial -> connecting -> connected`.
- If you lose connection, you will see `reconnecting -> connected`.
- If you fail to authenticate, you will see `disconnected`.

## `room.getStatus()`

This will replace the current `room.getConnectionState()` API.

```ts
// ❌ Old synchronous API
room.getConnectionState()  // "closed" | "open" | "authenticating" | "connecting" | "unavailable" | "failed"

// ✅ New/proposed synchronous API
room.getStatus()  // "initial" | "connecting" | "connected" | "reconnecting" | "disconnected"
```

Note that the new API will no longer let you programmatically distinguish the "authenticating" from "connecting" state, but I don't think it matters at all to an end user? Getting a notification of a change from `authenticating` → `connecting` isn't really actionable for any app.

### React hook: `useStatus`

We could then turn this into a React hook which would be used for displaying the current connection status, for example in a badge.
```tsx
const emoji = { initial: "⚪", connecting: "🟡", connected: "🟢", /* ... */ };

function Component() {
  const status: Status = useStatus();

  return <div>{emoji[status]}</div>;
}
```

## `room.subscribe("lost-connection")`

A new "convenience" subscription which would be used primarily for toast notifications. The callback will only be called if a connection is ever lost. This subscription will not ever be called on an initial successful connection.

```ts
type LostConnectionEvent =
[...]

PR #926, opened 2023-06-07
Fix Suspense boundary bug

nvie:
> **Warning** Intended to land in the `beta` branch.

Fixes the bug from #924. This was found while working on [the RFC for high-level status subscription](https://github.com/liveblocks/liveblocks/issues/878). I've fixed this first now, as a prereq. Some of this will get refactored again in the upcoming PR for that RFC, but I first wanted to ensure this bug was slayed before making more changes.

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -630,13 +630,7 @@ function makeIdFactory(connectionId: number): IdFactory {
   return () => `${connectionId}:${count++}`;
 }
 
-function isConnectionSelfAware(
-  connection: Connection
-): connection is typeof connection &
-  (
-    | { sessionInfo: SessionInfo; lastSessionInfo?: never }
-    | { sessionInfo?: never; lastSessionInfo: SessionInfo }
-  ) {
+function hasSessionInfo(connection: Connection): boolean {
   return (
     connection.status === "open" ||
     connection.status === "connecting" ||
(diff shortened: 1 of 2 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -882,16 +890,29 @@ export function createRoom<
...
-        userInfo: managedSocket.token.parsed.info,
-        userId: managedSocket.token.parsed.id,
-        isReadOnly: isStorageReadOnly(managedSocket.token.parsed.scopes),
+        sessionInfo,
       });
     } else {
-      context.connection.set({ status: newStatus });
+      context.connection.set({
+        status: newStatus,
+        lastSessionInfo: sessionInfo,
```
nvie, line 914: This was the crucial missing piece that caused the bug. When the connection dropped, the status was set to `status: "unavailable"`[^1] and the `sessionInfo` from the previously-`"open"` connection would be lost. Here, we're retaining it, but clearly marking it as "last" session info.

[^1]: Soon: `"reconnecting"`.

packages/liveblocks-core/src/room.ts
```diff
@@ -626,10 +630,18 @@ function makeIdFactory(connectionId: number): IdFactory {
   return () => `${connectionId}:${count++}`;
 }
 
-function isConnectionSelfAware(
```
jrowny, line 629: ha, I'm glad we got rid of this function before Skynet takes over.

jrowny (approved):

PR #930, opened 2023-06-07
Skip over `"authenticating"` connection state

nvie:
> **Warning** Stacked on top of #926, both of which are intended to land in the `beta` branch.

In preparation of making implementation of the new higher-level `Status` values (from [this RFC](https://github.com/liveblocks/liveblocks/issues/878)) easier, this PR makes the room's internal stop using the explicit `"authenticating"` state. This will make implementating the new/proposed connection states a lot easier.

Turns out the `"authenticating"` state was not really carrying any meaning anymore after the state machine rewrite anyway.

### Before (1.0 and lower)
Before this PR, the `connection` state changes would look like this for the happy path:
`closed` -> `authenticating` -> `connecting` -> `open`

### After (soon 1.1 and higher)
After this PR, the `connection` state changes will look like this for the happy path:
`closed` -> `connecting` -> `open`

After this PR, `room.getConnectionState()` will never be `"authenticating"` anymore. In cases where it previously was `"authenticating"` it will now be `"connecting"` instead. There was no behavior coupled to this state, so it's a nice cleanup, and it will allow us to map the new states more cleanly.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -156,7 +156,7 @@ describe("room / auth", () => {
       );
 
       room.connect();
-      await waitUntilStatus(room, "authenticating");
+      await waitUntilStatus(room, "connecting");
       await waitFor(() => consoleErrorSpy.mock.calls.length > 0);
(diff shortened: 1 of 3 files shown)


nvie:
Closing this PR, will just make it part of a bigger PR that this was prepwork for.

PR #934, opened 2023-06-08
Higher level status subscriptions API

nvie:
This introduces a new API to get the room's status using much more user-friendly status names than what we currently offer. **This work is intended to land in the 1.1 release.**

```ts
// ❌ Current/deprecated statuses
room.getConnectionState();  // "closed" | "authenticating" | "connecting" | "open" | "unavailable" | "failed"

// ✅ New/simpler statuses
room.getStatus();  // "initial" | "connecting" | "connected" | "reconnecting" | "disconnected"
```

These should be more user friendly and suitable to display in an application, in the form of a status badge, much like we do in DevTools, too.

After switching to the new statuses, I was able to do some more internal cleanups as a result, which you can see in the diff.

This PR partially implements https://github.com/liveblocks/liveblocks/issues/878:
- [x] A new `room.getStatus()` accessor (to replace `room.getConnectionState()`, which is now deprecated)
- [x] A new `room.subscribe('status')` event (to replace `room.subscribe('connection')`, which is now deprecated)
- [x] For React, a new `useStatus()` hook, which can be used to easily display a connectivity badge
- [ ] **(won't be part of this PR)** The proposed `room.subscribe("reconnect")` subscription API
- [ ] **(won't be part of this PR)** The proposed `useReconnectListener()` hook

I may implement those last two in a separate PR, as I think this change is nice to merge in as-is already.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -178,7 +178,7 @@ describe("room / auth", () => {
     );
 
     room.connect();
-    await waitUntilStatus(room, "failed");
+    await waitUntilStatus(room, "disconnected");
     expect(consoleErrorSpy).toHaveBeenCalledWith(
       "Unauthorized: reason not provided in auth response (403 returned by POST /mocked-api/403)"
     );
@@ -198,7 +198,7 @@ describe("room / auth", () => {
     );
 
     room.connect();
-    await waitUntilStatus(room, "failed");
+    await waitUntilStatus(room, "disconnected");
     expect(consoleErrorSpy).toHaveBeenCalledWith(
       "Unauthorized: wrong key type (401 returned by POST /mocked-api/401-with-details)"
     );
@@ -281,7 +281,7 @@ describe("room", () => {
     room.connect();
     expect(room.getConnectionState()).toBe("connecting");
     expect(room.getStatus()).toEqual("connecting");
-    await waitUntilStatus(room, "open");
+    await waitUntilStatus(room, "connected");
     expect(room.getConnectionState()).toBe("open"); // This API will be deprecated in the future
     expect(room.getStatus()).toEqual("connected");
   });
@@ -293,7 +293,7 @@ describe("room", () => {
     await waitUntilStatus(room, "connecting");
     expect(wss.receivedMessages).toEqual([]);
 
-    await waitUntilStatus(room, "open");
+    await waitUntilStatus(room, "connected");
     expect(wss.receivedMessages).toEqual([
       [
         {
@@ -310,7 +310,7 @@ describe("room", () => {
     room.updatePresence({ x: 0 });
     room.connect();
 
-    await waitUntilStatus(room, "open");
+    await waitUntilStatus(room, "connected");
     expect(wss.receivedMessages).toEqual([
       [
         {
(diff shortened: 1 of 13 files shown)


packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -156,7 +156,7 @@ describe("room / auth", () => {
       );
 
       room.connect();
-      await waitUntilStatus(room, "authenticating");
+      await waitUntilStatus(room, "connecting");
```
nvie, line 159: Note that "authenticating" and "connecting" are no longer considered two different states. In the new API, there will only be "connecting" as to the user it's uninteresting (and unactionable) to know if the machine is internally authenticating or connecting. It's all part of an attempt to establish a connection.

packages/liveblocks-core/src/index.ts
```diff
@@ -14,6 +14,10 @@
 export type { Client } from "./client";
 export { createClient } from "./client";
 export type { BaseAuthResult, Delegates } from "./connection";
+export type {
+  LegacyConnectionStatus as ConnectionStatus,
```
nvie, line 18: While renamed internally everywhere, we keep the existing public type around as-is.

packages/liveblocks-react/src/factory.tsx
```diff
@@ -214,6 +215,13 @@ export function createRoomContext<
     return room;
   }
 
+  function useStatus(): Status {
+    const room = useRoom();
+    const subscribe = room.events.status.subscribe;
+    const getSnapshot = room.getStatus;
+    return useSyncExternalStore(subscribe, getSnapshot, getSnapshot);
```
CTNicholas, line 222: Your new favourite API! So simple here 👍

packages/liveblocks-redux/src/index.ts
```diff
@@ -50,15 +52,14 @@ type LiveblocksContext<
    */
   readonly isStorageLoading: boolean;
   /**
-   * Connection state of the room
+   * @deprecated Old-style connection state of the room. Prefer using the newer
+   * connection statuses.
```
CTNicholas, line 56: Should we tell people the new API is at `room.status`?

packages/liveblocks-react/src/factory.tsx
nvie: It's so good indeed! 💯

packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -1598,41 +1604,48 @@ describe("room", () => {
...
+
+        const ws1 = wss.last;
+        ws1.accept();
+        await waitUntilStatus(room, "connected");
+        expect(room.getConnectionState()).toBe("open"); // This API will be deprecated in the future
+        expect(room.getStatus()).toEqual("connected");
+
+        room.reconnect();
+        expect(room.getConnectionState()).toBe("connecting");
+        expect(room.getStatus()).toEqual("connecting");
```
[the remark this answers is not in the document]
CTNicholas: Gotcha!

CTNicholas (approved):
Great work 🥳

packages/liveblocks-redux/src/index.ts
nvie: Yes we should! Fixed in b2fd48220a71d45d21f2420a21d7eeccbe322c3f.

PR #938, opened 2023-06-13
Fix DevTools to work with upcoming 1.1 beta, too

nvie:
Follow-up on #927. Storage/Presence views would only be rendered when status is `"open"`. This needed to be update for the new status values.

### DIFF
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -118,7 +118,8 @@ If you are using CommonJS, downgrade `node-fetch` to v2.
 
 ### WebSocket throttle [#createClientThrottle]
 
-By default, the client throttle the WebSocket messages sent to 100 milliseconds.
+By default, the client throttles the WebSocket messages sent to 100
+milliseconds.
 
 It is possible to override that configuration with the `throttle` option.
 
--- a/packages/liveblocks-devtools/src/devtools/tabs/presence.tsx
+++ b/packages/liveblocks-devtools/src/devtools/tabs/presence.tsx
@@ -13,7 +13,11 @@ export function Presence({ className, ...props }: ComponentProps<"div">) {
   const others = useOthers();
   const presence = useMemo(() => (me ? [me, ...others] : others), [me, others]);
 
-  if (currentStatus === "open") {
+  if (
+    currentStatus === "connected" ||
+    currentStatus === "open" || // Same as "connected", but only sent by old clients (prior to 1.1)
+    currentStatus === "reconnecting"
+  ) {
     if (presence.length > 0) {
       return (
         <div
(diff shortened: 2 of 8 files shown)


docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -118,7 +118,8 @@ If you are using CommonJS, downgrade `node-fetch` to v2.
 
 ### WebSocket throttle [#createClientThrottle]
 
-By default, the client throttle the WebSocket messages sent to 100 milliseconds.
+By default, the client throttles the WebSocket messages sent to 100
+milliseconds.
```
nvie, line 122: Just a minor typo in the documentation, unrelated to this change. Including it here, because I didn't feel like opening an entire PR for it.

packages/liveblocks-devtools/src/devtools/tabs/presence.tsx
```diff
@@ -13,7 +13,11 @@ export function Presence({ className, ...props }: ComponentProps<"div">) {
   const others = useOthers();
   const presence = useMemo(() => (me ? [me, ...others] : others), [me, others]);
 
-  if (currentStatus === "open") {
+  if (
+    currentStatus === "connected" ||
+    currentStatus === "open" || // Same as "connected", but only sent by old clients (prior to 1.1)
+    currentStatus === "reconnecting"
```
nvie, line 19: Now that we can differentiate between "connecting" and "reconnecting", it's perfectly fine to render the Storage/Presence views while still reconnecting! It's a much nicer experience than to see the views jump back to "loading" again while reconnecting, like it was doing before.

packages/liveblocks-devtools/turbo.json
```diff
@@ -2,12 +2,16 @@
   "$schema": "https://turborepo.org/schema.json",
   "extends": ["//"],
   "pipeline": {
+    "build": {
+      "outputs": ["dist/chrome-**"]
```
nvie, line 6: I've specified the exact build outputs for these commands, because they were interfering with one another in Turbo. For example, when you built Firefox, then Chrome, then Turborepo would think that the last successful build from Chrome included the Firefox assets, and restore them when re-running `turbo build`, overwriting the last successful Firefox output 😬

packages/liveblocks-devtools/turbo.json
```diff
@@ -2,12 +2,16 @@
...
+    },
     "dev:firefox": {
       "dependsOn": ["^build"],
       "cache": false
     },
     "build:firefox": {
-      "dependsOn": ["^build"]
+      "dependsOn": ["^build"],
+      "outputs": ["dist/firefox-**"]
```
nvie, line 14: Ditto here.

CTNicholas (commented):
Awesome!

CTNicholas (approved):

PR #939, opened 2023-06-14
Add support for Starter Kit in deploy-checking script

marcbouchenoire:
This PR attempts to make our deploy-checking script more generic to support other things than examples (by supporting any path), it would only add support for the Starter Kit for now.

When/if we merge this, right before, we'll update the "ignore" commands for all projects in Vercel. Because of that, all existing branches will have to rebase on `main` to get the new script and pass their builds.

### DIFF
--- a/scripts/should-rebuild-on-vercel.sh
+++ b/scripts/should-rebuild-on-vercel.sh
@@ -26,12 +26,12 @@ err () {
 }
 
 usage () {
-    err "usage: should-example-rebuild-on-vercel.sh [-hv] <example-name>"
+    err "usage: should-rebuild-on-vercel.sh [-hv] <project-path>"
     err
-    err "Returns a 0 or 1 exit code, indicating whether the given example should be redeployed."
+    err "Returns a 0 or 1 exit code, indicating whether the given project should be redeployed."
     err
     err "Options:"
-    err "-p    Also trigger on updates to packages dir (default is examples dir only)"
+    err "-p    Also trigger on updates to packages dir (default is project dir only)"
     err "-v    Be verbose"
     err "-h    Show this help"
 }
@@ -56,12 +56,10 @@ elif [ "$#" -gt 1 ]; then
     exit 4
 fi
 
-EXAMPLE="${1%/}"
+PROJECT="${1%/}"
 
-if [ ! -d "examples/$EXAMPLE" ]; then
-    err "Unknown example: $EXAMPLE"
-    err "Valid examples are:"
-    ls examples/ | cat >&2
+if [ ! -d "$PROJECT" ]; then
+    err "Invalid path: $PROJECT"
     exit 5
 fi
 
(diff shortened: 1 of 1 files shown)


scripts/should-rebuild-on-vercel.sh
```diff
@@ -95,9 +93,9 @@ get_all_changed_files () {
         const data = require("./diff-since-main.json");
 
         // If some commits on this branch have the phrase "[no vercel]" or
-        // "[skip examples]" or some combination like that in their message, skip
+        // "[skip deploy]" or some combination like that in their message, skip
```
nvie, line 96: LOL, I forgot we had this! Need to remember, because this is useful to save carbon emissions! 🌍

scripts/should-rebuild-on-vercel.sh
```diff
@@ -114,31 +112,32 @@ get_all_changed_files () {
...
 
 #
-# We're only interested in checking the examples and packages folders, ignore
+# We're only interested in checking some folders, ignore
 # any other changed files.
 #
 get_interesting_changed_files () {
     get_all_changed_files \
       | grep -Ee '^examples/' \
+      | grep -Ee '^starter-kits/' \
```
nvie, line 121: This won't work in a grep pipe. It's like doing

```getAllFiles
  .filter(f => f.startsWith('examples/'))
  .filter(f => f.startsWith('starter-kits/'))
```

That will always give 0 results.

You should say the equivalent of:

```getAllFiles
  .filter(f => f.startsWith('examples/') || f.startsWith('starter-kits/'))
```

That can be done using:

```suggestion
      | grep -Ee '^(examples|starter-kits)/' \
```

scripts/should-rebuild-on-vercel.sh
```diff
@@ -56,12 +56,10 @@ elif [ "$#" -gt 1 ]; then
     exit 4
 fi
 
-EXAMPLE="${1%/}"
+PROJECT="${1%/}"
```
nvie, line 59: This change is fine, as long as you update all the existing invocations to include the `examples/` prefix in the argument.

nvie: Unrelated, but in case you're wondering, the `"${1%/}"` syntax means: take the value of `$1` (the first argument to the script), and strip off the `/` suffix, if there is any. It means that the script will work the same whether you pass `examples/foo` or `examples/foo/` as the argument.

nvie (commented):
Think I spotted one bug, but otherwise this change looks fine to me! 👍

scripts/should-rebuild-on-vercel.sh
marcbouchenoire: Same 😅

scripts/should-rebuild-on-vercel.sh
marcbouchenoire: Thanks for the fix and the explanation!

scripts/should-rebuild-on-vercel.sh
marcbouchenoire: > This change is fine, as long as you update all the existing invocations to include the examples/ prefix in the argument.

Yep I'll update the paths for all existing examples before merging, and update the Notion doc for future examples.

nvie (approved):

PR #941, opened 2023-06-14
Fix future reconnection bug

nvie:
I found a small suboptimalness in our reconnecting logic that would be a (small) bug in the near future (when our new authentication method would land).

When connecting a socket, we should not fall back to re-authenticate in cases where for example the room is full. We just have to back off and try again with the same token. This is differently for cases where there is an explicit "Not Allowed" (4001) code sent by the server (which explicitly asks to re-auth), or an _unexpected_ error (which _could_ be a re-auth case which we cannot detect, so we have to err on the safe end and re-auth anyway).

Similarly, the machine wasn't responding to 4999 codes yet (which should put the client in "disconnected" state explicitly) when coming back from the WebSocket server.

Please note that these test cases address _future_ issues. Our WebSocket server currently does not refuse connections with 4001 Not Allowed or 4999 Close Without Retry just yet, but it will in the (very) near future. The client is adjusted in this PR to respond appropriately in these cases.

### DIFF
--- a/packages/liveblocks-core/src/connection.ts
+++ b/packages/liveblocks-core/src/connection.ts
@@ -304,6 +304,10 @@ function isCloseEvent(
   return !(error instanceof Error) && error.type === "close";
 }
 
+/**
+ * Whether this is a Liveblocks-specific close event (a close code in the 40xx
+ * range).
+ */
 function isCustomCloseEvent(
   error: IWebSocketEvent | Error
 ): error is IWebSocketCloseEvent {
@@ -769,21 +773,25 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
           }; // Should not retry, give up
         }
 
-        // XXX Handle 4001 specially here? Must re-auth!
+        // Server instructs us to reauthenticate
+        if (e.event.code === 4001 /* Not Allowed */) {
+          return {
+            target: "@auth.backoff",
+            effect: [increaseBackoffDelay, logCloseEvent(e.event)],
+          };
+        }
 
         // If this is a custom Liveblocks server close reason, back off more
         // aggressively, and emit a Liveblocks error event...
-        if (e.event.code >= 4000 && e.event.code <= 4100) {
+        if (isCustomCloseEvent(e.event)) {
           return {
             target: "@connecting.backoff",
             effect: [
               increaseBackoffDelayAggressively,
               logCloseEvent(e.event),
-              (_, { event }) => {
-                if (event.code >= 4000 && event.code <= 4100) {
-                  const err = new LiveblocksError(event.reason, event.code);
-                  onLiveblocksError.notify(err);
-                }
+              () => {
+                const err = new LiveblocksError(e.event.reason, e.event.code);
+                onLiveblocksError.notify(err);
               },
             ],
           };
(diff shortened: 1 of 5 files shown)


packages/liveblocks-core/src/connection.ts
```diff
@@ -261,10 +263,55 @@ function log(level: LogLevel, message: string) {
...
+    console.warn(
+      `Connection to Liveblocks websocket server closed (code: ${event.code}). Retrying in ${ctx.backoffDelay}ms.`
+    );
+  };
+}
+
+const logPermanentClose = log(
+  LogLevel.WARN,
+  "Connection to WebSocket closed permanently. Won't retry."
+);
```
nvie, line 293: This just DRYs up logs that already exist, but were inlined in the machine's state definitions below.

packages/liveblocks-core/src/connection.ts
```diff
@@ -261,10 +263,55 @@ function log(level: LogLevel, message: string) {
...
+
+/**
+ * Whether this is a Liveblocks-specific close event (a close code in the 40xx
+ * range).
+ */
+function isCustomCloseEvent(
+  error: IWebSocketEvent | Error
+): error is IWebSocketCloseEvent {
+  return isCloseEvent(error) && error.code >= 4000 && error.code < 4100;
+}
```
nvie, line 313: Similar DRY up of custom code handling. This was duplicates a few times before.

packages/liveblocks-core/src/connection.ts
```diff
@@ -724,29 +767,29 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
...
           }; // Should not retry, give up
         }
 
+        // Server instructs us to reauthenticate
+        if (e.event.code === 4001 /* Not Allowed */) {
+          return {
+            target: "@auth.backoff",
+            effect: [increaseBackoffDelay, logCloseEvent(e.event)],
+          };
+        }
```
nvie, line 780: This is an addition that fixes the bug. All WS status codes just retry connecting (with an aggressive backoff), but 4001 specifically is an instruction from the server to get a new token, because the current one is not authorized explicitly. Note that in this case we transition to `@auth.backoff` instead of `@connecting.backoff`, which we do in all other cases.

packages/liveblocks-core/src/connection.ts
```diff
@@ -605,44 +653,39 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
...
+        // Allowed).
+        if (isCustomCloseEvent(err) && err.code !== 4001 /* Not Allowed */) {
+          return {
+            target: "@connecting.backoff",
+            effect: [
+              increaseBackoffDelayAggressively,
+              logPrematureErrorOrCloseEvent(err),
+            ],
+          };
+        }
```
nvie, line 683: Similarly, these two cases address bugs. If the connection is refused[^1] with a 4999, we stop the machine and disconnect permanently. If we get a 4001, we re-authenticate. In all other cases, we can just retry after a backoff delay.

[^1]: This is what I've started to call the pattern of the server accepting-then-immediately-closing the WS connection with a close code. This verb matches the function name we already use in the backend.

CTNicholas (approved):

PR #943, opened 2023-06-15
Fix bug where storage won't always load after reconnection

### LINKED ISSUE #914: Storage won't always load after reconnection
### Describe the bug

When a room is full, a new client will not be able to connect to the room, and the client will keep retrying to connect (using slow backoff delays). If a slot in the room frees up, the client will automatically reconnect and continue loading. However, **only presence will be loaded, but storage will not**.

### To Reproduce

1. Spin up a room
2. Open many tabs until the max number of people in the room is hit
3. Open one more tab - this one will not load at all (no presence & no storage) — this is expected
4. Close one of the existing tabs
5. Wait for the last opened tab to automatically reconnect and start loading
6. Notice how only presence gets loaded and storage won't

We found this bug while stress-testing the new connection manager that's in the making (currently `1.1.0-fsm4`). However, it turns out this bug isn't introduced there, and exists as far back as Liveblocks 0.15.x (at least, perhaps even earlier).

### Expected behavior

Both Presence and Storage will load.

### Illustrations

@CTNicholas captured this beautifully in a Loom video.

<a href="https://www.loom.com/share/93a99388d5914dbcbb41314238c832d5"></a>

nvie:
Fixes #914.

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -1000,6 +1000,12 @@ export function createRoom<
     // room connection happens, we won't know the connection ID here just yet.
     context.idFactory = makeIdFactory(sessionInfo.id);
 
+    // XXX This needs to be fixed! We _also_ need to fetch if storage was
+    // _attempted_ to load before, but it failed to do so!
+    // XXX However, not if it happened _too close_ to the last attempt (because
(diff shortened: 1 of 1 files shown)


nvie:
Please still review this PR + leave comments! 🙏 
Happy to address anything in hindsight.

PR #945, opened 2023-06-16
docs: Update links in schema-lang README to correct ones

RickyC0626:
This fixes the links in `schema-lang/README.md` to point to the correct paths.
Before, the path linked to `schema-lang` twice, so it became `schema-lang/schema-lang/{path}`.


nvie:
Thanks for fixing these links (and for checking out our schema language)! 🙏

PR #949, opened 2023-06-16
Fix @liveblocks/react lint script

RickyC0626:
## Description

Linting fails in package `@liveblocks/react` when running `npm run lint` due to the lint path being `src/;`. 

### Before
```
@liveblocks/react:lint: 
@liveblocks/react:lint: > @liveblocks/react@1.0.12 lint
@liveblocks/react:lint: > eslint src/; npm run check:exports
@liveblocks/react:lint:
@liveblocks/react:lint: Oops! Something went wrong! :(
@liveblocks/react:lint:
@liveblocks/react:lint: ESLint: 8.38.0
@liveblocks/react:lint:
@liveblocks/react:lint: No files matching the pattern "src/;" were found.
@liveblocks/react:lint: Please check for typing mistakes in the pattern.
@liveblocks/react:lint:
@liveblocks/react:lint: npm ERR! Lifecycle script `lint` failed with error:
@liveblocks/react:lint: npm ERR! Error: command failed
@liveblocks/react:lint: npm ERR!   in workspace: @liveblocks/react@1.0.12
```

### After
```
@liveblocks/react:lint: 
@liveblocks/react:lint: > @liveblocks/react@1.0.12 lint
@liveblocks/react:lint: > eslint src/ && npm run check:exports
@liveblocks/react:lint:
@liveblocks/react:lint: > @liveblocks/react@1.0.12 check:exports
@liveblocks/react:lint: > tsc scripts/check-factory-exports.ts && node scripts/check-factory-exports.js
@liveblocks/react:lint:
```

PR #950, opened 2023-06-16
Fix format & lint scripts for all `package.json`

RickyC0626:
## Description

Semicolon is apparently not a valid command separator in Windows, so `npm run format` will fail when running locally. To preserve the original behavior of running prettier even if eslint fails, the command is wrapped like so:
```json
"format": "(eslint --fix src/ || true) && prettier --write src/"
```

### DIFF
--- a/packages/liveblocks-client/package.json
+++ b/packages/liveblocks-client/package.json
@@ -12,7 +12,7 @@
   "scripts": {
     "dev": "tsup --watch",
     "build": "tsup",
-    "format": "eslint --fix src/ && prettier --write src/",
+    "format": "(eslint --fix src/ || true) && prettier --write src/",
     "lint": "eslint src/",
     "test": "jest --silent --verbose --color=always --passWithNoTests",
     "test:watch": "jest --silent --verbose --color=always --passWithNoTests --watch"
--- a/packages/liveblocks-client/package.json
+++ b/packages/liveblocks-client/package.json
@@ -12,7 +12,7 @@
   "scripts": {
     "dev": "tsup --watch",
     "build": "tsup",
-    "format": "eslint --fix src/; prettier --write src/",
+    "format": "eslint --fix src/ && prettier --write src/",
     "lint": "eslint src/",
     "test": "jest --silent --verbose --color=always --passWithNoTests",
     "test:watch": "jest --silent --verbose --color=always --passWithNoTests --watch"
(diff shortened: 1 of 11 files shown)


nvie:
Hi @RickyC0626 – thank you for contributing! Unfortunately, the behavior you're trying to fix here is intentional, see [this commit](https://github.com/liveblocks/liveblocks/commit/51f5a1f28d43ae003a3e8c996326e0cdb145ae8f) where it was deliberately introduced. It was done so that even if there were linting errors, the prettier step would still (also) run. We didn't just flip the commands, because some auto-fixes by ESLint would also require prettier to re-run.

That said, looking at the "before" error you reported in #949, it looks like your shell isn't splitting the command on the `;`, and passing `src/;` as the first param to ESlint. I have never seen a shell that would do that. What shell are you using by any chance? Curious to see if adding a space between the `/` and the `;` would work to resolve this issue for you?

RickyC0626:
> Hi @RickyC0626 – thank you for contributing! Unfortunately, the behavior you're trying to fix here is intentional, see [this commit](https://github.com/liveblocks/liveblocks/commit/51f5a1f28d43ae003a3e8c996326e0cdb145ae8f) where it was deliberately introduced. It was done so that even if there were linting errors, the prettier step would still (also) run. We didn't just flip the commands, because some auto-fixes by ESLint would also require prettier to re-run.
> 
> That said, looking at the "before" error you reported in #949, it looks like your shell isn't splitting the command on the `;`, and passing `src/;"` as the first param to ESlint (instead of the expected `"src/"` – without the `;`). I have never seen a shell that would do that. What shell are you using by any chance? Curious to see if adding a space between the `/` and the `;` would work to resolve this issue for you?

Ahh I see, that's quite interesting. I'm running a Git bash terminal on Windows 11. I've tried adding a space between `/` and `;`, but that didn't seem to work either. I tried other terminals such as Powershell and Command Prompt, but still the same error occurs. Perhaps this is a Windows-only thing.

RickyC0626:
Perhaps modifying the script to this will keep the script's desired behavior, but also make it work on Windows.
```json
```

nvie:
Oh wow, TIL'ed that `;` isn't a command separator on Windows 🤯
Feel free to close #949, and update this PR to use your last proposed command instead. That will still work on all Unix-based platforms, and it will be nice if it works on Windows too. Thank you, @RickyC0626! 🙏

nvie (approved):

nvie:
Thanks, @RickyC0626 🙏 !

PR #951, opened 2023-06-19
Update config-generation for 1.1

CTNicholas:

### DIFF
--- a/packages/create-liveblocks-app/bin/templates/init/config-generation.ts
+++ b/packages/create-liveblocks-app/bin/templates/init/config-generation.ts
@@ -142,7 +142,7 @@ function allHooks() {
   useCanRedo,
   useMutation,
   useStatus,
-  useLostConnectionListener`;
+  useLostConnectionListener,`;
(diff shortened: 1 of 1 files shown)


packages/create-liveblocks-app/bin/templates/init/config-generation.ts
```diff
@@ -140,7 +140,9 @@ function allHooks() {
   useRedo,
   useCanUndo,
   useCanRedo,
-  useMutation,`;
+  useMutation,
+  useStatus,
+  useLostConnectionListener,`;
```
nvie, line 145: Just double-checking: will this also add it to the suspense block?

nvie (approved):

packages/create-liveblocks-app/bin/templates/init/config-generation.ts
CTNicholas: > Just double-checking: will this also add it to the suspense block?

Yep!

PR #953, opened 2023-06-20
Make DevTools future-compatible

nvie:
Due diligence, after the 1.1 release from yesterday. This PR stops relying on the exported `LegacyConnectionStatus` type in DevTools. This type will be removed in a future version of Liveblocks, but in DevTools we will always have to remain compatible with those old status names.


nvie:
@CTNicholas Normally I would ask Marc to review this, but since he's out until next week, would you remind taking a look? It's only small.

marcbouchenoire (approved):

PR #955, opened 2023-06-21
Add unit test for patching `LiveList` state using `set` update type

### LINKED ISSUE #293: Add a unit test for the bug uncovered by #292
https://github.com/liveblocks/liveblocks/pull/292#discussion_r887722211

RickyC0626:
## Description

Fixes #293. This test ensures expected behavior of `legacy_patchImmutableObject` when patching `LiveList` state via `set` update. Redux and Zustand packages are expected to continue using the legacy patch functions until a reimplementation using `.toImmutable()` is done.

Local testing shows successful coverage.

image

### DIFF
--- a/packages/liveblocks-core/src/__tests__/immutable.test.ts
+++ b/packages/liveblocks-core/src/__tests__/immutable.test.ts
@@ -914,6 +914,41 @@ describe("legacy_patchImmutableObject", () => {
     });
   });
 
+  test("replace an element in Array/LiveList", () => {
+    const state = {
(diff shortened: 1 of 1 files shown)


nvie (approved):
Wow, @RickyC0626 — this is amazing! 😍 I can confirm this is the correct unit test. I reverted my fix from https://github.com/liveblocks/liveblocks/pull/292#discussion_r887722211 briefly, and the test does indeed fail. Thanks so much for taking the time to write this test; super appreciated! 🙏

RickyC0626:
@nvie Thanks for the quick review!

PR #956, opened 2023-06-22
Clean up exported auth types

nvie:
Cleans up the AuthToken types exported by `@liveblocks/core`. Some backend systems have been relying on these type definitions, but since these are versioned (older client versions could have different type definitions), this setup wasn't ideal for backend systems to consume, because they would always have to account for other in-the-wild versions of these shapes.

In addition, this PR now stops the runtime validation of the following fields:

- `appId`
- `roomId`
- `maxConnectionsPerRoom`

Those fields are intended for the backend only, and the client has no business validating them. After this PR, those fields are treated as opaque, and simply passed through to the backend.

Because we introduced this runtime check a long time ago, we cannot touch or change any of these fields without breaking old clients, unfortunately — see this [overview sheet](https://docs.google.com/spreadsheets/d/1mP6WAdvlkX6GpFjGZ0gvImFbcK-NaUvFxTeflKyYGEE/edit?usp=sharing) I put together.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/_behaviors.ts
+++ b/packages/liveblocks-core/src/__tests__/_behaviors.ts
@@ -9,17 +9,17 @@
  */
 
 import { StopRetrying } from "../connection";
-import type { RichToken } from "../protocol/AuthToken";
+import type { ParsedAuthToken } from "../protocol/AuthToken";
 import type { RoomDelegates } from "../room";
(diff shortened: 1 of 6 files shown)


jrowny (approved):
😎 Down with richtokens, long live parsedtokens!

FlowFlorent (approved):

PR #957, opened 2023-06-22
Add types to constants

jrowny:
Adds a few message types to use in upcoming features.

### DIFF
--- a/packages/liveblocks-core/src/protocol/ClientMsg.ts
+++ b/packages/liveblocks-core/src/protocol/ClientMsg.ts
@@ -9,6 +9,10 @@ export enum ClientMsgCode {
   // For Storage
   FETCH_STORAGE = 200,
   UPDATE_STORAGE = 201,
+
+  // For YJS support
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/protocol/ClientMsg.ts
```diff
@@ -72,3 +80,13 @@ export type UpdateStorageClientMsg = {
 export type FetchStorageClientMsg = {
   readonly type: ClientMsgCode.FETCH_STORAGE;
 };
+
+export type FetchYDocClientMsg = {
+  readonly type: ClientMsgCode.FETCH_YDOC_UPDATE;
+  readonly vector: string;
```
nvie, line 86: Perhaps leave a small comment here about what should go in this `vector` value?

nvie (approved):
Looks good! Just a couple minor requests, but nothing blocking.

nvie (approved):
:shipit:

PR #960, opened 2023-06-23
Add NPM package linting checks to CI

nvie:
This PR adds two amazing tools to our CI setup that I've been using manually to verify NPM package configuration since a while:

- https://publint.dev/ (by @bluwy)
- https://arethetypeswrong.github.io/ (by @andrewbranch)

These tools helps lint our package.json files and published type definitions—_before_ publishing them! I recently learned that ATTW now also has a CLI, like publint has had for a while. The beauty is not only that they error when there is an issue, but they explain in great detail _why_ that setup is incorrect (or may lead to surprising results). Super useful information to learn about all the edge cases you may not be thinking about.


packages/liveblocks-client/package.json
```diff
@@ -14,6 +14,7 @@
     "build": "tsup",
     "format": "(eslint --fix src/ || true) && prettier --write src/",
     "lint": "eslint src/",
+    "lint:package": "npx -y publint@latest --strict && npx -y @arethetypeswrong/cli@latest /tmp/$( npm pack -q --pack-destination /tmp )",
```
nvie, line 17: We may consider installing these tools as top-level dev dependencies:

https://github.com/liveblocks/liveblocks/blob/main/package.json#L18-L25

But I've deliberately set this up to always use the latest version, because these tools are still in development, and if new best-practices arise, we'd always see the latest recommendations here. If we think it's not a good idea, I'm happy to pin this dependency explicitly too.

nvie: This morning, the latest version of `attw` broke when running on Node 16. To avoid breaking our builds, I've pinned the versions explicitly anyway, see d53d4ca. Let's just upgrade them regularly to stay up to date instead.

jrowny (approved):
looks good! Will add this to the new yjs package too

nvie:
> looks good! Will add this to the new yjs package too

@jrowny I already did on your branch—see https://github.com/liveblocks/liveblocks/blob/0f84d4183eda874bd16231cb51fa927cc133fcd0/packages/liveblocks-yjs/package.json#L17. Should work as soon as you back-merge `main` into your branch.

jrowny:
> > looks good! Will add this to the new yjs package too
> 
> @jrowny I already did on your branch—see
> 
> https://github.com/liveblocks/liveblocks/blob/0f84d4183eda874bd16231cb51fa927cc133fcd0/packages/liveblocks-yjs/package.json#L17
> 
> . Should work as soon as you back-merge `main` into your branch.

even better! thanks!

andrewbranch:
By the way, the change to attw that introduced the Node 16 break was npm packing in-place. You can now do `attw --pack` instead of `attw /tmp/$( npm pack -q --pack-destination /tmp )` if you like 😄

PR #965, opened 2023-06-29
Ignore events sent to the state machine after it has been stopped

nvie:
This PR fixes a bug that was reported in our Discord channel yesterday.


stevenfabre (approved):

CTNicholas (approved):

nvie:
Thanks for the quick approval. Will roll a release for this now.
@jrowny Please do give this a look in hindsight if you want, as you're most knowledgable in this area.

PR #967, opened 2023-06-29
Update `@liveblocks/node` APIs

nvie:
Implements the [ideas proposed here](https://www.notion.so/liveblocks/New-liveblocks-node-APIs-656345fffd574df7976fa4494f3b6114).


packages/liveblocks-node/src/PermissionSet.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const MAX_PERMS_PER_SET = 10;
+
+/**
+ * Assign this to a room or room pattern if you want to grant the user
+ * read-only permissions for this room.
+ */
+export const READ_ACCESS: readonly Permission[] = Object.freeze([
+  "room:read",
+  // "room:presence:write",  // XXX Should this be included in the READ_ACCESS set?
```
nvie, line 25: @ofoucherot What do you think of this? Should we include the `room:presence:write` permission in the `READ_ACCESS` preset? This may be a bit confusing because it would mean there is a write permission included something called `READ_ACCESS`, but… I still think it would make for the best default. (Write access to your own sandbox aka Presence seems like the most useful default here.) We could add an alternative preset that is truly, completely, read-only, but I think the majority wouldn't use it and these presets try to optimize for the majority.

packages/liveblocks-node/src/PermissionSet.ts
```diff
@@ -0,0 +1,146 @@
...
+ * permissions for this room.
+ */
+export const FULL_ACCESS: readonly Permission[] = Object.freeze([
+  "room:write",
+  "comments:write",
+
+  // XXX Are these implied automatically if you have the two write permissions above? If so, it would be nice to keep the JWT payload as small as possible.
+  // "room:read",
+  // "room:presence:write",
+  // "comments:read",
```
nvie, line 40: @ofoucherot I have a similar question here.

packages/liveblocks-node/src/PermissionSet.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      perms = new Set<Permission>();
+      this._permissions.set(roomId, perms);
+      return perms;
+    }
+  }
+
+  public allow(
+    roomIdOrPattern: string,
+    newPerms: readonly Permission[] = FULL_ACCESS
```
nvie, line 105: @ofoucherot Also, what do you think of this default (and the [related unit test](https://github.com/liveblocks/liveblocks/blob/5a779cd91cc930fa437bd608422f0bc20a9af451/packages/liveblocks-node/src/__tests__/PermissionSet.test.ts#L77-L89))? Is it convenient to make it the default, or should we remove this and ask people to be explicit about it?

I still kinda dislike that we have to import the presets from the global module, i.e.

```ts
import { FULL_ACCESS } from ...
```

Perhaps we can put these on the `PermissionSet` class as static members instead?

```ts
const perms = new PermissionSet();

// Alternative 1
perms.allow('my-room', PermissionsSet.FULL_ACCESS);

// Alternative 2
perms.allow('my-room', PermissionsSet.presets.FULL_ACCESS);

// Alternative 3
perms.allow('my-room', perms.FULL_ACCESS);
```

Do you have a preference for any of those alternatives, or shall we just keep what we have now?

packages/liveblocks-node/src/PermissionSet.ts
```diff
@@ -0,0 +1,146 @@
...
+ *
+ *    p
+ *      .allow('abc:*',   FULL_ACCESS)
+ *      .allow('abc:123', READ_ACCESS)
+ *
+ * Here, room `abc:123` would have full access. The second .allow() call only
+ * _adds_ read permissions, but that has no effect since full access
+ * permissions were already added to the set.
+ */
+export class PermissionSet {
```
nvie, line 78: Also, I'm not particularly happy with the name `PermissionSet` yet. I previously had `Permissions`, but I wanted to avoid a name clash with the built-in [Permissions API](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API). Perhaps we should just use `Permissions` anyway and not worry about it too much?

packages/liveblocks-node/src/PermissionSet.ts
ofoucherot: > Write access to your own sandbox aka Presence seems like the most useful default here
I agree! It should be the default.

packages/liveblocks-node/src/PermissionSet.ts
ofoucherot: Write implies read yes, you don't need to put both scopes.

packages/liveblocks-node/src/PermissionSet.ts
ofoucherot: I prefer to ask people to be explicit about it, it's permission, it's best if they know exactly what accesses they are giving.

I prefer alternative 2
`perms.allow('my-room', PermissionsSet.presets.FULL_ACCESS);`

packages/liveblocks-node/src/PermissionSet.ts
ofoucherot: What about `LiveblocksPermission`?

nvie: Could work. Would you prefer:

```typescript
import { Liveblocks, LiveblocksPermissions } from …

const liveblocks = new Liveblocks()

// Inside handler
const permissions = new LiveblocksPermissions()
``` 

or maybe do this pattern?

```typescript
import { Liveblocks } from …

const liveblocks = new Liveblocks()

// Inside handler
const permissions = new liveblocks.Permissions()
``` 

The second one seems a bit cleaner, not? Even though getting the class from the Liveblocks instance may not be too common? I don’t hate it.

nvie:
I've implemented the [new API idea](https://www.notion.so/liveblocks/Proposal-New-liveblocks-node-APIs-656345fffd574df7976fa4494f3b6114?pvs=4#d13d8ea263874dde94ad805b26d59985) in 10e5a4e448ecf801ba9f1b4f140b4374fbde4e58. What do you think?

packages/liveblocks-node/src/PermissionSet.ts
nvie: I make this argument required in 02fa3a940352834f012d450dde74c0b55ac7df15 👍 
Also, in the latest iteration, the syntax now is:

```ts
const session = liveblocks.prepareSession();
session.allow('my-room', session.FULL_ACCESS);
```

This way, we don't need to import any globals and it's all very easy to work with 👍

nvie:
Whoops! Renaming the branch accidentally closed this PR. Will reopen. Sorry!

PR #969, opened 2023-06-30
Upgrade all examples to 1.1.1

nvie:
This upgrades all examples to use 1.1.1, which was published yesterday. I skipped all the Yjs examples, since those are still in beta and there is no official 1.1.1 version for those yet.


nvie:
@CTNicholas I forget, is there somewhere inside `packages/create-liveblocks-app` we need to bump the version, too?

CTNicholas (approved):

CTNicholas:
> @CTNicholas I forget, is there somewhere inside `packages/create-liveblocks-app` we need to bump the version, too?

Nope, it just copies from the other repos!

nvie:
Ohhh, right — I got confused, because I thought you had to update it for 1.1 recently, but now I realize t was because we added a new hook, not because the version was bumped! 👍

PR #973, opened 2023-07-01
Update nextjs.mdx to reflect change from --react to --framework react

mxstbr:


CTNicholas (approved):
Thanks for catching this @mxstbr!

PR #979, opened 2023-07-04
CI fix: TypeScript errors should fail builds

nvie:
It's possible to have TypeScript errors, but `tsup` would still build successfully. Most TypeScript errors are caught by `tsup`, but if there are errors in places that don't end up in the build output — i.e. when you only have TypeScript errors in test files, then this could still pass. This makes out build commands resilient against these cases.

### DIFF
--- a/packages/liveblocks-client/package.json
+++ b/packages/liveblocks-client/package.json
@@ -11,7 +11,7 @@
   ],
   "scripts": {
     "dev": "tsup --watch",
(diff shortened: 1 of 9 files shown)


nvie:
Actually, nevermind. This will make the builds slower (which can add up when you need to run 3 or more nested Turbo builds). Besides, Jest will catch those kinds of TypeScript issues for us after all, just in a different CI phase.

PR #980, opened 2023-07-04
New auth manager

ofoucherot:
Intended to be part of [release 1.2](https://github.com/liveblocks/liveblocks/pull/975).

---

Implements new `auth-manager`. 
`auth-manager` is responsible for providing an auth value, which is either a public API key or a JWT token.

TODO:
- [ ] Auth manager tests.
- [ ] Fix room tests.
- [ ] Integrate with protocol v7 changes (public key auth, actorId).


packages/liveblocks-core/src/room.ts
```diff
@@ -934,15 +918,19 @@ export function createRoom<
...
-    const token = managedSocket.token?.parsed;
-    if (token !== undefined && token !== lastToken) {
-      context.sessionInfo.set({
-        id: token.actor,
-        userInfo: token.info,
-        userId: token.id,
-        isReadOnly: isStorageReadOnly(token.scopes),
-      });
-      lastToken = token;
+    if (managedSocket.authValue?.type === "secret") {
```
ofoucherot, line 921: @nvie We will have to merge this part with your changes.

packages/liveblocks-core/src/__tests__/_behaviors.ts
```diff
@@ -36,15 +36,15 @@ function makeToken(actor: number, scopes: string[]): ParsedAuthToken {
  * observe the effects of the Room taking actions.
  */
 export function defineBehavior(
-  authBehavior: AuthBehavior,
+  authValue: AuthValue,
```
nvie, line 39: This was deliberately taking a "behavior" instead of a single value. By letting this be a function, you would be able to set up a unit test that would [fail authorization](https://github.com/liveblocks/liveblocks/blob/3f3ce1baeaef330d7247525105059850e702d083/packages/liveblocks-core/src/__tests__/connection.test.ts#L14), or even fail _re_-authorization for example, i.e. pass it a function that would be stateful and return a different token on the second call.

I'm doing that a lot for the socket behavior already, for example [this unit test](https://github.com/liveblocks/liveblocks/blob/3f3ce1baeaef330d7247525105059850e702d083/packages/liveblocks-core/src/__tests__/room.test.ts#L324-L328), which sets up a test where the socket connection will fail to connect only on the second try. For some reason I didn't do it for the auth behavior yet, apparently, but that's mostly an oversight on my part than that it was intentional.

PR #981, opened 2023-07-05
New guide: Reauthenticate without reloading the page

CTNicholas:


guides/pages/reauthenticate-without-reloading-the-page-or-losing-state.mdx
```diff
@@ -0,0 +1,36 @@
+---
+meta:
+  title: "Reauthenticate without reloading the page or losing state"
+  description:
+    "Learn how to reauthenticate the current room without refreshing the page"
+---
+
+By calling
+[room.reconnect](/docs/api-reference/liveblocks-client#Room.reconnect) you can
+reconnect and reauthenticate your collaborative application without refreshing
```
nvie, line 10: My main question reading this would be: why would you want to reauthenticate your application? Perhaps good to add a few words on why you would want to do this in the first place?

nvie (approved):
Writing these quick guides is so nice. Thank you! 🙌

guides/pages/reauthenticate-without-reloading-the-page-or-losing-state.mdx
CTNicholas: Thanks for the recommendation, yes! Added

PR #984, opened 2023-07-06
Add proper sync status

jrowny:
- Adds sync property to server message when the full doc is being sync'd from the server


packages/liveblocks-core/src/room.ts
```diff
@@ -1713,7 +1718,8 @@ export function createRoom<
           }
 
           case ServerMsgCode.UPDATE_YDOC: {
-            eventHub.ydoc.notify(message.update);
+            const { isSync, update } = message;
+            eventHub.ydoc.notify({ update, isSync });
```
nvie, line 1722: Could we save the unpacking here, or did you do this deliberately?

```suggestion
            eventHub.ydoc.notify(message);
```

nvie (approved):
This looks good to me, simple enough!

I do want to call out that we have a similar API for Storage, which might be good to know in case you've never seen it. Maybe it would be nice to expose a similar API for YDoc sync status? Doesn't have to be part of this PR, but if that's something we would like to offer, it may be good to do it now before we need to start deprecation cycles, etc etc:

https://liveblocks.io/docs/api-reference/liveblocks-client#Room.getStorageStatus

nvie (approved):
:shipit:

PR #985, opened 2023-07-06
Fix a small export bug in `@liveblocks/core`

nvie:
Fixes a bug where `toPlainLson` was not exported correctly at the top-level.
Also upgrade publint to the latest version.

### DIFF
--- a/packages/liveblocks-core/src/index.ts
+++ b/packages/liveblocks-core/src/index.ts
@@ -35,7 +35,8 @@ export type {
   LiveObjectUpdate,
   StorageUpdate,
 } from "./crdts/StorageUpdates";
-export type { ToImmutable, toPlainLson } from "./crdts/utils";
+export type { ToImmutable } from "./crdts/utils";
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/index.ts
```diff
@@ -35,7 +35,8 @@ export type {
   LiveObjectUpdate,
   StorageUpdate,
 } from "./crdts/StorageUpdates";
-export type { ToImmutable, toPlainLson } from "./crdts/utils";
+export type { ToImmutable } from "./crdts/utils";
+export { toPlainLson } from "./crdts/utils";
```
nvie, line 39: @jrowny Sorry that I overlooked this when I reviewed #871. I didn't spot that this was a `type`-level export statement, which gets removed from the final bundle and is a TypeScript-only construct. Of course, the `toPlainLson` function needs to end up in the bundle, so I made this a normal `export` here.

CTNicholas (approved):
Nice!

PR #997, opened 2023-07-12
Detect multiple copies of Liveblocks

nvie:
This PR adds a new feature to all of our packages: detecting multiple copies of Liveblocks in the same bundle. This is enabled by placing a marker on the `globalThis` object when a module is imported, and if such marker already exists for that package, throw an error (when in development mode) or log the error to the console (when in production mode).

## Detecting multiple instances

This situation can arise in different circumstances:

- Two versions of Liveblocks are used by multiple dependencies in your project (for example when not correctly declaring those as peerDependencies)
- Two instances of the same version are included in the bundle, due to dual-package hazard

#### Error dialog shown

Screen Shot 2023-07-12 at 12 08 55@2x

#### Error page in the docs

Screen Shot 2023-07-12 at 12 32 41@2x

## Detect cross-linking

Also, this PR adds a check that will throw if there is cross-linking between versions going on somehow, for example, when you have `@liveblocks/core` and `@liveblocks/client` in the same bundle, but on different versions. This doesn't normally happen, but can happen with HMR when you `npm install` a package while a server is running.

#### Error dialog shown

Screen Shot 2023-07-12 at 12 08 36@2x

#### Error page in the docs

Screen Shot 2023-07-12 at 12 32 26@2x


nvie:
With this PR landed, I think we can re-enable our dual ESM/CJS builds again and publish that as 1.1.5.

GuillaumeSalles:
Implementation and error pages are :chef-kiss: !

I want to try to play with it a bit more to make sure they're are no way to break this. Especially with SSR. I want to make sure we don't create false positives. I'll accept the PR when I'm done with the manual testing

stevenfabre (commented):
I love the DX of putting links in the console + the guides you wrote corresponding to those links.

Great DX comes from "little" (not so little) details like that. It's great to see that you always think DX first. Thank you.

nvie:
@GuillaumeSalles Appreciate the effort—lmk if you find a false positive! 🙏

nvie:
Because the whole point of this protection is to be able to ship dual ESM/CJS again, I've re-enabled dual packaging on this branch.

@GuillaumeSalles For your testing convenience, the following packages exist on NPM that you can use if you want to try out various things locally:

- `v1.1.5-test1` contains an (earlier) version of the detection introduced by this PR, but is **CJS-only**
- `v1.1.5-test2` is **100% identical** to `v1.1.5-test1` (except the version number)
- `v1.1.5-test3` contains the latest version of the detection implementation, the "non-enumerable" fix from yesterday[^1], but is still **CJS-only**
- `v1.1.5-test4` is **100% identical** to `v1.1.5-test3`, but is **dual-packaged ESM + CJS**
- `v1.1.5-test5` is **100% identical** to `v1.1.5-test4` (except the version number)

This effectively makes `v1.1.5-test{4,5}` the official release candidate. You can use these 5 versions to easily test various upgrade path behaviors locally.

[^1]: I only mention the "non-enumerable" fix from #998 as a recognition point in case you want to inspect the bundle output and search for things. It's not a relevant change for this PR otherwise.

GuillaumeSalles (approved):
I couldn't manage to create a false positive! 

I still have a feeling that something will break with Jest or some weird SSR setup, but let's see if we get report issue in the coming days

PR #998, opened 2023-07-12
Make room internals non-enumerable

nvie:
This solves the Immer incompatibility issue in a more robust way, since Immer will not freeze non-enumerable properties.


GuillaumeSalles (approved):
Kind of crazy that we have to do this, but LGTM!

PR #1001, opened 2023-07-13
Fix error links

nvie:
Fix the error links in the next version. These broken links have been fixed with a redirect too, so no urgency to get this published.


stevenfabre (approved):
Nice. We need to remember to remove the redirects once this lands in a release.

PR #1002, opened 2023-07-14
Fix fast LiveList.insert + LiveList.set bug

GuillaumeSalles:
### Description

This PR ensures that insert acknowledge op is ignored if the item has been deleted by a local `LiveList.set`.

I took the opportunity to add comments and do some cleanups.

**Before merging this, we need to make sure e2e tests are passing on public and private repositories. (We might need to fix our tests setup to work with ESM)**

_Review commit by commit recommended_

#### How to test it

Upgrade our block based editor example to use `1.2.2-test3` and try to reproduce the behavior described in #731. This should also fix the "ghosting bug" mentioned by Fermat and Arcol.

#### Related issue(s)

- #628
- #731
- #1051


packages/liveblocks-core/src/crdts/LiveList.ts
```diff
@@ -870,7 +892,7 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
     this._pool?.assertStorageIsWritable();
     if (index < 0 || index > this._items.length) {
       throw new Error(
-        `Cannot insert list item at index "${index}". index should be between 0 and ${this._items.length}`
+        `Cannot insert list item at index "${index}". index should be between 0 and ${this._items.length}`
```
GuillaumeSalles, line 895: These error messages included some invisible characters between `"` and `$`

packages/liveblocks-core/src/crdts/__tests__/LiveList.test.ts
```diff
@@ -1122,6 +1122,66 @@ describe("LiveList", () => {
...
+          opId: "1:0",
+          type: OpCode.CREATE_REGISTER,
+          parentId: "0:1",
+          parentKey: FIRST_POSITION,
+          data: "A",
+        },
+      ]);
+
+      expectStorage({
+        items: ["B"],
```
GuillaumeSalles, line 1164: Before the fix, items were equal to `["A","B"]` here.

jamesbvaughan:
Hi @GuillaumeSalles, we believe that we're hitting the error that this PR intends to fix. Do you know when we should expect more update here?

We're happy to help test out the fix if you can publish an updated `-test` release of the NPM packages.

PR #1008, opened 2023-07-18
Remove testing-library as a dependency from various examples

nvie:
@marcbouchenoire Not sure if there is a gotcha with this, but I ran in to some peer dependency resolution conflicts where `testing-library` required React < 18, whereas these examples use React 18. I wondered why testing-library was a dependency for these examples in the first place, and… I don't think it's used? Possible included because it was in the template these examples are spawned from originally?

Either way, removing the dependency resolved the conflict obviously.

Let me know if we need to keep this!


marcbouchenoire (approved):
I think that's just CRA boilerplate remnants, at least no gotchas to my knowledge.

PR #1009, opened 2023-07-18
Make examples play better with Turbo when linked locally

nvie:


examples/react-dashboard/package.json
```diff
@@ -12,6 +12,7 @@
     "recharts": "^2.1.13"
   },
   "scripts": {
+    "dev": "react-scripts start",
```
nvie, line 15: Basically just aliasing `"dev"` to whatever `"start"` is for these examples. This makes them work nicely with Turborepo when linked locally, so Turbo triggers the necessary (local) builds first.

e2e/next-sandbox/turbo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "test": {
+      "cache": false
+    }
+  }
+}
```
nvie, line 9: The point of the E2E test is that it's testing the backend. If the backend changes, but the front-end does not, Turbo will think nothing changed, and serve the cache. This makes no sense for the E2E tests like it does for normal unit tests.

marcbouchenoire (approved):

CTNicholas (approved):

PR #1013, opened 2023-07-19
Improved the error message on createClient package, when both publicApiKey & authEndpoint are not defined

dev-badace:
I would like to inform you that I have made a modification to the error message that appears during the "createClient" process when both the publicKeyApi and authEndpoint parameters are missing. The reason behind this modification is that I recently encountered a situation where a user in the Liveblock's Discord community encountered this error and was unable to identify the root cause.

In order to assist users in troubleshooting such issues more effectively, I have made the error message more informative. This enhancement will provide clearer guidance to users when they encounter this specific error scenario.

I believe that improving the error message in this manner will contribute to a smoother user experience and help to minimize confusion and frustration.

Thankx, have a nice day.

### DIFF
--- a/packages/liveblocks-core/src/__tests__/client.node.test.ts
+++ b/packages/liveblocks-core/src/__tests__/client.node.test.ts
@@ -43,12 +43,12 @@ describe("createClient", () => {
     [
       undefined,
       undefined,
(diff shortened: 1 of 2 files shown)


nvie:
Thanks for your contribution, @dev-badace – this will be published with the next release (via #1072)! 🙏

PR #1024, opened 2023-07-25
Fixes a regression in our Suspense hooks

nvie:
This fixes a bug where I accidentally introduced a regression to our Suspense hooks on the 1.2 release branch, unsuspending on the wrong condition 🙈

What's worse is that I could introduce this issue without _any_ test noticing.

So this PR:
1. First adds E2E tests for our Suspense hooks, basically copies of the existing e2e tests/pages, but adjusted to use Suspense (e2e tests fail)
2. Then fixes the stupid bug in 4f4c63ba2fcd88e0883c2b562f53bdfc48e7b492 (e2e pass again)

### DIFF
--- a/e2e/next-sandbox/package.json
+++ b/e2e/next-sandbox/package.json
@@ -4,7 +4,7 @@
   "scripts": {
     "dev": "next dev --port 3007",
     "start": "next start",
-    "test": "playwright test"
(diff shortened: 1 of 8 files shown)


GuillaumeSalles (approved):
Nice catch!

PR #1025, opened 2023-07-25
Make `TOKEN_EXPIRED` code 4109, change client logic based on error ranges

nvie:
Part of [Release 1.2](https://github.com/liveblocks/liveblocks/pull/975).
Client-side counterpart to https://github.com/liveblocks/liveblocks-cloudflare/pull/469.

---

This changes a couple things:
- `TOKEN_EXPIRED` is now 4109
- Close code range `40xx` will now _disconnect_ the client (and emit an `"error"` event, i.e. what can be observed with `room.subscribe("error")`)
- Close code range `41xx` will now _reauthorize_ after a (normal) backoff delay. One exception is `TOKEN_EXPIRED` (4019), which will be the only error code that will reauthorize _without_ a backoff delay.
- Close code range `42xx` will now _retry without reauthorizing_ after an (aggressive) backoff delay.

This updates all the unit tests to reflect this new behavior.

### DIFF
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -529,17 +529,35 @@ const unsubscribe = room.subscribe("connection", (status) => {
 
 ### Room.subscribe("error") [#Room.subscribe.error]
 
-Subscribe to potential room connection errors.
+Subscribe to unrecoverable room connection errors. This event will be emitted
+right before the client disconnects and won’t try reconnecting again.
 
 Returns an unsubscribe function.
 
 ```ts
 const unsubscribe = room.subscribe("error", (error) => {
-  if (error.code === 4005) {
-    // Maximum concurrent connections per room exceeded.
+  switch (error.code) {
+    case -1:
+      // Authentication error
+      break;
+
+    case 4001:
+      // Could not connect because you don't have access to this room
+      break;
+
+    case 4005:
+      // Could not connect because room was full
+      break;
+
+    default:
+      // Unexpected error
+      break;
   }
 });
 ```
+
+You can use this event to trigger a “Not allowed” screen/dialog. If you want to
+offer retrying connecting, call `room.reconnect()`.
 
 ### Room.subscribe("history") [#Room.subscribe.history]
 
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -556,8 +556,8 @@ const unsubscribe = room.subscribe("error", (error) => {
 });
 ```
 
-You can use this event to trigger a “Not allowed” screen/dialog. If you want to
-offer retrying connecting, call `room.reconnect()`.
+You can use this event to trigger a “Not allowed” screen/dialog. If you’d like to
+retry connecting, call `room.reconnect()`.
 
 ### Room.subscribe("history") [#Room.subscribe.history]
 
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -139,10 +139,15 @@ describe("room / auth", () => {
         }),
       }
     );
-
     room.connect();
+
+    let err = {} as any;
+    room.events.error.subscribeOnce((e) => (err = e));
+
     await waitUntilStatus(room, "disconnected");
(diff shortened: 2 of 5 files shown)


packages/liveblocks-core/src/connection.ts
```diff
@@ -492,7 +490,10 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
         if (failedEvent.reason instanceof StopRetrying) {
           return {
             target: "@idle.failed",
-            effect: log(LogLevel.ERROR, failedEvent.reason.message),
+            effect: [
+              log(LogLevel.ERROR, failedEvent.reason.message),
+              fireErrorEvent(failedEvent.reason.message, -1),
```
nvie, line 495: @GuillaumeSalles Do you have a suggestion for the `-1` error code to use here? These aren't websocket codes, but the `code` param is mandatory on the `"error"` event. I used -1 here as a placeholder for now, but making it optional would technically be a breaking change.

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -529,18 +529,36 @@ const unsubscribe = room.subscribe("connection", (status) => {
...
+Subscribe to unrecoverable room connection errors. This event will be emitted
+right before the client disconnects and won’t try reconnecting again.
 
 Returns an unsubscribe function.
 
 ```ts
 const unsubscribe = room.subscribe("error", (error) => {
-  if (error.code === 4005) {
-    // Maximum concurrent connections per room exceeded.
+  switch (error.code) {
```
GuillaumeSalles, line 539: I'm considering typing error as it's always manually thrown by us to use an enum to avoid the switch with random numbers, what do you think? No need to do that right now

packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -417,12 +466,16 @@ describe("room", () => {
...
+    let err = {} as any;
+    room.events.error.subscribeOnce((e) => (err = e));
+
     await waitUntilStatus(room, "connected");
 
     // Closing this connection will trigger a reconnection...
     wss.last.close(
       new CloseEvent("close", {
         code: WebsocketCloseCodes.CLOSE_WITHOUT_RETRY,
+        reason: "wha'er",
```
GuillaumeSalles, line 478: Ahaha I like how the reason is becoming shorter and shorter :D

GuillaumeSalles (approved):
Nice! I'm feeling way more confident about our new system with this coverage! Thank you for making sure everything is robust before your well deserved vacations 🌴

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -529,18 +529,36 @@ const unsubscribe = room.subscribe("connection", (status) => {
...
+
+    default:
+      // Unexpected error
+      break;
   }
 });
 ```
 
+You can use this event to trigger a “Not allowed” screen/dialog. If you want to
+offer retrying connecting, call `room.reconnect()`.
```
CTNicholas, line 560:
```suggestion
You can use this event to trigger a “Not allowed” screen/dialog. If you’d like to
retry connecting, call `room.reconnect()`.
```

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -529,18 +529,36 @@ const unsubscribe = room.subscribe("connection", (status) => {
 
 ### Room.subscribe("error") [#Room.subscribe.error]
 
-Subscribe to potential room connection errors.
+Subscribe to unrecoverable room connection errors. This event will be emitted
```
CTNicholas, line 532:
```
unrecoverable
```
Thanks for adding this 👍

CTNicholas (approved):
I think just a little change to the wording!

docs/pages/api-reference/liveblocks-client.mdx
nvie: Maybe. Would have to think about it. I wouldn't want to advertise the `WebsocketCloseCodes` enum publicly. Maybe a different one? Not doing it now anyway, let's punt on this for later.

packages/liveblocks-core/src/connection.ts
GuillaumeSalles: Related to [this](https://github.com/liveblocks/liveblocks/pull/1025#discussion_r1273705977), I think that if we type our error and add a new enum (different from `WebsocketCloseCodes`)  containing the codes, `-1` would be less visible.

PR #1036, opened 2023-08-01
Docs improvements for 1.2 @liveblocks/node page

CTNicholas:
There's some other small changes too, such as a short description at the top of every API page.

### DIFF
--- a/docs/pages/platform/upgrading/1.2.mdx
+++ b/docs/pages/platform/upgrading/1.2.mdx
@@ -96,9 +96,9 @@ with `authorize`, is still supported but will eventually be deprecated.
 ## Access tokens [#access-tokens]
 
 Access tokens are the new recommended way to authenticate, because they’re easy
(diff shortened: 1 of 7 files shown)


docs/pages/platform/upgrading/1.2.mdx
```diff
@@ -96,9 +96,9 @@ with `authorize`, is still supported but will eventually be deprecated.
 ## Access tokens [#access-tokens]
 
 Access tokens are the new recommended way to authenticate, because they’re easy
-to manage from your custom back end. They follow the analogy of a “hotel room
-key card”. Anyone that has a key card can enter any room that the card gives
-access to. It’s easy to give out these key cards right from your back end.
+to manage from your custom back end. They follow the analogy of a _hotel key
```
stevenfabre, line 99: If we change this copy here, we likely want to change it in the authentication tooltip as well

stevenfabre (approved):
Much better. Thanks for improving those pages!

PR #1040, opened 2023-08-03
Update lexical nextjs example guide

CTNicholas:
- [x] Now uses /app , our new auth, and now has avatars
- [x] Updated the guide to reflect this
- [x] Fixed a small bug preventing deploys with Slate and Tiptap


stevenfabre (approved):
Awesome! Thanks for moving that example over to the `/app` folder.

PR #1045, opened 2023-08-03
Mention `/pages` endpoints in important places, improve permissions guide

CTNicholas:
It's okay to use `/app` endpoints up front, and in blog posts, but on reflection we should also have collapsible `/pages` endpoint examples in important places.

Also improvements to the permissions guide, as it didn't use our new auth in a couple of places.


stevenfabre (approved):
I like that pattern of showing those in a collapsed way.

Nice improvement. Thank you Chris.

PR #1048, opened 2023-08-04
Fix session allow pattern length

GuillaumeSalles:


CTNicholas (approved):
Ah nice, thanks for fixing!

PR #1052, opened 2023-08-07
Improve block text performance by orders of magnitude 

CTNicholas:
Enhancements to improve performance of the block text editor when using large documents


marcbouchenoire (approved):
LGTM ✅

PR #1057, opened 2023-08-09
Support native `Headers` in `verifyRequest`

CTNicholas:
It'd be nice if the native `Headers` object was supported in [`WebhookHandler.verifyRequest`](https://liveblocks.io/docs/api-reference/liveblocks-node#verifyRequest), then it'd work with [Next.js](https://nextjs.org/docs/app/building-your-application/routing/route-handlers#headers) `/app` and other modern frameworks.

This would enable both of these:
```ts
export function POST(request: Request) {
  const event = webhookHandler.verifyRequest({
    headers: request.headers,
    rawBody: "",
  });

  // ...
}
```

```ts
import { headers } from "next/headers";

export function ServerComponent() {
  const event = webhookHandler.verifyRequest({
    headers: headers(),
    rawBody: "",
  });

  // ...
}
``

### DIFF
--- a/packages/liveblocks-node/src/webhooks.ts
+++ b/packages/liveblocks-node/src/webhooks.ts
@@ -62,15 +62,13 @@ export class WebhookHandler {
   private verifyHeaders(headers: IncomingHttpHeaders | Headers) {
     const sanitizedHeaders: IncomingHttpHeaders = {};
 
-    if (headers.constructor === Headers) {
+    if (headers instanceof Headers) {
       for (const [key, value] of headers) {
         sanitizedHeaders[key.toLowerCase()] = value;
       }
     } else {
       Object.keys(headers).forEach((key) => {
-        sanitizedHeaders[key.toLowerCase()] = (headers as IncomingHttpHeaders)[
-          key
-        ];
+        sanitizedHeaders[key.toLowerCase()] = headers[key];
       });
     }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/liveblocks-node/src/webhooks.ts [hunk 1]
```diff
@@ -58,11 +59,18 @@ export class WebhookHandler {
...
    * Verifies the headers and returns the webhookId, timestamp and rawSignatures
    */
-  private verifyHeaders(headers: IncomingHttpHeaders) {
+  private verifyHeaders(headers: IncomingHttpHeaders | Headers) {
     const sanitizedHeaders: IncomingHttpHeaders = {};
-    Object.keys(headers).forEach((key) => {
-      sanitizedHeaders[key.toLowerCase()] = headers[key];
-    });
+
+    if (headers instanceof Headers) {
```
GuillaumeSalles: Not sure `instanceof Headers` works in a node environment. I don't think `Headers` is globally available in node. We should probably add an additional check.

CTNicholas: Good shout, `Headers` have only been supported in Node.js since `18.0.0`. Added a check.

FlowFlorent: Nice, also feel free to add an example with `Headers` in the `WebhookRequest` type if you think it helps!

FlowFlorent (approved):
one suggestion, otherwise looks good! 🚀

packages/liveblocks-node/src/webhooks.ts
FlowFlorent: @GuillaumeSalles they do but on the off chance, I'd rather be safe

CTNicholas: Done @FlowFlorent!

PR #1061, opened 2023-08-11
Fix Suspense option in create-liveblocks-app

marcbouchenoire:
This PR fixes the Suspense option when specifying a framework and adds comments by default.


CTNicholas (approved):
Thanks Marc!

PR #1070, opened 2023-08-17
Fix: Comments examples

marcbouchenoire:


CTNicholas (approved):
Adding `ClientSideSuspense` instead of `Suspense` fixed the strange error, and made builds work again.

I'm not sure if we were using `Suspense` for a particular reason, but I figured we want this working today, so I swapped it out.

PR #1072, opened 2023-08-17
Improve error messages in client

nvie:
Improve error messages by making them more helpful and user-friendly. Based on @dev-badace's contribution from #1013, but extended to improve more error messages, and with merge conflicts resolved that were introduced when we released 1.2.

@dev-badace: Thanks for your contribution! I've opened this PR instead because I could not push to your branch directly.


CTNicholas (approved):
Good improvement, thanks for starting this @dev-badace!

PR #1082, opened 2023-08-23
Fix code in "Modify storage from server" guide

CTNicholas:
The old code had a problem with storage synchronising.

### DIFF
--- a/guides/pages/how-to-modify-liveblocks-storage-from-the-server.mdx
+++ b/guides/pages/how-to-modify-liveblocks-storage-from-the-server.mdx
@@ -134,11 +134,7 @@ export async function modifyStorage(
 
     // Make storage adjustments in a batch, so they all happen at once
     room.batch(() => {
-      try {
-        storageChanges(root);
-      } catch (error) {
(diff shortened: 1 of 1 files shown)


guides/pages/how-to-modify-liveblocks-storage-from-the-server.mdx
```diff
@@ -134,16 +134,23 @@ export async function modifyStorage(
...
-        if (status === "synchronized") {
-          serverClient.leave(roomId);
-          resolve();
-        }
-      });
+      try {
+        storageChanges(root);
+      } catch (error) {
+        console.log("Problem modifying Liveblocks Storage from Node.js", error);
+      }
```
nvie, line 141: Not sure if that console log should be in here. Ideally people would put that in the `storageChanges()` helper if they're interested in catching things there? Feel free to leave it off.

nvie (approved):
Nice work! I had no idea this pattern was based on an example/guide of our own — that's great! Thanks for adding this fix in!

PR #1086, opened 2023-08-24
Various example changes

marcbouchenoire:
This PR:
- Harmonizes all examples' `package.json`
	- The package name isn't important but it can still cause a conflict with Turborepo when two packages have the same one (e.g. `@liveblocks/react-comments` was both the published package and the vanilla example)
	- I did this change in https://github.com/liveblocks/liveblocks/tree/comments but it's now causing all examples to deploy every time we commit because they technically have one file changed compared to `main`
- Ports the vanilla React comments example from https://github.com/liveblocks/liveblocks/tree/comments


CTNicholas (approved):
Looks good!

PR #1088, opened 2023-08-24
Fix bug where entering a new room could potentially initialize the undo stack incorrectly

nvie:


CTNicholas (approved):
Thanks for making the PR!

packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -916,6 +914,33 @@ describe("room", () => {
...
+          items: [["root", { type: CrdtType.OBJECT, data: {} }]],
+        })
+      );
+    });
+
+    room.connect();
+    const storage = await room.getStorage();
+    expect(storage.root.toObject()).toEqual({ foo: 1234 });
+    //                                        ^^^ Added by the client, from initialStorage
+    expect(room.history.canUndo()).toBe(false);
```
nvie, line 941: Here is the check that failed, which demonstrates the bug.

nvie:
The last commit makes the new test pass again ✅ – and thus fixes the bug. Unfortunately, the Vercel integration failed again, so CI shows ❌ incorrectly :(

PR #1090, opened 2023-08-24
Upgrade all examples to 1.2.3

nvie:
This upgrade fixes the "undo stack" initialization bug in our examples, see #1088.


CTNicholas (approved):
Thank you!

PR #1093, opened 2023-08-25
Comments: Hide @ button when not using mentions

marcbouchenoire:
This PR hides the @ button in `<Composer />` when not using mentions. It also shuffles a few internal things to make exposing private methods/properties from `@liveblocks/react` to `@liveblocks/react-comments` easier through `useRoomContextBundle`.


GuillaumeSalles (approved):
Nice! No API is the best API :)

PR #1099, opened 2023-08-30
add string to userId description

jrowny:
Small doc update to make it more explicit that userId must be a string when using identifyUser.


nvie (approved):

CTNicholas (approved):
Thanks!

PR #1116, opened 2023-09-05
Comments

marcbouchenoire:
This PR brings Comments to `main`.

Comments is currently in private beta, [learn more](https://liveblocks.io/docs/products/comments).


packages/liveblocks-node/src/client.ts
```diff
@@ -1,3 +1,4 @@
+import type { CommentData, ThreadData } from "@liveblocks/core";
```
FlowFlorent, line 1:
```suggestion
/**
 * NOTE: only types should be imported from @liveblocks/core.
 * This is because this package is made to be used in Node.js, and
 * @liveblocks/core has browser-specific code.
 */
import type { CommentData, ThreadData } from "@liveblocks/core";
```

packages/liveblocks-react/src/factory.tsx
```diff
@@ -109,14 +136,94 @@ function makeMutationContext<
...
+    hasWarnedIfNoResolveUser = true;
+  }
+}
+
+let hasWarnedAboutCommentsBeta = false;
+function warnIfBetaCommentsHook() {
+  if (!hasWarnedAboutCommentsBeta && process.env.NODE_ENV !== "production") {
+    console.log(
+      `%cWarning: The comments API is still in beta and may change in the future. Please report any issues you encounter at https://github.com/liveblocks/liveblocks/issues/new/choose`,
+      "color: #FFA500; font-weight: bold;"
```
nimeshnayaju, line 184: The following warning is displayed (once) when any of the following hooks are used:

- `useThreads`
- `useThreadsSuspense`
- `useCreateThread`
- `useEditThreadMetadata`
- `useCreateComment`
- `useEditComment`
- `useDeleteComment`

I have used `console.log` instead of `console.warn` because the behavior of console.warn()—including whether it shows a stack trace or not—can vary between different environments (e.g., browsers, Node.js, etc.) and is not directly configurable using standard JavaScript or APIs. This means you can't disable the stack trace in the warning if the environment decides to include one. Any thoughts on this @marcbouchenoire?

```
Warning: The Comments API is still in beta and may change in the future. Please report any issues you encounter at https://github.com/liveblocks/liveblocks/issues/new/choose 
```

Using `console.log` like described in the method (doesn't display stack trace):
Screenshot 2023-09-06 at 10 43 52 PM

Using `console.warn` (displays stack trace):

packages/liveblocks-react/src/factory.tsx
```diff
@@ -109,14 +136,93 @@ function makeMutationContext<
...
+    hasWarnedIfNoResolveUser = true;
+  }
+}
+
+let hasWarnedAboutCommentsBeta = false;
+function warnIfBetaCommentsHook() {
+  if (!hasWarnedAboutCommentsBeta && process.env.NODE_ENV !== "production") {
+    console.warn(
+      "The comments API is still in beta and may change in the future. Please report any issues you encounter at https://github.com/liveblocks/liveblocks/issues/new/choose"
+    );
```
nimeshnayaju, line 184: The following warning is displayed (once) when any of the following hooks are used:

- `useThreads`
- `useThreadsSuspense`
- `useCreateThread`
- `useEditThreadMetadata`
- `useCreateComment`
- `useEditComment`
- `useDeleteComment`

I am using `console.warn` but it seems the behavior of console.warn()—including whether it shows a stack trace or not—can vary between different environments (e.g., browsers, Node.js, etc.) and is not directly configurable using standard JavaScript or APIs. Any thoughts on this @nvie @marcbouchenoire ?

```
Warning: The Comments API is still in beta and may change in the future. Please report any issues you encounter at https://github.com/liveblocks/liveblocks/issues/new/choose 
```

nimeshnayaju: I am open to replacing the console message to say something else (Add link to docs, etc). Open to feedback @marcbouchenoire @nvie @FlowFlorent

marcbouchenoire: I tweaked the message to give context to non-beta users who would try to use Comments instead.

nimeshnayaju (approved):

PR #1120, opened 2023-09-07
Drop long-deprecated old array methods

nvie:
Long ago, we added `.count` and `.toArray()` helpers on the others array. These were deprecated with the release of 0.19. This now finally removes them.


marcbouchenoire (approved):
🧹

PR #1123, opened 2023-09-07
Fix Yjs docs params not rendering + document new `format` query param for Get Storage

jrowny:
The parameters were one level too high and subsequently don't show up in the docs. This PR fixes that and also fixes the response of the PUT call which just returns status.


CTNicholas (approved):
Thanks for the fix!

nvie:
I've fixed the issues here, added documentation for the new `?format=json` flag (for `GET /rooms/:roomId/storage`) and reordered some fields (basically moving `"parameters"` and `"description"` fields upward, so they're easier to edit). I think this is now okay to merge.

CTNicholas:
Thanks for the extra fixes and new docs @nvie, looks good!

PR #1125, opened 2023-09-07
Clear history API

### LINKED ISSUE #1117: Clear Room History
**Is your feature request related to a problem? Please describe.**

Can we be allowed to clear room history (undo/redo stack)? Similarly to a computer's file directory, a user in our product can navigate to multiple UI levels of the same project. Currently the undo stack will build up multiple actions from one level, and once the user changes levels the undo stack will affect offscreen elements. The ability to reset history would allow us to ensure the undo stack is relevant only to the view the user is currently on. 

**Describe the solution you'd like**

Please expose a function to clear room history adjacent to `history.undo()` and `history.redo()`. Perhaps `history.clear()`

**Additional context**

We are using the REACT library.

nvie:
This adds a new `room.history.clear()` API, which allows users to explicitly reset the undo/redo stack, which is useful for some applications.

Fixes #1117.
Fixes https://github.com/liveblocks/liveblocks.io/issues/1473.


packages/liveblocks-core/src/room.ts
```diff
@@ -2082,6 +2087,11 @@ export function createRoom<
     flushNowOrSoon();
   }
 
+  function clear() {
+    context.undoStack.length = 0;
```
jrowny, line 2091: does it matter that this is mutable, vs. setting to `[]`?

jrowny (approved):

packages/liveblocks-core/src/room.ts
nvie: Both should work. In this case, I used this mutable version because of the inherent mutable nature of the stack — so it's similarly mutating a single array instance, just like `.pop()` and `.push()` and our other `.length` mutation is doing. To make this extra clear, I've made these `readonly` properties on `context` now in f81715c0ebc8be23d0ca81a1b79f030baaad4047. This ensures we can never accidentally store a reference to an "old stack" somewhere.

marcbouchenoire (approved):
Nice 🤌

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -674,6 +674,18 @@ room.history.redo();
 // room.history.canRedo() is false
 ```
 
+### Room.history.clear
+
+Clears the undo and redo stacks. Explicitly clearing history resets the ability
+to undo beyond the current document state.
```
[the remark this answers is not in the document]
nvie: Oh nice, I like this addition!

PR #1127, opened 2023-09-08
Fix @liveblocks/react bundle

marcbouchenoire:
This PR adds `nanoid` as a dependency as it should have been.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# v1.3.1
+
+### `@liveblocks/react`
+
+- Fix bug that would cause Node.js-specific imports to be present in the final
+  bundle
+
 # v1.3.0
 
 v1.3.0 marks the initial release of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ### `@liveblocks/react`
 
-- Fix bug that would cause Node.js-specific imports to be present in the final
-  bundle
+- Fix missing dependency declaration
 
 # v1.3.0
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -19187,6 +19187,7 @@
       "dependencies": {
         "@liveblocks/client": "1.3.0",
         "@liveblocks/core": "1.3.0",
+        "nanoid": "^4.0.2",
         "use-sync-external-store": "^1.2.0"
       },
       "devDependencies": {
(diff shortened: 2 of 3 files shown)


package-lock.json
```diff
@@ -19187,6 +19187,7 @@
       "dependencies": {
         "@liveblocks/client": "1.3.0",
         "@liveblocks/core": "1.3.0",
+        "nanoid": "^4.0.2",
```
nvie, line 19190: Thinking about this… I think maybe we need to make this a `peerDependencies` instead, just to facilitate people that already may be using (a different version of) `nanoid` in their frontend projects?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# v1.3.1
+
+### `@liveblocks/react`
+
+- Fix bug that would cause Node.js-specific imports to be present in the final
+  bundle
```
nvie, line 6: Maybe just say

```suggestion
- Fix missing dependency declaration
```

nvie (approved):
Approved… but with one consideration.

package-lock.json
marcbouchenoire: Do `peerDependencies` auto-install nowadays?

nvie: They do these days!

Screen Shot 2023-09-08 at 11 16 10@2x

nvie: (And that makes a lot of sense.)

packages/liveblocks-react/package.json
```diff
@@ -35,10 +35,10 @@
   "dependencies": {
     "@liveblocks/client": "1.3.0",
     "@liveblocks/core": "1.3.0",
-    "nanoid": "^4.0.2",
     "use-sync-external-store": "^1.2.0"
   },
   "peerDependencies": {
+    "nanoid": "^3 || ^4",
```
nvie, line 41: Oh, that's clever. That lets people decide which one to use! I like it 👍

marcbouchenoire: The issue I have with that is that if `^4` is picked (e.g. automatically), our CJS bundle might be incorrect depending on the final bundler or lack of it. I think the ideal setup is to add it as a normal dependency as `"nanoid": "^3.3.6"`.

nvie (approved):
Yesss, this looks great now! 👍

PR #1130, opened 2023-09-11
Document connectionId in events

CTNicholas:

### DIFF
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -376,9 +376,13 @@ const unsubscribe = room.subscribe(
 
 ### Room.subscribe("event") [#Room.subscribe.event]
 
-Subscribe to events broadcasted by [`Room.broadcastEvent`][].
+Subscribe to events broadcast by [`Room.broadcastEvent`][].
 
 Takes a callback that is called when a user calls [`Room.broadcastEvent`][].
+Provides the `event` along with the `connectionId` of the user that sent the
+message. If an event was sent from the
+[Broadcast to a room](/docs/api-reference/rest-api-endpoints#post-broadcast-event)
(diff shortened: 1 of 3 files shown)


docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -376,9 +376,13 @@ const unsubscribe = room.subscribe(
...
 ### Room.subscribe("event") [#Room.subscribe.event]
 
-Subscribe to events broadcasted by [`Room.broadcastEvent`][].
+Subscribe to events broadcast by [`Room.broadcastEvent`][].
 
 Takes a callback that is called when a user calls [`Room.broadcastEvent`][].
+Provides the `event` along with the `connectionId` of the user that sent the
+message. If an event was sent from the
+[Broadcast to a room](/docs/api-reference/rest-api-endpoints#post-broadcast-event)
+REST API, `connectionId` will be `-1`.
```
nvie, line 385: I was wrong about this earlier! It's not `-1`, but it's going to be `undefined` when coming from the server. I'll fix the TypeScript definition for this in our client 👍

nvie (approved):
Thanks for documenting this — this will surely help! 👍

docs/pages/api-reference/liveblocks-client.mdx
nvie: I'm tempted to fix this in the client, by returning `-1` instead of `undefined` in this case. It may be a little more explicit that way.

CTNicholas: We've just agreed that `-1` is better, and the change will be made.

nvie: The server-side bug fix PR: https://github.com/liveblocks/liveblocks-cloudflare/pull/520

nvie (approved):

PR #1131, opened 2023-09-12
Implement Auto-Linking Feature in Rich Text Editor for Comments

nimeshnayaju:
### Summary 
This Pull Request introduces auto-linking feature to the rich text editor used in our Comments component, implemented using [Slate.js](https://www.slatejs.org/examples/richtext). 

### Details
The PR modifies the editor by adding a custom Slate.js plugin (called `withAutoLinks`). This plugin modifies the `normalizeNode` and `deleteBackward` function of the editor in order to ensure any url entered in the editor is recognized correctly and automatically create an auto-link node around the url. The key functionalities added by the plugin are:

- **Auto-Link Creation:** When a user enters or pastes a URL that matches the prescribed regex, it will automatically wrap the URL in a "link" node, providing immediate hyperlinking without the user needing to perform additional actions.
- **Link Correction:** If any modifications invalidate an existing "link" node, the plugin will automatically unwrap the node, reverting it back to a standard text node. This ensures that the links remain valid even after surrounding text changes.

> Note: There isn't any UI change in the rich text editor despite the addition of a new plugin. The plugin operates in the background, automatically creating auto-link nodes for applicable texts. However, there is a noticeable UI change in the comment component that displays the published comments. This component now renders clickable hyperlinks whenever a URL is detected.

### DIFF
--- a/packages/liveblocks-react-comments/src/primitives/Comment/index.tsx
+++ b/packages/liveblocks-react-comments/src/primitives/Comment/index.tsx
@@ -9,6 +9,8 @@ import type {
   CommentRenderMentionProps,
 } from "./types";
 import { isCommentBodyMention } from "./utils";
+import { isComposerBodyAutoLink } from "../../slate/plugins/auto-links";
+import type { ComponentPropsWithSlot } from "../../types";
 
 const COMMENT_MENTION_NAME = "CommentMention";
 const COMMENT_BODY_NAME = "CommentBody";
@@ -35,6 +37,18 @@ function CommentDefaultRenderMention({ userId }: CommentRenderMentionProps) {
 const CommentMention = forwardRef<HTMLSpanElement, CommentMentionProps>(
   ({ children, asChild, ...props }, forwardedRef) => {
     const Component = asChild ? Slot : "span";
+
+    return (
+      <Component {...props} ref={forwardedRef}>
+        {children}
+      </Component>
+    );
+  }
+);
+
+const AutoLink = forwardRef<HTMLAnchorElement, ComponentPropsWithSlot<"a">>(
+  ({ children, asChild, ...props }, forwardedRef) => {
+    const Component = asChild ? Slot : "a";
 
     return (
       <Component {...props} ref={forwardedRef}>
@@ -80,6 +94,19 @@ const CommentBody = forwardRef<HTMLDivElement, CommentBodyProps>(
                       ) : null;
                     }
 
+                    if (isComposerBodyAutoLink(inline)) {
+                      return (
+                        <AutoLink
+                          href={toAbsoluteURL(inline.href)}
+                          target="_blank"
+                          rel="noopener noreferrer nofollow"
+                          key={index}
+                        >
+                          {inline.href}
+                        </AutoLink>
+                      );
+                    }
+
                     // <code><s><em><strong>text</strong></s></em></code>
                     let children: ReactNode = inline.text;
 
@@ -115,6 +142,17 @@ const CommentBody = forwardRef<HTMLDivElement, CommentBodyProps>(
 if (process.env.NODE_ENV !== "production") {
   CommentBody.displayName = COMMENT_BODY_NAME;
   CommentMention.displayName = COMMENT_MENTION_NAME;
+}
+
+function toAbsoluteURL(url: string): string | undefined {
+  // Check if the URL already contains a scheme
+  if (url.startsWith("http://") || url.startsWith("https://")) {
+    return url;
+  } else if (url.startsWith("www.")) {
+    // If the URL starts with "www.", prepend "https://"
+    return "https://" + url;
+  }
+  return;
 }
 
 // NOTE: Every export from this file will be available publicly as Comment.*
--- a/packages/liveblocks-react-comments/src/slate/plugins/auto-links.ts
+++ b/packages/liveblocks-react-comments/src/slate/plugins/auto-links.ts
@@ -1,0 +1,316 @@
+import {
+  Editor,
+  Node,
+  NodeEntry,
+  Text,
+  Transforms,
+  Element,
+  Path,
+  Range,
+} from "slate";
+import type { ComposerBodyAutoLink } from "../../types";
+
+/**
+ * This implementation is inspired by Lexical's AutoLink plugin.
+ * Additional modifications and features were added to adapt it to our specific needs.
+ *
+ * Original Lexical AutoLink plugin can be found at [Lexical's Github Repository](https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/LexicalAutoLinkPlugin.ts)
+ */
+export function withAutoLinks(editor: Editor): Editor {
+  const { isInline, normalizeNode, deleteBackward } = editor;
+
+  editor.isInline = (element) => {
+    return element.type === "auto-link" ? true : isInline(element);
+  };
+
+  editor.normalizeNode = (entry) => {
+    const [node, path] = entry;
+
+    if (Text.isText(node)) {
+      const parentNode = Node.parent(editor, path);
+
+      if (isComposerBodyAutoLink(parentNode)) {
+        const parentPath = Path.parent(path);
+        handleLinkEdit(editor, [parentNode, parentPath]);
+      } else {
+        handleLinkCreate(editor, [node, path]);
+        handleNeighbours(editor, [node, path]);
+      }
+    }
+
+    normalizeNode(entry);
+  };
+
+  editor.deleteBackward = (unit) => {
+    deleteBackward(unit);
+    const { selection } = editor;
+    if (!selection) return;
+
+    if (!Range.isCollapsed(selection)) return;
+
+    const [match] = Editor.nodes(editor, {
+      at: selection,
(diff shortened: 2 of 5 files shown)


packages/liveblocks-react-comments/src/primitives/Comment/index.tsx
```diff
@@ -80,6 +94,19 @@ const CommentBody = forwardRef<HTMLDivElement, CommentBodyProps>(
                       ) : null;
                     }
 
+                    if (isComposerBodyAutoLink(inline)) {
+                      return (
+                        <AutoLink
+                          href={toAbsoluteURL(inline.href)}
+                          target="_blank"
+                          rel="noopener noreferrer nofollow"
+                          key={index}
```
nimeshnayaju, line 103: Do we want to provide styling option via `className` or the render prop similar to the mention component? @marcbouchenoire

.github/CODEOWNERS
```diff
@@ -2,7 +2,8 @@
 /packages/   @nvie
 /packages/liveblocks-devtools/   @nvie @marcbouchenoire
 /packages/create-liveblocks-app/   @CTNicholas
-/packages/liveblocks-node/   @nvie @ofoucherot @FlowFlorent 
+/packages/liveblocks-node/   @nvie @ofoucherot @FlowFlorent
+/packages/liveblocks-react-comments/   @marcbouchenoire
```
nvie, line 6: I was auto-assigned to review this PR because I'm a code owner for the entire `/packages` directory, but I've opted-out for the `/packages/liveblocks-react-comments` one for now, and made @marcbouchenoire the code owner, so he'll get invited automatically whenever someone touches a file in there going forward (instead of me).

packages/liveblocks-react-comments/src/primitives/Comment/index.tsx
```diff
@@ -117,5 +144,16 @@ if (process.env.NODE_ENV !== "production") {
...
+  // Check if the URL already contains a scheme
+  if (url.startsWith("http://") || url.startsWith("https://")) {
+    return url;
+  } else if (url.startsWith("www.")) {
+    // If the URL starts with "www.", prepend "https://"
+    return "https://" + url;
+  }
+  return;
+}
+
```
marcbouchenoire, line 157: We can move this to `/src/primitives/Comment/utils.ts`.

packages/liveblocks-react-comments/src/primitives/Comment/index.tsx
marcbouchenoire: Via `renderLink` (which would be the same render prop for both auto and later custom links), given that this is in a primitive where we should always give access to the markup if possible (contrary to the default components).

packages/liveblocks-react-comments/src/slate/plugins/auto-links.ts
```diff
@@ -0,0 +1,316 @@
...
+};
+
+const handleNeighbours = (editor: Editor, entry: NodeEntry<Text>) => {
+  const [node, path] = entry;
+  const text = node.text;
+
+  const previousSibling = Editor.previous(editor, { at: path });
+
+  if (previousSibling && isComposerBodyAutoLink(previousSibling[0])) {
+    if (/^\.[a-zA-Z0-9]+/.test(text)) {
```
marcbouchenoire, line 295: We should probably define this regex once with the other ones (`URL_REGEX` and  `PUNCTUATION_OR_SPACE`) above.

marcbouchenoire (changes requested):
This works as expected! Thanks for going all the way, it feels like a correct implementation.

Regarding no UI in the editor: why shouldn't we indicate them visually if they're correctly created on the fly?

I only tested the playground (thanks for making it, it's so helpful for things like this), but I'm not seeing any serialization logic: how do links make their way into `CommentBody`? We'll need to handle them within `composerBodyToCommentBody`/`commentBodyToComposerBody`/etc and type them in `@liveblocks/core`. For the shape they would take in `CommentBody`, I think we could have a single "link" type with an `auto` property? Given that `CommentBody` is public and we have full control over it, having two different "link"/"auto-link" types there feels internal-ish.

packages/liveblocks-react-comments/src/primitives/Comment/index.tsx
nimeshnayaju: Done! 

I am not completely certain how I feel about this method. If our customers are using the high level `Comment` component, then it works all right, but if customers intend to customize the `Link` component via the `renderLink` prop, we are expecting them to build their own implementation of this method if they want to ensure the links are absolute, which should be most of the time. 
An alternative could be providing both absolute url and text in the comment body but this will interfere with our upcoming potential design for custom links which will have the same two properties. What do you think @marcbouchenoire?

packages/liveblocks-react-comments/src/slate/plugins/auto-links.ts
nimeshnayaju: Good idea! Done!

packages/liveblocks-react-comments/src/components/Composer.tsx
```diff
@@ -203,6 +205,21 @@ function ComposerMentionSuggestions({
...
+
+  return (
+    <ComposerPrimitive.Link
+      href={href}
+      target="_blank"
+      rel="noopener noreferrer nofollow"
+      className="lb-composer-link"
+    >
+      {children}
+    </ComposerPrimitive.Link>
```
nimeshnayaju, line 219: The `children` here is apparently quite important. If I instead render `{url}` here, Slate.js throws this error when a link node is rendered. 

```Cannot resolve a Slate point from DOM point```

This might make the `renderLink` feature a little confusing for developers because if they do not render the `children` prop, they editor errors out. Might have to spend a little more time on this.

Comment (marcbouchenoire) [reply packages/liveblocks-react-comments/src/components/Composer.tsx:219]: Yeah `children` contains invisible DOM nodes required by Slate, but it can be solved (similarly to `Composer.Mention`) by having an intermediate "wrapper" component that passes the invisible children so that the user-provided component doesn't have to. https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react-comments/src/primitives/Composer/index.tsx#L160

File: packages/liveblocks-react-comments/src/primitives/Comment/index.tsx (hunk 3 above)
Comment (marcbouchenoire) [reply packages/liveblocks-react-comments/src/primitives/Comment/index.tsx]: I tweaked the way both `renderLink` props work so that instead of `url` we provide `href` and `text`: we convert the URL to absolute and pass it as `href` so that users don't have to handle it, and for auto-links we provide the original URL as `text` (which means that the users' implementations won't need to change when custom links arrive later on: `<a href={href}>{text}</a>` will remain valid).

File: packages/liveblocks-react-comments/src/components/Composer.tsx (hunk 5 above)
Comment (marcbouchenoire) [reply packages/liveblocks-react-comments/src/components/Composer.tsx:216]: I went ahead and replicated what we do with `Composer.Mention`: https://github.com/liveblocks/liveblocks/pull/1131/commits/6edb9f43b26967410128645a69cb736b169ad09e#diff-a4264574a0ff5a9b06f3691aa0544b8e8825b8fd990403f6988f7d3563c67551R205-R207

Comment (marcbouchenoire) [reply packages/liveblocks-react-comments/src/components/Composer.tsx:216]: The change above didn't work because links are editable, so I reverted to passing `children`. But for consistency, we now pass a `children` prop instead of `text` in both `renderLink` props. That also means `<a {...props} />` is now possible given the prop naming.

File: packages/liveblocks-react-comments/src/primitives/Comment/index.tsx (hunk 3 above)
Comment (marcbouchenoire) [reply packages/liveblocks-react-comments/src/primitives/Comment/index.tsx]: (`text` is now `children`: https://github.com/liveblocks/liveblocks/pull/1131#discussion_r1324317386)

**Review by marcbouchenoire — CHANGES_REQUESTED** [review]
I pushed a few tweaks and added a few comments/requests about some other tiny changes. It's almost done! 🤏

File: packages/liveblocks-react-comments/src/primitives/index.ts [hunk 6]
```diff
@@ -17,6 +19,7 @@ export type {
   ComposerSubmitProps,
   ComposerSuggestionsListItemProps,
   ComposerSuggestionsListProps,
+  ComposerLinkProps,
```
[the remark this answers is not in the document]
nimeshnayaju: Nice catch! Thanks

docs/pages/api-reference/liveblocks-react-comments.mdx
```diff
@@ -320,7 +320,12 @@ import { Button } from "@/my-design-system";
 ```tsx title="Anatomy"
 import { Comment } from "@liveblocks/react-comments/primitives";
 
-() => <Comment.Body renderMention={() => <Comment.Mention />} />;
+() => (
+  <Comment.Body
+    renderMention={() => <Comment.Mention />}
+    renderLink={(props) => <Comment.Link {...props} />}
```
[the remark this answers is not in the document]
nimeshnayaju: Makes sense! Thanks for clarifying!

marcbouchenoire (approved):

PR #1132, opened 2023-09-12
Throw when encountering a cached token from the backend

nvie:
Fixes an issue that could put the client into an unhealthy reconnecting loop if caching is implemented by customers in the `/api/liveblocks-auth` endpoint. The Liveblocks client already implements its own caching, so it will not hit the backend unnecessarily with requests. If it does, it expects a fresh new token (with an `iat` date that's equal to the current date).

Starting with this PR, we expicitly forbid caching in the `/api/liveblocks-auth` endpoint (we should document this). This was previously implied, but not made explicit enough. If the backend returns a cached token (issued before in the past), the client will now throw this error:

> _"The same Liveblocks auth token was issued from the backend before. Caching Liveblocks tokens is not supported."_

It's expected that developers/users will never see this error, unless they have explicitly implemented caching in their `/api/liveblocks-auth` endpoint. If so, the solution is to remove this caching layer from `/api/liveblocks-auth` and always making a fresh request to the Liveblocks servers for every token request.

[Context](https://liveblocks.slack.com/archives/C02PZL7QAAW/p1694512460266229?thread_ts=1694455495.336149&cid=C02PZL7QAAW)

### DIFF
--- a/packages/liveblocks-core/src/auth-manager.ts
+++ b/packages/liveblocks-core/src/auth-manager.ts
@@ -84,9 +84,6 @@ export function createAuthManager(
       if (token.parsed.k === TokenKind.ID_TOKEN) {
         // When ID token method is used, only one token per user should be used and cached at the same time.
         return token;
-      } else if (token.parsed.k === TokenKind.SECRET_LEGACY) {
-        // Legacy tokens are not cached.
-        return undefined;
       } else if (token.parsed.k === TokenKind.ACCESS_TOKEN) {
         for (const [resource, scopes] of Object.entries(token.parsed.perms)) {
           if (
@@ -180,8 +177,11 @@ export function createAuthManager(
         (token.parsed.exp - token.parsed.iat) -
         BUFFER;
 
-      tokens.push(token);
-      expiryTimes.push(expiresAt);
+      // Legacy tokens should not get cached
+      if (token.parsed.k !== TokenKind.SECRET_LEGACY) {
+        tokens.push(token);
+        expiryTimes.push(expiresAt);
+      }
 
       return { type: "secret", token };
     } finally {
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/auth-manager.ts
```diff
@@ -180,8 +177,11 @@ export function createAuthManager(
         (token.parsed.exp - token.parsed.iat) -
         BUFFER;
 
-      tokens.push(token);
-      expiryTimes.push(expiresAt);
+      // Legacy tokens should not get cached
+      if (token.parsed.k !== TokenKind.SECRET_LEGACY) {
+        tokens.push(token);
+        expiryTimes.push(expiresAt);
+      }
```
nvie, line 196: @ofoucherot @marcbouchenoire I made a subtle change here while I was at it — see commit 5b3ec426d3e980b8a8a5b4f2b4615c44072315ca. Could you double-check that change, please?

I realized that legacy tokens would still get _written_ to the cache, but then ignored when trying to _read_ from the cache later. This is fine behavior as long as you _only_ have legacy tokens in the cache. But if you somehow would have a mix of token types in there, if a legacy token is written to the cache, it couldn't be garantueed that ID or Access tokens could still get read back from the cache, due to the way the read loop was constructed. (I removed those troublesome lines L87-89.)

Can you confirm this was indeed a potential bug and it's better to not even store those secret legacy tokens in the cache in the first place, like this commit is doing?

ofoucherot: A client should never get a mix of legacy and new tokens. And yes, I confirm it's better not to store the legacy tokens.

jrowny (approved):
it looks good to me!

PR #1135, opened 2023-09-15
Fix potential `Error: "undefined" is not a valid event name`

nvie:
A user [reported a bug on Discord](https://discord.com/channels/913109211746009108/913158542565994606/1151980952046809180) where if you'd just do:

```tsx
function MyComponent() {
  const foo = useMap("foo");
}
```

On an empty room where `"foo"` is _not_ initialized, then this runtime error would get thrown:

```
Error: "undefined" is not a valid event name
```

This is caused by the hook trying to create a subscription on the live node it finds at `root.foo`, but in this case, no such root node exists, and as a result it tried to subscribe to `undefined`.

This is only a bug in our legacy `use{Map,Object,List}` hooks. The newer/preferred `useStorage` hook is **not affected**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### `@liveblocks/react`
 
 - Fix confusing `Error: "undefined" is not a valid event name` error when using
-  the (deprecated) `useMap()`, `useObject`, or `useList()` hooks on
+  the (deprecated) `useMap()`, `useObject()`, or `useList()` hooks on
   uninitialized storage values.
 
 # v1.3.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ### `@liveblocks/react`
 
 - Fix confusing `Error: "undefined" is not a valid event name` error when using
-  `useMap()`, `useObject`, or `useList()` on uninitialized storage values.
+  the (deprecated) `useMap()`, `useObject`, or `useList()` hooks on
+  uninitialized storage values.
 
 # v1.3.3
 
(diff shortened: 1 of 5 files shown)


packages/liveblocks-react/src/factory.tsx
```diff
@@ -617,50 +618,48 @@ export function createRoomContext<
...
           rerender();
         }
       }
 
-      let unsubscribeCrdt = room.subscribe(
-        liveValue as any /* AbstractCrdt */, // TODO: This is hiding a bug! If `liveValue` happens to be the string `"event"` this actually subscribes an event handler!
-        rerender
-      );
-      const unsubscribeRoot = room.subscribe(
-        root as any /* AbstractCrdt */, // TODO: This is hiding a bug! If `liveValue` happens to be the string `"event"` this actually subscribes an event handler!
```
nvie, line 648: This also fixes this (theoretic) bug we already knew about. In a way, it's a special case of the same bug. No longer!

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# v1.3.4
+
+### `@liveblocks/react`
+
+- Fix confusing `Error: "undefined" is not a valid event name` error when using
+  `useMap()`, `useObject`, or `useList()` on uninitialized storage values.
```
marcbouchenoire:
```suggestion
  `useMap()`, `useObject()`, or `useList()` on uninitialized storage values.
```

marcbouchenoire (approved):
Much better! I just left a tiny comment about `CHANGELOG.md`.

CTNicholas (approved):
Ah, nice one!

PR #1137, opened 2023-09-15
Officially mark `useList`, `useMap`, `useObject` as deprecated

nvie:
I noticed we never officially marked them as deprecated, although we've no longer been recommending their use since the release of 0.18:

Screen Shot 2023-09-15 at 16 21 29@2x

Users may not realize these helpers are deprecated and still import/use them (like [here](https://discord.com/channels/913109211746009108/1151980952046809180/1152212910425264228)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# v1.3.5
+
+### `@liveblocks/react`
+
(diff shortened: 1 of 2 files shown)


CTNicholas (approved):
Not sure how we missed this! Thanks for the change, looks good.

PR #1138, opened 2023-09-18
Make E2E tests more robust

nvie:
Using room IDs based on the first two digits of the Git SHA creating at most 256 room copies in this test account that will keep being recycled. The aim is to reduce the chance of conflicts when two E2E CI run at the same time.


marcbouchenoire (approved):
God's work 🙏

CTNicholas (approved):
Nice! Thanks for the fix 🙌

PR #1142, opened 2023-09-20
Implement request deduping and handle race conditions during mutations and revalidation.

nimeshnayaju:
### Description
This PR updates the current data fetching solution for threads with a solution inspired by [swr](https://swr.vercel.app).

1. Implements request deduping during revalidation on polling or changes in comment events via WebSocket.
2. Handle race conditions that may happen due to:
    - Mutation overlapping with another mutation.
    - Mutation overlapping with a revalidation request. https://github.com/liveblocks/liveblocks.io/issues/1524
    - Revalidation request overlapping with another revalidation request.


packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -1,4 +1,26 @@
-/// <reference types="react/experimental" />
+/**
+MIT License
```
nimeshnayaju, line 2: The revalidation logic is based on [swr](https://swr.vercel.app)'s revalidation implementation. I've added credits here!

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -93,166 +118,262 @@ export type RoomThreads<TThreadMetadata extends BaseMetadata> =
...
+  timestamp: number;
+};
+
+type MutationInfo = {
+  startTime: number;
+  endTime: number;
+};
+
+class StateManager<Data = any> extends EventTarget {
+  private _cache: State<Data> | undefined; // Stores the current cache state (threads)
```
nimeshnayaju, line 136: Chose to use the `EventTarget` Web API for the `StateManager` class as it aligns well with its browser-only usage context and doesn't require the additional methods provided by the `makeEventSource` API. However, I remain open to switching to `makeEventSource` if there is a stronger team preference for it.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# v1.3.6
+
+### `@liveblocks/react`
+
+- Implement request deduping and handle race conditions during mutations and
+  revalidation, inspired by SWR.
+
```
marcbouchenoire, line 7: It should be part of `v.1.3.5` since it hasn't been released yet. Also, I think we should mention Comments since it lives within `@liveblocks/react`, and maybe we can keep the SWR credit in the code.

```
- Deduplicate Comments requests and improve how race conditions are handled during mutations.
```

packages/liveblocks-react/src/comments/CommentsRoom.ts
marcbouchenoire: I'm not sure if the full license is necessary (especially given that the logic is going to evolve greatly over time with the remaining optimizations we have in mind), ~maybe just a normal comment explaining that this implementation is based on SWR and linking to it?~ I see that there's already another comment below linking to SWR directly.

packages/liveblocks-react/src/comments/CommentsRoom.ts
marcbouchenoire: I don't have strong preferences on these, maybe @nvie?

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -80,7 +105,7 @@ function createOptimisticId(prefix: string) {
   return `${prefix}_${nanoid()}`;
 }
 
-export type RoomThreads<TThreadMetadata extends BaseMetadata> =
+export type State<Data = any> =
```
marcbouchenoire, line 108: Given that this is only for one use-case, I don't know if we should make all types generic like SWR.

nvie, line 108: 1. The name `State` is very generic, even though this is a datastructure specifically used for threads only (due to the `threads` property in there). Maybe rename this to something more specific, or… do the opposite, and make this a truly generic type, using a `data` property here.
2. Try to avoid using default type params, i.e. avoid using `=` unless there is a good reason. Especially with generic types, you don't want to make it possible to use `const s = State;` without specifying the expected payload, for the same reason that `const arr = Array;` would not be valid.
3. Avoid using `any`s at all costs, especially as a library author. They infect, often unintentionally, many parts of our user's code bases for no good reason. Most of the time, just use `unknown` (or `never` if you don't anticipate using the key), but in this case, as per point (2), I would just remove it entirely.

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -93,166 +118,262 @@ export type RoomThreads<TThreadMetadata extends BaseMetadata> =
...
-  // When all mutations are finished, we pull the source of truth from the backend
-
-  let numberOfMutations = 0;
-  function endMutation() {
-    numberOfMutations--;
-    if (numberOfMutations === 0) {
-      void revalidateThreads();
-    }
-  }
+type RequestInfo<Data = any> = {
```
nvie, line 125: Same, avoid using `= any` here.

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -93,166 +118,262 @@ export type RoomThreads<TThreadMetadata extends BaseMetadata> =
...
+  fetcher: Promise<Data>;
+  timestamp: number;
+};
+
+type MutationInfo = {
+  startTime: number;
+  endTime: number;
+};
+
+class StateManager<Data = any> extends EventTarget {
```
nvie, line 135: Ditto.

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -93,166 +118,262 @@ export type RoomThreads<TThreadMetadata extends BaseMetadata> =
...
-        const nextRealtimeClientConnected = status === "connected";
+  let timestamp = 0;
 
-        if (nextRealtimeClientConnected !== realtimeClientConnected) {
-          realtimeClientConnected = nextRealtimeClientConnected;
-          pollingHub.threads.restart(getPollingInterval());
-        }
-      });
+  let refCount = 0; // Reference count for the number of components with a subscription (via the `subscribe` function)
+  let disposer: (() => void) | undefined; // Disposer function for the `comments` event listener
```
marcbouchenoire, line 204: I think we should name these differently to be more explicit about what they are/do.

marcbouchenoire (changes requested):
Nice! I added a few comments, and I think maybe @nvie could take a look if possible? Not necessarily at the logic itself but at the code style.

packages/liveblocks-react/src/comments/CommentsRoom.ts
nvie: The code already ships, and it's used everywhere else, so I'd personally prefer that. Practically speaking, in most usage contexts, it's easier to use too because it returns the unsubscribe function directly, so where you use it, you don't have to go through the ceremony of naming the subscribe function and then making sure to pass the exact same function reference to the unsubscribe function later. And lastly, using the event source API ensures type-safety nicely. It's impossible to call `eventSource.notify(...)` with a param that isn't of the expected type, unlike doing `.dispatchEvent(...)`, which you can pass any event (type). So IMO, code maintenance is much easier with the event source API — it's why it exists.

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -93,166 +118,262 @@ export type RoomThreads<TThreadMetadata extends BaseMetadata> =
...
 
-    pollingHub.threads.resume();
+  removeEventListener(
+    type: string,
+    callback: EventListenerOrEventListenerObject | null,
+    options?: boolean | EventListenerOptions | undefined
+  ): void {
+    super.removeEventListener(type, callback, options);
   }
+}
```
nvie, line 187: These are likely needed to enforce type-safety of the params here since they're just calling `super`'s method, right? None of that code is needed if you use the event source API instead, and it's type-safe automatically.

packages/liveblocks-react/src/comments/CommentsRoom.ts
nvie: Since this isn't literally Vercel's code you're using here, but just borrowing ideas (which is originally the idea from [HTTP RFC 5861](https://datatracker.ietf.org/doc/html/rfc5861)), I don't think this copyright comment is technically appropriate here? I'd suggest replacing the comment with a shout-out or attribution of inspiration to the SWR project, and maybe include a link for reference, but avoid using a copyright.

nvie (commented):
Left a few thoughts on form. I didn't review that actual caching code/feature here just yet.

CHANGELOG.md
nvie: Yeah, this is in part my bad. I used to add `# v1.3.5 (not released yet)` in the heading which made that easy to recognize to avoid just this, but I've stopped doing that because these days it's a bit too annoying to open a PR right before a release just to remove that `(not released yet)` part of the heading, getting CI to pass, getting approval, etc. Maybe we could set things up in a way that would allow us to have a free-to-merge pass on PRs that only change README or CHANGELOG files and still remain SOC2-compliant?

marcbouchenoire: I also stopped doing this (I used to add it as `# Unpublished`) for the same reason, it would be nice to have a way to quickly merge these types of PRs yeah. 🙏

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -97,153 +99,239 @@ export type RoomThreads<TThreadMetadata extends BaseMetadata> =
...
+
+function createThreadsManager<TThreadMetadata extends BaseMetadata>() {
+  let cache: RoomThreads<TThreadMetadata> | undefined; // Stores the current cache state (threads)
+  let request: ThreadsRequestInfo<TThreadMetadata> | undefined; // Stores the currently active revalidation request
+  let mutation: MutationInfo | undefined; // Stores the start and end time of the currently active mutation
+
+  const eventSource = makeEventSource<
+    RoomThreads<TThreadMetadata> | undefined
+  >();
+
```
nimeshnayaju, line 120: Updated to use `EventSource` API. One of the reasons why I opted to use `EventTarget` Web API was that I wanted to use standard Web APIs, so anyone reading the code in future won't have to learn a new API if they already knew the `EventTarget` API but I have to agree that this `EventSource` API provides a much better experience. Thanks for the suggestion @nvie

packages/liveblocks-react/src/comments/CommentsRoom.ts
nimeshnayaju: I started with wanting to build a generic cache manager but ended up pivoting to only threads. I agree the name is too generic. I've reverted the naming to use what we originally used:

```
export type RoomThreads<TThreadMetadata extends BaseMetadata> =
  | {
      isLoading: true;
      threads?: never;
      error?: never;
    }
  | {
      isLoading: false;
      threads?: never;
      error: Error;
    }
  | {
      isLoading: false;
      threads: ThreadData<TThreadMetadata>[];
      error?: never;
    };
```

packages/liveblocks-react/src/comments/CommentsRoom.ts
nimeshnayaju: Done!

packages/liveblocks-react/src/comments/CommentsRoom.ts
nimeshnayaju: Just added a comment related to this: https://github.com/liveblocks/liveblocks/pull/1142#discussion_r1333316264

> Updated to use `EventSource` API. One of the reasons why I opted to use `EventTarget` Web API was that I wanted to use standard Web APIs, so anyone reading the code in future won't have to learn a new API if they already knew the `EventTarget` API but I have to agree that this `EventSource` API provides a much better experience. Thanks for the suggestion @nvie

CHANGELOG.md
```diff
@@ -1,10 +1,17 @@
...
 - Officially mark `useList()`, `useMap()`, and `useObject()` as deprecated in
   JSDoc comments (we stopped recommending them since the release of 0.18)
 
+# v1.3.5
+
+### `@liveblocks/react`
+
+- Deduplicate Comments requests and improve how race conditions are handled
+  during mutations.
+
```
nvie, line 14: v1.3.5 hasn't been published yet (although it's on `main` already). Sorry for the confusion!

```suggestion
# v1.3.5

### `@liveblocks/react`

- Officially mark `useList()`, `useMap()`, and `useObject()` as deprecated in
  JSDoc comments (we stopped recommending them since the release of 0.18)

### `@liveblocks/react`

- Deduplicate Comments requests and improve how race conditions are handled
  during mutations.

```

nvie (approved):
Nice work, @nimeshnayaju! 💯

marcbouchenoire (approved):
Great work! 🥳

Issue #1143, opened 2023-09-21
Comment hooks (`useThreads`, `useEditThreadMetadata`, etc) don't work as expected on non-suspense mode


nimeshnayaju:
**Describe the bug**

When using comments on non-suspense, `useThreads` returns an object with `isLoading` property always set to true.

**To Reproduce**

1. Export non-suspense `useThreads` hook.
2. Use this hook inside a component.
3. Log the object returned by this hook. This object will always have a `isLoading` property set to true.

**Expected behavior**
`useThreads` should return an object with isLoading initially set to true but would later resolve and have a data property set to the list of threads.

PR #1150, opened 2023-09-22
Fix `useThreads`, `useEditThreadMetada`, etc comment hooks  not working on non-suspense export

nimeshnayaju:
`useThreads`, `useEditThreadMetada`, etc comment hooks were not working on non-suspense export because the room reference inside the hook  used was an outdated one.

Fixes: https://github.com/liveblocks/liveblocks/issues/1143

### DIFF
--- a/packages/liveblocks-react/src/factory.tsx
+++ b/packages/liveblocks-react/src/factory.tsx
@@ -296,9 +296,9 @@ export function createRoomContext<
       setRoom(room);
 
       return () => {
-        const commentsRoom = commentsRooms.get(roomId);
+        const commentsRoom = commentsRooms.get(room);
         if (commentsRoom) {
-          commentsRooms.delete(roomId);
+          commentsRooms.delete(room);
         }
         client.leave(roomId);
       };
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react/src/factory.tsx
```diff
@@ -296,9 +296,9 @@ export function createRoomContext<
       setRoom(room);
 
       return () => {
-        const commentsRoom = commentsRooms.get(roomId);
+        const commentsRoom = commentsRooms.get(room);
```
nvie, line 299: The `roomId` is a string, whereas the `room` is a Room instance. How did TypeScript accept this before?

packages/liveblocks-react/src/factory.tsx
```diff
@@ -843,15 +843,18 @@ export function createRoomContext<
...
 
   function getCommentsRoom(
     room: Room<TPresence, TStorage, TUserMeta, TRoomEvent>
   ) {
-    let commentsRoom = commentsRooms.get(room.id);
+    let commentsRoom = commentsRooms.get(room);
     if (commentsRoom === undefined) {
       commentsRoom = createCommentsRoom(room, commentsErrorEventSource);
-      commentsRooms.set(room.id, commentsRoom);
+      commentsRooms.set(room, commentsRoom);
```
nimeshnayaju, line 857: `commentsRooms` map now uses the room reference instead of the room id to identify a comment room. This ensures that whenever the room reference changes, a new comment room is created. 

This wasn't true when the room id was used to identify a comment room. Say, you had a room with id `nextjs-threads`, and you deleted this room and created a new room with the same id. Because the id hadn't changed between the two rooms, our comment room would still references the old room (that was destroyed).

packages/liveblocks-react/src/factory.tsx
marcbouchenoire: The `commentsRoom` map used to be `Map<string, CommentsRoom>` before this PR. https://github.com/liveblocks/liveblocks/pull/1150/files#r1334019235

marcbouchenoire (approved):
Great find! 🙏

PR #1152, opened 2023-09-25
Support HTTP fallback with any token type

nvie:
This fixes the HTTP fallback so it works with any token type, not just our legacy single-room tokens (which are no longer recommended since 1.2). Also makes it work with public key auth'ed clients.

This is the client counterpart to https://github.com/liveblocks/liveblocks-cloudflare/pull/540, which will need to be merged first.

Fixes https://github.com/liveblocks/liveblocks-cloudflare/issues/536.

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -706,6 +706,7 @@ type StaticSessionInfo = {
 
 type DynamicSessionInfo = {
   readonly actor: number;
+  readonly nonce: string;
   readonly scopes: string[];
 };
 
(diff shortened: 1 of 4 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -1140,27 +1142,44 @@ export function createRoom<
...
+    const fetcher = config.polyfills?.fetch || /* istanbul ignore next */ fetch;
+    return fetcher(url.toString(), {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        Authorization: `Bearer ${authTokenOrPublicApiKey}`,
+      },
+      body: JSON.stringify({ nonce, messages }),
+    });
+  }
```
nvie, line 1166: Key change here is twofold:

- Using `/v2/c/` now, instead of `/v2/` (L1154)
- Sending the `nonce` in the payload, alongside `messages` (previously `messages` _was_ the payload) (L1164)

marcbouchenoire (approved):

PR #1153, opened 2023-09-26
Inline custom shell script into package command

nvie:
…and rely on Turbo to deal with the dependencies efficiently.

This avoids build issues on Windows systems.


CTNicholas (approved):
Thank you! It's working now 🙂

PR #1156, opened 2023-09-27
Comments: Harmonize Suspense API

marcbouchenoire:
This PR harmonizes the Suspense version of `useThreads` and `useUser`.

Both Suspense versions will now:
- Throw on error
- Return the same object as their non-Suspense versions (but typed with `isLoading: false` and `error?: never`)

### DIFF
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -1130,6 +1130,9 @@ import { useThreads } from "./liveblocks.config";
 const { threads, error, isLoading } = useThreads();
 ```
 
+👉 A [Suspense version][] of this hook is also available, which will never
+return a loading state and will throw when there's an error.
+
 ### useUser
 
 Returns user info from a given user ID.
@@ -1139,6 +1142,9 @@ import { useUser } from "./liveblocks.config";
 
 const { user, error, isLoading } = useUser("user-id");
 ```
+
+👉 A [Suspense version][] of this hook is also available, which will never
+return a loading state and will throw when there's an error.
 
 <Banner title="Providing a resolver function">
 
--- a/packages/liveblocks-react/src/comments/CommentsRoom.ts
+++ b/packages/liveblocks-react/src/comments/CommentsRoom.ts
@@ -7,7 +7,6 @@ import type {
   Json,
   JsonObject,
   LsonObject,
-  Resolve,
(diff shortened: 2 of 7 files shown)


packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -99,6 +100,10 @@ export type RoomThreads<TThreadMetadata extends BaseMetadata> =
       error?: never;
     };
 
+export type RoomThreadsSuspense<TThreadMetadata extends BaseMetadata> = Resolve<
+  Extract<RoomThreads<TThreadMetadata>, { isLoading: false; error?: never }>
```
nvie, line 104: This construct is unnecessarily hard to read and might be hard to maintain. `RoomThreads` is a union of three members, of which you're excluding two here. Why not name those members explicitly, and just refer to [the one member](https://github.com/liveblocks/liveblocks/blob/168610b68f879da0f977864fa52b0d56c9221c88/packages/liveblocks-react/src/comments/CommentsRoom.ts#L87-L91) you're interested in here by a named type? I think it will make this much simpler to follow. You can even drop the `Resolve` helper too, and we can use "jump to definition".

nvie (approved):
Nice cleanup! 🙏

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -1130,6 +1130,9 @@ import { useThreads } from "./liveblocks.config";
 const { threads, error, isLoading } = useThreads();
 ```
 
+👉 A [Suspense version][] of this hook is also available, which will never
```
CTNicholas, line 1133: I need to remove all these emoji! But yes, this is consistent with the rest of the page, so we'll leave it as it is for now, thanks 🙂

CTNicholas (approved):
Docs look good!

packages/liveblocks-react/src/comments/CommentsRoom.ts
marcbouchenoire: Agree! Changed in https://github.com/liveblocks/liveblocks/pull/1156/commits/c03ef95d1608fd317ddb61d92d494d03e8e7bec7.

GuillaumeSalles (approved):

nvie (approved):
Excellent! 👍

PR #1157, opened 2023-09-27
Add `user` property to broadcast events

nvie:
This adds the sending `user` to the payload of each broadcast event.

The `user` will be `null` if the event is broadcast from the server, or if the connection ID somehow isn't recognized (yet or no longer) as another user in the room. The latter situation can happen in three edge cases:
- If the broadcast event is received before client A has received its `ROOM_STATE` message.
- If the broadcast event is received by client A before client B's initial `UPDATE_PRESENCE` message has reached client A.
- If client B already left the room by the time client A receives this broadcast event.

## TODO

- [x] Fix failing tests
- [x] Update [`Room.broadcastEvent`](https://liveblocks.io/docs/api-reference/liveblocks-client#Room.broadcastEvent) docs
- [x] Update `useBroadcastEvent` React hook similarly
- [x] Update documentation for [`Room.subscribe("event")`](https://liveblocks.io/docs/api-reference/liveblocks-client#Room.subscribe.event)
- [x] Update documentation for [`useBroadcastEvent`](https://liveblocks.io/docs/api-reference/liveblocks-react#useBroadcastEvent)

### DIFF
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -592,12 +592,20 @@ return `null`.
 Returns a callback that lets you broadcast custom events to other users in the
 room.
 
-```ts
+```ts highlight="3-5"
 import { useBroadcastEvent } from "./liveblocks.config";
 
+// On client A
 const broadcast = useBroadcastEvent();
-
 broadcast({ type: "EMOJI", emoji: "🔥" });
+
+// On client B
+useEventListener(({ event, user }) => {
+  //                       ^^^^ Will be Client A
+  if (event.type === "EMOJI") {
+    // Do something
+  }
+});
 ```
 
 ### useEventListener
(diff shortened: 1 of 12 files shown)


docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -592,12 +592,20 @@ return `null`.
...
+```ts highlight="3-5"
 import { useBroadcastEvent } from "./liveblocks.config";
 
+// On client A
 const broadcast = useBroadcastEvent();
-
 broadcast({ type: "EMOJI", emoji: "🔥" });
+
+// On client B
+useEventListener(({ event, user }) => {
```
CTNicholas, line 603:
```suggestion
useEventListener(({ event, user, connectionId }) => {
```

Even though `user.connectionId` is the same, shouldn't we still show every option? 

Though if we're deprecating `{ connectionId }` soon, we should add a deprecation message to the comments in the package.

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -608,16 +616,27 @@ the user that sent the message. If an event was sent from the
...
+```ts highlight="7-13"
 import { useEventListener } from "./liveblocks.config";
 
-useEventListener(({ connectionId, event }) => {
+// On client A
+const broadcast = useBroadcastEvent();
+broadcast({ type: "EMOJI", emoji: "🔥" });
+
+// On client B
+useEventListener(({ event, user }) => {
```
CTNicholas, line 627: Same as above!

packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -1541,9 +1541,40 @@ describe("room", () => {
...
+        SOCKET_AUTOCONNECT_BUT_NO_ROOM_STATE
+      );
+      room.connect();
 
       wss.onConnection((conn) => {
+        conn.server.send(
+          serverMessage({
+            type: ServerMsgCode.ROOM_STATE,
+            actor: 2,
+            nonce: "nonce-for-actor-2",
```
CTNicholas, line 1556: > nonce

This word has a very different meaning in the UK 😂

CTNicholas (approved):
Looks good, just a quick comment!

packages/liveblocks-core/src/__tests__/room.test.ts
nvie: So I learned… after shipping this 🙈 To be clear, its intended meaning is [from cryptography](https://en.wikipedia.org/wiki/Cryptographic_nonce).

docs/pages/api-reference/liveblocks-react.mdx
nvie: I don't think we want to (or should) deprecate the `connectionId`. There still is a niche use case for it: if it's `>= 0`, but `user` is still `null`, you could know it's not a server event, but the user is not yet (or no longer) in the room. Depending on the app you may want to ignore those events, yet not ignore server events.

nvie: You convinced me that it's better to should all the available fields in the example though — will add 👍

PR #1158, opened 2023-09-27
Comments: Document `CommentsConfig`

marcbouchenoire:
`CommentsConfig` is currently only mentioned in the `Overrides` section but not fully documented.


CTNicholas (approved):
Nice!

PR #1160, opened 2023-09-28
Support using `@liveblocks/node` on Edge runtimes

nvie:
This PR is an attempt to fix this issue:

Screen Shot 2023-09-28 at 09 20 04@2x

Rather than adding a package `@liveblocks/edge`, this PR tries to make `node-fetch` an optional dependency instead, so it just works on all envs, including Edge runtimes. This change is made rather "dry", and it's not tested on all combinations of environments.

@CTNicholas I may need some help to verify the following:

- [x] ~It should not work on Node < 18 when `node-fetch` is *not* installed~  We decided to keep `node-fetch` as a dependency after all, so this is impossible
- [x] ~Things JustWork™ on Node >= 18 when `node-fetch` is *not* installed~
- [x] Things JustWork™ on Node < 18
- [x] Things JustWork™ on Node >= 18
- [x] Things JustWork™ on Edge runtimes (to test/replicate, see [these instructions](https://liveblocks.slack.com/archives/C01TDLHJ9R6/p1695922638908799?thread_ts=1695320999.608849&cid=C01TDLHJ9R6))


packages/liveblocks-node/src/utils.ts
```diff
@@ -1,3 +1,9 @@
+export async function fetchPolyfill(): Promise<typeof fetch> {
+  return typeof globalThis.fetch !== "undefined"
+    ? globalThis.fetch
+    : ((await import("node-fetch")).default as unknown as typeof fetch);
```
jrowny, line 4: neat

jrowny (approved):

GuillaumeSalles (approved):
🚀

PR #1173, opened 2023-09-29
Update changelog

nvie:
Rewrite changelog for 1.4 to show the diff between 1.3 and 1.4, not 1.3.6 and 1.4.


nvie:
Keeping the changelog as-is (showing the diff since 1.3.6), but updating the GitHub release notes to show the diff since 1.3 instead 👍

PR #1176, opened 2023-09-29
Add new `backgroundKeepAliveTimeout` client option

nvie:
Example usage:

```tsx
const client = createClient({
  authEndpoint: "/api/auth",
  backgroundKeepAliveTimeout: 300000,  // e.g. 5 minutes
});
```

Without setting it, the client will forever try to keep the connection to the Liveblocks server alive — aka exactly like today's behavior.

When specifying a number of milliseconds here, clients are allowed to go into a dorment "zombie state", where they are disconnected from the server for as long as they remain in the background, but they are ready to be woken up as soon as the tab becomes visible again. This way, those background tabs won't count towards the MSU metrics.

### How this works

Getting this exactly right around all edge cases was tricky, but here is how the solution works:

- There is a new `@idle.zombie` state in the state machine, which will externally be observable as "reconnecting" status (but not doing anything)
- The trigger to decide whether to go into this state is whenever we would otherwise send a heartbeat to the server. When the client is ready to perform a heartbeat, it does the following checks:
  - Is a `backgroundKeepAliveTimeout` value configured?
  - Is the browser tab _currently_ in the background?
  - Has it continuously been in the background for _at least_ `backgroundKeepAliveTimeout` milliseconds?
  - Are there no pending changes? (i.e. `getStorageStatus() !== 'synchronizing'`)

Only if all of those conditions apply, the client may transition to "zombie state". Effectively it just disconnects the socket and waits for the browser tab to become visible again.

Fixes https://github.com/liveblocks/liveblocks.io/issues/1561.

## TODO

- [x] Documentation
- [ ] Unit tests


packages/liveblocks-core/src/__tests__/_behaviors.ts
```diff
@@ -59,11 +59,13 @@ export function defineBehavior(
...
   const wss = new MockWebSocketServer();
   const createSocket = () => socketBehavior(wss);
+  const canZombie = () => false;
 
-  const delegates: RoomDelegates = {
+  const delegates: Delegates<AuthValue> = {
     authenticate: jest.fn(authenticate),
     createSocket: jest.fn(createSocket),
-    //            ^^^^^^^ Allow observing these calls in tests
+    canZombie: jest.fn(canZombie),
```
jrowny, line 67: I think I just found my new band name.

packages/liveblocks-core/src/connection.ts
```diff
@@ -767,18 +771,31 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
...
     },
   };
 
+  const maybeHeartbeat: Target<Context, Event | BuiltinEvent, State> = () => {
+    // If the browser tab isn't visible currently, ask the application if going
+    // zombie is fine
+    const doc = typeof document !== "undefined" ? document : undefined;
+    const canZombie =
+      doc?.visibilityState === "hidden" && delegates.canZombie();
+    return canZombie ? "@idle.zombie" : sendHeartbeat;
```
jrowny, line 785: why do we want zombies to not heartbeat?

jrowny: oh, nm, I see that zombie means it's actually disconnected, for some reason in my head I was thinking this was the state when focus was lost but it wasn't disconnected.

packages/liveblocks-core/src/room.ts
```diff
@@ -893,8 +894,40 @@ export function createRoom<
       ? options.initialStorage(config.roomId)
       : options.initialStorage;
 
+  const doc = typeof document !== "undefined" ? document : undefined;
+  let inBackgroundSince: number | null = null;
+
+  function onVisibilityChange() {
+    if (doc?.visibilityState === "hidden") {
```
jrowny, line 901: nit, but maybe consider early out unless config.backgroundKeepAliveTimeout is defined.

packages/liveblocks-core/src/connection.ts
```diff
@@ -767,18 +771,31 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
...
-    .addTimedTransition("@ok.connected", HEARTBEAT_INTERVAL, sendHeartbeat)
+    .addTimedTransition("@ok.connected", HEARTBEAT_INTERVAL, maybeHeartbeat)
     .addTransitions("@ok.connected", {
-      NAVIGATOR_OFFLINE: sendHeartbeat, // Don't take the browser's word for it when it says it's offline. Do a ping/pong to make sure.
+      NAVIGATOR_OFFLINE: maybeHeartbeat, // Don't take the browser's word for it when it says it's offline. Do a ping/pong to make sure.
       WINDOW_GOT_FOCUS: sendHeartbeat,
     });
 
+  machine.addTransitions("@idle.zombie", {
+    WINDOW_GOT_FOCUS: "@connecting.backoff", // When in zombie state, the client will try to wake up automatically when the window regains focus
```
jrowny, line 796: I can't seem to figure out how does this get triggered?

jrowny: oh, I see, WINDOW_GOT_FOCUS already existed, not part of this PR. I answer my own questions 😂

jrowny (approved):
looks good, but two pieces of feedback:
- do you think we should warn people who set this variable in a non-document env, like node?
- the config value is set in milliseconds, but the resolution is really up to the ping interval right? This probably isn't critical to the functionality, but I can imagine we're going to get a ticket at some point stating that this is triggering at 35 seconds when they set it to 31.25 seconds. Maybe a note in the documentation is enough.

packages/liveblocks-core/src/__tests__/_behaviors.ts
nvie: 🤘🧟‍♂️🎸

packages/liveblocks-core/src/connection.ts
nvie: Correct! Instead of transitioning within the `@ok.*` group to "awaiting pong" state, "going zombie" transitions to the `@idle.zombie` state, which leaves the `@ok.*` state group and as a result the connection will get terminated (the `@ok.*` state group's exit function will handle that).

packages/liveblocks-core/src/room.ts
nvie: I wanted to keep this logic generic and since it's just keeping track of a number, I don't think it's worth polluting it here. In a75a08ab8c0bb96c1eeae06990efed9f08d85af0, I've refactored this thing out into a helper function that's independent of the room at all, so we don't have to pollute the function with unnecessary local variables/functions that only complicate the already-large `createRoom` function.

nvie:
> do you think we should warn people who set this variable in a non-document env, like node?

That's actually quite a nice idea. I've added this in 0f50b48948431f6a2f0edbce14f7fa1a7f720ce6 — thanks 👍

> the config value is set in milliseconds, but the resolution is really up to the ping interval right?

Correct! It is definitely not precise. The reason I kept this as a value in mililseconds is mainly just for consistency with our other values. We do guarantee that it will never happen _sooner_ than this configured value. I think we can just document that though.

In practice, if _t_ is the "time in the background", it will happen anywhere between:  
`config.backgroundKeepAliveTimeout` <= _t_ <= `config.backgroundKeepAliveTimeout` + 30 seconds

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -178,6 +178,34 @@ that this also affects when `others` are reset to an empty array after a
...
+```ts
+import { createClient } from "@liveblocks/client";
+
+const client = createClient({
+  /* ... other options ... */
+  backgroundKeepAliveTimeout: 15 * 60 * 1000 /* 15 minutes */,
+});
+```
+
+We recommend setting this value to at least multiple minutes.
```
nvie, line 207: @CTNicholas Would you mind reviewing/finetuning/simplifying these docs a bit for me? Feel free to directly push your improvements/suggestions to this branch! 👍

CTNicholas (approved):
Great change!

nvie:
Unit testing this is pretty tricky for this one in the current state… I've been thinking about how to best test this, but it will most likely require me to make some bigger changes to the way the client gets those DOM's visibilitychange events. Rather than reading them from a global `document`, I'll likely have to dependency-inject that into the client in order to simulate a minimal DOM environment in there, and to be able to do timing-based tests on those.

Do we think it's worth doing that rewrite as part of this PR?

PR #1177, opened 2023-09-29
Fix "ghosting bug" caused by using tuples in Zustand and Redux

GuillaumeSalles:
### Description

Fixes the "ghosting bug" caused by using tuples in Zustand.

If a Zustand state contains arrays used as tuples, they are backed by `LiveList` under the hood. Updating the value of a tuple right after creating it can cause [LiveList.insert + LiveList.set conflicts](https://github.com/liveblocks/liveblocks/pull/1002) issues. This PR is not an attempt to fix the `LiveList.insert` + `LiveList.set` issue. This PR specifically fixes the issue for Zustand and Redux users by serializing the newly created `LiveList` children as `set` operations instead of `inserts`.

I locally updated the `zustand-whiteboard` example to use tuples instead of objects for coordinates and added a feature to insert on click to reproduce the issue. If a coordinates tuple contains more than 2 numbers, I show an error in the console.

### Before

https://github.com/liveblocks/liveblocks/assets/7273466/2b9424f2-6027-4f60-9be9-b5c853aff1a0

### After

https://github.com/liveblocks/liveblocks/assets/7273466/9ef9f8ba-8d3b-4484-aafa-1b67064e269f

### DIFF
--- a/packages/liveblocks-core/src/crdts/LiveList.ts
+++ b/packages/liveblocks-core/src/crdts/LiveList.ts
@@ -134,7 +134,16 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
     ops.push(op);
 
     for (const item of this._items) {
-      ops.push(...item._toOps(this._id, item._getParentKeyOrThrow(), pool));
+      const parentKey = item._getParentKeyOrThrow();
+      const childOps = HACK_addIntentAndDeletedIdToOperation(
+        item._toOps(this._id, parentKey, pool),
+        undefined
+      );
+      const childOpId = childOps[0].opId;
+      if (childOpId !== undefined) {
+        this._unacknowledgedSets.set(parentKey, childOpId);
+      }
+      ops.push(...childOps);
     }
 
     return ops;
--- a/package-lock.json
+++ b/package-lock.json
@@ -22689,7 +22689,7 @@
       }
     },
     "packages/create-liveblocks-app": {
-      "version": "1.4.0",
+      "version": "1.4.1-test1",
       "license": "Apache-2.0",
       "dependencies": {
         "ansi-colors": "^4.1.3",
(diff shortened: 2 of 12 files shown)


packages/liveblocks-core/src/crdts/LiveList.ts
```diff
@@ -134,7 +134,16 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
     ops.push(op);
 
     for (const item of this._items) {
-      ops.push(...item._toOps(this._id, item._getParentKeyOrThrow(), pool));
+      const parentKey = item._getParentKeyOrThrow();
+      const childOps = HACK_addIntentAndDeletedIdToOperation(
```
nvie, line 138: Had to dig up to understand what this function was doing exactly again. It adds the **set** intent and removes any potential `deletedId` value, specifically from the **first** op in the list. Maybe we can rename this function to make it a bit more clear what's happening (and why)?

packages/liveblocks-core/src/crdts/LiveList.ts
```diff
@@ -134,7 +134,16 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
...
+        item._toOps(this._id, parentKey, pool),
+        undefined
+      );
+      const childOpId = childOps[0].opId;
+      if (childOpId !== undefined) {
+        this._unacknowledgedSets.set(parentKey, childOpId);
```
nvie, line 144: Out of scope for this change, but I want to flag a slight worry here that it doesn't feel obvious that (or why) you need to mutate the `_unacknowledgedSets` structure after turning this into a set op above. Or mainly that it's now _this function's_ responsibility to do that. After looking at other code, I can understand it's the right thing, but the rationale isn't too obvious from the code yet, and I think how we update `_unacknowledgedSets` in various places may not be abstracted well-enough. It's mostly just very easy to make a mistake here — e.g. by forgetting to update this set.

But again, definitely out of scope for this PR.

nvie (approved):
🚀 This is a pretty clever approach! I don't think the code/change is very readable to the untrained eye, but that's maybe something for another PR. Thanks for the fix! 🙏

GuillaumeSalles:
I fully agree that this is not intuitive at all. This is a pretty hidden side effect. :/ I have ideas on how to restructure the `_toOps` + `HACK_addIntentAndDeletedIdToOperation` (we never use `HACK_addIntentAndDeletedIdToOperation` without using `_toOps`), but no time to do it right now.  I left a comment for now and try to come up with something in the coming days.

PR #1179, opened 2023-10-02
Comments: Performance improvements

marcbouchenoire:
This PR adds memoization to a couple of places:
- Our usage of the [`Intl` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl)
- The heaviest primitives (components)


nvie (approved):
🚤💨

PR #1180, opened 2023-10-02
Fix bug in `between()` edge case

nvie:
This PR fixes a bug in `between()`, where it could generate a position value that wasn't alphabetically between the two given values. Fast-check randomly found this counter example this morning! ❤️ 

`between("a", "a!")` would incorrectly return `"a0"` instead of `"a 0"`

### DIFF
--- a/packages/liveblocks-core/src/lib/__tests__/position.test.ts
+++ b/packages/liveblocks-core/src/lib/__tests__/position.test.ts
@@ -393,6 +393,9 @@ describe("makePosition", () => {
   test("before .1 lies .09", () =>
     expect(makePosition(undefined, ONE)).toBe(ZERO + NINE));
 
+  test("between .1 and .11 lies .105", () =>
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/lib/__tests__/position.test.ts
```diff
@@ -392,6 +393,9 @@ describe("makePosition", () => {
   test("before .1 lies .09", () =>
     expect(makePosition(undefined, ONE)).toBe(ZERO + NINE));
 
+  test("between .1 and .11 lies .105", () =>
+    expect(makePosition(ONE, asPos(ONE + ONE))).toBe(ONE + ZERO + MID));
```
nvie, line 397: The bug is clearly visible here: this previously returned the equivalent of `.15` instead of `.105`!

GuillaumeSalles (approved):
Nice catch! Thanks property based testing! :)

PR #1187, opened 2023-10-04
Fix a bug in `useOthers()`

nvie:
...that could lead to the warning "The result of getServerSnapshot should be cached to avoid an infinite loop".


marcbouchenoire (approved):

packages/liveblocks-react/src/factory.tsx
```diff
@@ -560,7 +569,7 @@ export function createRoomContext<
       [selector]
     );
 
-    const getServerSnapshot = React.useCallback((): Snapshot => null, []);
+    const getServerSnapshot = alwaysNull;
```
nvie, line 572: Avoiding `useCallback()` wrappers in all of our hooks might actually be a nice pattern to use for a library (I would not do this pattern in application code). I think we could rewrite most of these `useCallback`s and use a stable function reference instead, similar to this change.

I only changed it for the `() => null` factory here, but I think it could be nice just so React doesn't have to add all those callback functions to its registry and compare them on every render, etc.

PR #1188, opened 2023-10-04
Comments: Improve `CommentBody` (de-)serialization

marcbouchenoire:
This PR checks if a comment's body does exist instead of assuming it does, even with server validation it's still best not to assume too much from what we receive.


CTNicholas (approved):
Thanks for the change!

PR #1191, opened 2023-10-04
Improve E2E test suite

nvie:
Improve E2E test suite massively.

- Enable `playwright --ui` locally (run `turbo run test:ui`)
- Added background colors per client
- Added "tables" to print JSON-serialized values to the HTML, that can be easily read or awaited using test helpers `getJson` (read JSON value right now), `expectJson` (expect JSON value right now), or `waitForJson` (wait up to 2 seconds for the field to become the given value)
- Simplified, unified, cleaned up test assertions
- Remove all artificial sleeps - tests now always await an event or change, never arbitrary times
- Added render counts (useful when scrubbing back/forth through sessions)
- Showing what random action was/will be taken - makes scrubbing back through a recorded session much easier to understand
- Added FSM debug logging
- Added "padding" to inserted data items (even clients add left padding, odd clients add right padding). While this may sound stupid, it actually makes it much easier at a glance to see which client has inserted which items in the list. Also makes it much easier to tell apart `"48:I"` and `"49:I"` instantly, as they live in separate "columns" visually.
- Automatically generate unique room ID for each test
- Because tests are now all isolated, we can crank up to `--workers=8` by default

## TODO for later

I will open separate PRs for these later:
- [ ] Some tests are still marked `.skip()`. Those are the tests that are indicative of actual bugs we'll have to address.
- [ ] Offline tests need to be re-expressed differently. I'll rewrite those tomorrow, and re-enable them.
- [ ] I want to investigate options to expose private/internals controls for more precise "gated" flushes/receives, and make buttons to control those in the E2E apps.
- [ ] Fix bugs I found during this process.
- [ ] Add more tests to cover more areas, especially newer APIs that are pretty much untested still. For example, the Storage tests all use `useList`, `useObject`, etc and don't use the newer/preferred APIs.


GuillaumeSalles (approved):
I skimmed through a few tests file and it LGTM! With the demo you made on Wednesday, I feel confident to move forward with this. Amazing work :)

ofoucherot (approved):
Great work!

PR #1194, opened 2023-10-05
Fix room encoding bug

nvie:
This fixes a room encoding bug in old-style authorize URLs. (This was done correctly for all public APIs, but this encoding bug happened only when using the internal `liveblocksAuthorizeEndpoint` override option.)


ofoucherot (approved):
💯

marcbouchenoire (approved):

PR #1195, opened 2023-10-05
Add `client.logout()` method

nvie:
This PR introduces a new API on the Client:

```tsx
// Purges the auth token cache. If any rooms are still
// conneced, they are forced to reauthorize.
client.logout()
```

This is a solution for [this support question](https://discord.com/channels/913109211746009108/913158542565994606/1159209059573055498).

Since the Liveblocks `client` is a global that caches auth tokens, whenever you log out a person from your application, you'll need to clear that cache. Force-reloading the page in the client typically works for this, but for SPAs that don't force-reload, there is no way to clear our client's caches.

### DIFF
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -276,6 +276,18 @@ const room = client.getRoom("my-room");
 You’ll likely not need this API if you’re using the newer [`client.enterRoom`][]
 API.
 
+### Client.logout
+
+Purges any auth tokens from the client’s memory. If there are any rooms that are
+still connected, they will be forced to reauthorize.
+
+```ts
+client.logout();
+```
+
+Call this API if you have a single page application (SPA) and you want to log
+your user from your application.
+
 ## Room
 
 Room returned by [`client.enterRoom`][] (or [`client.getRoom`][]).
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -286,7 +286,7 @@ client.logout();
 ```
 
 Call this API if you have a single page application (SPA) and you want to log
-your user from your application.
+your user out.
 
 ## Room
 
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -374,7 +374,7 @@ export function createClient(options: ClientOptions): Client {
     // rooms will get reauthorized now that the auth cache is reset. If that
     // fails, they might disconnect.
     const IDLE_STATES: Status[] = ["initial", "disconnected"];
-    for (const room of roomsById.values()) {
+    for (const { room } of roomsById.values()) {
       if (!IDLE_STATES.includes(room.getStatus())) {
         room.reconnect();
       }
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -1,4 +1,5 @@
 import { createAuthManager } from "./auth-manager";
+import type { Status } from "./connection";
 import type { LsonObject } from "./crdts/Lson";
 import { linkDevTools, setupDevTools, unlinkDevTools } from "./devtools";
 import { deprecateIf } from "./lib/deprecation";
@@ -105,6 +106,14 @@ export type Client = {
    * @param roomId The id of the room
    */
   leave(roomId: string): void;
+
+  /**
+   * Purges all cached auth tokens and reconnects all rooms that are still
+   * connected, if any.
+   *
+   * Call this whenever you log out a user in your application.
+   */
+  logout(): void;
 };
 
 export type AuthEndpoint =
@@ -358,7 +367,23 @@ export function createClient(options: ClientOptions): Client {
     }
   }
 
+  function logout() {
+    authManager.reset();
+
+    // Reconnect all rooms that aren't idle, if any. This ensures that those
+    // rooms will get reauthorized now that the auth cache is reset. If that
+    // fails, they might disconnect.
+    const IDLE_STATES: Status[] = ["initial", "disconnected"];
+    for (const room of roomsById.values()) {
+      if (!IDLE_STATES.includes(room.getStatus())) {
+        room.reconnect();
+      }
+    }
+  }
+
   return {
+    logout,
+
     // Old, deprecated APIs
     enter,
     getRoom,
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -1,5 +1,5 @@
 import { createAuthManager } from "./auth-manager";
-import type { Status } from "./connection";
+import { isIdle } from "./connection";
 import type { LsonObject } from "./crdts/Lson";
 import { linkDevTools, setupDevTools, unlinkDevTools } from "./devtools";
 import { deprecateIf } from "./lib/deprecation";
@@ -373,9 +373,8 @@ export function createClient(options: ClientOptions): Client {
     // Reconnect all rooms that aren't idle, if any. This ensures that those
     // rooms will get reauthorized now that the auth cache is reset. If that
     // fails, they might disconnect.
-    const IDLE_STATES: Status[] = ["initial", "disconnected"];
     for (const { room } of roomsById.values()) {
-      if (!IDLE_STATES.includes(room.getStatus())) {
+      if (!isIdle(room.getStatus())) {
         room.reconnect();
       }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@ leave();
     room and an "unsubscribe function" to leave that room again. This newer API
     supports entering/leaving the same room multiple times, making it possible
     to connect to the same room from different parts of your application.
+  - `client.logout()` – Call this on the Liveblocks client when you log out a
+    user in your application. It will purge all auth tokens and force-leave any
+    rooms, if any are still connected.
 - Deprecated APIs:
   - `client.enter(roomId, options)`
   - `client.leave(roomId)`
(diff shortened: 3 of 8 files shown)


packages/liveblocks-react/src/factory.tsx
```diff
@@ -280,45 +319,44 @@ export function createRoomContext<
...
-    const [room, setRoom] = React.useState<
-      Room<TPresence, TStorage, TUserMeta, TRoomEvent>
-    >(() =>
-      client.enter(roomId, {
+    const [{ room }, setRoomLeavePair] = React.useState(() =>
+      stableEnterRoom(frozen.instanceId, roomId, {
         initialPresence: frozen.initialPresence,
         initialStorage: frozen.initialStorage,
-        shouldInitiallyConnect: frozen.shouldInitiallyConnect,
+        autoConnect: false,
```
jrowny, line 333: is this intentional? Should it be frozen.autoConnect?

packages/liveblocks-react/src/factory.tsx
```diff
@@ -280,45 +319,44 @@ export function createRoomContext<
...
+        initialStorage: frozen.initialStorage,
+        autoConnect: false,
+        unstable_batchedUpdates: frozen.unstable_batchedUpdates,
+      });
+
+      setRoomLeavePair(pair);
+      const { room, leave } = pair;
 
-      setRoom(room);
+      if (frozen.autoConnect) {
```
jrowny, line 349: oh, I see. but why does this need to happen here? I only worry that sometime in the future autoConnect takes on some additional meaning inside the client itself, but to the react library it means "just connect when mounted" and if those two meanings diverge it could be inconsistent.

packages/liveblocks-react/src/factory.tsx
```diff
@@ -280,45 +319,44 @@ export function createRoomContext<
...
-      const room = client.enter<TPresence, TStorage, TUserMeta, TRoomEvent>(
-        roomId,
-        {
-          initialPresence: frozen.initialPresence,
-          initialStorage: frozen.initialStorage,
-          shouldInitiallyConnect: frozen.shouldInitiallyConnect,
-          unstable_batchedUpdates: frozen.unstable_batchedUpdates,
-        }
-      );
+      const pair = stableEnterRoom(frozen.instanceId, roomId, {
```
jrowny, line 339: is the idea that this is hitting the cache? I don't totally understand why we have this in the setState as well as here.

packages/liveblocks-redux/src/index.ts
```diff
@@ -207,24 +214,33 @@ const internalEnhancer = <TState>(options: {
 
       const store = createStore(newReducer, initialState, enhancer);
 
-      function enterRoom(roomId: string) {
-        if (storageRoot) {
-          return;
+      function enterRoom(newRoomId: string): () => void {
+        if (lastRoomId === newRoomId) {
+          return lastLeaveFn!;
```
jrowny, line 219: would it make sense to make this consistent with core library and return `{ room, leave }` ?

packages/liveblocks-core/src/__tests__/client.node.test.ts
```diff
@@ -29,12 +29,12 @@ function atobPolyfillMock(data: string): string {
 
 function enterAndLeave(options: ClientOptions) {
   const client = createClient(options);
-  client.enter("room", { initialPresence: {} });
+  const { room: _, leave } = client.enterRoom("room", { initialPresence: {} });
```
jrowny, line 32: not sure if this is a big concern, but we're replacing the old methods here with the new ones, which leaves the old ones untested. Should we leave a legacy test in place until we're ready to remove them?

jrowny (approved):
This looks great! huge improvement, love the return leave style. I left some comments for my own understanding and a few minor suggestions.

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -749,14 +813,39 @@ Get the storage status.
...
+### Room.connect
+
+Connect the local room instance to the Liveblocks server. Does nothing if the
+room is already connecting, reconnecting or connected.
+
+```ts
+room.connect();
+```
+
+We don’t recommend using this API directly.
```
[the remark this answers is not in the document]
nvie: Since this is part of the diff of #1226 that this PR is stacked on top of, I'll fix this in that other PR 👍

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -179,14 +180,14 @@ two RoomProviders with the same room ID in your React tree.
...
-  to Liveblocks servers when the RoomProvider is rendered. By default equals to
-  `typeof window !== "undefined"`, meaning the RoomProvider tries to connect to
-  Liveblocks servers only on the client side.
-
-The `initialPresence`, `initialStorage` and `shouldInitiallyConnect` props are
-ignored after the first render, so changes to the initial value argument won’t
-have an observable effect.
+- `autoConnect` (optional) - Whether or not the room should automatically
+  connect to Liveblocks servers when the RoomProvider is mounted. By default
+  equals to `typeof window !== "undefined"`, meaning the RoomProvider tries to
```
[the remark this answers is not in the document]
CTNicholas: Good change, thanks for the recommendation!

packages/liveblocks-core/src/client.ts
```diff
@@ -248,24 +327,70 @@ export function createClient(options: ClientOptions): Client {
...
+    }
+  }
+
+  function logout() {
+    authManager.reset();
+
+    // Reconnect all rooms that aren't idle, if any. This ensures that those
+    // rooms will get reauthorized now that the auth cache is reset. If that
+    // fails, they might disconnect.
+    const IDLE_STATES: Status[] = ["initial", "disconnected"];
```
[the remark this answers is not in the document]
nvie: Ah, yes — done in 6d1a98f8ffa4642f72cb547209c1bc08457c84fc!

e2e/next-sandbox/test/multi.test.ts
```diff
@@ -0,0 +1,228 @@
...
+    await page.click("#mount_1");
+
+    await waitForJson(page, "#socketStatus_1", "connected");
+    const connId = (await getJson(page, "#connectionId_1")) as number;
+
+    await page.click("#unmount_1");
+    await page.click("#mount_1");
+    await waitForJson(page, "#socketStatus_1", "connected");
+
+    await expectJson(page, "#connectionId_1", connId + 1);
```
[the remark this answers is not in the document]
nvie: Yeah, this would work in the "expectJson" cases, but it would not work in the "waitForJson" cases, where I would also would like to add the ability to provide a predicate function, where checks like that would be much easier to express.

Imagine:

```tsx
await expectJson(page, "#connectionId_1", (value) => value > connId);
await waitForJson(page, "#connectionId_1", (value) => value > connId);
```

This is something I want to implement, but haven't done yet. I'd prefer to keep this out of this PR, but cycle back to implementing that and rewriting these expressions everywhere!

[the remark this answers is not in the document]
nvie: Tracking this in #1239.

packages/liveblocks-redux/src/index.ts
nvie: Don’t think it’s needed, since `enterRoom()` will just mutate the Zustane store, where `room` already exists. Maybe by the same argument, we could even let it return `void` like before, since you could just call `leaveRoom()` later. Internally it will keep a reference to the leave function already. We don’t have to expose it I think—thanks!

packages/liveblocks-react/src/factory.tsx
nvie: Yep! In a classic React provider implementation, typically on the first render the context value is still `null`, and then, on mount you would populate the actual context value.

However; we want to populate the `room` with a value on the first render so that we can continue rendering the component tree which may use `useRoom()` which does not expect to see a `null` value, ever.

First, if externally the RoomProvider’s `id` prop get changed, we’ll also want to run the leave function for the old room and connect to the new room. This is what the useEffect and the setState combo are for.

In the most common use case, the room ID prop won’t change, so the following sequence happens:

- Initial render on SSR: `room` has a value (and its state is “initial”)
- Initial hydration render on client: `room` has a value (and its state is “initial”)
- Effect is scheduled for room ID: setter is called, which would normally trigger a re-render, but since the value of `room` is retrieved from cache, its value won’t change and we spare ourselves an extra render. At the same time… we have the opportunity to grab a handle on the `leave` function we didn’t grab on the initial render, and register it for the unmount (or when the room changes)!

This construct already existed before my changes, btw. The previous/current implementation works the same way.

packages/liveblocks-react/src/factory.tsx
nvie: Yep, see below!

packages/liveblocks-react/src/factory.tsx
nvie: I've cleaned this up and documented the rationale for this a bit better in dd3be7b9a1403cacd85db17ff062675eb84f7ff5 (as part of the other PR). Hopefully this makes it a bit more obvious. The `autoConnect` feature was added mainly to support the React SSR use case, and to facilitate people that may want to use Liveblocks in "local only" mode (i.e. when you want to nest the same component tree, but not make it multiplayer).

packages/liveblocks-redux/src/index.ts
nvie: Fixed in 3026a119eeba57c0f376d2178b5f5652255179f8 (in the other PR).

packages/liveblocks-core/src/__tests__/client.node.test.ts
nvie: Great point! I realized this new behavior wasn't tests in unit test form yet, so I added a test and added a similar test for the old API in 81edfb325bcc7e6f97f01adee4b62a4faf4f471f. Thanks for catching that!

PR #1206, opened 2023-10-09
Docs: Add note to primitives API reference about common error

marcbouchenoire:
We might still want to make a lengthier guide in the future about module resolution but since it's a pretty common error we should at least have a small note with the fix in situ.


CTNicholas (approved):
Thanks for writing up the note!

PR #1207, opened 2023-10-09
Fix `useStatus()` SSR response

nvie:
This small "fix" will make the `useStatus()` hook return `"connecting"` in SSR responses (instead of `"initial"`), to match the most typical same value that would happen after hydration/initial render in the client. After all, in 95% of cases, the client will also be `"connecting"` on the first render of `useStatus`, and not `"initial"`.

Technically this isn't really a "proper fix", because there can still be cases that will cause a mismatch. However, such cases should be contrived or at least are by far in the minority.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# v1.4.6
+
+### `@liveblocks/react`
+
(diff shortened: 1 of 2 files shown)


CTNicholas (approved):
Ah, good change!

PR #1211, opened 2023-10-09
Fix offline E2E tests

nvie:
In #1191, I still kept the "offline" tests disabled as a lot of that has fundamentally changed in 1.1. This PR rewrites the offline test app to be account for a lot more possible scenarios, and adds E2E tests for all of those situations.

https://github.com/liveblocks/liveblocks/assets/83844/6a04986e-15e2-423a-8d52-2f2fc365a09e


GuillaumeSalles (approved):
Nice! Way cleaner!

PR #1214, opened 2023-10-09
Comments: Improve portals

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1544 by adding a common class (`lb-portal`) to all portaled elements to e.g. change their `z-index`.

(I looked into exposing our portal but even if we'd do that, I discovered that the one from Radix is exposed but doesn't accept/forward any DOM props so we can't add a class or data attribute to it. But it does copy the computed `z-index` from the element within it, so we mimic that behavior in our own portal and add `lb-portal` to all elements which are direct children of portals. Which means that setting a `z-index` on `lb-portal` will give the expected result.)


docs/pages/api-reference/liveblocks-react-comments.mdx
```diff
@@ -526,11 +526,11 @@ The components displayed within the editor.
...
-| `MentionSuggestions` | <code>ComponentType&lt;ComposerEditorMentionProps&gt;</code>            | The component used to display mention suggestions.<br/>Defaults to a list of the suggestions’ `userId`. |
-| `Link`               | <code>ComponentType&lt;ComposerEditorMentionSuggestionsProps&gt;</code> | The component used to display links.<br/>Defaults to the link’s `children` property.                    |
+| Component            | Type                                                         | Description                                                                                             |
+| -------------------- | ------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------- |
+| `Mention`            | <code>ComponentType&lt;ComposerEditorMentionProps&gt;</code> | The component used to display mentions.<br/>Defaults to the mention’s `userId` prefixed by an @.        |
+| `MentionSuggestions` | <code>ComponentType&lt;ComposerEditorMentionProps&gt;</code> | The component used to display mention suggestions.<br/>Defaults to a list of the suggestions’ `userId`. |
```
CTNicholas, line 532:
```suggestion
| `MentionSuggestions` | <code>ComponentType&lt;ComposerEditorMentionSuggestionProps&gt;</code> | The component used to display mention suggestions.<br/>Defaults to a list of the suggestions’ `userId`. |
```

I think this was a typo

CTNicholas (approved):
Thanks for adding this! Left a quick comment

docs/pages/api-reference/liveblocks-react-comments.mdx
marcbouchenoire: Ah yep! I found the one below but missed this one. Thanks!

PR #1219, opened 2023-10-11
Comments: Add `onThreadDelete`

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1598.

- **Breaking:** Rename `onEdit` and `onDelete` on `<Comment />` to `onCommentEdit` and `onCommentDelete` respectively
- Add `onThreadDelete` to `<Thread />`
- Add support for generic metadata type to `<Thread />`

Regarding the last point, components don't have access to the types given to `createRoomContext` since they don't come from it (they're just normal imports). To offer a better way than having to manually type `thread` within `onThreadDelete`, `<Thread />` now supports a generic type. It's opt-in so it can be used only when needed.

```tsx
import type { MyThreadMetadata } from "../liveblocks.config.ts"

<Thread<MyThreadMetadata> onThreadDelete={(thread) => console.log(thread)}  />
// thread: ThreadData<MyThreadMetadata>
```

It looks weird but if it's opt-in there's no harm? I could also add the same thing to `<Composer />` so that a `metadata` prop can be specified which would remove the need to resort to `onComposerSubmit` (while still being possible of course) just to set custom metadata when creating comments.

### DIFF
--- a/packages/liveblocks-core/src/comments/types/ThreadData.ts
+++ b/packages/liveblocks-core/src/comments/types/ThreadData.ts
@@ -4,14 +4,14 @@ import type { CommentData } from "./CommentData";
 /**
  * Represents a thread of comments.
  */
-export type ThreadData<ThreadMetadata extends BaseMetadata = never> = {
+export type ThreadData<TThreadMetadata extends BaseMetadata = never> = {
   type: "thread";
   id: string;
   roomId: string;
   createdAt: string;
   updatedAt?: string;
   comments: CommentData[];
-  metadata: [ThreadMetadata] extends [never]
+  metadata: [TThreadMetadata] extends [never]
     ? Record<string, never>
-    : ThreadMetadata;
+    : TThreadMetadata;
 };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@
 - Export `ComposerSubmitComment` type from root too, in addition to
   `/primitives`.
 - Add `onThreadDelete` to `Thread`.
-- Add support for specifying a custom `ThreadMetadata` type on `Thread`.
+- Add `metadata` to `Composer` to attach custom metadata to new threads.
+- Add support for specifying a custom `ThreadMetadata` type on `Thread` and
+  `Composer`.
 - **Breaking (beta):** `Comment`’s `onEdit` and `onDelete` were renamed to
   `onEditComment` and `onDeleteComment` respectively.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v1.4.6 (not released yet)
+# v1.4.6
 
 ### `@liveblocks/react`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 - Export `ComposerSubmitComment` type from root too, in addition to
   `/primitives`.
+- Add `onThreadDelete` to `Thread`.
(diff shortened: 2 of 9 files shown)


CTNicholas (approved):
Fantastic! Thanks for adding it so quickly. Changes look good to me 👍

packages/liveblocks-core/src/comments/types/ThreadData.ts
```diff
@@ -4,14 +4,14 @@ import type { CommentData } from "./CommentData";
 /**
  * Represents a thread of comments.
  */
-export type ThreadData<ThreadMetadata extends BaseMetadata = never> = {
+export type ThreadData<TThreadMetadata extends BaseMetadata = never> = {
```
nvie, line 7: Not really feedback for this PR directly, but general. Would it be possible to rename this type to:

```suggestion
export type Thread<TMeta extends BaseMetadata = never> = {
```

I think that would be just as clear, yet much shorter. Or is `Thread` already taken as a type elsewhere?

nvie (approved):
Looking good to me 👍 ! I had one suggestion to shorten the type names, if possible.

packages/liveblocks-core/src/comments/types/ThreadData.ts
marcbouchenoire: We've looked at this before and we couldn't make TypeScript accept both `Thread`/`Comment` types and `Thread`/`Comment` components. (And I think even if we did you get into issues quickly with auto-imports/ESLint etc?)

And regarding `TThreadMetadata`, I'm not a fan of naming the same generics/values differently throughout different places, e.g. having `TMeta` here but it's not possible in `createRoomContext` so `TThreadMeta` there and we refer to it in plain text as "thread metadata" because the property is called `metadata`. I think it gets confusing quickly and prevents easy understanding at a glance.

nvie: > We've looked at this before and we couldn't make TypeScript accept both `Thread`/`Comment` types and `Thread`/`Comment` components. (And I think even if we did you get into issues quickly with auto-imports/ESLint etc?)

I see! Which of these types do will you be working with mostly as a user? For example, if the typical type you import from our library is this (`ThreadData`) type, then maybe we can export the component types as `ThreadComponent` instead, and rename this type to just `Thread`? I don't expect users to often use the `ThreadComponent` type in normal code, right? In that case, there is no conflict, and the most commonly needed type is nice and short.

> And regarding `TThreadMetadata`, I'm not a fan of naming the same generics/values differently throughout different places, e.g. having `TMeta` here but it's not possible in `createRoomContext` so `TThreadMeta` there and we refer to it in plain text as "thread metadata" because the property is called `metadata`. I think it gets confusing quickly and prevents easy understanding at a glance.

Oh, I see. Yeah, that makes sense if the same type variable appears elsewhere: better to keep those consistent in that case 👍

marcbouchenoire: > I see! Which of these types do will you be working with mostly as a user? For example, if the typical type you import from our library is this (ThreadData) type, then maybe we can export the component types as ThreadComponent instead, and rename this type to just Thread? I don't expect users to often use the ThreadComponent type in normal code, right? In that case, there is no conflict, and the most commonly needed type is nice and short.

Oh sorry, I meant that `Thread`/`Comment` are the actual components so even more useful than the `ThreadData`/`CommentData` types, which is why we decided to suffix the types.

PR #1223, opened 2023-10-12
Use unstable streaming api to fetch initial storage

GuillaumeSalles:
Instead of getting the initial storage from the websocket, we're using th `storage-stream` API to reduce the memory usage on the server.

```
createClient({
   unstable_streamData: true
   /* ... */
})
```

**⚠️ Warning ⚠️**

`unstable_streamData` and `storage-stream` APIs are not final and will change in the short future

### Demo

https://github.com/liveblocks/liveblocks/assets/7273466/197326ab-21d7-4284-98ab-7b1f9791057b


packages/liveblocks-core/src/room.ts
```diff
@@ -1156,6 +1157,23 @@ export function createRoom<
     comments: makeEventSource<CommentsEventServerMsg>(),
   };
 
+  async function streamFetch(authTokenOrPublicApiKey: string, roomId: string) {
+    const baseUrl = new URL(config.liveblocksServer);
+    baseUrl.protocol = "https";
+    const url = new URL(
+      `/v2/c/rooms/${encodeURIComponent(roomId)}/storage-stream`,
```
nvie, line 1164: Agreed that we don't need the `-stream` suffix here.

packages/liveblocks-core/src/room.ts
```diff
@@ -1156,6 +1157,23 @@ export function createRoom<
     comments: makeEventSource<CommentsEventServerMsg>(),
   };
 
+  async function streamFetch(authTokenOrPublicApiKey: string, roomId: string) {
+    const baseUrl = new URL(config.liveblocksServer);
+    baseUrl.protocol = "https";
```
nvie, line 1162: Unrelated to this PR, but I'm doing similar hacking elsewhere! Should we redefine the "liveblocks server" config to be a value like `https://api.liveblocks.io` and not `wss://api.liveblocks.io/v7`?

packages/liveblocks-core/src/room.ts
```diff
@@ -2005,11 +2018,39 @@ export function createRoom<
...
+      return;
+    }
+    const result = await streamFetch(
+      managedSocket.authValue.type === "public"
+        ? managedSocket.authValue.publicApiKey
+        : managedSocket.authValue.token.raw,
+      config.roomId
+    );
+    const items = (await result.json()) as IdTuple<SerializedCrdt>[];
+    processInitialStorage({ type: ServerMsgCode.INITIAL_STORAGE_STATE, items });
```
nvie, line 2041: To be ahead of race conditions later, when this code grows more complicated, we may want to also think about cancelling any in-flight HTTP requests, or otherwise re-validating its mandate when its response arrives in the client.

Areas where this might get tricky in the future as this code grows more complex is when disconnecting/reconnecting a client while such an HTTP request is under way. In theory, you can send the HTTP request, disconnect/reconnect, send the HTTP request again, and the HTTP requests arriving back out of order, or multiple times, which may lead to unforeseen edge cases.

nvie (approved):
Looks good at first sight, but I may have a few reservations when it comes to edge case handling of these. May not be a direct issue right now, but personally think we should not embrace HTTP for this too enthusiastically just yet, and definitely keep this under the `unstable_` flag for now indeed.

packages/liveblocks-core/src/room.ts
GuillaumeSalles: Answered on slack

nvie: Created an issue for this to track: #1230

PR #1224, opened 2023-10-12
Add E2E test to verify thread creation and display

nimeshnayaju:
### Description
This PR adds two E2E basic tests to verify thread display and creation. The long term goal is to add more tests overtime.

### DIFF
--- a/e2e/next-sandbox/test/comments/index.test.ts
+++ b/e2e/next-sandbox/test/comments/index.test.ts
@@ -11,7 +11,7 @@ import {
 
 const TEST_URL = "http://localhost:3007/comments";
 
-test.describe("Threads", () => {
+test.describe("Comments", () => {
   let pages: [Page, Page];
 
   test.beforeEach(async ({}, testInfo) => {
--- a/e2e/next-sandbox/test/comments/index.test.ts
+++ b/e2e/next-sandbox/test/comments/index.test.ts
@@ -1,0 +1,42 @@
+import type { Page } from "@playwright/test";
+import { test } from "@playwright/test";
+
+import {
+  genRoomId,
+  getJson,
+  preparePages,
+  waitForJson,
+  waitUntilEqualOnAllPages,
+} from "../utils";
+
+const TEST_URL = "http://localhost:3007/comments";
+
+test.describe("Threads", () => {
+  let pages: [Page, Page];
+
+  test.beforeEach(async ({}, testInfo) => {
+    const room = genRoomId(testInfo);
+    pages = await preparePages(`${TEST_URL}?room=${encodeURIComponent(room)}`);
+  });
+
+  test.afterEach(() => Promise.all(pages.map((page) => page.close())));
+
+  test("verify A and B display same number of threads after threads are loaded", async () => {
+    await waitForJson(pages, "#isLoading", false);
+    await waitUntilEqualOnAllPages(pages, "#numOfThreads");
+  });
+
+  test("verify thread creation on B is broadcasted correctly to A", async () => {
+    const [page1, page2] = pages;
+
+    await waitForJson(pages, "#isLoading", false);
+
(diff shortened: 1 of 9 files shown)


e2e/next-sandbox/test/comments/index.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  test.afterEach(() => Promise.all(pages.map((page) => page.close())));
+
+  test("verify A and B display same number of threads after threads are loaded", async () => {
+    await waitForJson(pages, "#isLoading", false);
+    await waitUntilEqualOnAllPages(pages, "#numOfThreads");
+  });
+
+  test("verify thread creation on B is broadcasted correctly to A", async () => {
+    const [page1, page2] = pages;
```
nimeshnayaju, line 30: This tests for the broadcasting of thread/comment creation. We use event broadcasting to also relay information about thread metadata update, comment update, and comment deletion, so this test provides validity to those actions too. That being said, it might still be helpful to have a test item for each of the actions.

e2e/next-sandbox/test/comments/index.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  test.beforeEach(async ({}, testInfo) => {
+    const room = genRoomId(testInfo);
+    pages = await preparePages(`${TEST_URL}?room=${encodeURIComponent(room)}`);
+  });
+
+  test.afterEach(() => Promise.all(pages.map((page) => page.close())));
+
+  test("verify A and B display same number of threads after threads are loaded", async () => {
+    await waitForJson(pages, "#isLoading", false);
+    await waitUntilEqualOnAllPages(pages, "#numOfThreads");
```
nimeshnayaju, line 26: This tests for the initial loading of thread and verifies that threads are loaded correctly and all clients see the same number of threads.

nvie: When I try to run it locally (against the DEV servers), I'm seeing this error message:

Screen Shot 2023-10-13 at 17 35 20@2x

nvie (approved):
Thanks for extending the E2E test suite, @nimeshnayaju 🙌 ! Glad to see them get a bit more love! 😍 

I tried running them locally, but it somehow didn't work for me directly. Maybe I'm missing an env var for comments locally?

e2e/next-sandbox/test/comments/index.test.ts
nimeshnayaju: That is probably happening because,`serverEndpoint` need to be updated to point at the dev comments server while testing in dev.

https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-core/src/room.ts#L2256-L2258

```ts
  const commentsApi = createCommentsApi(config.roomId, delegates.authenticate, {
    serverEndpoint: "https://api.liveblocks.io/v2",
  });
```

marcbouchenoire (approved):
LGTM! 🧑‍🍳💋

e2e/next-sandbox/test/comments/index.test.ts
nvie: Ohh, good that you mention this. We were thinking to unify our various env var configs (client, comments, node) into just one new config var named `baseUrl`, which would replace `liveblocksServer`, `liveblocksAuthorizeEndpoint`, `liveblocksBaseUrl`, and maybe also this `serverEndpoint`. The idea is that we strip the `/v2` part from this config value, and that all three existing variables could then be _derived_ from the one `baseUrl` config then.

Any objections against me doing this soon, @marcbouchenoire @nimeshnayaju ?

nvie: I created a ticket for this one: https://github.com/liveblocks/liveblocks/issues/1230

PR #1226, opened 2023-10-12
Support for multi room providers

### LINKED ISSUE #1081: Support Redux/Zustand: Handle multiple `RoomProvider`
Ensure we can use multiple `RoomProvider` connected to the same room. This should also enable us to use `RoomProvider` while using `@liveblocks/zustand` / `@liveblocks/redux`

Two parts:
- Support multiple RoomProviders for the same room ID in different component trees in the same app
- Support nesting RoomProviders (or prevent that explicitly)

nvie:
Fixes #1081 and #1232.

This enables a new client API for entering and leaving rooms, which we'll start recommending over the existing APIs. But the existing APIs remain working exactly how they are today.

```tsx
// Old APIs we'll no longer be recommending (but that will remain working)
const room = client.enter("my-room", options);
client.getRoom("my-room");
client.leave("my-room");
```

```tsx
// New API we'll be recommending instead
const { room, leave } = client.enterRoom("my-room", options);
leave();
```

## TODO

- [x] Set merge base to `release-1.5` branch
- [x] Fix a memory leak with this implementation
- [x] Make it work with and without StrictMode
- [x] Make `.connect()` and `.disconnect()` APIs public on Room
- [x] Rename enter option `shouldInitiallyConnect` → `autoConnect`
- [x] Update React implementation
- [x] Update Redux implementation
- [x] Update Zustand implementation
- [x] Add E2E tests
- [x] Simplify ticket internals
- [x] Can we also simplify the internals of the per-Provider cache, and put those at the `Outer` level instead of the global closure? **Update:** Yes we can—fixed in a2978a2d20297e9cbbb3066ba8609dd41f235c06!
- [x] Deprecate no-longer-used `roomId` arguments
  - [x] In Redux
  - [x] In Zustand
  - [x] Double-check this decision: [is it really the best DX](https://github.com/liveblocks/liveblocks/pull/1226#discussion_r1362113080)? **Update:** No it isn't, we're going to simply technically break the API in TypeScript here, because the DX is still much better than the alternative.
- [x] Test this out on real examples
- [x] Documentation
  - [x] Document new Client `.enterRoom()` API
  - [x] Add "no longer recommended" badge to Client's `.enter()` and `.leave()` methods
  - [x] `Room.connect()` now public
  - [x] `Room.disconnect()` now public
  - [x] Rewrite all references to `shouldInitiallyConnect` → `autoConnect`
  - [x] Remove limitation about multiple room providers
  - [x] Rewrite/double-check code samples changes
  - [x] Rewrite/double-check guides
  - [x] #1237
  - [x] Check docs fully @CTNicholas


nvie:
@nimeshnayaju Would also like your review on this PR, given that this is closely related to [your origin story](https://github.com/liveblocks/liveblocks/issues/730) at Liveblocks 😉

marcbouchenoire (approved):
Can't wait to have this, great work on the API!

packages/liveblocks-core/src/client.ts
```diff
@@ -248,24 +275,100 @@ export function createClient(options: ClientOptions): Client {
...
+    TUserMeta extends BaseUserMeta = BaseUserMeta,
+    TRoomEvent extends Json = never,
+  >(
+    roomId: string,
+    options: EnterOptions<TPresence, TStorage>
+  ): {
+    room: Room<TPresence, TStorage, TUserMeta, TRoomEvent>;
+    leave: () => void;
+  } {
+    const ticket = createTicketForRoom(roomId, options);
```
GuillaumeSalles, line 294: Implementation details, so not blocking at all. Feel free to merge as is.

Is it possible to simplify the implementation by storing a map of roomId and array of leave callbacks instead of tickets `Map<string, Array<() => void>>`? What we're trying to do here is close the `EventSource` pattern so my intuition is telling me it could be modeled in the same way.

packages/liveblocks-redux/src/index.ts
```diff
@@ -353,14 +376,10 @@ function enterRoom(roomId: string): {
   };
 }
 
-function leaveRoom(roomId: string): {
+function leaveRoom(): {
```
GuillaumeSalles, line 379: Is it a breaking change?

packages/liveblocks-zustand/src/index.ts
```diff
@@ -41,12 +41,12 @@ export type LiveblocksContext<
...
-  readonly enterRoom: (roomId: string) => void;
+  readonly enterRoom: (roomId: string) => () => void;
   /**
-   * Leaves a room and stops sync it with zustand state.
-   * @param roomId The id of the room
+   * Leaves the currently entered room and stops sync it with zustand state, if
+   * any. If enterRoom was not called before, this is a no-op.
    */
-  readonly leaveRoom: (roomId: string) => void;
+  readonly leaveRoom: () => void;
```
GuillaumeSalles, line 49: Is it a breaking change?

CHANGELOG.md
```diff
@@ -1,3 +1,56 @@
...
+
+- New APIs:
+  - `client.enterRoom(roomId, options)` – enters the room and return both the
+    room and an "unsubscribe function" to leave that room again. This newer API
+    supports entering/leaving the same room multiple times, making it possible
+    to connect to the same room from different parts of your application.
+- Deprecated APIs:
+  - `client.enter(roomId, options)`
+  - `client.leave(roomId)`
+- Renamed enter option: `shouldInitiallyConnect` → `autoConnect`. Its meaning or
```
GuillaumeSalles, line 33: I like that name! :)

GuillaumeSalles (commented):
Glad that we're finally going to support multi `RoomProviders`.

Other than the breaking changes, just one question for you. Do you feel confident enough about this API or is there any risk that we'll find any edge cases (weird SSR behavior, `Strict` mode) that would force us to change it again? If you're feeling confident, I'll trust your guts :) 
If not, one possibility is to not document this API until 1.5 and just use it internally to give us a bit more breathing room to make sure there is no unexcepted edge cases.

packages/liveblocks-react/src/factory.tsx
```diff
@@ -238,22 +235,64 @@ export function createRoomContext<
...
+      origLeave();
+      roomCache.delete(key);
+    };
+
+    roomCache.set(key, rv);
+    return rv;
+  }
+
+  function RoomProviderOuter(props: RoomProviderProps<TPresence, TStorage>) {
+    // This "Outer" hack mostly exists to make ID generation work well in
```
GuillaumeSalles, line 274: Oof 😬

packages/liveblocks-redux/src/index.ts
nvie: Nope! Only at the type level it is. TypeScript will not like it if you call `leaveRoom("my-room")` but at runtime it works. I considered doing this instead:

```tsx
function leaveRoom(
  /** @deprecated No longer needed */
  _roomId?: string
) {
}
```

But outside of TypeScript, this isn't a breaking change.

packages/liveblocks-zustand/src/index.ts
nvie: Same here. Removing the last param from a function is not a breaking change, semver-wise.

packages/liveblocks-react/src/factory.tsx
nvie: Yep! I tried really hard doing it without it, but could not make it part of the same state/scope by definition, since all state will get blown away by the second render call in StrictMode. The only way to have a stable identifier is to keep that state outside of the component, unfortunately.

packages/liveblocks-core/src/client.ts
nvie: Hmm. Maybe yeah, I'd like that simplification. I'll give it a try!

nvie:
> Do you feel confident enough about this API or is there any risk that we'll find any edge cases (weird SSR behavior, `Strict` mode) that would force us to change it again?

I feel confident about this API, yes, and I don't anticipate that it would have to change again soon, no. It still passes all of our unit and E2E tests, for one thing. That said though, I still want to rewrite all of our examples to use this new pattern, which is on my TODO list before I'll merge this PR anyway. I think that will be the definitive test, but I don't expect any caveats here any longer. (That's why I marked the PR as "ready for review".)

packages/liveblocks-core/src/client.ts
nvie: Look at this [beautiful cleanup of the internals](https://github.com/liveblocks/liveblocks/commit/bf55fb4a73421a4c0f8ef7e999fd30ead425db74) — thanks for the suggestion! ❤️

packages/liveblocks-zustand/src/index.ts
nvie: As agreed after a Slack discussion, I changed the type definition to deprecate these overloads instead of removing the params, so technically it's no longer a breaking change: https://github.com/liveblocks/liveblocks/commit/0f3d3bb27ea025660365e386d09df28527872141

However, I'm honestly unsure if this leads to a better DX this way 🤔 I find the IDE hints not _that_ clear, and you have to dive into the type definitions to get the necessary information, and honestly the types aren't that clear. Maybe #staypragmatic here and not care too deeply in this case? Seeing a single/simple function definition without a room ID seems much clearer to see that it's no longer needed.

@GuillaumeSalles What do you think?

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -199,34 +200,85 @@ The second argument is a configuration for the presence and storage.
...
+
+### Client.enter
+
+Enters a room and returns its local `Room` instance.
+
+<Banner title="Deprecated since 1.5" type="warning">
+
+We no longer recommend using this API, because it will cause issues if you have
+multiple component trees in your application that connect to the same room.
+Prefer the new and improved [`client.enterRoom`][] API.
```
CTNicholas, line 226:
```suggestion
We no longer recommend this API, because issues will occur if you 
simultaneously connect to the same room with multiple instances.
We recommend using [`client.enterRoom`][] instead.
```

Wow, this is really hard to word! Does this work?

Please update this elsewhere on the page!

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -749,14 +801,40 @@ Get the storage status.
...
 room.reconnect();
 ```
 
+### Room.disconnect
+
+Disconnect the local room instance from the Liveblocks server. The room instance
+will remain functional (for example, it will still allow local presence or
+storage mutations), but since there is no connectivity any longer, those changed
+will not be persisted or synchronized until the room instance is reconnected to
+the server again.
```
CTNicholas, line 830:
```suggestion
Disconnect the local room instance from the Liveblocks server. The room instance
will remain functional (for example, it will still allow local presence or
storage mutations), but since it’s no longer connected, changes
will not be persisted or synchronized until the room instance is reconnected again.
```

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -179,14 +180,14 @@ two RoomProviders with the same room ID in your React tree.
   mutable by every client. Must either contain Live structures (e.g.
   `new LiveList()`, `new LiveObject({ a: 1 })`, etc.) or be serializable to
   JSON.
-- `shouldInitiallyConnect` (optional) - Whether or not the room should connect
-  to Liveblocks servers when the RoomProvider is rendered. By default equals to
+- `autoConnect` (optional) - Whether or not the room should immediately connect
```
CTNicholas, line 183:
```suggestion
- `autoConnect` (optional) - Whether or not the room should automatically connect
```
If we write "immediately" people might assume there's a delay sometimes, so switching for "automatically" here.

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -179,14 +180,14 @@ two RoomProviders with the same room ID in your React tree.
...
+  to Liveblocks servers when the RoomProvider is mounted. By default equals to
   `typeof window !== "undefined"`, meaning the RoomProvider tries to connect to
   Liveblocks servers only on the client side.
 
-The `initialPresence`, `initialStorage` and `shouldInitiallyConnect` props are
-ignored after the first render, so changes to the initial value argument won’t
-have an observable effect.
+The `initialPresence`, `initialStorage` and `autoConnect` props are ignored
+after the first render, so changes to the initial value argument won’t have an
+observable effect.
```
CTNicholas, line 190:
```suggestion
effect.
```
People don't need to know

CTNicholas (approved):
Left a couple of comments on the docs—you're always great at writing them @nvie!

I'll check the rest of the docs now.

docs/pages/api-reference/liveblocks-client.mdx
nvie: Love it!

docs/pages/api-reference/liveblocks-client.mdx
nvie: Thanks, this is much better! 👍

packages/liveblocks-react/src/factory.tsx
```diff
@@ -238,22 +235,64 @@ export function createRoomContext<
...
+    // StrictMode. See this issue https://github.com/facebook/react/issues/27103#issuecomment-1763359077
+    const instanceId = React.useRef(`p${++lastInstanceId}`).current; // Like useId(), but will also work in React<18
```
nimeshnayaju, line 276: This is maybe a nitpick and I know we do this in other places (e.g. [useInitial](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react/src/lib/use-initial.ts) hook)  too but React discourages reading the `ref.current` during rendering. 

https://react.dev/reference/react/useRef#caveats

> Do not write or read ref.current during rendering, except for [initialization.](https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents) This makes your component’s behavior unpredictable.

CTNicholas (approved):
Think I've updated it everywhere in the docs now 👍

packages/liveblocks-react/src/factory.tsx
nimeshnayaju: I don't think I understand this correctly. From my understanding, using an external variable with `useRef` also results in a different id during second render.

```tsx
let lastInstanceId = 0;

function Test() {
  const instanceIdRef = useRef(`p${++lastInstanceId}`);

  console.log(lastInstanceId, instanceIdRef.current);

  return null;
```

This also logs:
```
1 p1
2 p2
```

packages/liveblocks-react/src/factory.tsx
```diff
@@ -238,22 +235,64 @@ export function createRoomContext<
...
+
+  const RoomContext = React.createContext<TRoom | null>(null);
+
+  const roomCache = new Map<string, { room: TRoom; leave: () => void }>();
+
+  /**
+   * A cached version of .enterRoom(), where we cache each room/leave pair on
+   * a per-RoomProvider instance basis.
+   */
+  function stableEnterRoom(
```
nimeshnayaju, line 248: @nvie Do you mind explaining the significance of `roomCache` in more detail and how this functions differently from the `roomsById` variable inside [client.ts](https://github.com/liveblocks/liveblocks/pull/1226/files#diff-f7fafe86f1919de5119b3b0c680a8c12d5cd72b72c39cab803147a64faceb1e2) file?

nimeshnayaju:
The way I understand this new API works in summary is: a room can be entered (leased) multiple times by different parties (e.g. react, vanilla) and can be left, but the room is only destroyed (websocket is disconnected, etc) when the last party has left the room. I've definitely missed various details here. Is that the right way to think of this change @nvie ?

nimeshnayaju (approved):

packages/liveblocks-zustand/src/index.ts
GuillaumeSalles: Makes sense, let's just put it in the release notes

GuillaumeSalles (approved):

packages/liveblocks-react/src/factory.tsx
nvie: Correct! It's a little bit hard to explain, but I'll try.

At first I tried doing this:

```tsx
function Component() {
  const instanceIdRef = useRef(`p${++lastInstanceId}`);
  const [{room, leave}, /* setter */] = useState(() => stableEnterRoom(instaceIdRef, ...));
}
```

In StrictMode, among more things, "renders" (eg executing the functional component as a function) are performed twice, and `useState` hook initializer functions are called twice.

In this case, that means the following happens:

- 1st render of `Component()`:
  - `lastInstanceId` becomes `1`
  - `useRef` produces `"p1"`
  - `instanceIdRef` is `"p1"`
  - 1st execution of `useState` initializer calls: `stableEnterRoom("p1", ...)` -> result discarded
  - 2nd execution of `useState` initializer calls: `stableEnterRoom("p1", ...)` -> result returned as `{ room, leave }`
  - Let's call this `leave` function instance F1
  - **All local state and render result from this 1st render is now discarded**
- 2nd render of `Component()`:
  - `lastInstanceId` becomes `2`
  - `useRef` produces `"p2"`
  - `instanceIdRef` is `"p2"`
  - 1st execution of `useState` initializer calls: `stableEnterRoom("p2", ...)` -> result discarded
  - 2nd execution of `useState` initializer calls: `stableEnterRoom("p2", ...)` -> result returned as `{ room, leave }`
  - Let's call this `leave` function instance F2

At this point, we have a problem. The room is now "claimed" twice, and we have created two instances of `leave` unsubscribe callbacks (F1 and F2), but we haven't been able to "capture" the first instance (F1). Since we are not able to ever call F1, the room will never get freed, since it is never fully released. From the client's perspective, it's still in use.

---

Now let's move the generation of the ID to the outer component, and structure it like this:

```tsx
function Outer() {
  const instanceIdRef = useRef(`p${++lastInstanceId}`);
  return <Inner instanceIdRef={instanceIdRef} />;
}

function Inner({ instanceIdRef }) {
  const [{room, leave}, /* setter */] = useState(() => stableEnterRoom(instaceIdRef, ...));
}
```

- 1st render of `Outer()`:
  - `lastInstanceId` becomes `1`
  - `useRef` produces `"p1"`
  - Function produces JSX tree `<Inner instanceIfRef="p1" />`
  - **All local state and render result from this 1st render is now discarded**
- 2nd render of `Outer()`:
  - `lastInstanceId` becomes `2`
  - `useRef` produces `"p2"`
  - Function produces JSX tree `<Inner instanceIfRef="p2" />`

- 1st render of `Inner()` (called with "p2"):
  - 1st execution of `useState` initializer calls: `stableEnterRoom("p2", ...)` -> result discarded
  - 2nd execution of `useState` initializer calls: `stableEnterRoom("p2", ...)` -> result returned as `{ room, leave }`
  - Let's call this `leave` function instance F1
  - **All local state and render result from this 1st render is now discarded**
- 2nd render of `Inner()` (called with "p2"):
  - 1st execution of `useState` initializer calls: `stableEnterRoom("p2", ...)` -> result discarded
  - 2nd execution of `useState` initializer calls: `stableEnterRoom("p2", ...)` -> result returned as `{ room, leave }`
  - This is the same instance as F1

There now only exists exactly one instance of a `leave` function (F1). The key difference with the first approach is that the `Inner()` component will never get rendered with "p1" as the prop, and therefore this poses no issue in StrictMode.

In the second example's timeline, you can see that in total, 4 calls to `stableEnterRoom()` are being made, all with "p2" as the argument. The purpose of `stableEnterRoom()` is to cache the result of calling `enterRoom()` in the client so that when called for the same room ID (and by the same provider instance), that it will return the same, single `{ room, leave }` pair on every call. This is crucial in StrictMode, because if we did _not_ cache this result, it would create 4 `leave` functions/claims. This isn't problematic in and of itself of course, but… in StrictMode, we would not be able to capture 2 of those, because StrictMode would discard those results.

I realize this is complicated to explain, but hopefully this helps to understand and provides context!

packages/liveblocks-react/src/factory.tsx
nvie: Wow, this is new information to me! I don't understand why that would be discouraged, but thanks for bringing this to my attention! I'll dig a bit deeper tomorrow, and see what the recommended approach would be.

nvie:
> The way I understand this new API works in summary is: a room can be entered (leased) multiple times by different parties (e.g. react, vanilla) and can be left, but the room is only destroyed (websocket is disconnected, etc) when the last party has left the room.

Spot on!

packages/liveblocks-react/src/factory.tsx
nimeshnayaju: I believe the recommended way is to use `state` if something needs to be read or written to during rendering. There is slightly more information about this in the `Pitfall` section under [useRef documentation](https://react.dev/reference/react/useRef).

packages/liveblocks-react/src/factory.tsx
nimeshnayaju: Thank you for the detailed explanation! Feels slightly hacky to do this but I understand the need to do this. Thank you for ensuring the changes are thorough! Very much appreciated ❤️

nvie: I agree that it _is_ slightly hacky. One reason this must be done is because our `client.enter("my-room")` was pure before (it returned the same room however often you would call it with the same room ID argument). This is no longer the case. Our new `client.enterRoom("my-room")`, when called multiple times, will return the equivalent of this:

- `{ room: R1, leave: F1 }`
- `{ room: R1, leave: F2 }`
- `{ room: R1, leave: F3 }`
- `{ room: R1, leave: F4 }`

So even though the _room_ is stable, each call produces a new leave function (and thus the entire object returned is unstable). By caching these results, by doing that on a per-RoomProvider instance basis, we make the function pure again, which is what React expects.

nvie: Writing this comment just made me realize a thing — we don't have to keep the state at the factory-closure level outside of React and keep cache keys as `p<instance-id>:<room-id>`! Instead, we can replace the generation of the instance ID ref and instead keep a cache there, local to the RoomProvider instance, with cache keys of just `<room-id>`. This way, it's just kept in local React state, solves the same problem, and makes leaking memory less of a concern, should there be a bug somewhere (because if the component is unmounted, React will clear the state that keeps the cache). Much nicer!

This is simplified in a2978a2d20297e9cbbb3066ba8609dd41f235c06 now!

Thanks for asking the right questions here!

packages/liveblocks-react/src/factory.tsx
nvie: @nimeshnayaju I hope my explanation in [this comment](https://github.com/liveblocks/liveblocks/pull/1226#discussion_r1363332040) makes this easier to follow? I've also attempted to explain the rationale in the comment block in a2978a2d20297e9cbbb3066ba8609dd41f235c06.

PR #1233, opened 2023-10-16
Comments: CSS improvements and fixes

marcbouchenoire:
This PR improves and fixes various CSS-related things for `@liveblocks/react-comments`:

### Public
- Remove `@layer` and instead use `:where()` to control all specificities (`0,1,1` for pseudo-selectors and `0,1,0` for everything else)
   - We need to control specificities so that default styles can be overridden without having to use layers or `!important`
 - Move tokens from `:root` to `.lb-root`
   - Relying on `:root` created multiple issues: breaking the cascade (setting e.g. `--lb-accent` on anything below `<html>` didn't work as expected because the other dependant tokens were still coming from `<html>`, now the way to customize any token is to do it on `.lb-root`), not working well with shadow DOM, etc.
 - Fix reactions and links on Safari 

### Internal
- Replace Sass by PostCSS to simplify build pipeline
- Run Stylelint during build to avoid issues

### To do
- [x] Update CHANGELOG.md
- [x] Publish beta release
- [x] Test with beta
  - [x] Test @CTNicholas' example (from https://github.com/liveblocks/liveblocks.io/issues/1525)
  - [x] Test @nimeshnayaju's example (from https://github.com/nimeshnayaju/liveblocks-yjs-comments)
  - [x] Test on Windows

This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1603.
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1525.


CTNicholas (approved):
Nice work! Looks fixed

PR #1234, opened 2023-10-16
Comments: Fix `userIds` type in `resolveUsers`

marcbouchenoire:
This PR fixes an important issue where the `userIds` property within `ResolveUsersArgs` was incorrectly typed as `string` instead of `string[]`.

We use `URLSearchParams` in the examples and because that worked, we missed this obvious mistake earlier.


CTNicholas (approved):
Thanks for the quick fix, just saw Discord!

PR #1238, opened 2023-10-18
Comments examples: Add error boundaries

marcbouchenoire:
Since we're using Suspense in all examples, we should also catch errors.


CTNicholas (approved):

CTNicholas:
👍

PR #1240, opened 2023-10-18
Standardize Liveblocks base URL config

### LINKED ISSUE #1230: Standardize Liveblocks base URL config
We currently have these (hidden) config options:

- `liveblocksServer` (in client)
- `serverEndpoint` (in client, for Comments)
- `liveblocksBaseUrl` (in our Node client)
- `liveblocksAuthorizeEndpoint` (in our Node client)

These are all URL _paths_ for each specific use case, but they all typically must be in sync and point to the same server backend in order to work.

We should deprecate all these different config options and replace them with a single new one:

- `baseUrl: "https://api.liveblocks.io"`

…and then _derive_ all of these various concrete URLs from that base value.

### Context

See also, this [comment](https://github.com/liveblocks/liveblocks/pull/1223) and this [discussion](https://liveblocks.slack.com/archives/C0540DQAT9P/p1697137428655689) that followed it.

nvie:
Fixes #1230.

Unifies all of our various config options in the client and node packages, in favor of a single `baseUrl` config which defaults to the value `"https://api.liveblocks.io"` and from which all relevant paths are derived, whether those are the WebSocket address, the authorize endpoint, the comments API paths, etc.

## TODO

- [x] Remove all references of:
  - [x] `NEXT_PUBLIC_LIVEBLOCKS_SERVER`
  - [x] `LIVEBLOCKS_SERVER`
  - [x] `LIVEBLOCKS_AUTHORIZE_ENDPOINT`
- [x] Remove all options:
  - [x] `liveblocksBaseUrl`
  - [x] `liveblocksAuthorizeEndpoint`
  - [x] `liveblocksServer`
  - [x] `serverEndpoint`

After the 1.5 release, we can clean up these GitHub Action secrets/vars:
- [ ] Remove `E2E_TEST_LIVEBLOCKS_SERVER` secret
- [ ] Remove `E2E_TEST_LIVEBLOCKS_AUTHORIZE_ENDPOINT` secret
- [ ] Remove `E2E_TEST_LIVEBLOCKS_SERVER` var


packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -58,7 +58,7 @@ const defaultRoomConfig: RoomConfig = {
   roomId: "room-id",
   throttleDelay: THROTTLE_DELAY,
   lostConnectionTimeout: 99999,
-  liveblocksServer: "wss://live.liveblocks.io/v7",
+  baseUrl: "https://api.liveblocks.io",
```
jrowny, line 61: Should this use the same DEFAULT_BASE_URL from constants as __tests__/_utils.ts?

jrowny (approved):

packages/liveblocks-core/src/__tests__/room.test.ts
nvie: Yeah, looks like I missed it — thanks for catching that 👍

PR #1241, opened 2023-10-19
WIP: Make E2E tests for Comments a bit more robust

nvie:
Our Comments™ E2E tests can be pretty slow to run, sometimes causing the `waitForJson` call to timeout. The default timeout is set to 5 seconds, which should always be enough when expressing tests, IMHO. If you actually need to wait more than 5 seconds for something to happen, there probably is a better way to express the test.

This PR is an attempt to ensure the tests spend less time waiting, making them more robust, and faster at the same time.


e2e/next-sandbox/test/comments/index.test.ts
```diff
@@ -22,21 +22,25 @@ test.describe("Comments", () => {
...
 
+    await page1.click("#create-thread");
     await page2.click("#create-thread");
-    await waitForJson(page1, "#numOfThreads", numOfThreads + 1);
+    await page2.click("#create-thread");
+    await waitForJson(page1, "#numOfThreads", n + 3, { timeout: 15_000 });
 
     await page2.click("#delete-comment");
-    await waitUntilEqualOnAllPages(pages, "#numOfThreads");
+    await waitForJson(pages, "#numOfThreads", n + 3 - 1, { timeout: 15_000 });
```
nvie, line 44: I could have written `n + 2` here, but hopefully this makes it a bit more readable.

e2e/next-sandbox/test/comments/index.test.ts
```diff
@@ -22,21 +22,25 @@ test.describe("Comments", () => {
...
+    await waitForJson(pages, "#isLoading", false, { timeout: 15_000 });
+    await waitUntilEqualOnAllPages(pages, "#numOfThreads", { interval: 250 });
   });
 
   test("verify thread creation on B is broadcasted correctly to A", async () => {
     const [page1, page2] = pages;
 
-    await waitForJson(pages, "#isLoading", false);
+    await waitForJson(pages, "#isLoading", false, { timeout: 15_000 });
+    await waitUntilEqualOnAllPages(pages, "#numOfThreads", { interval: 250 });
```
nvie, line 33: This waitUntilEqualOnAllPages sync step is crucial, before reading the initial `n` value on the line below. If you read it before you've ensured they are in sync, then the initial value could be different, which causes the rest of the computation to get weird.

e2e/next-sandbox/test/comments/with-suspense.test.ts
```diff
@@ -22,21 +22,25 @@ test.describe("Comments", () => {
...
 
+    await page1.click("#create-thread");
     await page2.click("#create-thread");
-    await waitForJson(page1, "#numOfThreads", numOfThreads + 1);
+    await page2.click("#create-thread");
+    await waitForJson(pages, "#numOfThreads", n + 3, { timeout: 15_000 });
 
     await page2.click("#delete-comment");
-    await waitUntilEqualOnAllPages(pages, "#numOfThreads");
+    await waitForJson(pages, "#numOfThreads", n + 3 - 1, { timeout: 15_000 });
```
nvie, line 44: Hmm. Actually, even with a 15 second timeout, this line keeps failling. Maybe there is an actual issue here after all? It looks like it's pretty common that this comment deletion would not reach the other end of the line!

nvie: I'd encourage you to download the Playwright report ZIP file from CI and open it locally to inspect the full trace there. You can see all the details, including recorded network traffic, etc. Let me know if you need help doing that, happy to help.

nimeshnayaju: The tests seem to pass most of the times on my machine locally. But, the tests failing/passing is inconsistent on GitHub. Can't seem to understand properly why this may be happening

nvie: One way I found pretty effective to capture this and get a good trace of it locally is to run the following:

```fish
while env HEADLESS=1 turbo run test -- comments; end
```

This will run the test suite in a loop until it fails. This one just took me 3 runs to get:

Screen Shot 2023-10-20 at 08 38 31@2x

Maybe there is an actual bug here that it's trying to tell us about after all, instead of just infra slowness?

marcbouchenoire (approved):

PR #1242, opened 2023-10-19
Add new hook: `useOthersListener()`

nvie:
Fixes https://github.com/liveblocks/liveblocks.io/issues/1582.

First and foremost, this PR introduces a new React hook:

```tsx
function Component() {
  useOthersListener(({ type, user, others }) => {
    switch (type) {
      case "enter":
        // `user` has entered the room
        break;

      case "leave":
        // `user` has left the room
        break;

      case "update":
        // Presence for `user` was updated
        break;

      case "reset":
        // Others list was emptied
        break;
    }
  });
}
```

However, it does a few extra things.

### Type-level tests
We've sparingly used `tsd` before, but I've always wanted to add more of these, as they are an excellent way to ensure we're not accidentally changing out outer API if we make an internal refactoring and our outer types would reference them as a "DRY" way of avoiding duplication.

You can see I'm explicitly testing the APIs for the `room.subscribe("others", (others, event) => ...)` callback [here](https://github.com/liveblocks/liveblocks/blob/c1e56b03518dffcb7cb7df3a6db82394d7536421/packages/liveblocks-client/test-d/room-api.test-d.ts#L30-L50).

### Changing the "others" event type
I've used the opportunity here to make one final change to the `room.events.others` event hub type that I've always wanted to do to make it consistent with our other event types. Since this API isn't public/documented yet, I think this is fine to do now and I don't consider it a breaking change. This small change makes the API more intuitive and match all of our other event types better.

Here is the outer change of affected APIs, as a Liveblocks user would use them:

```tsx
// ❌ New/future/undocumented API (before this PR)
room.events.others.subscribe(({ event, others }) => { /* ... */ });

// New/future/undocumented API (after this PR)
room.events.others.subscribe(({ type, user, others }) => { /* ... */ });
//                            ^^^^^^^^^^^^^^^^^^^^^^
//                            This object itself is the `event` itself!

// ✅ New hook from this PR is consistent with that API
useOthersListener(({ type, user, others }) => { /* ... */ });

// ✅ The classic room.subscribe API is untouched (no breaking change)
room.subscribe("others", (others, event) => { /* ... */ });
//                        ^^^^^^^^^^^^^ NOTE!
//                        No unpacking! This has always been the only odd
//                        event that sends two args instead of one event
```

### DIFF
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -281,8 +281,9 @@ are:
 
 - `enter` – A user has entered the room.
 - `leave` – A user has left the room.
-- `reset` – The others list has been emptied. This is the first event that occurs
-  when the room is entered. It also occurs when you’ve lost connection to the room.
+- `reset` – The others list has been emptied. This is the first event that
+  occurs when the room is entered. It also occurs when you’ve lost connection to
+  the room.
 - `update` – A user’s presence data has been updated.
 
 ```ts
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -276,7 +276,7 @@ The possible value are: `initial`, `connecting`, `connected`, `reconnecting`, or
 
 ### useOthersListener
 
-Calls the given callback when an "others" event happens. Possible event types
+Calls the given callback when an “others” event occurs. Possible event types
 are:
 
 - `enter` – A new User entered the room. The others list grew by one.
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -302,7 +302,7 @@ function App() {
         break;
 
       case "reset":
-        // Others list was emptied
+        // Others list has been emptied
         break;
     }
   });
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -298,7 +298,7 @@ function App() {
         break;
 
       case "update":
-        // Presence for `user` was updated
+        // Presence for `user` has updated
         break;
 
       case "reset":
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -287,22 +287,23 @@ are:
 - `update` – Presence data of one of the Users was updated.
 
 ```ts
-import { toast } from "my-preferred-toast-library";
-
 function App() {
   useOthersListener(({ type, user, others }) => {
     switch (type) {
       case "enter":
-        toast.info(`${user.info.name} has joined the room`);
+        // `user` has entered the room
         break;
 
       case "leave":
-        toast.info(`${user.info.name} has left the room`);
+        // `user` has left the room
+        break;
+
+      case "update":
+        // Presence for `user` was updated
         break;
 
       case "reset":
-      case "update":
-        /* Do nothing */
+        // Others list was emptied
         break;
     }
   });
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -274,6 +274,41 @@ const status = useStatus();
 The possible value are: `initial`, `connecting`, `connected`, `reconnecting`, or
 `disconnected`.
 
+### useOthersListener
+
+Calls the given callback when an "others" event happens. Possible event types
+are:
+
+- `enter` – A new User entered the room. The others list grew by one.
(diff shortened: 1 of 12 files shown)


CTNicholas (approved):
Great work @nvie! I like this hook 🙂. I think everything looks good, I've just reworded a couple of things to make them sound more like I wrote them (my way of trying to keep our docs consistent).

packages/liveblocks-react/src/factory.tsx
```diff
@@ -550,6 +551,19 @@ export function createRoomContext<
...
+  function useOthersListener(
+    callback: (event: OthersEvent<TPresence, TUserMeta>) => void
+  ) {
+    const room = useRoom();
+    const savedCallback = useLatest(callback);
+
+    React.useEffect(
+      () =>
+        room.events.others.subscribe((event) => savedCallback.current(event)),
+      [room, savedCallback]
```
nimeshnayaju, line 563: **Nitpick:** I am guessing we save this callback (as `savedCallback`) so that the effect doesn't rerun when the callback changes. But, I think in this case, it might be okay (not a big deal) to let the effect rerun when the callback changes. Also, I do not expect this callback to change on every render, and if it does become a bother for user that this effect is rerunning multiple times, they could apply this optimization on their own end (i.e., save the callback themselves before passing to the `useOthersListener` hook). 

And, I know it's okay to keep using `useLatest` here but I see the opportunity to use less hooks, which would mean less things to understand for someone new joining or looking at the code.

nimeshnayaju (approved):
🚀🚀🚀🚀

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -274,6 +274,42 @@ const status = useStatus();
 The possible value are: `initial`, `connecting`, `connected`, `reconnecting`, or
 `disconnected`.
 
+### useOthersListener
+
+Calls the given callback when an "others" event happens. Possible event types
+are:
```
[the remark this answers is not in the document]
nvie: Shoot! I tried, I swear!

nvie:
> I've just reworded a couple of things to make them sound more like I wrote them (my way of trying to keep our docs consistent).

@CTNicholas Thanks for the improvements! 🙏  I'll apply all the suggestions now.
Absolutely feel free to make these edits directly in the MDX files next time and commit them to the branch. I find the GitHub UI for applying these suggestions rather annoying, and at the end you often need to re-run Prettier anyway.

packages/liveblocks-react/src/factory.tsx
nimeshnayaju: Thanks for the clarification @nvie . It is a difficult balance to make. I think this experimental hook will help solve some of the issues once released and I think you might find it interesting - [useEffectEvent](https://react.dev/learn/separating-events-from-effects)

nvie: Yeah, I recently read about it! Not sure if I'm a fan of the API yet. I definitely see the problem it solves, but I also think using our `useLatest` construct, and relying on `ref.current` isn't that much worse. At least there are fewer caveats and special cases with this. No new concepts to learn if you already understand callbacks, effects, and refs.

What worries me a bit is how specially the result of `useEffectEvent` is (going to be) treated. It should never be used outside of an effect (i.e. doing `setTimeout(onSomething, 100)` are considered wrong, and linters will have to start catching that).

Also, the `exhaustive-deps` ESLint rule will treat it specially, as if it's a normal callback function you would have to put in the dep array, except… if the variable is the result of a call to `useEffectEvent`, you may not. I think with custom hooks already, but with more special cases like this, `exhaustive-deps` will stop scaling over time.

Comparing their example:

```tsx
function ChatRoom({ roomId, theme }) {
  const onConnected = useEffectEvent(() => {
  //                  ^^^^^^^^^^^^^^ 1️⃣
    showNotification('Connected!', theme);
  });

  useEffect(() => {
    const connection = createConnection(serverUrl, roomId);
    connection.on('connected', () => {
      onConnected();
//    ^^^^^^^^^^^^^ 2️⃣
    });
    connection.connect();
    return () => connection.disconnect();
  }, [roomId]);
  // ...
```

vs

```tsx
function ChatRoom({ roomId, theme }) {
  const onConnected = useLatest(() => {
  //                  ^^^^^^^^^ 1️⃣
    showNotification('Connected!', theme);
  });

  useEffect(() => {
    const connection = createConnection(serverUrl, roomId);
    connection.on('connected', () => {
      onConnected.current();
//    ^^^^^^^^^^^^^^^^^^^ 2️⃣
    });
    connection.connect();
    return () => connection.disconnect();
  }, [roomId, /* (optionally) onConnected */]);
//            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 3️⃣
  // ...
```

To me I don't think we're that far apart. But maybe I'm missing something here, so take it with a grain of salt.

Issue #1000, opened 2023-07-13
"Cannot set parent: node already has a parent"


nvie:
I created this ticket as a reference for anyone running into the following error:

```
Uncaught Error: Cannot set parent: node already has a parent
```

![image](https://github.com/liveblocks/liveblocks/assets/83844/000d07e3-4bcf-42e3-a18c-f920329597df)

## The issue

You can run into this runtime error when using `@liveblocks/client` or `@liveblocks/react`. This error is an internal integrity check that should never be surfaced to Liveblocks users, yet some people have seen it. It typically happens when you pass `initialStorage` like so:

```tsx
const initialStorage = {
  animals: new LiveList(["🦊", "🐍", "🐻‍❄️"])
}

// If you use @liveblocks/client
client.enter(..., {
  initialStorage,
});

// If you use @liveblocks/react
<RoomProvider id="..." initialStorage={initialStorage}>...</RoomProvider>
```

The cause of the problem is that, on line 1, `initialStorage` is a global instance that instantiates the new LiveList (or LiveObject, or LiveMap) only once, and keeps a reference to the same instance, which gets reused between rooms when you leave/re-enter the room. This is problematic, because every Live data structure can only be bound to a room once, after that it "belongs" to that room instance. An attempt to reuse it in a different room later will fail — hence the "already has a parent" error message.

## The fix

If you run into this issue, the fix is to pass a _function_ as the `initialStorage` value, so new/unbound instances are created every time:

```tsx
// ✅ Turn this into a factory!
const initialStorageFn = () => ({
  animals: new LiveList(["🦊", "🐍", "🐻‍❄️"])
});

// If you use @liveblocks/client
client.enter(..., {
  initialStorage: initialStorageFn,
});

// If you use @liveblocks/react
<RoomProvider id="..." initialStorage={initialStorageFn}>...</RoomProvider>
```

## The real fixes

- [ ] We should make the error message clearer and less obtuse
- [ ] We should recommend the factory pattern in our documentation and examples by default
- [ ] Maybe even deprecate passing a value directly, and always requiring a function?

Issue #560, opened 2022-11-10
[BUG] Cannot set parent: node already has a parent


chaitanyya:
**Describe the bug**

We use Liveblocks on a react app with live `text input` updates. The initial contents of this `text input` are fetched from a param value in the URL.
**Eg.**
`/play/<text-input-id>`

This `text-input-id` is then used to retrieve associated data from the database. There is also a dropdown input that lets you select valid `text-input-id` on selecting, we update the route history as well `/play/<new-text-input-id>`

Everything works as expected, however, there is a strange edge case where Liveblock fails with the following error:

```
index.js:194 Uncaught Error: Cannot set parent: node already has a parent
    at LiveObject._setParentLink (index.js:194:1)
    at LiveObject.update (index.js:1953:1)
    at LiveObject.set (index.js:1849:1)
    at createOrUpdateRootFromMessage (index.js:2747:1)
    at index.js:3247:1
    at doNotBatchUpdates (index.js:2637:1)
    at WebSocket.onMessage (index.js:3205:1)
```

**To Reproduce**

1. Got to `/play/<text-input-id>`
2. Inspect that the initial data is loaded correctly.
3. Change the ID from the dropdown to a new ID. The route is now updated to `/play/<new-text-input-id>`
4. Inspect that the initial data is loaded correctly.
5. Switch back to the old ID. The route is now updated to `/play/<text-input-id>`
6. Check the console for the above error.

**Expected behavior**

The input value should be updated to the old value fetched from the database.

**Environment (please complete the following information):**

- react - v18.2.0
- @liveblocks/client - ^0.18.5",
- @liveblocks/react - ^0.18.5",

Happy to provide any additional context to help understand the root cause here.

nvie:
Hello @chaitanyya — thanks for writing up this bug report. That error message is an internal error that should never happen under normal circumstances. We'd be interested in understanding how this code path gets triggered as much as you do.

To that end, do you happen to have a small reproduction of this issue, for example, a minimal example that showcases this bug on, say, [Codesandbox](https://codesandbox.io/), by any chance? Once we have a way to replicate it, it'll be easier for us to take a deeper look. Thank you so much!

chaitanyya:
Thanks for the update; I'll prepare a minimal example based on our implementation and share it asap.

chaitanyya:
Hey, I may have found the issue here. What is happening is the `room_id` changes when the route changes and a new id is provided to `<RoomProvider />` component.

[...]

PR #1245, opened 2023-10-20
Add `.clone()` method to Live structures

nvie:
Fixes https://github.com/liveblocks/liveblocks.io/issues/1594, #560, and #1000.

This adds a `.clone()` method to `LiveMap`, `LiveList`, and `LiveObject`, to make a deep clone of those values that are not attached to the pool, and thus are suitable for injection into another document, or in a different part of the Storage tree. Previously, users have been using this community-provided [code snippet](https://discord.com/channels/913109211746009108/913157273046646795/944290231065149461).

This method can be useful to implement features like "duplicate selected layers" in a drawing app, for example.

Usage:

```tsx
const shapes = root.get("shapes");
const shape1 = shapes.get("shape1");
shapes.set("shape2", shape1.clone());
```

## TODO

- [x] [~Only clone when needed, and issue a warning when we do~](https://github.com/liveblocks/liveblocks/pull/1245/files#r1367162560) **Update:** decided not to
- [x] Unit test this
- [x] Check if this can be a good default solution to avoid #1000
- [x] Documentation

### DIFF
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -959,7 +959,7 @@ describe("room", () => {
     const initialPresence = {};
     const initialStorage = {
       list: new LiveList([13, 42]),
-      map: new LiveMap([["a", 1]]),
+      map: new LiveMap([["a" as string, 1 as number]]),
       obj: new LiveObject({ x: 0 }),
     };
 
@@ -987,7 +987,18 @@ describe("room", () => {
       });
       room.connect();
       try {
-        await room.getStorage();
+        const { root } = await room.getStorage();
+        expect(root.toImmutable()).toEqual({
+          list: [13, 42],
+          map: new Map([["a", 1]]),
+          obj: { x: 0 },
+        });
+
+        // Now start mutating these Live structures (the point being that those
+        // mutations should not be observable in the second loop)
+        root.get("list").push(13);
+        root.get("map").set("b", 2);
+        root.get("obj").set("x", 42);
       } finally {
         room.destroy();
       }
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -1,4 +1,5 @@
 import { nn } from "..";
+import { LiveMap } from "../crdts/LiveMap";
 import type { AuthValue } from "../auth-manager";
 import { StopRetrying } from "../connection";
 import { DEFAULT_BASE_URL } from "../constants";
@@ -952,6 +953,45 @@ describe("room", () => {
     expect(storage.root.toObject()).toEqual({ foo: 1234 });
     //                                        ^^^ Added by the client, from initialStorage
     expect(room.history.canUndo()).toBe(false);
+  });
+
+  test("missing storage keys are properly initialized using initialStorage, even when they are already part of the storage tree", async () => {
+    const initialPresence = {};
+    const initialStorage = {
+      list: new LiveList([13, 42]),
+      map: new LiveMap([["a", 1]]),
+      obj: new LiveObject({ x: 0 }),
+    };
+
+    // Create the same room instance twice (using the same, global,
+    // initialStorage instance)
+    for (let i = 0; i < 2; i++) {
+      const { room, wss } = createTestableRoom(
+        initialPresence,
+        undefined,
+        undefined,
+        undefined,
+        initialStorage
+      );
+      wss.onConnection((conn) => {
+        conn.server.send(
+          serverMessage({
+            type: ServerMsgCode.INITIAL_STORAGE_STATE,
+            items: [["root", { type: CrdtType.OBJECT, data: {} }]],
+            //                                              ^^
+            //                                   NOTE: Storage is initially
+            //                                   empty, so initialStorage keys
+            //                                   will get added
+          })
+        );
+      });
+      room.connect();
+      try {
+        await room.getStorage();
+      } finally {
+        room.destroy();
+      }
+    }
   });
 
   test("storage should be initialized properly", async () => {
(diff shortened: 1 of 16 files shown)


packages/liveblocks-core/src/crdts/LiveList.ts
```diff
@@ -21,26 +21,10 @@ import type { LiveNode, Lson } from "./Lson";
...
-    }
-  | {
-      index: number;
-      item: Lson;
-      type: "set";
-    };
+  | { type: "insert"; index: number; item: Lson }
+  | { type: "delete"; index: number }
+  | { type: "move"; index: number; previousIndex: number; item: Lson }
+  | { type: "set"; index: number; item: Lson };
```
nvie, line 27: Just formatting, no changes made here.

packages/liveblocks-core/src/crdts/LiveMap.ts
```diff
@@ -52,13 +52,12 @@ export class LiveMap<
...
-      const mappedEntries: Array<[TKey, LiveNode]> = [];
-      for (const entry of entries) {
-        const value = lsonToLiveNode(entry[1]);
-        value._setParentLink(this, entry[0]);
-        mappedEntries.push([entry[0], value]);
+      const mappedEntries: [TKey, LiveNode][] = [];
+      for (const [key, value] of entries) {
+        const node = lsonToLiveNode(value);
+        node._setParentLink(this, key);
+        mappedEntries.push([key, node]);
```
nvie, line 59: No changes made here, just stylistic.

packages/liveblocks-core/src/crdts/__tests__/_arbitraries.ts
```diff
@@ -0,0 +1,33 @@
...
+    liveMap: fc
+      .array(fc.tuple(key, tie("lson")))
+      .map((pairs) => new LiveMap(pairs as [string, Lson][])),
+    liveObject: fc
+      .array(fc.tuple(key, tie("lson")))
+      .map(
+        (pairs) => new LiveObject(Object.fromEntries(pairs as [string, Lson][]))
+      ),
+  })
+);
```
nvie, line 33: These fast-check arbitraries will generate arbitrarily deep recursive LSON trees.

packages/liveblocks-core/src/crdts/__tests__/cloning.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+          // Clone "a" to "b"
+          root.set("a", data);
+          root.set("b", data.clone());
+
+          const imm = root.toImmutable();
+          expect(imm.a).toEqual(imm.b);
+        }
+      )
+    ));
```
nvie, line 84: Here's the simplest test explained:

First, we use the `liveStructure` Arbitrary (see above) to generate randomized LiveMap/LiveList/LiveObject structures. (Some deep, some shallow, some simple, some complex, etc.)

Then, we inject it into the root, under key `"a"`. Next, we clone the value and inject it under key "b". Note that inserting the same value into the tree in different places is illegal and would lead to an error, e.g.

```tsx
root.set("a", data);
root.set("b", data);  // ❌ Would error!
```

But not when you clone it—then it's fine:

```tsx
root.set("a", data);
root.set("b", data.clone());  // ✅ Fine!
```

Finally, we check that the cloned tree will produce the same immutable value for keys "a" and "b".

packages/liveblocks-core/src/room.ts
```diff
@@ -1341,7 +1342,7 @@ export function createRoom<
     const stackSizeBefore = context.undoStack.length;
     for (const key in context.initialStorage) {
       if (context.root.get(key) === undefined) {
-        context.root.set(key, context.initialStorage[key]);
+        context.root.set(key, cloneLson(context.initialStorage[key]));
```
nvie, line 1345: @GuillaumeSalles What do you think of doing this when initializing `initialStorage`? This would avoid the error from #1000 from ever showing up, even if people set it up using a global instance accidentally. I think this is pretty nice, but not sure if maybe you see a caveat with this?

packages/liveblocks-core/src/__tests__/room.test.ts
```diff
@@ -954,6 +955,45 @@ describe("room", () => {
...
+        );
+      });
+      room.connect();
+      try {
+        await room.getStorage();
+      } finally {
+        room.destroy();
+      }
+    }
+  });
```
nvie, line 995: This test fails on `main`, but passes with [this fix](https://github.com/liveblocks/liveblocks/pull/1245/files#diff-582bfabb50c03434ec253eb453124d78711e62f2afac7270fd5813ab638f90cf) applied.

packages/liveblocks-core/src/room.ts
nvie: We just chatted about this, and decided to change this slightly to the following behavior:

- We do a quick check to see if `context.initialStorage[key]` is a Live structure
  - If so, we check to see if it has a parent node set (i.e. it's part of a tree already)
     - If it has a parent, we `.clone()` it, and issue a warning that this can be indicative of a bug in the way the app is configured, and that the user should always use fresh/new instances
     - If not, great, just use it
  - If not, it's a JSON value, so no need to clone anything

nvie: Will implement this next week.

jrowny (approved):
looks good!

packages/liveblocks-core/src/room.ts
nvie: Now that I started implementing this, I'm having second thoughts about this solution, because doing the optional cloning leads to situations that are maybe even more confusing, potentially.

Here's such a footgun:

- First, empty/new room A is entered with `initialStorage={initialStorage}`
- Keys are missing, so the global `initialStorage` values are assigned, no cloning
- User now just uses the room, mutating things as they go, populating the document with whatever content. Since no clone wat taken, this will also mutate the global `initialStorage` instance (unintentionally)
- Now, the user joins empty/new room B with `initialStorage={initialStorage}`
- Keys are missing, but since global `initialStorage` values are detected to already have a parent, a clone is taken, but… the clone will now contain all the mutations made in room A!

This makes me conclude that the optional cloning tactic is maybe the worst solution even.

Therefore, I think that always cloning (i.e. what this PR implemented) is still be best pragmatic solution that leads to the fewest surprises — and reflects best how the prop is intended. We can document that whatever you pass as `initialStorage` will be considered a "template".

cc @GuillaumeSalles

PR #1249, opened 2023-10-20
Comments: Prevent nested `lb-root` classes

marcbouchenoire:
Now that tokens are declared on `.lb-root`, having nested `.lb-root` classes means that overriding some tokens on `.lb-thread` for example, won't affect `<Comment />` and `<Composer />` since they also have `.lb-root` classes, which will reset the tokens, that's not great. This PR removes `.lb-root` from components when they are nested.

**Before:**
```
Thread (.lb-root)
├── Comment (.lb-root)
│   └── Composer (.lb-root)
└── Composer (.lb-root)
Comment (.lb-root)
Composer (.lb-root)
```

**After:**
```
Thread (.lb-root)
├── Comment
│   └── Composer
└── Composer
Comment (.lb-root)
Composer (.lb-root)
```

### DIFF
--- a/packages/liveblocks-react-comments/src/components/Comment.tsx
+++ b/packages/liveblocks-react-comments/src/components/Comment.tsx
@@ -109,6 +109,11 @@ export interface CommentProps extends ComponentPropsWithoutRef<"div"> {
   /**
    * @internal
    */
+  root?: boolean;
+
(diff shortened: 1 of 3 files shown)


CTNicholas (approved):
Looks good 👍

PR #1259, opened 2023-10-25
Upgrade all dev dependencies

nvie:
This PR upgrades all dev dependencies (linters, build tools, etc). I went through the same hoops earlier today while prepping extraction of the server library in the Cloudflare repo, and figured I might just backport these changes here too.

### DIFF
--- a/packages/liveblocks-yjs/package.json
+++ b/packages/liveblocks-yjs/package.json
@@ -27,7 +27,7 @@
     "build": "tsup && cp dist/index.d.ts dist/index.d.mts",
     "format": "eslint --fix src/; prettier --write src/",
     "lint": "eslint src/",
-    "lint:package": "publint --strict && attw --pack",
+    "lint:package": "publint --strict && attw --pack --ignore-rules cjs-only-exports-default",
     "test": "jest --silent --verbose --color=always",
     "test:watch": "jest --silent --verbose --color=always --watch"
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -559,10 +559,11 @@
       }
     },
     "node_modules/@babel/plugin-syntax-jsx": {
-      "version": "7.21.4",
-      "license": "MIT",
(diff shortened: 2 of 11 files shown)


packages/liveblocks-yjs/package.json
```diff
@@ -27,7 +27,7 @@
     "build": "tsup && cp dist/index.d.ts dist/index.d.mts",
     "format": "eslint --fix src/; prettier --write src/",
     "lint": "eslint src/",
-    "lint:package": "publint --strict && attw --pack",
+    "lint:package": "publint --strict && attw --pack --ignore-rules cjs-only-exports-default",
```
nvie, line 30: @jrowny This is an annoying one. By upgrading `@arethetypeswrong/cli`, it started issuing this new warning. While `@liveblocks/yjs` is structured like pretty much all other packages, it uses a default export:

```ts
import LiveblocksProvider from "@liveblocks/yjs";
```

The warning points to [an extensive description of the (potential) issue](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/CJSOnlyExportsDefault.md) and a [precise table](https://andrewbranch.github.io/interop-test/#synthesizing-default-exports-for-cjs-modules) to point out when this would be an issue.

I did not really dig in to understand, and decided to just ignore this warning for now, but… could we not just avoid using default exports here and just rely on named exports, like we do for all other packages? These potential headaches/support issues/caveats might just be completely avoided if we simply switch to this instead:

```ts
import { LiveblocksProvider } from "@liveblocks/yjs";
```

marcbouchenoire (approved):

packages/liveblocks-yjs/package.json
nvie: Going to ignore this for now and get this PR merged, since really nothing changed here except we now know this may cause issues, but would love to hear your thoughts on this, @jrowny.

PR #1261, opened 2023-10-26
Fix `backgroundKeepAliveTimeout` option bounds check

nvie:
Fixes this issue, where a valid option value would get rejected, due to a logic bug in the bounds check:

CleanShot 2023-10-26 at 09 20 00


marcbouchenoire (approved):

packages/liveblocks-core/src/client.ts
```diff
@@ -425,8 +425,7 @@ function checkBounds(
   if (
     typeof value !== "number" ||
     value < min ||
-    max === undefined ||
-    value > max
+    (max !== undefined && value > max)
```
nvie, line 428: 🙈

packages/liveblocks-core/src/client.ts
```diff
@@ -20,7 +20,7 @@ const MIN_THROTTLE = 16;
 const MAX_THROTTLE = 1_000;
 const DEFAULT_THROTTLE = 100;
 
-const MIN_BACKGROUND_KEEP_ALIVE_TIMEOUT = 60_000;
+const MIN_BACKGROUND_KEEP_ALIVE_TIMEOUT = 15_000;
```
nvie, line 23: 🙈🙈

PR #1262, opened 2023-10-27
Add comments back to starter kit

CTNicholas:


marcbouchenoire (approved):
`v1.5.2` is now out so you can update to fix the `resolveUsers` issue.

PR #1263, opened 2023-10-27
Fix `resolveUsers` return type

marcbouchenoire:
This PR fixes the return type of `resolveUsers` which wasn't updated to an array when it changed from `resolveUser` to `resolveUsers`.


CTNicholas (approved):
Thanks!

PR #1266, opened 2023-10-27
Remove CreateRootObjectOp

GuillaumeSalles:
We're no longing supporting the ws protocol that relied on `CreateRootObjectOp`, so we can safely remove it.

### DIFF
--- a/packages/create-liveblocks-app/package.json
+++ b/packages/create-liveblocks-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-liveblocks-app",
-  "version": "1.5.2",
+  "version": "1.5.3-test1",
   "description": "An installer for Liveblocks projects, including various examples and a Next.js starter kit. Liveblocks is the all-in-one toolkit to build collaborative products like Figma, Notion, and more.",
   "license": "Apache-2.0",
   "bin": "dist/index.cjs",
--- a/packages/liveblocks-core/src/crdts/liveblocks-helpers.ts
+++ b/packages/liveblocks-core/src/crdts/liveblocks-helpers.ts
@@ -208,18 +208,18 @@ export function getTreesDiffOperations(
           });
           break;
         case CrdtType.OBJECT:
-          ops.push(
-            crdt.parentId
-              ? {
-                  type: OpCode.CREATE_OBJECT,
-                  id,
-                  parentId: crdt.parentId,
-                  parentKey: crdt.parentKey,
-                  data: crdt.data,
-                }
-              : // Root object
-                { type: OpCode.CREATE_OBJECT, id, data: crdt.data }
-          );
+          if (crdt.parentId === undefined || crdt.parentKey === undefined) {
+            throw new Error(
+              "Internal error. Cannot serialize storage root into an operation"
+            );
+          }
+          ops.push({
+            type: OpCode.CREATE_OBJECT,
+            id,
+            parentId: crdt.parentId,
+            parentKey: crdt.parentKey,
+            data: crdt.data,
+          });
           break;
         case CrdtType.MAP:
           ops.push({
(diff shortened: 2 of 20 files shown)


packages/liveblocks-core/src/protocol/Op.ts
```diff
@@ -24,7 +24,7 @@ export type Op =
   | SetParentKeyOp // Only for lists!
   | DeleteObjectKeyOp;
 
-export type CreateOp = CreateRootObjectOp | CreateChildOp;
+export type CreateOp = CreateChildOp;
```
nvie, line 27: Since `CreateChildOp` only existed to exclude the root op type, we can also rename all instances to `CreateOp` now, as all members left have a parent now 👌

GuillaumeSalles: 👍

packages/create-liveblocks-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "create-liveblocks-app",
-  "version": "1.5.2",
+  "version": "1.5.3-test1",
```
nvie, line 3: Side-note: I find these auto-bumps on PRs a little bit annoying. If you do multiple of them, they make merging back to main harder, especially when there are multiple PRs that are issuing test releases. Maybe we should change the job to not push those bump commits to the PR. Instead, it can create that commit, but not on a branch, tag that (dangling) commit, and publish it. Same thing, except the PR isn't polluted.

packages/liveblocks-core/src/crdts/liveblocks-helpers.ts
```diff
@@ -208,18 +208,18 @@ export function getTreesDiffOperations(
...
-                  data: crdt.data,
-                }
-              : // Root object
-                { type: OpCode.CREATE_OBJECT, id, data: crdt.data }
-          );
+          if (crdt.parentId === undefined || crdt.parentKey === undefined) {
+            throw new Error(
+              "Internal error. Cannot serialize storage root into an operation"
+            );
+          }
```
nvie, line 215: Any reason to add this runtime check here? TypeScript should have us covered, I think. Won't our linter nag about these checks being useless?

nvie (approved):
So fresh, so clean! ✨ 🧼 ✨

nvie:
@GuillaumeSalles Btw, if you're going to publish this an an internal change, would you mind cherry-picking [this commit](https://github.com/liveblocks/liveblocks/commit/f75b090f66958bb4c9d4f402b9717a7611e8b8f8) into this branch, and include that? Just two tiny little utils that I'm using in multiple projects now. Would be nice to have in `@liveblocks/core` as well.

packages/create-liveblocks-app/package.json
GuillaumeSalles: I agree

packages/liveblocks-core/src/crdts/liveblocks-helpers.ts
GuillaumeSalles: `SerializedRootObject` does not have a `parentId` or a `parentKey`, but it can never be sent from the client. The root is initialized from the server. That's why we never have a `CreateRootObjectOp`.

GuillaumeSalles:
> @GuillaumeSalles Btw, if you're going to publish this an an internal change, would you mind cherry-picking [this commit](https://github.com/liveblocks/liveblocks/commit/f75b090f66958bb4c9d4f402b9717a7611e8b8f8) into this branch, and include that? Just two tiny little utils that I'm using in multiple projects now. Would be nice to have in @liveblocks/core as well.

Done! I didn't cherry pick though, I updated the files manually

nvie:
> _Done! I didn't cherry pick though, I updated the files manually_

🙏

PR #1271, opened 2023-10-30
Y subdoc

jrowny:
Adds subdoc support to Yjs package


packages/liveblocks-yjs/src/doc.ts
```diff
@@ -0,0 +1,106 @@
+import { detectDupes } from "@liveblocks/core";
+import { Base64 } from "js-base64";
+import { Observable } from "lib0/observable";
+import * as Y from "yjs";
+
+import { PKG_FORMAT, PKG_NAME, PKG_VERSION } from "./version";
+
+detectDupes(PKG_NAME, PKG_VERSION, PKG_FORMAT);
```
jrowny, line 8: this only needs to be on index.ts, right @nvie ?

GuillaumeSalles: My understanding is that it should only be in `index.ts`

GuillaumeSalles (approved):

e2e/next-sandbox/test/utils.ts
```diff
@@ -154,6 +154,25 @@ export async function waitForJson(
...
+  expectedValue: string,
+  options?: { timeout?: number }
+) {
+  const pages = Array.isArray(oneOrMorePages)
+    ? oneOrMorePages
+    : [oneOrMorePages];
+
+  return Promise.all(
+    pages.map((page) =>
+      expect(page.locator(selector)).toContainText(expectedValue, {
```
nvie, line 169: `.toContainText` will do substring matching, whereas `.toHaveText` will do ful string matching. Typically I prefer the latter because it's more strict, but are you maybe using substring matching here deliberately for subdocs?

packages/liveblocks-yjs/src/awareness.ts
```diff
@@ -0,0 +1,125 @@
...
+  lastUpdated: number;
+};
+
+/**
+ * This class will store Yjs awareness in Liveblock's presence under the __yjs key
+ * IMPORTANT: The Yjs awareness protocol uses ydoc.clientId to reference users
+ * to their respective documents. To avoid mapping Yjs clientIds to liveblock's connectionId,
+ * we simply set the clientId of the doc to the connectionId. Then no further mapping is required
+ */
+export class Awareness extends Observable<unknown> {
```
nvie, line 28: Didn't read this code closely, because I assume they're just code moves? Lmk if there are any changes in here I should pay attention to.

e2e/next-sandbox/test/ydoc/subdoc.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    await waitForJson(pages, "#sync", true);
+
+    // clear text from page 1
+    await page1.click("#clear");
+    await waitForJson(pages, "#text", "{}");
+
+    // insert subdoc on page 2
+    await page2.click("#insert");
+
+    await page1.waitForTimeout(500); // todo replace with waiting for subdoc to load
```
nvie, line 41: Yeah, it will be nicer to have an API or observable property to check if a specific (individual) subdoc has been loaded. That will be a nice API to use for users anyway, in building real apps.

shared/eslint-config/index.js
```diff
@@ -29,6 +29,7 @@ module.exports = {
     "@typescript-eslint/no-inferrable-types": "off",
     "no-constant-condition": "off",
     "@typescript-eslint/no-non-null-assertion": "off", // Because we have a custom no-restricted-syntax rule for this
+    "@typescript-eslint/no-base-to-string": "off", // Too many false positives for Yjs objects
```
nvie, line 32: I've centrally disabled this lint rule for all of our projects (in `shared/eslint-config`). I previously also added a couple of these ignores on a case-by-case basis, but the lint rule isn't smart enough to not report these for Y objects, where this is not an issue, so it will just be annoying and not actually helpful.

nvie (approved):
Amazing! 😍 Also thanks for adding these E2E tests – loved seeing them, and they will be super helpful. I'll take a look at the CI + packaging issues, and get this published.

packages/liveblocks-yjs/src/doc.ts
nvie: Correct! It must be loaded exactly once, at the root of the module.

e2e/next-sandbox/test/comments/with-suspense.test.ts
```diff
@@ -11,6 +11,12 @@ import {
 
 const TEST_URL = "http://localhost:3007/comments/with-suspense";
 
+// These tests sometimes fail on CI, because some operations aren't coming
+// through timely enough (or not at all) on the other end. We're still figuring
+// out what's the cause of this.
+// eslint-disable-next-line @typescript-eslint/unbound-method
+const skipOnCI = process.env.CI ? test.skip : test;
```
nvie, line 18: When I skipped this test on CI (in #1241), apparently I forgot to make this change in this test file too, and since the test just happened to pass, I missed it. I should have updated this file as part of that PR, so including it now here, although unrelated to this change.

PR #1280, opened 2023-11-02
Log details about `FORBIDDEN` errors

marcbouchenoire:
This PR logs details (if available) about `FORBIDDEN` errors before they're sent to `useErrorListener`.

error

### DIFF
--- a/packages/liveblocks-core/src/comments/index.ts
+++ b/packages/liveblocks-core/src/comments/index.ts
@@ -1,4 +1,5 @@
 import type { AuthValue } from "../auth-manager";
+import type { JsonObject } from "../lib/Json";
 import type { BaseMetadata } from "./types/BaseMetadata";
 import type { CommentBody } from "./types/CommentBody";
 import type { CommentData } from "./types/CommentData";
@@ -58,6 +59,16 @@ export type CommentsApi<TThreadMetadata extends BaseMetadata> = {
   }): Promise<CommentData>;
 };
 
+export class CommentsApiError extends Error {
+  constructor(
+    public message: string,
+    public status: number,
+    public details?: JsonObject
+  ) {
+    super(message);
+  }
+}
+
 export function createCommentsApi<TThreadMetadata extends BaseMetadata>(
   roomId: string,
   getAuthValue: () => Promise<AuthValue>,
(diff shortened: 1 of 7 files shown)


packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -344,9 +360,10 @@ export function createCommentsRoom<TThreadMetadata extends BaseMetadata>(
 
     mutate(room.editThreadMetadata({ metadata, threadId }), {
       optimisticData,
-    }).catch((err: Error) => {
+    }).catch((err: CommentsApiError) => {
+      const error = handleCommentsApiError(err);
```
nvie, line 364: TypeScript will force you to use `err: unknown` in normal catch statements, and for good reason. Since anything can be thrown at any time, there is no way to guarantee that at runtime this value will actually be a `CommentsApiError` instance.

Therefore, the safer way to write this is:
```suggestion
    }).catch((err: unknown) => {
      if (!(err instanceof CommentsApiError)) { throw e; }
      const error = handleCommentsApiError(err);
```

…or some variation thereof.

nvie: As a general principle (maybe even a lint rule?), the `err` argument of a `.catch()` handler must always we typed as `unknown`.

packages/liveblocks-core/src/comments/index.ts
```diff
@@ -58,6 +59,16 @@ export type CommentsApi<TThreadMetadata extends BaseMetadata> = {
...
 
+export class CommentsApiError extends Error {
+  constructor(
+    public message: string,
+    public status: number,
+    public details?: JsonObject
+  ) {
+    super(message);
+  }
+}
```
nvie, line 70: Not a super big fan of using custom error classes in JS, because they add complexity to error handlers elsewhere and are hard to maintain over time. At the same time, I also don't have a great better suggestion to use right now. But maybe I'll want to unify the error handling across the board at one point in the future.

At the very least, can we make sure such custom error class is a private API and this class won't show up in the `*.d.ts` files, so we're free to refactor this later without users relying on this implementation detail? You can ensure that by adding an `/** @internal */` JSDoc comment.

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -174,6 +175,21 @@ function createThreadsManager<TThreadMetadata extends BaseMetadata>() {
...
+function handleCommentsApiError(err: CommentsApiError): Error {
+  const message = `Request failed with status ${err.status}: ${err.message}`;
+
+  // Log details about FORBIDDEN errors
+  if (err.details?.error === "FORBIDDEN") {
+    const detailedMessage = [message, err.details.suggestion, err.details.docs]
+      .filter(Boolean)
+      .join("\n");
+
+    console.error(detailedMessage);
```
nvie, line 187: 👍

nvie (approved):
Nice! Having this show up is important and helpful to users – it will save a lot of confusion! 👍

PR #1283, opened 2023-11-03
Add room, room storage, schema validation, and Yjs methods in `@liveblocks/node`

nimeshnayaju:
### Description

This PR adds the following methods to `@liveblocks/node` package.

  - [x] Room
    - [x] `getRooms`
    - [x] `createRoom`
    - [x] `getRoom`
    - [x] `updateRoom`
    - [x] `deleteRoom`
    - [x] `getActiveUsers`
    - [x] `broadcastMessage`
 - [x] Storage
    - [x] `getStorageDocument`
    - [x] `initializeStorageDocument`
    - [x] `deleteStorage`
- [x] Yjs
    - [x] `getYjsDocument`
    - [x] `sendYjsBinaryUpdate`
    - [x] `getYjsDocumentAsBinaryUpdate`
- [x] Schema Validation
    - [x] `createSchema`
    - [x] `updateSchema`
    - [x] `deleteSchema`
    - [x] `getSchemaByRoomId`
    - [x] `attachSchema`
    - [x] `detachSchema`
    
**Docs**
- https://github.com/liveblocks/liveblocks/pull/1298


packages/liveblocks-node/src/client.ts
```diff
@@ -48,6 +53,46 @@ type ThreadParticipants = {
   participantIds: string[];
 };
 
+type Permission = "room:write" | "room:read" | "room:presence:write";
```
FlowFlorent, line 56: it's missing the comment permissions:
https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-core/src/protocol/AuthToken.ts#L5

Is it also possible to bring in from @liveblocks/core?

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +273,512 @@ export class Liveblocks {
...
+    }
+
+    const res = await this.get(`/${path}`);
+
+    if (!res.ok) {
+      const text = await res.text();
+      throw new ApiError(res.status, text);
+    }
+
+    return res.json();
```
FlowFlorent, line 329:
```suggestion
    return await res.json();
```

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +273,512 @@ export class Liveblocks {
...
+      userAccesses,
+      metadata,
+    });
+
+    if (!res.ok) {
+      const text = await res.text();
+      throw new ApiError(res.status, text);
+    }
+
+    return res.json();
```
FlowFlorent, line 367:
```suggestion
    return await res.json();
```

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +273,512 @@ export class Liveblocks {
...
+   */
+  public async getRoom(roomId: string): Promise<RoomInfo> {
+    const res = await this.get(`/v2/rooms/${roomId}`);
+
+    if (!res.ok) {
+      const text = await res.text();
+      throw new ApiError(res.status, text);
+    }
+
+    return res.json();
```
FlowFlorent, line 383:
```suggestion
    rreturn await res.json();
```

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +273,512 @@ export class Liveblocks {
...
+      userAccesses,
+      metadata,
+    });
+
+    if (!res.ok) {
+      const text = await res.text();
+      throw new ApiError(res.status, text);
+    }
+
+    return res.json();
```
FlowFlorent, line 419:
```suggestion
   return await res.json();
```

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +273,512 @@ export class Liveblocks {
...
+    roomId: string
+  ): Promise<RoomUser<T>[]> {
+    const res = await this.get(`/v2/rooms/${roomId}/active_users`);
+
+    if (!res.ok) {
+      const text = await res.text();
+      throw new ApiError(res.status, text);
+    }
+
+    return res.json();
```
FlowFlorent, line 452:
```suggestion
    return await res.json();
```

packages/liveblocks-node/src/client.ts
nimeshnayaju: It's true that we have types for `room:presence:write`, `comments:write` and `comments:read` but the decoder used to validate permissions in our backend only checks for the following:

```
"room:write" | "room:read" | "room:presence:write"
```

See: https://github.com/liveblocks/liveblocks-cloudflare/blob/main/packages/liveblocks-cloudflare/src/api/lib/decoders.ts#L32-L49

packages/liveblocks-node/src/client.ts
nimeshnayaju: I am not sure if this makes any difference in terms of the function's behavior or the return type. That being said, I understand that using await can sometimes make error handling clearer and stack traces more accurate. @nvie Do you have a preference?

packages/liveblocks-node/src/client.ts
```diff
@@ -70,10 +115,10 @@ export class Liveblocks {
 
   /** @internal */
   private async post(
-    path: `/${string}`,
+    relativeURL: string,
     json: Record<string, unknown>
   ): Promise<Response> {
-    const url = urljoin(this._baseUrl, path);
+    const url = new URL(relativeURL, this._baseUrl);
```
nimeshnayaju, line 121: @nvie You'll notice `urljoin` has been replaced with `new URL()` here. I was getting some weird errors when using `urljoin` and making a fetch request because it was encoding characters like `?` but the API handler wasn't  reading it correctly.

Here is a reproduction of the issue:  [Playground link](https://www.typescriptlang.org/play?ssl=8&ssc=1&pln=8&pc=71#code/GYVwdgxgLglg9mABCATgGwFZxmAFAIwEMBnAUwFV0AuRYqFHAc0QB9FyAlAGQBpEAHQlAAWNOgzCMAlGPpNEAbwBQiRBAR1k6RAF5EYUgHd23AiQropAbhVa0AOkEiwhALaldAocJuqUpKFQkVAcoOABlOUlcayUAXyUldTBiODRSezQ4RlwQrBxcAHJhKCh+YioAekrCfhhMmAA3UnwsiABrYnt4Qr5CysaAJkqUODhXYgB+NBhXGCgdQcKpKSSNNIysnINjTi4igeHR8amZuYWlvpKyiura+pnm1rgOrp6VoA)

I also think using `URL` directly simplifies things here. Let me know what you think?

nimeshnayaju:
I've tried to keep the wrapper methods fairly simple. Here is the snippet from `getRoom` method - by checking `res.ok` and throwing an error for non-2xx responses, it communicates API-related issues to the consumer. Network errors (like problems with the fetch request itself) are not explicitly caught here, which means they will be thrown as exceptions. This will allow consumers of the function to differentiate between API errors and network problems and handle them accordingly.

```ts
public async getRoom(roomId: string): Promise<RoomInfo> {
    const res = await this.get(`/v2/rooms/${roomId}`);

    if (!res.ok) {
      const text = await res.text();
      throw new ApiError(res.status, text);
    }

    return res.json();
}
```

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +273,518 @@ export class Liveblocks {
...
+   * @param params.key (optional) If provided, returns only a single key’s value, e.g. doc.get(key).toJSON().
+   * @param params.type (optional) Used with key to override the inferred type, i.e. "ymap" will return doc.get(key, Y.Map).
+   * @returns A JSON representation of the room’s Yjs document.
+   */
+  public async getYjsDocument(
+    roomId: string,
+    params: {
+      format?: boolean;
+      key?: string;
+      type?: string;
```
nimeshnayaju, line 584: @jrowny The [REST documentation on Yjs](https://liveblocks.io/docs/api-reference/rest-api-endpoints#get-rooms-roomId-ydoc) mentions only `formatting`, `key`, and `type` as accepted parameters but the backend seems to also accept `guid`. I am guessing `guid` is related to sub-documents. Wondering if you know `guid` will also be documented in the REST documentation soon and if we want to include it here after the addition.

Same question for the `sendYjsBinaryUpdate` function ([REST documentation](https://liveblocks.io/docs/api-reference/rest-api-endpoints#put-rooms-roomId-ydoc))

packages/liveblocks-node/src/client.ts
```diff
@@ -48,6 +53,46 @@ type ThreadParticipants = {
...
+};
+
+type RoomUser<Info> = {
+  type: "user";
+  id: string | null;
+  connectionId: number;
+  info: Info;
+};
+
+type YJson =
```
GuillaumeSalles, line 79: I believe we can use `JsonObject` from `@liveblocks/core` instead of `YJson`.

packages/liveblocks-node/src/client.ts
```diff
@@ -48,6 +53,46 @@ type ThreadParticipants = {
   participantIds: string[];
 };
 
+type Permission = "room:write" | "room:read" | "room:presence:write";
+type RoomAccesses = Record<string, Permission[]>;
+type RoomMetadata = Record<string, string | string[]>;
+
+type RoomInfo = {
```
GuillaumeSalles, line 60: We need to re-export these types so our users can use them in their own functions.

packages/liveblocks-node/src/client.ts
```diff
@@ -299,21 +878,26 @@ export class Liveblocks {
...
+      const text = await res.text();
+      throw new ApiError(res.status, text);
     }
 
-    return body;
+    return res.json() as Promise<CommentData>;
+  }
+}
+
+export class ApiError extends Error {
```
GuillaumeSalles, line 896: I saw this [comment](https://github.com/liveblocks/liveblocks/pull/1280#discussion_r1379827639) about custom errors on a different PR. So I'm preemptively going to give my opinion on this just in case @nvie jumps in :D 

Like @nvie, I'm also not a fan of extending `Error` to add extra data to an error. However, it seems to be the most common and recommended pattern in the JS ecosystem. So I like what you did here. I agree with exporting a custom error, but maybe `HttpError` instead? Usually, I also add a `type` field so the consumer can use a switch and not have to use `instanceOf`.

We can keep it as `@internal` until we have a plan to expose custom errors at large. My only requirement for that plan is that we don't fight the ecosystem/language for purity sake (e.g. force people to handle errors to read the data).

GuillaumeSalles (commented):
That's really good @nimeshnayaju! Thank you so much for all that documentation <3 Can't wait to ship this!

I mainly focused on the external API surface and not the implementation. I'll let other people in the team review that part.

packages/liveblocks-node/src/client.ts
nvie: Ahhh yes. Originally, we followed the ESLint rule `no-return-await` for this — so we were deliberately avoiding the `return await` construct, because it would be less performant. (Return-await would create a new promise, which would only resolve in the next tick. So basically a trade-off of stack trace info for a bit more performance. And, of course, code consistency.)

I just read up a bit more on the latest state of this, and it looks like V8 recently added an optimization for this, making the `return await promise` _faster_ than `return promise`, so the default nowadays is the best of both worlds: more performance _and_ better stack traces.

Screen Shot 2023-11-06 at 10 28 28@2x

I think this means we should start discouraging `return promise` in favor of `return await promise` indeed, so you're making the right call here! Let's do a pass where we change our default inside `shared/eslint-config` (and in our other repos), and change this across the board in a follow-up PR.

packages/liveblocks-node/src/client.ts
```diff
@@ -88,8 +133,38 @@ export class Liveblocks {
   }
 
   /** @internal */
-  private async get(path: `/${string}`): Promise<Response> {
-    const url = urljoin(this._baseUrl, path);
+  private async put(
+    relativeURL: string,
+    json: Record<string, unknown>
```
nvie, line 138: Consider using `JsonObject` (from `@liveblocks/core` here). The `unknown` type could contain values that aren't JSON-serializable, which would not be the user's intention.

packages/liveblocks-node/src/client.ts
```diff
@@ -266,22 +849,18 @@ export class Liveblocks {
...
-    const body = await (resp.json() as Promise<ThreadParticipants>);
-
-    if (resp.status !== 200) {
-      throw {
-        status: resp.status,
-        ...body,
-      };
+    if (!res.ok) {
+      const text = await res.text();
+      throw new ApiError(res.status, text);
```
nvie, line 860: Could this generic error handling not be part of the `.get()`, `.post()`, etc methods? It looks like this needs to be handled the same everywhere?

packages/liveblocks-node/src/client.ts
nvie: +1 to changing the name to `HttpError`, because… that's what it seems to be.

Adding a `type` field won't work with error handling, because the only reasonable way to handle errors in TypeScript is by catching a value as `unknown`, i.e.

```ts
try {
  ...
} catch (e: unknown) {
  if (e instanceof HttpApi) { ... }     // alternative 1
  if (e.type === 'HttpError') { ... }   // alternative 2
  if (isHttpError(e)) { ... }           // alternative 3
  if (client.isError(e)) { ... }        // alternative 4
}
```

Therefore, alternative 1 works, but it means we'll have to expose the HttpError class as an implementation detail and we're stuck with maintaining it over time.

Alternative 2 that @GuillaumeSalles suggested won't be nice DX in TypeScript, because `e` is `unknown`, and TypeScript won't let you read the `.type` field off of it.

I'd personally be in favor of using alternative 3 or 4. In alternative 3, we could exposing a helper function to be imported `isHttpError()` (or similarly named). Alternative 4 is basically the same, but by exposing it on the `client` that's already in scope for the user, they can avoid the extra import.

```ts
export function isHttpError(e: unknown): e is HttpError {
  return e instanceof HttpError;
}
```

Using this function will let TypeScript refine the value, so you can access members afterwards:

```ts
try {
  ..
} catch (e: unknown) {
  if (client.isError(e)) {
    e.status  // ✅ number
    e.message  // ✅ string
  }
}
```

What do you think?

nvie (commented):
Looking great, @nimeshnayaju! I left a bunch of suggestions, but I think you're really close here. This is going to be extremely helpful to our users! 🙌

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +273,518 @@ export class Liveblocks {
...
+    }
+
+    return res.json() as Promise<RoomUser<T>[]>;
+  }
+
+  /**
+   * Boadcasts an event to a room without having to connect to it via the client from @liveblocks/client.
+   * @param roomId The id of the room to broadcast the event to.
+   */
+  public async broadcastMessage(
```
CTNicholas, line 465: Shouldn't this one be called `broadcastEvent`? Then it'd be in line with our other APIs.

CTNicholas (commented):
Excited to use this, thanks for working on it already!

packages/liveblocks-node/src/client.ts
nimeshnayaju: That is correct! Thanks for catching this. I've updated the function name to `broadcastEvent`

packages/liveblocks-node/src/client.ts
nimeshnayaju: Good point. Would `Json` be more appropriate here, instead of `JsonObject`?

```ts
declare type Json = JsonScalar | JsonArray | JsonObject;
declare type JsonScalar = string | number | boolean | null;
declare type JsonArray = Json[];
declare type JsonObject = {
    [key: string]: Json | undefined;
};
```

packages/liveblocks-node/src/Session.ts
```diff
@@ -1,14 +1,20 @@
...
 const ALL_PERMISSIONS = Object.freeze([
   "room:write",
   "room:read",
   "room:presence:write",
-  "comments:write",
-  "comments:read",
+  // TODO: Add these permissions back once the backend is set up to handle them
+  // "comments:write",
+  // "comments:read",
 ] as const);
```
nimeshnayaju, line 18: @nvie I have commented out `comments:write` and `comments:read` from `Sessions.ts` and added a comment to add the permissions back once the backend is set up to handle them.

packages/liveblocks-node/src/utils.ts
```diff
@@ -94,3 +94,26 @@ export function urljoin(
...
+export type URLSafeString = Brand<string, "URLSafeString">;
+
+/**
+ * Builds a URL where each "hole" in the template string will automatically be
+ * encodeURIComponent()-escaped, so it's impossible to build invalid URLs.
+ */
+export function url(
+  strings: TemplateStringsArray,
+  ...values: string[]
+): URLSafeString {
```
nimeshnayaju, line 115: @nvie This is genius. I love it!!

packages/liveblocks-node/src/client.ts
```diff
@@ -299,21 +736,27 @@ export class Liveblocks {
...
 
-    const body = await (resp.json() as Promise<CommentData>);
+  /**
+   * Checks whether the error is an HttpError.
+   * @param error The error to check.
+   * @returns Whether the error is an HttpError.
+   */
+  public isHttpError(error: unknown): error is HttpError {
+    return error instanceof HttpError;
+  }
```
nimeshnayaju, line 744: Renamed `ApiError` to `HttpError` and created this helper method inside the class to check if the error is `HttpError`. Additionally, `HttpError` isn't exported as a part of the `@liveblocks/node` package @nvie

packages/liveblocks-node/src/client.ts
```diff
@@ -69,36 +117,84 @@ export class Liveblocks {
   }
 
   /** @internal */
-  private async post(
-    path: `/${string}`,
-    json: Record<string, unknown>
-  ): Promise<Response> {
+  private async post(path: URLSafeString, json: Json): Promise<Response> {
```
nimeshnayaju, line 120: Updated both `post` and `put` methods to accept json of type `Json`. This required updating the types for a few other things that relied on this method, e.g., `userInfo` in the `CreateSessionOptions` type that is used by `prepareSession` method was previously typed as `unknown` but is now typed as `Json`.

https://github.com/liveblocks/liveblocks/blob/270650251d1f7af389427ac866047715a370f91a/packages/liveblocks-node/src/Session.ts#L100

packages/liveblocks-node/src/client.ts
```diff
@@ -48,6 +57,48 @@ type ThreadParticipants = {
   participantIds: string[];
 };
 
+export type RoomPermission =
+  | ["room:write"]
+  | ["room:read", "room:presence:write"];
```
nimeshnayaju, line 62: @nvie What I've done for now is create this `RoomPermission` type. I know `Session` also makes use of `Permission` type at some places and has the complete set of permissions. I haven't updated those for the sake of keeping this PR not too broad. But, I think it's worth taking a look at how our backend handles permission and updating these code accordingly.

packages/liveblocks-node/src/client.ts
nvie: I suggested `JsonObject` here because you had a `Record` type here already. If you need to allow scalars and arrays here too, you can widen it to `Json`, sure. If not needed yet, I’d suggest starting with the strictest type that works first. We can always loosen it later if needed (but we cannot easily stricter it later).

packages/liveblocks-node/src/client.ts
nvie: Really nice! I wanted to suggest this yesterday, but refrained from it because I didn’t want to be the nagging Puritan. Again, like I mentioned above, starting with the strictest types possible is a great starting position. Loosening is always possible for free, when it comes to input params.

packages/liveblocks-node/src/client.ts
nvie: Seeing this reminds me that we should probably also stricten the input type for `_userInfo` in TypeScript here too. I just made a backend change that will get deployed soonish, where the shape of the passed-in `userInfo` will be enforced to be of [`IUserInfo`](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-core/src/protocol/BaseUserMeta.ts#L3-L12) type (which is a strict subset of JSON).

Specifically:
- It's a JSON object
- If that object has a `name` field, it _must_ be a string
- If that object has a `avatar` field, it _must_ be a string
- All other fields are free to use and can be free-form JSON

Since the backend is going to enforce that shape soon, it may be good to include this strictening of this input type here with your changes as well.

packages/liveblocks-node/src/client.ts
jrowny: yes, guid is a new option that I did not document yet, all it does is specify which subdoc to load. I will add that rest documentation soon.

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +293,377 @@ export class Liveblocks {
...
+    document: PlainLsonObject
+  ): Promise<PlainLsonObject> {
+    const res = await this.post(url`/v2/rooms/${roomId}/storage`, document);
+    return (await res.json()) as Promise<PlainLsonObject>;
+  }
+  /**
+   * Deletes all of the room’s Storage data and disconnect all users from the room if there are any.
+   * @param roomId The id of the room to delete the storage from.
+   */
+  public async deleteStorage(roomId: string): Promise<void> {
```
CTNicholas, line 512: Missed this the other day! Can we make this one `deleteStorageDocument`? Our docs currently have the wrong title—this should match the other two Storage functions.

nvie: Also maybe mention explicitly that it will **not** delete the YDoc if it exists.

packages/liveblocks-node/src/client.ts
nimeshnayaju: This is very interesting! Thank you for looking into this.

packages/liveblocks-node/src/client.ts
nimeshnayaju: Done! Updated the method name to `deleteStorageDocument` and updated the comment to say `Note that this does not delete the Yjs document in the room if one exists.`. Thanks @nvie @CTNicholas 

https://github.com/liveblocks/liveblocks/blob/ea2b50212f51c08eba046da122ed4c45b01438c4/packages/liveblocks-node/src/client.ts#L511C1-L517C4


```ts
/**
 * Deletes all of the room’s Storage data and disconnect all users from the room if there are any. Note that this does not delete the Yjs document in the room if one exists.
 * @param roomId The id of the room to delete the storage from.
 */
public async deleteStorageDocument(roomId: string): Promise<void> {
  await this.delete(url`/v2/rooms/${roomId}/storage`);
}
```

nimeshnayaju:
@nvie How do you suggest we go about merging this PR? @marcbouchenoire is working on making few extra changes to the node package for a separate issue. Because we treat minor versions as important milestones and often write a blog post about the new changes, we were curious if it makes more sense to wait to release till we have both changes (and potentially more from others if any) ready to merge.

marcbouchenoire:
I'm especially curious about how to proceed with this: do we keep this PR open (and create a `1.8` milestone) to not block `main` from doing any patch releases? Do we create a dedicated `1.8` PR?

nvie:
Sounds like a 1.8 release that combines both of your work + publish a blog post would be nice. I'd suggest starting a new empty branch/PR from `main` and name it `release-1.8`, and then change the merge target for both of your PRs to that new branch. After that's done, and the tests are all passing here, you can just merge that into the release 1.8 branch.
We can also keep a checklist of things to do in the PR description, including blog post, documentation, upgrade guide (if applicable) etc. Lmk if that isn't clear or if you have any questions.

packages/liveblocks-node/src/client.ts
```diff
@@ -48,6 +58,48 @@ type ThreadParticipants = {
...
+  | string
+  | undefined
+  | unknown[];
+
+export type Schema = {
+  id: string;
+  name: string;
+  version: number;
+  createdAt: string;
+  updatedAt: string;
```
nvie, line 99: I think it will be nicer if we'd convert these back into Date instances for our users.

```suggestion
  createdAt: Date;
  updatedAt: Date;
```

This means we can no longer directly return `(await res.json()) as Schema`, but we'd have to convert these fields explicitly. Doing so at least prevents us from having to "fix" this type/return value later in change that will be a breaking change at that point.

If we have more exported types here with dates in them, I'd suggest to do the same for all.

packages/liveblocks-node/src/client.ts
```diff
@@ -69,36 +121,84 @@ export class Liveblocks {
...
+    const res = await fetch(url, {
       method: "POST",
       headers,
       body: JSON.stringify(json),
     });
+    if (!res.ok) {
+      const text = await res.text();
+      throw new HttpError(res.status, text);
+    }
+    return res;
```
nvie, line 141: With these generic `.post()`, `.get()`, `.put()`, etc helpers, you're either throwing the response if it's a non-200 response, or _returning_ the response. However, this isn't obvious from the call site perspective if you look at the return type, because it just says `Response`. So from the call site it's unclear that those responses will only ever be 2xx responses, and `res.ok` will never be `true` on the returned value.

I'd suggest to therefore return the result of `res.json()` here and change the return type for these helpers to `Promise<Json>` instead, to help make that intent and behavior more clear just by looking at the types.

packages/liveblocks-node/src/client.ts
```diff
@@ -157,11 +257,11 @@ export class Liveblocks {
       | string // Shorthand for userId
       | Identity,
     options?: {
-      userInfo: unknown;
+      userInfo: Json;
```
nvie, line 260: This should also be `IUserInfo` type (the backend will throw if you'd pass it anything else).

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +294,379 @@ export class Liveblocks {
...
+      startingAfter?: number;
+      metadata?: RoomMetadata;
+      userId?: string;
+      groupIds?: string;
+    } = {}
+  ): Promise<{
+    nextPage: string | null;
+    data: RoomInfo[];
+  }> {
+    const path = url`v2/rooms`;
```
nvie, line 322: It will work, but let's make all URL paths absolute everywhere, by letting them all start with a `/`. It also helps hugely with the greppability of the codebase if we do that:

```suggestion
    const path = url`/v2/rooms`;
```

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +294,379 @@ export class Liveblocks {
...
+    }
+    if (params.groupIds) {
+      queryParams.set("groupIds", params.groupIds);
+    }
+    if (params.metadata) {
+      Object.entries(params.metadata).forEach(([key, val]) => {
+        const value = Array.isArray(val) ? val.join(",") : val;
+        queryParams.set(`metadata.${key}`, value);
+      });
+    }
```
nvie, line 342: With the introduction of the new helper, you should be able to write this more succinctly (but as safely), using the following:

```suggestion
    const queryParams = {
      limit: params.limit,
      startingAfter: params.startingAfter,
      userId: params.userId,
      groupIds: params.groupIds,

      // "Flatten" {metadata: {foo: "bar"}} to {"metadata.foo": "bar"}
      ...Object.fromEntries(
        Object.entries(params.metadata ?? {}).map(([key, val]) => {
          const value = Array.isArray(val) ? val.join(",") : val;
          return [`metadata.${key}`, value];
        })
      ),
    };
```

packages/liveblocks-node/src/utils.ts
```diff
@@ -94,3 +94,26 @@ export function urljoin(
   }
   return url.toString();
 }
+
+declare const brand: unique symbol;
+
+type Brand<T, TBrand extends string> = T & { [brand]: TBrand };
```
nvie, line 100: The `Brand` util is exported in `@liveblocks/core` since recently, so feel free to DRY that up!

nvie (approved):
Left a few last pieces of feedback, but once addressed, this all looks good to me. Can I say the end result here is absolutely gorgeous — amazing work @nimeshnayaju! Our users will love it, I'm sure. 😍

marcbouchenoire:
@nimeshnayaju One thing I forgot yesterday is that we should update `CHANGELOG.md` and add all the new methods to the docs, maybe with subheadings (e.g. "Comments", "Storage", etc) to divide them nicely? cc: @CTNicholas

CTNicholas:
> @nimeshnayaju One thing I forgot yesterday is that we should update `CHANGELOG.md` and add all the new methods to the docs, maybe with subheadings (e.g. "Comments", "Storage", etc) to divide them nicely? cc: @CTNicholas
> 
> 

Yes! I'll write up the docs later today.

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +294,379 @@ export class Liveblocks {
...
+   * @param params.startingAfter (optional) A cursor used for pagination. You get the value from the response of the previous page.
+   * @param params.userId (optional) A filter on users accesses.
+   * @param params.metadata (optional) A filter on metadata. Multiple metadata keys can be used to filter rooms.
+   * @param params.groupIds (optional) A filter on groups accesses. Multiple groups can be used.
+   * @returns A list of rooms.
+   */
+  public async getRooms(
+    params: {
+      limit?: number;
+      startingAfter?: number;
```
CTNicholas, line 313: I think `startingAfter` should be a `string`, could you check?

packages/liveblocks-node/src/client.ts
```diff
@@ -194,6 +294,379 @@ export class Liveblocks {
...
+      metadata?: RoomMetadata;
+      userId?: string;
+      groupIds?: string;
```
CTNicholas, line 316: We should accept an array of strings here, right?

```ts
// Currently not possible
const groupIds = ["design", "engineering"];
liveblocks.getRooms({ groupIds });
```

packages/liveblocks-node/src/client.ts
nvie: > The Comments functions in `@liveblocks/node` return [`ThreadData`](https://github.com/liveblocks/liveblocks/blob/a7aef1d0a0cf8b5bada932805ef52d00cccc77ef/packages/liveblocks-core/src/comments/types/ThreadData.ts#L11) which uses a `string`. This same type is also returned on the front end by `useThreads`.

Is this something can we can still change for the comments use case?

Keeping and communicating the semantics of fields in APIs is really important and sometimes the name of those fields just aren't enough to really know what values are legal. While technically the fields _must_ travel over the wire as strings (thanks JSON spec), our `@liveblocks/node` library is in the unique position to translate them back to a semantic value before returning them to the user.

This way, the type that users get exposed to can simply keep using Date reliably, and we don't have to expose our users to the implementation details of our serialization format directly.

CTNicholas:
I've noticed there's an inconsistency with the placement of `roomId` in the params:

Comments functions:
```ts
liveblocks.getThreads({ roomId: "my-room-name" });
liveblocks.getComment({ roomId: "my-room-name", threadId: "my-thread-id", commentId: "my-comment-id" });
```

New functions:
```ts
liveblocks.getRoom("my-room-name");
liveblocks.createRoom("my-room-name", { defaultAccesses: [] });
liveblocks.getRooms({ limit: 20 });
```

It would be awesome if this were consistent through every method. I think I'd prefer using `{ roomId }` for everything, though have no strong feelings either way. Anyone else got a preference?

nvie:
@CTNicholas I’d prefer to keep the API’s as they exist right now. I’m personally not bothered that there is a difference between these, and think they both picked the param style that best fits them already.

Avoiding object params where possible keeps the API’s simplest to learn, use, and understand I think. For example, if you require an object param, users can be surprised and may expect there to be more than one potential option to specify. Especially if an API needs only one ID, and the one of the main resource being queried, it’s going to be pretty obvious. For example, in `.getRoom("abc")` it’s pretty obvious that’s a room ID.

In the `.getThreads()` and `.getComments()` examples, the main ID to pass is not immediately obvious, or there are more than one possible ID to pass, with no natural (or easy) order to them, so there the object pattern is a better fit, because it helps you pass the right IDs correctly.

That’s my £0.02.

CTNicholas:
Ah really! I figured since they're all room-based APIs (except `getRooms`) it would make sense to use the same pattern for `roomId`, e.g. always pass it as the first argument. I'm not too fussed though 👍

packages/liveblocks-node/src/client.ts
GuillaumeSalles: What are the benefits of exposing a function that does `instanceof` instead of the user doing `instanceof` directly?

Doing `instanceof` for customer error handling is the most common and recommended pattern. Expected by developers of all levels.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#handling_a_specific_error_type

nvie: Two benefits:

- We'd not have to commit to the name `HttpError` just yet and can keep it an implementation detail. If we want to change the custom error name later, it's not a breaking change. (This doesn't hold for the _fields_ in the error of course.)
- Users don't have to import a custom error type just to be able to handle it. Reason I suggested using putting a method on the client is that that variable is already in scope, no extra import needed.

To me, it feels quite wrong to be exporting a non-standard `HttpError` class from a (Liveblocks-specific) library, mostly. If all library authors did this, then as a user you'd be importing and handling a ton of different "HttpError"s from different libraries 😬

nvie: I mean, I could live with alternative 1 if we think it's more intuitive. In a way, this suggestion is a way to punt on the decision of how we want to do error handling until we have an API we are happy with and willing to commit to in the long term.

packages/liveblocks-node/src/client.ts
nimeshnayaju: It is indeed a string. Nice catch!

packages/liveblocks-node/src/client.ts
nimeshnayaju: The backend still decodes it as a string however. We could accept an array and then join the array to convert to string before passing it to the request. Is that what you you referring to too @CTNicholas ?

nimeshnayaju:
> Avoiding object params where possible keeps the API’s simplest to learn, use, and understand I think. For example, if you require an object param, users can be surprised and may expect there to be more than one potential option to specify. Especially if an API needs only one ID, and it’s the one of the main resource being queried, it’s going to be pretty obvious. For example, in .getRoom("abc") it’s pretty obvious that’s a room ID.

I feel similarly. Although for `getThreads

packages/liveblocks-node/src/client.ts
```diff
@@ -48,6 +58,48 @@ type ThreadParticipants = {
...
+  info: Info;
+};
+
+export type YJson =
+  | {
+      [x: string]: unknown;
+    }
+  | string
+  | undefined
+  | unknown[];
```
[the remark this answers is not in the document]
nimeshnayaju: I did miss that. Thanks for reminding! 💟

packages/liveblocks-node/src/client.ts
nimeshnayaju: I like this and makes sense! I do have some concerns though but not major ones: this assumes that all responses will be in JSON format and if the server sends a response in a different format, this will fail or will need additional handling. I noticed inside `identifyUser`, we make a `POST` request and then parse the response body as text before returning it. 

https://github.com/liveblocks/liveblocks/blob/ea2b50212f51c08eba046da122ed4c45b01438c4/packages/liveblocks-node/src/client.ts#L272-L284

This was one of the reasons why I initially opted out of checking `res.status` inside the generic `POST`, `GET` methods so that public methods calling these internal methods had more flexibility in terms of what they did with the response but there was repetition involved with checking for `res.status` for each method, which also wasn't desirable. 

Another minor concern is that returning only JSON means that we lose access to other details of the response, such as headers, status code, or status text, which is completely fine with the current setup, but something to keep in mind.

What are your thoughts @nvie ?

packages/liveblocks-node/src/client.ts
nimeshnayaju: Agreed. Nice to keep it consistent and makes sense to make this an absolute url.

packages/liveblocks-node/src/client.ts
nimeshnayaju: This one I did to demonstrate how we can also use `URLSearchParams` with the new `GET` method. There is another place I use the `Record<string, string | number | null | undefined>` type instead but makes sense to keep it consistent in all places.

packages/liveblocks-node/src/client.ts
nvie: OMG, you're right @nimeshnayaju! I wonder why this is… 🤔 

https://github.com/liveblocks/liveblocks-cloudflare/blob/2ee8995fc5646e71bf7af5a8c3a2a17412568f64/packages/liveblocks-cloudflare/src/api/v2/get-rooms.ts#L77-L86

cc @FlowFlorent @ofoucherot Any idea why this historically isn't taking in this value as an array of strings instead?

nvie:
> Do you have opinion on the signature of this method specifically

I'd keep it consistent within the realm of Comments™. So it's very reasonable to me that this is taking it as `{ roomId: string }` here. So I'd suggest to keep the signatures as they are, because I think they're great already 👍

packages/liveblocks-node/src/client.ts
nvie: Yeah, it's a tough trade-off to make.

All _success_ responses from our API should (and will always) be in JSON format I think? If CF has a disruption, they may return a HTML-based 503 page, in which case there is no JSON payload, but that case is handled by your `res.ok` check already. I think all of our success responses should just always return JSON, to make handling this easy.

Although some of our endpoints may return an empty response body (e.g. in the case of an [HTTP 204](https://github.com/liveblocks/liveblocks-cloudflare/blob/2ee8995fc5646e71bf7af5a8c3a2a17412568f64/packages/liveblocks-cloudflare/src/api/v2/get-rooms.ts#L77-L86)), then JSON parsing would still fail. Maybe we should simply use our `tryParseJson` helper instead (from `@liveblocks/core`), and use this instead:

```tsx
return tryParseJson(await res.text()) ?? null;
```

I don't think we actually need any response info (from headers or status fields). All information this client should return to the user is coming from the response payload. If this changes in the future, we could expose a "raw" version that would retain the rich response info and maybe let the error handling be dealt with by the user, but for now I'd say YAGNI.

packages/liveblocks-node/src/client.ts
GuillaumeSalles: > We'd not have to commit to the name HttpError just yet and can keep it an implementation detail. If we want to change the custom error name later, it's not a breaking change. (This doesn't hold for the fields in the error of course.)

If we do this, we're committing to a function name instead of class name, so I don't think this argument is valid.

> Users don't have to import a custom error type just to be able to handle it. Reason I suggested using putting a method on the client is that that variable is already in scope, no extra import needed.

It's more important to have an expected pattern than avoid an import. 

- Stripe is exporting `StripeError`s for [error handling](https://github.com/stripe/stripe-node/wiki/Error-Handling#typescript)

- Slack is exporting [custom errors](https://www.npmjs.com/package/@slack/web-api?activeTab=code) under `dist/errors.d.ts`

- Resend is exporting [ResendError](https://www.npmjs.com/package/resend?activeTab=code) under `build/src/error.d.ts`

I'm not using a lot of REST client but this is the pattern I always encountered/expected since I used typescript. Happy to look into other REST API wrappers if you think I'm currently biased.

(They all also include a code to let the user do a `switch` after checking the type with `instanceof`. That's what I meant in my original message.)

If I see a `StripeError` in my console logs, without looking at the docs, my reflex is to try to import `StripeError` and not look for method on the client to check the type of the error.

> To me, it feels quite wrong to be exporting a non-standard HttpError class from a (Liveblocks-specific) library, mostly. If all library authors did this, then as a user you'd be importing and handling a ton of different "HttpError"s from different libraries 😬

That's why I went for `LiveblocksError` in the client (following Stripe pattern at the time) to avoid any collision (even if I never exported it).

packages/liveblocks-node/src/client.ts
nimeshnayaju: I like the idea of converting the strings to Date before returning. Like @nvie said, this allows the library to manage the complexity of serialization and deserialization while providing a more intuitive and useful interface to the users. How do you suggest we go about this if we do? `getThreads`, `getThread`, `getComment` currently exposes `createdAt` property as a string (and `useThreads` from our react package does so too). Having them also return a Date instead would mean breaking change, I believe. Do we want to do it for all methods (including `getRoom`, `getRooms`) in this PR or a different PR?

packages/liveblocks-node/src/client.ts
nimeshnayaju: That is true. All our success responses do return JSON and the `tryParseJson` should gracefully deal with cases where we send an empty response body. With this change, how do you suggest we update the `identifyUser`. Do we do something like this instead?

Before:
```ts
try {
  // this.post() used to return Promise<Response>
  const resp = await this.post(path, {
    userId,
    groupIds,
  });

  return {
    status: normalizeStatusCode(resp.status),
    body: await resp.text(),
  };
}
```

After:
```ts
try {
  // this.post() now returns Promise<Json | null>
  const json = await this.post(path, {
    userId,
    groupIds,
  });

  return {
    status: 200,
    body: JSON.stringify(json),
  };
}
```

packages/liveblocks-node/src/client.ts
nvie: A'ight, you convinced me. Let's go with `LiveblocksError` in that case 👍

packages/liveblocks-node/src/client.ts
nvie: I think a breaking change is legitimized to fix an existing issue with this format (and since comments is still in beta, I think it's fine), as long as we document it, and tackle all the instances at once. I think most users won't run into problems with it.

packages/liveblocks-node/src/client.ts
nvie: Sure, that works! I wouldn't worry about the `JSON.stringify(JSON.parse(...))` that's happening there effectively perf-wise — if that's what you mean. You can stringify/parse a typical response many 100000's times in one second, it's nothing compared to the async/networking overhead here.

I think this code is really readable this way 👍

nimeshnayaju: Hmmm. I looked more into this and I think this might not work for all cases. For e.g. when requesting for (binary) Yjs document, the server sends binary data. See [link](https://github.com/liveblocks/liveblocks-cloudflare/blob/2ee8995fc5646e71bf7af5a8c3a2a17412568f64/packages/liveblocks-cloudflare/src/room.ts#L608-L617).

```ts
const update = Y.encodeStateAsUpdate(doc);
return new Response(update, {
  headers: { "Content-Type": "application/octet-stream" },
  status: 200,
});
```

This needs to be read as `res.arrayBuffer()` from the client. This is making me more inclined towards keeping the generic `post`, `get` internal methods only do the fetching and letting the methods that consume this method deal with the rest of the logic as they see fit despite the repetition that is involved (e.g. handling non-2xx responses) until we come up with an approach that fits our situation well. Also, these are all implementation details, so we can always go back and fine tune these later.

So, something like:

```ts
private async post(path: URLSafeString, json: Json): Promise<Response> {
  const url = urljoin(this._baseUrl, path);
  const headers = {
    Authorization: `Bearer ${this._secret}`,
    "Content-Type": "application/json",
  };

  const fetch = await fetchPolyfill();
  return await fetch(url, {
    method: "POST",
    headers,
    body: JSON.stringify(json),
  });
}

public async getRoom(roomId: string): Promise<RoomInfo> {
  const res = await this.get(url`/v2/rooms/${roomId}`);

  if (!res.ok) {
    const text = await res.text();
    throw new HttpError(res.status, text);
  }

  return (await res.json()) as Promise<RoomInfo>;
}
```

packages/liveblocks-node/src/client.ts
nimeshnayaju: What I have done for now is for non-comments related methods (like `getRoom`, `createRoom`, `getRooms`, `getSchema`, `createSchema`), the properties like `createdAt`, etc that are returned as ISO strings from server are converted to `Date` objects before being returned. 

The same isn't true for comments related methods yet because I noticed a lot of dependency of comment types in other packages and think that change warrants a separate PR of its own.

https://github.com/liveblocks/liveblocks/blob/a7243a992d377e01223dcab1db49a6862d450327/packages/liveblocks-node/src/client.ts#L400-L422

nimeshnayaju:
@nvie Made some final changes including removing the non 2xx response status checking responsibility from the internal `post`, `get`, `delete`, `put` methods, which means this check is now done explicitly by each methods. I think I'm okay with this for the time being. I have also removed the `isHttpError` helper method and renamed `HttpError` to `LiveblocksError` based on your discussion with @GuillaumeSalles. 

I have also made sure methods that previously returned ISO string are now morphed to `Date` objects before being returned!

packages/liveblocks-node/src/client.ts
marcbouchenoire: I agree with making a separate PR for that, and regarding doing the same in the front-end packages, given that the `Date` constructor and our `<Timestamp />` primitive both accept `string` and `Date` (and `number`), the breaking change transition could be very minimal.

nvie:
> Made some final changes including removing the non 2xx response status checking responsibility […]  
> I have also removed the `isHttpError` helper method and renamed `HttpError` to `LiveblocksError` […]  
> I have also made sure methods that previously returned ISO string are now morphed to `Date` objects […]

You're amazing, @nimeshnayaju — you made the right calls and handled all the feedback beautifully! Super happy with the end result here! 😍

CTNicholas:
Could you check that returned (and related) types are also being exported from `@liveblocks/client`?

I've noticed that the `PlainLson` types [aren't currently exported](https://github.com/liveblocks/liveblocks/blob/a7243a992d377e01223dcab1db49a6862d450327/packages/liveblocks-client/src/index.ts#L7), and it would be helpful to have all of those now we have the Storage functions. Same for `Schema`!

CTNicholas:
I've written up the docs (took ages!). If anyone could review them, that would be awesome: https://github.com/liveblocks/liveblocks/pull/1298

nimeshnayaju:
@CTNicholas @nvie I am curious too how we want to go about the exports. It does feel like the exports are coming from different places now but it's difficult to understand where they should be coming from.

nvie:
> @CTNicholas @nvie I am curious too how we want to go about the exports. It does feel like the exports are coming from different places now but it's difficult to understand where they should be coming from.

We (should be) maintaining a stable outer API surface area for `@liveblocks/client`. Any new export you introduce there is a liability to maintain going forward.

Exports put in `@liveblocks/core` should only be relied on by Liveblocks employees. It's 100% private, and we're free to change these types at any type, as we see fit. No customer should ever directly depend on `@liveblocks/core`, only on `@liveblocks/client`. This is [documented in the README](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-core/README.md#liveblockscore), but it may not be obvious enough. Sometimes I want to rename this package to a more ugly name for this reason.

I have not yet looked at your PR from the angle of what new types or new surface area is being introduced here (except all those new client methods of course). If we're exposing new types to users that previously wasn't the case, please let me know. If we need to expose types to users (because they are being used in newly exposed signatures), then we should be exposing them properly through `@liveblocks/client` (as re-exports). And from that point onwards, we're responsible to no longer change them and support them going forward. We should always make sure we're introducing those deliberately and intentionally.

nvie:
@nimeshnayaju Also relevant and a bit of a "here be dragons" thing that's good to know:

https://github.com/liveblocks/liveblocks/blob/b0e50fc94f6075117afde249ba422dd25ef3e97c/packages/liveblocks-node/src/client.ts#L1-L5

This is the reason why `@liveblocks/core` is a [_dev_ dependency](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-node/package.json#L40) and not a production dependency. (It's not a mistake!)

CTNicholas:
@nvie We need to export these types so people can handle Storage, but they're currently just in `@liveblocks/core`:
https://github.com/liveblocks/liveblocks/blob/6a35394a430c313580264212324d2dadacecb08e/packages/liveblocks-core/src/types/PlainLson.ts#L47

Used here:
https://github.com/liveblocks/liveblocks/pull/1283/files#diff-b7ffaad9c487753169585109f7bec10f73f1ee571f9f424f894f5966f0cac4b8R555

packages/liveblocks-node/src/client.ts
FlowFlorent: This is because groupIds in this case is coming from a query parameter
ex: `/v2/rooms?groupIds=marketing,engineering`

We could take in an array and transform it to a query parameter string, that would be the easiest for the user I think.

nimeshnayaju:
@nvie We seem to currently only export the `Liveblocks` class and type `LiveblocksOptions` from our `client.ts` file, which probably isn't enough.

https://github.com/liveblocks/liveblocks/blob/b0e50fc94f6075117afde249ba422dd25ef3e97c/packages/liveblocks-node/src/index.ts#L2-L3

We have introduced some new types including `RoomPermission`, `RoomAccesses`, `RoomMetadata`, `RoomInfo`, `RoomUser<Info>` and `Schema`. 

https://github.com/liveblocks/liveblocks/blob/b0e50fc94f6075117afde249ba422dd25ef3e97c/packages/liveblocks-node/src/client.ts#L61-L80


We also make use of some exports from `@liveblocks/core` which we may need to re-export from `@liveblocks/node`. 
https://github.com/liveblocks/liveblocks/blob/b0e50fc94f6075117afde249ba422dd25ef3e97c/packages/liveblocks-node/src/client.ts#L6-L13

nvie:
If it’s just the PlainLson types, then we should be able to simply re-export them from `@liveblocks/core`. Since this does mean we make it public for the first time officially, we should make them harder to change unintentionally. We could use `tsd` to enforce that, so that if we change those types inside `core`, the unit tests inside `node` would start failing.

Let me know if you need my help writing those type-level tests! You can find an example of such tests [here](https://github.com/liveblocks/liveblocks/blob/6a35394a430c313580264212324d2dadacecb08e/packages/liveblocks-client/test-d/room-api.test-d.ts#L4) or [here](https://github.com/liveblocks/liveblocks-cloudflare/blob/main/packages/liveblocks-server/test-d/server.test-d.ts). They look a bit weird, but they are super useful! We don’t have enough of these, but I want to change that.

CTNicholas:
Not sure on standard practices when it comes to libraries, but we could add a `name` property to `LiveblocksError` for a more obvious console error. Before and after:

image

packages/liveblocks-node/test-d/client.test-d.ts
```diff
@@ -0,0 +1,15 @@
...
+
+async () => {
+  const client = new Liveblocks({ secret: "sk_xxx" });
+
+  expectType<PlainLsonObject>(await client.getStorageDocument("document-id"));
+  expectType<PlainLsonObject>(
+    await client.getStorageDocument("document-id", "plain-lson")
+  );
+  expectType<JsonObject>(
+    await client.getStorageDocument("document-id", "json")
```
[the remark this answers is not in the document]
nvie: This is perfect! Think of these as unit tests, but for our type definitions. What’s expressed here is that if you call `.getStorageDocument()` it will return a value of type `PlainLsonObject` (exactly the same as from `core`. If this function or return type would ever change, then this “unit test” can fail in the future, which then might indicate a breaking change.

The value of these tests grow over time, as we approach 100% type coverage with them. They are especially useful for people making changes in `@liveblocks/core`, where it may be unclear what the effect on `@liveblocks/node`’s types are.

packages/liveblocks-node/tsconfig.json
```diff
@@ -6,5 +6,5 @@
     "target": "es2020",
     "module": "es2020"
   },
-  "include": ["src/**/*"]
+  "include": ["src/**/*", "test-d"]
```
nvie, line 9: TypeScript itself should not know about this dir I think. The way you run tsd is through `turbo run test:types`, which first builds to `dist/`, and tsd will run against the produced `dist/index.d.mts` file. Including it here will make TypeScript look at these files, which is not necessary I believe.

nimeshnayaju: Makes sense! and done!

PR #1284, opened 2023-11-03
Docs: Compress visuals

marcbouchenoire:
This PR compresses most visuals (images and videos) used in the docs. These can be served as is in the pages so we should keep them as small as possible, and it also helps downloading less stuff whenever we build the site (since it fetches them).


CTNicholas (approved):
Was thinking about this the other day, thanks!

I was also thinking of adding something like [`sharp`](https://www.npmjs.com/package/sharp) to our build process, but I think it would probably slow builds down too much.

PR #1286, opened 2023-11-06
Extend `urljoin` with ability to handle query params safely

nvie:
Adds a third param to `urljoin` to safely yet conveniently build URLs with query params.

### DIFF
--- a/packages/liveblocks-node/src/utils.ts
+++ b/packages/liveblocks-node/src/utils.ts
@@ -70,7 +70,7 @@ function toURLSearchParams(
 ): URLSearchParams {
   const result = new URLSearchParams();
   for (const [key, value] of Object.entries(params)) {
-    if (value != null) {
+    if (value !== undefined && value !== null) {
       result.set(key, value.toString());
     }
   }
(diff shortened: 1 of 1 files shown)


nimeshnayaju (approved):
This is brilliant! Thanks for doing this!

packages/liveblocks-node/src/utils.ts
```diff
@@ -42,11 +42,55 @@ export function normalizeStatusCode(statusCode: number): number {
...
+ * Notice how the number is converted to its string representation
+ * automatically and the `null`/`undefined` values simply don't end up in the
+ * URL.
  */
-export function urljoin(baseUrl: string | URL, path: string): string {
-  const url = new URL(baseUrl);
-  url.pathname = path;
+function toURLSearchParams(
+  params: Record<string, string | number | null | undefined>
+): URLSearchParams {
```
nimeshnayaju, line 70: In future (if we have the use case), I believe, we could also accept an array value and use `append` instead of `set` appropriately so that we are able set the same parameter multiple times. 

Example - construction of urls like: `https://liveblocks.io/123?groupId=1&groupId=2`

nvie: > _“In future (if we have the use case), I believe, we could also accept an array value and use append instead of set appropriately so that we are able set the same parameter multiple times.”_

Yep! Feel free to update accordingly when there is the need for that!

PR #1288, opened 2023-11-06
Comments: Various improvements

marcbouchenoire:
This PR adds a few small improvements to Comments:

- Remove unnecessary calculations in "subtle" tokens (since their `index` is 0)
- Improve relative date formatting again: we recently switched from the `"short"` style to the `"narrow"` one ("1 hour ago" became "1h ago") but in some locales that resulted in weird formatting like in French where it formatted as "-1 h". I made a quick sandbox (https://xhqmwd.csb.app/) to compare all locales and identified where the `"narrow"` style was broken but not the `"short"` one, and made it so that we use the `"short"` one for these locales only, otherwise we keep `"narrow"`.
- I like our setup where we don't explicitly set fonts and only inherit from the context, I think it's easier that way compared to forcing an arbitrary list and forcing people (who already have a specific font that they want to apply to Comments as well) to specify their font a second time just for us. That said, I think we can do it for monospace fonts used in inline code blocks, especially since the default `monospace` doesn't look great. I've set it with a specificity of `0,0,0` which means that it will override the browser's default, but anything will be able to override it, even something like `code { font-family: "My Code Font" }` so I think it's an ideal way to set a better default without added complexity.
  - **Before:** Before
  - **After:** After

### DIFF
--- a/packages/liveblocks-react-comments/src/primitives/Timestamp.tsx
+++ b/packages/liveblocks-react-comments/src/primitives/Timestamp.tsx
@@ -80,12 +80,38 @@ function formatShortDate(date: Date, locale?: string) {
   return capitalize(formatter.format(date));
 }
 
+// Some locales' relative formatting can be broken (e.g. "-1h") when using the narrow style.
+const localesWithBrokenNarrowRelativeFormatting = [
+  "br",
+  "fr",
+  "nb",
+  "nn",
+  "no",
+  "ro",
+  "sv",
+];
+
 /**
  * Formats a date relatively.
  */
 function formatRelativeDate(date: Date, locale?: string) {
-  const formatter = relativeTimeFormat(locale, {
-    style: "narrow",
+  let resolvedLocale: string;
+
+  if (locale) {
+    resolvedLocale = locale;
+  } else {
+    const formatter = relativeTimeFormat();
+
+    resolvedLocale = formatter.resolvedOptions().locale;
+  }
+
+  const isBrokenWhenNarrow = localesWithBrokenNarrowRelativeFormatting.some(
+    (locale) =>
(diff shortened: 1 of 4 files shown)


packages/liveblocks-react-comments/src/primitives/Timestamp.tsx
```diff
@@ -80,12 +80,38 @@ function formatShortDate(date: Date, locale?: string) {
   return capitalize(formatter.format(date));
 }
 
+// Some locales' relative formatting can be broken (e.g. "-1h") when using the narrow style.
+const localesWithBrokenNarrowRelativeFormatting = [
```
CTNicholas, line 84: It feels strange to do this instead of using `short` everywhere! 

But I trust there's a reason behind using `narrow` where we can, and that you've already considered the options 👍.

packages/liveblocks-react-comments/src/styles/index.css
```diff
@@ -639,6 +639,12 @@
   }
 }
 
+/* 0,0,0 specificity to inherit any styles applied to `code` elements */
```
CTNicholas, line 642: Here's before and after on Windows

image

CTNicholas (approved):

CTNicholas:
Nice one!

packages/liveblocks-react-comments/src/primitives/Timestamp.tsx
marcbouchenoire: I think it's just because those locales don't support `narrow` correctly (yet), from the browsers' side. The alternative to cherry-picking these locales like I did (using xhqmwd.csb.app) is to use `short` for all locales but it's a bit unfortunate to give up on "1h ago" (and go back to "1 hour ago") just because of a few locales imo.

PR #1293, opened 2023-11-07
Fix Video Comments example

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1677.

The issue was triggered because the Suspense version of `useUser` was used in a couple of places without any `<Suspense />` boundaries around it.


CTNicholas (approved):
Ahh, missing `Suspense`! Thanks for identifying the cause and making the fix ❤

PR #1294, opened 2023-11-08
Comments: Fix composer focus issues

marcbouchenoire:
This PR fixes an issue that could happen sometimes in Chrome/Safari, and always in Firefox: focusing the composer's editor would focus it as expected but then wouldn't create a selection in it meaning that it was impossible to interact before clicking on it manually again (to create a selection).

Fixes https://github.com/liveblocks/liveblocks.io/issues/1721.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### `@liveblocks/react-comments`
 
+- Fix `Composer` focus issues.
 - Improve relative date formatting for some locales. (e.g. the `"fr"`` locale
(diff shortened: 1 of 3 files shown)


nvie (approved):
Not sure what the difference is between "focusing" and "selecting" but it looks like this is a Slate-specific verb/meaning? Either way, thanks for fixing it! 🙌

nvie:
TIL something can be focused but not have a selection! I always thought that a component always has a selection, but maybe an empty one. I didn't know selection could be absent as a separate state from empty.

PR #1295, opened 2023-11-08
Comments: Logical properties

marcbouchenoire:
This PR adds linting rules (and applies them) to force us to only use [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values) to fully support RTL and other writing modes/directions, we already used them but it's too easy to forget about it.


packages/liveblocks-react-comments/src/styles/index.css
```diff
@@ -816,27 +816,27 @@
   display: flex;
   gap: calc(0.75 * var(--lb-spacing));
   align-items: center;
-  height: var(--lb-comment-avatar-size);
+  block-size: var(--lb-comment-avatar-size);
```
CTNicholas, line 819: I've been thinking about languages with vertical writing systems would like to display Comments, and it feels unusual, e.g. a composer with an input that's far taller than it is wide. But if they're typing in vertically, I guess it makes sense!

CTNicholas (approved):
Looks good!

packages/liveblocks-react-comments/src/styles/index.css
marcbouchenoire: We're applying vertical-specific rules to avoid adding linting exceptions but I think for now we shouldn't recommend vertical writing modes, I don't think we can say we fully support them like we do for RTL.

PR #1298, opened 2023-11-08
Docs for typed endpoint functions and `CommentBody` functions

CTNicholas:
Docs API reference for:
-  https://github.com/liveblocks/liveblocks/pull/1283
- https://github.com/liveblocks/liveblocks/pull/1301

**TODO**
- [x] Rooms
- [x] Storage
- [x] Yjs
- [x] Schema validation
- [x] `CommentBody` functions
- [x] Update email notification guide
- [ ] Check for any API changes in PRs before merging

PR #1301, opened 2023-11-10
Add `CommentBody` utilities in `@liveblocks/node`

marcbouchenoire:
This PR adds two utilities to `@liveblocks/node` to make working with the `CommentBody` format easier:
  - `getMentionedIdsFromCommentBody(body)` - Get a list of all mentioned IDs from a `CommentBody`.
  - `stringifyCommentBody(body, options)` - Convert a `CommentBody` to a string, either as plain text, HTML, or Markdown. It supports resolving mention IDs similarly to `@liveblocks/react` and overriding each element to control the formatting.

In addition to converting to a string, I explored other formats like https://github.com/portabletext/portabletext and https://github.com/resendlabs/react-email but that quickly became a scope creep so we'll start small, and a string should cover most use-cases.

Any feedback on the API/naming/exporting is welcome.

Fixes https://github.com/liveblocks/liveblocks.io/issues/1642 and https://github.com/liveblocks/liveblocks.io/issues/1645.


CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+
+### `@liveblocks/node`
+
+- Add utilities to work with the `CommentBody` format from Comments:
+  - `getMentionIdsFromCommentBody(body)` - Get a list of all mention IDs from a
+    `CommentBody`.
+  - `stringifyCommentBody(body, options)` - Convert a `CommentBody` to a string,
+    either as plain text, HTML, or Markdown. It supports resolving mention IDs
+    similarly to `@liveblocks/react` and overriding each element to control the
+    formatting.
```
nvie, line 11: May be nice to include a link to the documentation right in these release notes too.

nimeshnayaju:
Nice work @marcbouchenoire ! ❤️

nimeshnayaju:
My understanding is that exports from `@liveblocks/node` are primarily intended for server-side (Node.js) use, and aren't expected to be utilized in a browser environment. @nvie  Could you confirm if this understanding is accurate? However, the methods in question, such as those for processing and rendering comments, do not appear to rely on Node.js-specific APIs and seem universally applicable to both Node.js and browser environments. Is there a particular reason these methods are part of the @liveblocks/node package, given their apparent platform-agnostic nature? @marcbouchenoire

marcbouchenoire:
> However, the methods in question, such as those for processing and rendering comments, do not appear to rely on Node.js-specific APIs and seem universally applicable to both Node.js and browser environments. Is there a particular reason these methods are part of the @liveblocks/node package, given their apparent platform-agnostic nature?

Yeah it's tricky, because those could definitely have valid client-side use-cases (e.g. in tooltips, in a notification inbox, etc) and as is they will work but not sure people will like the idea of importing something client-side from a package with "node" in its name. One alternative would be to have them in `@liveblocks/core` and export them from both `@liveblocks/client` and `@liveblocks/node`? It does make everything more complex and blurs the line between client and Node.js code within `@liveblocks/core` (which I think @nvie is against).

CTNicholas:
It could be helpful to have an `index` given to you in `elements.paragraph`. Say I'm writing an email, and I want the first paragraph to be a larger text size, yet I know I can only use simple formatting to achieve this.

```ts
const stringComment = stringifyCommentBody(comment.body, {
  elements: {
    paragraph: ({ element, children }, index) =>
      `<p class="${index === 0 ? "large" : "normal"}>${children}</p>`,
  }
})
```

You could also use it for other things, such as inserting a drop cap.

CTNicholas:
I'm adding the docs for this in this PR, since they're modifying the same file:
- https://github.com/liveblocks/liveblocks/pull/1298

marcbouchenoire:
@CTNicholas I haven't found time to review the docs yet but thanks for working on them! And good idea regarding providing an index to elements, I just added that to all of them.

nimeshnayaju (approved):
Looks great! Thanks for working on this @marcbouchenoire

PR #1307, opened 2023-11-22
Fix bug in `toPlainLson` around falsey values

### LINKED ISSUE #1304: toPlainLson converts falsy values on a LiveObject to empty string (`''`)
Here the falsy values always come out as empty strings:

https://github.com/liveblocks/liveblocks/blob/77971fc300ecf4b74843fc0992b36f5ff13e014d/packages/liveblocks-core/src/crdts/utils.ts#L54C9-L57C11

```
Object.entries(lson.toObject()).map(([key, value]) => [
          key,
          value ? toPlainLson(value) : "",
        ])
```

Instead of that this function should filter out undefined values and call toPlainLson on any other value, like this:

```
Object.entries(lson.toObject())
    .filter((entry): entry is [string, Lson] => entry[1] !== undefined)
    .map(([key, value]) => [key, toPlainLson(value)])
```

nvie:
This fixes the following incorrect output:

Screen Shot 2023-11-22 at 14 09 44@2x

Fixes #1304.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v1.8.1
+
+- Fixes a bug in `toPlainLson` helper (#1304)
+
 # v1.8.0
 
 This release adds all the REST APIs as fully typed methods, and utilities to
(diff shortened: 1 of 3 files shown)


marcbouchenoire (approved):
😮

PR #1308, opened 2023-11-22
Add options to only revalidate when browser is online and is in focus or is visible

nimeshnayaju:
### Description

This PR updates the revalidation logic to gracefully handle network loss and browser not in focus. Additionally, the revalidation logic has been decoupled from comments to make the code more maintainable and scalable. The main revalidation and mutation logic hasn't changed, however.

You'll notice a `revalidation.ts` file which now exports the following hooks:

1. `useRevalidateCache` - accepts a cache manager that must follow a certain schema and a fetcher that is called when revalidation is triggered.
2. `useMutate` - accepts a cache manager and a revalidation function that is called at the end of the mutation.
3. `useAutomaticRevalidation` - accepts a cache manager and a revalidation function that is called when certain revalidation conditions are met.


packages/liveblocks-react/src/comments/lib/revalidation.ts
```diff
@@ -0,0 +1,364 @@
...
+      if (refreshInterval === 0) return;
+
+      revalidationTimerId = window.setTimeout(() => {
+        const isOnline = isOnlineRef.current;
+        const isVisible = document.visibilityState === "visible";
+
+        // Only revalidate if the browser is online AND document is visible AND there are currently no errors, otherwise schedule the next revalidation
+        if (isOnline && isVisible && !manager.cache?.error) {
+          // Revalidate cache and then schedule the next revalidation
+          void revalidateCache(true).then(scheduleRevalidation);
```
nimeshnayaju, line 298: We only revalidate if the browser is online **AND** document is visible **AND** there are currently no errors, otherwise we schedule the next revalidation.

packages/liveblocks-react/src/comments/lib/revalidation.ts
```diff
@@ -0,0 +1,364 @@
...
+  /**
+   * Subscribe to online and offline events to trigger a revalidation when the browser comes back online.
+   * Note: There is a 'navigator.onLine' property that can be used to determine the online status of the browser, but it is not reliable (see https://bugs.chromium.org/p/chromium/issues/detail?id=678075).
+   */
+  useEffect(() => {
+    function handleIsOnline() {
+      isOnlineRef.current = true;
+      const isVisible = document.visibilityState === "visible";
+
+      if (revalidateOnReconnect && isVisible) void revalidateCache(true);
```
nimeshnayaju, line 323: We perform a revalidation when the browser comes back online (and if the document is visible)

packages/liveblocks-react/src/comments/lib/revalidation.ts
```diff
@@ -0,0 +1,364 @@
...
+  /**
+   * Subscribe to focus and visibility change events to trigger a revalidation when the document becomes visible.
+   */
+  useEffect(() => {
+    function handleFocusOrVisibilityChange() {
+      const isVisible = document.visibilityState === "visible";
+      const isOnline = isOnlineRef.current;
+
+      if (revalidateOnFocus && isVisible && isOnline) {
+        void revalidateCache(true);
```
nimeshnayaju, line 347: We perform a revalidation when the document becomes visible or the window is focused (and if the browser is online)

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -114,732 +118,577 @@ export type ThreadsStateSuccess<TThreadMetadata extends BaseMetadata> = {
...
 
-    // Update the cache with the optimistic data
-    manager.cache = {
-      isLoading: false,
-      threads: options.optimisticData,
-    };
+    const cache = useSyncExternalStore(
+      manager.subscribe,
+      () => manager.cache,
+      () => manager.cache
```
marcbouchenoire, line 167: Should we re-use those and more importantly `useCallback` it?

packages/liveblocks-react/src/comments/lib/create-cache-manager.ts
```diff
@@ -0,0 +1,49 @@
+import { makeEventSource } from "@liveblocks/core";
+
+import type {
+  Cache,
+  CacheManager,
+  MutationInfo,
+  RequestInfo,
+} from "./revalidation";
+
+export default function createCacheManager<Data>(): CacheManager<Data> & {
```
marcbouchenoire, line 10: Since they're so tied together, I think we should keep them in a single `cache-manager.ts` file.

packages/liveblocks-react/src/comments/CommentsRoom.ts
```diff
@@ -100,10 +104,10 @@ export type ThreadsStateLoading = {
...
 };
 
-export type ThreadsStateError = {
+export type ThreadsStateSuccessOrError<TThreadMetadata extends BaseMetadata> = {
   isLoading: false;
-  threads?: never;
-  error: Error;
+  threads: ThreadData<TThreadMetadata>[];
+  error?: Error;
 };
```
marcbouchenoire, line 111: Can there be a state with both `threads` and `error` now? If not, we should keep the error and success states separate since they can be used for narrowing (e.g. check if `threads` exist, if so, `error` will be typed as `never` as expected)

marcbouchenoire (changes requested):
Thanks for working on this! I added a few comments and I have a question regarding polling: could we act on it instead of checking for visibility/connectivity at each already scheduled poll? I think it might be simpler to pause polling when going offline/offscreen (we could still check for online/visible at each poll just in case), then start it again when getting online/visible. Or at least I think we should make `revalidateCache` restart the polling loop when called, because currently if you get online 5 seconds before your scheduled poll there will one revalidation for getting online and still the scheduled one only 5 seconds later. Maybe we should move towards a single state machine that is responsible for both revalidations/etc and polling, I think it could prevent unwanted scenarios like that.

packages/liveblocks-react/src/comments/CommentsRoom.ts
nimeshnayaju: That is correct. `threads` and `error` can both exist at the same time. I think this could provide better user experience to the end users - e.g., by keeping the existing threads data, users are not left with an empty or broken page if there's an error during revalidation. This ensures that the user experience remains uninterrupted. Also, since we have an error retry mechanism in place (and revalidation on browser 'online' events), and say network connection is lost only briefly, there won't be a drastic shift in the threads list.

Did you have a particular idea in mind with threads and error being mutually exclusive? @marcbouchenoire

packages/liveblocks-react/src/comments/lib/create-cache-manager.ts
nimeshnayaju: They are quite relevant to each other. However, the revalidation methods only require the cache manager to be a certain shape. It doesn't really care of the implementation of it. For instance, in this particular cache manager implementation, we have an extra `subscribe` property which isn't relevant at all to the revalidation logic but is only used by our `useThreads` hook.

That being said, I do think there is still a large enough coupling between the two (especially with how strict the shapes are), so it might make sense to keep the two together for now and think of segregating later in the future.

packages/liveblocks-react/src/comments/CommentsRoom.ts
marcbouchenoire: I just thought we didn't have that yet but I agree it's better this way 👍 I do think we could name the type something else, `ThreadsStateResolved` maybe.

packages/liveblocks-react/src/comments/CommentsRoom.ts
nimeshnayaju: 1️⃣ I think it might be okay for now to keep it inline for better readability despite the repetition. Also, the repeated code snippet is quite small, but might be worth keeping an eye on in future. Do you have preference for one over the other? 

2️⃣ I believe it is fine without memoizing. From what I understand, the value returned by selector function in `useSyncExternalStore` is what is required to be memoized or cached, not necessarily the function itself.

https://react.dev/reference/react/useSyncExternalStore#im-getting-an-error-the-result-of-getsnapshot-should-be-cached
> React will re-render the component if getSnapshot return value is different from the last time. This is why, if you always return a different value, you will enter an infinite loop and get this error.


https://github.com/reactwg/react-18/discussions/86
> Selectors no longer need to be memoized
>
> Inline selectors are a common feature of state libraries, especially those that provide a hook-based API, such as Redux’s useSelector.

packages/liveblocks-react/src/comments/lib/create-cache-manager.ts
marcbouchenoire: I don't think we should focus too much on making it generic tbh, we've (I've 🙈) tried in the past with `AsyncCache` but then use-case A needs batching, use-case B needs optimistic updates' rollbacks while use-case A doesn't, and use-case C (the threads use-case) needs specific optimizations (pagination, granularity, etc). If it's easier for us to build/optimize/maintain 2 or 3 use-case-specific caches I don't think it's too bad.

packages/liveblocks-react/src/comments/CommentsRoom.ts
marcbouchenoire: I tend to be overzealous with these things (cc: @nvie), you're right it's fine as is 👍

packages/liveblocks-react/src/comments/CommentsRoom.ts
nimeshnayaju: `ThreadsStateResolved` does sound more representative! I've added a small comment before the type to explain the type in slightly more detail.

packages/liveblocks-react/src/comments/lib/create-cache-manager.ts
nimeshnayaju: Agreed! My original intention was not to make the code more generic but more from the maintainability perspective. It was just moving the revalidation and mutation logic out of the `createCommentsRoom` method without updating any logic (except for the new network loss condition) so that the code is slightly more maintainable and easier to read. 

I agree it is a very hard balance though.

nimeshnayaju:
>I have a question regarding polling: could we act on it instead of checking for visibility/connectivity at each already scheduled poll? I think it might be simpler to pause polling when going offline/offscreen (we could still check for online/visible at each poll just in case), then start it again when getting online/visible. 

Good point! It's definitely more efficient to pause polling while browser is offline or document is not in focus. I have updated the code to do exactly that. Thanks for the pointer!

> Or at least I think we should make `revalidateCache` restart the polling loop when called, because currently if you get online 5 seconds before your scheduled poll there will be one revalidation for getting online and still the scheduled one only 5 seconds later. Maybe we should move towards a single state machine that is responsible for both revalidations/etc and polling, I think it could prevent unwanted scenarios like that.

What I've done for now is restart polling every time browser comes back online or document is visible again. So, if you come back online 5 seconds before the scheduled poll, it'll reset the polling timer. But, I agree this can be hard to maintain in the long term.

nimeshnayaju:
Added a few more tests! @marcbouchenoire Do you mind taking a look when you get a chance?

packages/liveblocks-react/src/comments/lib/create-cache-manager.ts
nimeshnayaju: I think the new tests really take advantage of this decoupling as the hooks can be tested independently without the extra overhead of setting up Liveblocks client, creating room context, etc. Here is the test file: [Link](https://github.com/liveblocks/liveblocks/blob/629fa11d181eb07cf01e0f3ad84efa872a6cf911/packages/liveblocks-react/src/__tests__/revalidation.test.tsx)

marcbouchenoire (approved):
Thanks for making the remaining changes, LGTM 👏

PR #1309, opened 2023-11-23
Fix history `pause` bug

nvie:
This fixes a bug where calling `room.history.pause()` more than once could lead to history loss.

This happens in the following sequence:

```tsx
room.history.pause();
/* mutation 1 */
room.history.pause();
/* mutation 2 */
room.history.resume();
```

The second "pause" call here will unintentionally lose the addition of the inverse operation of mutation 1 to the undo stack.

This PR fixes this by making any `.pause()` call beyond the first a no-op.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v1.8.1
+
+- Fix a bug where pausing history more than once can lead to history loss
+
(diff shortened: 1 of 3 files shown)


marcbouchenoire (approved):
🧹

PR #1310, opened 2023-11-24
Improve overlap between `@liveblocks/client` and `@liveblocks/node`

marcbouchenoire:
This PR marks `@liveblocks/core` as a normal dependency of `@liveblocks/node` which allows us to share more code/types between `@liveblocks/node` and `@liveblocks/client`, to improve DX (hybrid code/types should be importable from both browser and non-browser contexts).

`@liveblocks/core` was already Node.js-compatible so I just added a safety check in the `build` script, otherwise most of the work is around harmonizing exports between `@liveblocks/client` and `@liveblocks/node`:
- Harmonize exported types
- Move the new `CommentBody` utilities to `@liveblocks/core` and export them from both `@liveblocks/client` and `@liveblocks/node`

While I was there, I also added the `detectDupes` logic to the packages that didn't have it.

### To-do:
- [x] How do we document the `CommentBody` utilities if they're exported from both `@liveblocks/client` and `@liveblocks/node`? cc: @CTNicholas


packages/liveblocks-react-comments/src/version.ts
```diff
@@ -0,0 +1,6 @@
+declare const __VERSION__: string;
+declare const TSUP_FORMAT: string;
+
+export const PKG_NAME = "@liveblocks/yjs";
```
nvie, line 4: This should be:

```suggestion
export const PKG_NAME = "@liveblocks/react-comments";
```

packages/liveblocks-yjs/src/index.ts
```diff
@@ -12,7 +12,6 @@ import type * as Y from "yjs";
 import { Awareness } from "./awareness";
 import yDocHandler from "./doc";
 import { PKG_FORMAT, PKG_NAME, PKG_VERSION } from "./version";
-2;
```
nvie, line 15: Nice catch!

packages/liveblocks-core/src/comments/comment-body.ts
```diff
@@ -149,6 +148,10 @@ export type StringifyCommentBodyOptions<
   ) => PromiseOrNot<(TUserMeta["info"] | undefined)[] | undefined>;
 };
 
+function isSomething<T>(input: null | undefined | T): input is T {
+  return input !== null && input !== undefined;
+}
```
nvie, line 153: I assume this helper exists purely to be able to filter null elements from an array, and retain that knowledge statically in the types?

Sometimes I use a `.flatMap()` for use cases like this:

```ts
// ❌ Instead of…
someList
  .map(x => condition ? x : null)
  .filter(isSomething);

// ✅ Do a flat map instead
someList
  .flatMap(x => condition ? [x] : []);
```

This effectively does the same thing, and "maps and filters" the elements in a single pass, by either returning a 1-element-list, or an empty list. It is still 100% type safe, and avoids the need for helpers like this.

nvie (approved):
Nice!! Glad this is now much simpler and more intuitive again, as it should be! Thanks, @marcbouchenoire! 🙌

packages/liveblocks-react-comments/src/version.ts
marcbouchenoire: Oof, thanks!

packages/liveblocks-core/src/comments/comment-body.ts
marcbouchenoire: Yep exactly, it's purely here because of TS's `filter` limitations.

Ooh interesting, I'll switch to that!

PR #1315, opened 2023-11-24
Fix an invisible hidden space in string literal

nvie:
Removes an invisible space char from this string literal.


marcbouchenoire (approved):
🤔

PR #1316, opened 2023-11-26
types: fix `getActiveUsers` types

antoinewg:
### Fixing a bug

The types of `getActiveUsers` in the node client seem to be incorrect.

See [implementation](https://github.com/liveblocks/liveblocks/blob/b071a8049a877f6769505d48f4c6a22a3f16aee9/starter-kits/nextjs-starter-kit/lib/server/liveblocks/getActiveUsers.ts#L29) in next starter kit: we get data.data.

To reproduce, run script, after replacing `LIVEBLOCKS_SECRET_KEY`, and `roomid`.

```ts
import dotenv from "dotenv";

dotenv.config();

const main = async () => {
  const client = new Liveblocks({
    secret: process.env.LIVEBLOCKS_SECRET_KEY ?? "",
  });

  const activeUsers = await client.getActiveUsers("<roomid>");
  console.log(activeUsers);
};

main();

// prints
// {
//   data: [
//     {
//       type: 'user',
//       connectionId: <num>,
//       id: '<user id>',
//       info: [Object]
//     }
//   ]
// }

// but types says:
// const activeUsers: RoomUser<unknown>[]
```

Note: I'm changing the types and not the response to avoid breaking changes.


nimeshnayaju:
Hey @antoinewg, thank you for bringing this issue to our attention and for submitting the PR. I've reviewed the discrepancy you've identified between the implementation of getActiveUsers and its type definition. Your observation about the mismatch with the OpenAPI schema and the actual response structure is correct.

@nvie  I'd like to gather your thoughts on the following options:

1️⃣ We abide by the typescript type and have `getActiveUsers()` in fact return an `RoomUser[]` - this would mean we are doing `(await res.json().data`) on our end and if in future, we introduce some extra property on the API response other than just `data` (e.g. introduce pagination, although I am not sure we would), we may have to introduce a larger breaking change.

2️⃣ We could do what @antoinewg suggested and maintain the structure as close to the actual API response as possible. This is consistent with how `getRooms()` works, returning an object with a `data` property that contains an array.

nvie:
I think it makes sense to have all the top-level "list" endpoints return responses/types with a top-level `data` field that is an array. Then there is at least consistency.

I'd propose:
- All endpoints always return a JSON **object** at the top-level.
- All endpoints that return a **list** of resources, will have a top-level `data` key which will always be an array. Those endpoint _may_ return extra top-level fields for pagination, if pagination makes sense for that endpoint.
- All endpoints that return a **single** resource, can just return that resource directly as the top-level object.

To tie that back to this PR: in this case, I'd be in favor of doing what @antoinewg already did here: expose the `data` field here at the type level in the return type (aka I prefer solution 2️⃣) 👍 

@nimeshnayaju We should double-check all endpoints to see if they match the rule I suggested above. I think they all do once `getActiveUsers`'s type is updated.

nimeshnayaju:
Thanks for the input @nvie !

> We should double-check all endpoints to see if they match the rule I suggested above. I think they all do once `getActiveUsers`'s type is updated.

We also have the method `getThreadParticipants()` that is currently returns `{ participantIds: ThreadParticipants[]` } instead of `{ data: ThreadParticipants[] }` but there is discussion about potentially deprecating `getThreadParticipants()` for another method that should conform to the rule you suggested!

nvie:
> We also have the method `getThreadParticipants()` that is currently returns `{ participantIds: ThreadParticipants[]` } instead of `{ data: ThreadParticipants[] }` but there is discussion about potentially deprecating `getThreadParticipants()` for another method that should conform to the rule you suggested!

Nice! 👍

nimeshnayaju (approved):
Added a simple test to the PR. Thanks for reporting and working on this @antoinewg ! ❤️

nimeshnayaju:
@antoinewg I've cherry picked your commits from the PR into [this new PR ](https://github.com/liveblocks/liveblocks/pull/1322) (We were having some problems with correctly running the E2E tests on a fork). Thank you for submitting this PR again!

PR #1325, opened 2023-11-29
Add createComment method to node client

FlowFlorent:

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -4,6 +4,7 @@
  * @liveblocks/core has browser-specific code.
  */
 import type {
+  CommentBody,
   CommentData,
   IUserInfo,
   Json,
@@ -922,6 +923,38 @@ export class Liveblocks {
 
     const res = await this.get(
       url`/v2/rooms/${roomId}/threads/${threadId}/comments/${commentId}`
+    );
+    if (!res.ok) {
+      const text = await res.text();
+      throw new LiveblocksError(res.status, text);
+    }
+    return (await res.json()) as Promise<CommentData>;
+  }
+
+  /**
+   * Creates a comment.
+   * 
+   * @param params.roomId The room ID to create the comment in.
+   * @param params.threadId The thread ID to create the comment in.
+   * @returns The created comment.
+   */
+  public async createComment(params: {
+    roomId: string;
+    threadId: string;
+    comment: {
+      userId: string;
+      /**
+       * ISO 8601 date string
+       */
+      createdAt?: string;
+      body: CommentBody;
+    };
+  }): Promise<CommentData> {
+    const { roomId, threadId, comment } = params;
+
+    const res = await this.post(
+      url`/v2/rooms/${roomId}/threads/${threadId}/comments`,
+      comment
     );
     if (!res.ok) {
       const text = await res.text();


packages/liveblocks-node/src/client.ts
```diff
@@ -929,6 +930,38 @@ export class Liveblocks {
...
+   */
+  public async createComment(params: {
+    roomId: string;
+    threadId: string;
+    comment: {
+      userId: string;
+      /**
+       * ISO 8601 date string
+       */
+      createdAt?: string;
```
GuillaumeSalles, line 949: This should be a `Date`. We decided to transform all the ISO 8601 strings to `Date` when fetching so we should do the same when creating an object.

packages/liveblocks-node/src/client.ts
```diff
@@ -929,6 +930,38 @@ export class Liveblocks {
     }
     return (await res.json()) as Promise<CommentData>;
   }
+
+  /**
+   * Creates a comment.
+   * 
+   * @param params.roomId The room ID to create the comment in.
+   * @param params.threadId The thread ID to create the comment in.
+   * @returns The created comment.
```
nimeshnayaju, line 939: **Nitpick:** Should we also include the comment in this JSDoc comment? even a very brief one.

packages/liveblocks-node/src/client.ts
```diff
@@ -929,6 +931,70 @@ export class Liveblocks {
...
+   *
+   * @param params.roomId The room ID to create the comment in.
+   * @param params.threadId The thread ID to create the comment in.
+   * @returns The created comment.
+   */
+  public async createComment(params: {
+    roomId: string;
+    threadId: string;
+    comment: {
+      userId: string;
```
nimeshnayaju, line 946: I am brainstorming other names for this particular variable. What do you think of `data` instead of `comment`? Would it be more consistent to our other names? I know we use the term `data` on our returned value quite a lot - Examples:

``` ts
public async getThreads(...): Promise<{ data: ThreadData[] }>()


public async getRooms(...): Promise<{
  nextPage: string | null;
  data: RoomInfo[];
}>


public async getActiveUsers(...): Promise<{ data: RoomUser[] }
```

packages/liveblocks-node/src/client.ts
```diff
@@ -929,6 +931,205 @@ export class Liveblocks {
...
+
+    const res = await this.post(
+      url`/v2/rooms/${roomId}/threads/${threadId}/comments`,
+      comment
+    );
+    if (!res.ok) {
+      const text = await res.text();
+      throw new LiveblocksError(res.status, text);
+    }
+    return (await res.json()) as Promise<CommentData>;
```
nimeshnayaju, line 964: `CommentData` includes a `createdAt`, and may include `editedAt` and `deletedAt` field. The `createComment` currently doesn't convert them into a `Date` object before returning them. We should consider doing that, as we already do that for other methods like `getRoom`, `createRoom`, etc.

nimeshnayaju: @FlowFlorent I've created two helper methods: `convertToCommentData` and `convertToThreadData` that takes the response returned by the API and massages this object so that the date strings are formatted as Date objects, which are easier to consume by clients. These methods are currently used by `editComment`, `createThread`, `editThread`. We might want to do the same for `createComment`.

https://github.com/liveblocks/liveblocks/blob/143e3288f7c47e1929adb73cf6165a98258cfe94/packages/liveblocks-node/src/client.ts#L217-L236

https://github.com/liveblocks/liveblocks/blob/143e3288f7c47e1929adb73cf6165a98258cfe94/packages/liveblocks-node/src/client.ts#L245-L259

packages/liveblocks-node/src/client.ts
nimeshnayaju: I've updated the implementation of `createComment` to the following:

https://github.com/liveblocks/liveblocks/blob/666e47aca2c4365331421f7a50effaf3f44cc790/packages/liveblocks-node/src/client.ts#L1020-L1053

packages/liveblocks-node/src/client.ts
nimeshnayaju: I ended up updating the the `createComment` method to use the `convertToCommentData`. See https://github.com/liveblocks/liveblocks/pull/1325#discussion_r1413584082

nimeshnayaju:
I've implemented the rest of the methods and added tests for some. Could you please take a look when you are free @FlowFlorent ?

packages/liveblocks-node/src/client.ts
```diff
@@ -929,6 +1016,256 @@ export class Liveblocks {
...
+   * @param params.thread.comment.body The body of the comment.
+   * @returns The created thread. The thread will be created with the specified comment as its first comment.
+   */
+  public async createThread<
+    TThreadMetadata extends BaseMetadata = never,
+  >(params: {
+    roomId: string;
+    data: {
+      metadata?: [TThreadMetadata] extends [never]
+        ? Record<string, never>
```
nimeshnayaju, line 1128: @CTNicholas @nvie @FlowFlorent Any suggestion on the name of this variable. The information in this variable is generally used in the request body. I've currently named it `data` but we use the name `params` to describe request body in `createRoom` method, so I am worried about inconsistency here. Would appreciate your thoughts:

For context, this is the implementation of `createRoom` method:

```ts
public async createRoom(
  roomId: string,
  params: {
    defaultAccesses: RoomPermission;
    groupsAccesses?: RoomAccesses;
    usersAccesses?: RoomAccesses;
    metadata?: RoomMetadata;
  }
): Promise<RoomInfo> {
  const { defaultAccesses, groupsAccesses, usersAccesses, metadata } = params;

  const res = await this.post(url`/v2/rooms`, {
    id: roomId,
    defaultAccesses,
    groupsAccesses,
    usersAccesses,
    metadata,
  });

  ..
}
```

packages/liveblocks-node/src/client.ts
```diff
@@ -929,6 +1016,256 @@ export class Liveblocks {
...
+    if (!res.ok) {
+      const text = await res.text();
+      throw new LiveblocksError(res.status, text);
+    }
+
+    return this.convertToCommentData((await res.json()) as CommentDataOriginal);
+  }
+
+  /**
+   * Deletes a comment.
```
FlowFlorent, line 1091:
```suggestion
   * Deletes a comment. If there are no other comments in the thread, the thread is also deleted.
```

FlowFlorent:
@CTNicholas @pierrelevaillant As the functionality is new, it'd be good to release it with a`beta` tag in case we need to make a change.

What would be the best way to display the information?

Screenshot 2023-12-04 at 2 02 56 PM
Screenshot 2023-12-04 at 2 03 19 PM

packages/liveblocks-node/src/client.ts
FlowFlorent: This is really nice! Good call on the abstraction.

packages/liveblocks-node/src/client.ts
FlowFlorent: I would still go for `data` on other endpoints, rational being that we are nesting the data as part of a `params` object for the comment methods.

`params` = everything to achieve the call
`data`= data specific to the body

packages/liveblocks-node/src/client.ts
nimeshnayaju: Done! Nice addition

nimeshnayaju:
I've also updated the existing `getThreads`, `getThread`, and `getComment` methods to return the modified data (that converts any ISO date strings to `Date` objects). This will therefore be a breaking change, but I think it's a useful breaking change, and since comments is still in open beta, I believe it is warranted. @nvie @FlowFlorent @CTNicholas any thoughts?

https://github.com/liveblocks/liveblocks/blob/1d500a2947f56eeef60cbc91c8e209d2d1969403/packages/liveblocks-node/src/client.ts#L969-L983

packages/liveblocks-node/src/client.ts
```diff
@@ -84,6 +86,66 @@ type RoomInfoOriginal = Omit<RoomInfo, "lastConnectionAt" | "createdAt"> & {
...
+} & (
+    | {
+        body: CommentBody;
+        deletedAt?: never;
+      }
+    | {
+        body?: never;
+        deletedAt: Date;
+      }
+  );
```
nvie, line 116: This type setup seems a bit hard to maintain to me. I have a couple of suggestions to make maintaining this over time easier. Could we "flip this around", i.e. making this new `CommentData` (with possible Date values) the rich base type, and deriving the `CommentDataOriginal` (with strings) from it? That way, you go from a richer to a less rich type, which can be done using a generic transformation function that converts `Date` types to `string` types. The big benefit is that those types are more readable _and_ they don't require the maintenance of keeping the field union up to date, any time a new Date field would get added or removed. The same maintenance currently lies on the optional-or-required field information (on L100-101).

Suggestion:

- In `@liveblocks/core`, expose `CommentData` as a rich type that has `Date` values.
- In `@liveblocks/core`, also expose `CommentDataPlain`, which is defined as `CommentDataPlain = DatesToStrings<CommentData>`.
- Use `CommentDataPlain` in our Node endpoint responses.
- Use `CommentData` (with Dates) in our client's REST API responses.

Here's a TypeScript playground with an example of such helper type:

https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcBSBnCA7OBedLbAZQGMBDAG3KjgB9CcBBKKcxex7AeQCMArYKRgBuAFChIsBMlSYcZKjXxwMMKAEtsAc07YArgFtewWg14QIlYOVwMDlSuMnR4SFFxZsOBedgDaALrO4K4yHn58gsIqAN5w-gDWwIgAXKrqWtqB6X6c+tgAJsAAZlrAhXAAvuJiAPR1cAC0La1t7R2dXd09XWJi7qgAIuQwwBgAKhDEmToYADwTAHwqE3CgY0UYcCNjcAD8GZo6cOnx-gDScFpwyYgQJXATOTuj41Mzx9oLE5eBKzV+oM4AAlDSkAAWcTEcFhcFIUBsY0KTBg6V2wHEcLgxWsyNR+3RbyxcMR5EKOEoHGw42R6TJFOwVLgsWwEAA7kS9vZ9I4RNUgiTYdhyIZgOk1F8hddCukDMZTNLyNpgIS4PKTFBpWM1GrJVlOPqdEE9LynGIqkDZHAAAqWWgEDGTaazb7zMGQpbiBpwgB6+zEQA

packages/liveblocks-node/src/client.ts
```diff
@@ -84,6 +86,66 @@ type RoomInfoOriginal = Omit<RoomInfo, "lastConnectionAt" | "createdAt"> & {
   createdAt?: string;
 };
 
+type CommentReactionUser =
+  CommentDataOriginal["reactions"][number]["users"][number];
```
nvie, line 90: Maybe expose this type as a building block in `@liveblocks/core`, and use it in the definition of `CommentDataOriginal`, instead of defining it as one big type and then "reaching into" that big type to get the smaller type out and alias it. Keeping types as small as possible generally really helps with composability and avoids having to write types like this.

CTNicholas:
> @CTNicholas @pierrelevaillant As the functionality is new, it'd be good to release it with a`beta` tag in case we need to make a change. The tag will mainly be for endpoints doing mutations. Our GET endpoints are stable.
> 
> What would be the best way to display the information?

I'll figure this one out!

FlowFlorent:
Thank you @CTNicholas, such a quick turnaround!

packages/liveblocks-node/test-d/client.test-d.ts
```diff
@@ -1,15 +1,57 @@
...
+  expectType<CommentBodyBlockElement[] | undefined>(comment.body?.content);
+
+  const thread = await client.getThread<Record<string, string>>({
+    roomId: "room-id",
+    threadId: "th_threadId",
+  });
+  expectType<"thread">(thread.type);
+  expectType<string>(thread.id);
+  expectType<string>(thread.roomId);
+  expectType<Date>(thread.createdAt);
```
nvie, line 39: Nice! This is exactly the kind of test that will let a change inside **core** fail here in the **node** package if they're part of the public API, which may or may not be intentional! In the future, someone will be happy this test exists, I'm sure 😊

packages/liveblocks-node/src/__tests__/client.test.ts
```diff
@@ -157,7 +198,178 @@ describe("client", () => {
...
+            return HttpResponse.json(comment);
+          }
+
+          return HttpResponse.error();
+        }
+      )
+    );
+
+    const client = new Liveblocks({ secret: "sk_xxx" });
+    const res = await client.editComment({
```
nvie, line 226: I've not been part of this discussion, so maybe this has already been discussed at length and decided. Please feel free to discard this comment in that case!

I noticed the verb "edit" here. In an API, I'd personally prefer to use the more common verb "update". To me "edit" is more of a UI verb, e.g. you put the UI into edit mode, and then when you submit the changes, you _update_ the resource by making the API call.

(Again, feel free to disregard if this was already discussed/decided on.)

nvie (approved):
Nice improvements, @nimeshnayaju — this is looking great! 💯

packages/liveblocks-node/src/__tests__/client.test.ts
nimeshnayaju: We discussed this briefly in our check in yesterday with @marcbouchenoire , @FlowFlorent and myself. The reason we opted for the verb `edit` for the node methods is mostly to keep it consistent with our client hooks/functions (`useEditThreadMetadata`, `useEditComment`).

FlowFlorent:
@CTNicholas good catch! I will adjust it.

packages/liveblocks-core/src/comments/types/Reaction.ts
```diff
@@ -0,0 +1,9 @@
+import type { DateToString } from "./DateToString";
+
+export type Reaction = {
+  emoji: string;
+  createdAt: Date;
```
nimeshnayaju, line 5: @marcbouchenoire This type is slightly different from the `CommentReaction` type. What do you think we should name this? I feel like `Reaction` is too close to `CommentReaction` but there is a distinction between the two.

`Reaction`:
```ts
export type Reaction = {
  emoji: string;
  createdAt: Date;
  userId: string;
}
```

`CommentReaction`:
```ts
export type Reaction = {
  emoji: string;
  createdAt: Date;
  userId: CommentReactionUser[];
}

Comment (marcbouchenoire) [reply packages/liveblocks-core/src/comments/types/Reaction.ts:5]: I also feel like it should still include "comment", what do you think of something like `CommentIndividualReaction`/`CommentReactionIndividual`?

Comment (nimeshnayaju) [reply packages/liveblocks-core/src/comments/types/Reaction.ts:5]: Definitely better than just `Reaction`. We do have to be careful about the names because this type will be exported from the `node` package. @nvie Do you have any thoughts?

Comment (marcbouchenoire) [reply packages/liveblocks-core/src/comments/types/Reaction.ts:5]: My ideal choice would be something around `User` instead of `Individual`/`Single` but there's `CommentReactionUser` already (although it's not exported so we could rename this one to have a better name for the public `node` one)

Comment (nimeshnayaju) [reply packages/liveblocks-core/src/comments/types/Reaction.ts:5]: What I've done for now since our current `CommentReactionUser` is not exported, is move the type to inline:

```ts
export type CommentReaction = {
  emoji: string;
  createdAt: Date;
  users: {
    id: string;
  }[];
};
```

and renamed the new `Reaction` type to `CommentUserReaction`:

```ts
export type CommentUserReaction = {
  emoji: string;
  createdAt: Date;
  userId: string;
};
```

PR #1328, opened 2023-11-30
Use typed REST APIs in Yjs guide

CTNicholas:
This Yjs guide still used `fetch` instead of our typed functions.


marcbouchenoire (approved):
So much cleaner 🧹

PR #1329, opened 2023-11-30
Webhooks: add explicit error if body is not a string

FlowFlorent:
#### Description

Adds an explicit error when `rawBody` is not a string. It will ensure users have an easier way to debug message verification.

### DIFF
--- a/packages/liveblocks-node/src/webhooks.ts
+++ b/packages/liveblocks-node/src/webhooks.ts
@@ -32,8 +32,9 @@ export class WebhookHandler {
     const { webhookId, timestamp, rawSignatures } = this.verifyHeaders(headers);
 
     if (typeof rawBody !== "string") {
-      throw new Error(`Invalid body, must be a string, got "${typeof rawBody}" instead. 
-      It is likely that you need to JSON.stringify the body before passing it.`);
+      throw new Error(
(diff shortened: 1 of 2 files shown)


packages/liveblocks-node/src/webhooks.ts
```diff
@@ -27,13 +27,18 @@ export class WebhookHandler {
...
-    const { webhookId, timestamp, rawSignatures } = this.verifyHeaders(
-      request.headers
-    );
+    const { headers, rawBody } = request;
+
+    const { webhookId, timestamp, rawSignatures } = this.verifyHeaders(headers);
+
+    if (typeof rawBody !== "string") {
+      throw new Error(`Invalid body, must be a string, got "${typeof rawBody}" instead. 
+      It is likely that you need to JSON.stringify the body before passing it.`);
```
nvie, line 36: 1. In error messages, prefer not using `\n` since they can get cropped weirdly depending on where they display. Just using a space should be enough.
2. Maybe say which field this is checking in the error message, i.e. say "the `rawBody` field must be …".

nvie (approved):
Nice! Thanks for adding this check, it will cause a lot less confusion for users. 💯

packages/liveblocks-node/src/webhooks.ts
FlowFlorent: Thanks, will adjust!

PR #1332, opened 2023-12-06
Add force option to `link-example-locally.sh`

marcbouchenoire:
Sometimes I want to link an example even though I have some uncommitted changes. (e.g. I have a stash to switch an example to the DEV environment, and I want to both apply that stash and link the example)


nvie (approved):
Great for those who live on the edge and know what they're doing! 👌

PR #1349, opened 2023-12-12
Add support for query options on `useThreads`

nimeshnayaju:
Add an option to `useThreads` to filter threads based on their metadata.

## Definition

```typescript
useThreads(options?: { query?: { metadata?: Partial<TThreadMetadata> }});
```

TODOS:

- [x] Support passing no options
- [x] Improve polling
- [x] Fix tests that are failing due to change in cache manager signature
- [x] Support errors

Each `useThreads` can be passed filter options that is used to determine what threads the `useThreads` hook should return. We collect all these filter options in a map and make `getThreads` call for each option. Then, we normalize this list and replace the cache with the normalized list of threads. Any update to the cache is made via the `setCache` method inside the threads cache manager. Each time a `setCache` method is called, we also calculate threads for each filter  option and store these filtered threads for each filter option.

### High level architecture:

1. We have a main cache which stores all threads we are currently interested in. Subsets of this main cache is stored in the cache state of filter options.
2. Each filter option has its own cache state. This cache state is either in loading state or has data in it. Any data in the cache state of filter options are derived from the main cache.
3. The cache state of each filter option is initially set to loading state until the next revalidation or optimistic mutation updates the main cache.
4. Cache state for each filter option is calculated from the main cache on each update to the main cache.
5. We store each filter option in an array and during revalidation, we make a call to fetch threads using each of those filters.

### Possible improvements

- [ ] Currently, we store all filter options and make a `getThreads` for each option, which is okay but can be improved in certain cases. Consider the case where one of the filter options applied is `{ resolved: true }` and the other one applied is `{ resolved: true, page: 10 }`. In the current setup, we make two `getThreads` call one with each of the aforementioned filter options. But, notice how the filter `{ resolved: true, page: 10 }` is a subset of the filter `{ resolved: true }`, so we could have only made one single call to `getThreads` with filter `{ resolved: true }`. For this, we will need a function, let's call it `isSubset(subset, superset)` that checks if an option filter is a subset of another.
- [ ] Similar to the above mentioned improvement, but slightly different. During initial rendering, we set the cache state of a filter option to `{ isLoading: true }`. But, say we already have mounted a component that calls `useThreads` with `{ resolved: true }` filter option and we have all those threads in the main cache. Now, say after clicking a button, a new component is rendered that calls `useThreads` with `{ resolved: true, page: 10 }` filter. Ideally, we don't want to set the cache state for this filter option to `{ isLoading: true }` because this filter option is a subset of `{ resolved: true }`, which we already have data for. For this optimization as well, we also need the `isSubset` method.
- [ ] Each sub-cache associated to a filter option is initially set to `{ isLoading: true }` and this state is updated to show filtered threads for the filter option on the next call to `setCache`, which normally happens after the initial revalidation (on mount) is completed. But, there can be cases when the `setCache` can be triggered by a mutation before the revalidation is complete, which results in an empty threads list for the filter, until a new revalidation is successful (which happens immediately after a mutation)
- [x] Better support for suspense. The suspense `useThreads` hook currently doesn't delete the query info after completely removing the component from React tree. I don't think this will cause any UX problems other than a less optimal fetcher call but this is something we should be wary of.


marcbouchenoire:
@nimeshnayaju @GuillaumeSalles I asked this in the past but it's buried deep in Slack: should we name the option `query` instead of something more fitting like `filter`? To me, `query` sounds like it will map to all options available on the query string but I assume for things like pagination in the future we'll make the API more intuitive than asking for query parameters.

nimeshnayaju:
With the new changes, the global fetcher (collection of `getThreads` for each filter option) is only called during polling and revalidation after a mutation. During initial revalidation of `useThreads`, the filter that is passed as option is used, and this request is deduped too, which makes the experience slightly smoother and faster in general.

GuillaumeSalles (approved):
🚀  Amazing job @nimeshnayaju

PR #1351, opened 2023-12-13
Comments: Unread status and inbox notifications

marcbouchenoire:
This PR will contain the work to visualize unread comments within threads, the cache refactoring necessary to prepare for inbox notifications (which are tightly related to unread status), and inbox notifications components.

### To-do

> Searchable `TODO` comments: `// [comments-unread] TODO:`


packages/liveblocks-react/src/types.ts
```diff
@@ -83,6 +80,122 @@ export type UserState<T> =
...
+
+export type ThreadsState<TThreadMetadata extends BaseMetadata> =
+  | ThreadsStateLoading
+  | ThreadsStateResolved<TThreadMetadata>;
+
+export type InboxNotificationsStateLoading = {
+  isLoading: true;
+  inboxNotifications?: never;
+  error?: never;
+  loadMore: () => void;
```
nimeshnayaju, line 151: Should `loadMore` be returned when the inbox notifications query is still loading?

```tsx
export type InboxNotificationsStateLoading = {
  isLoading: true;
  inboxNotifications?: never;
  error?: never;
  loadMore: () => void;
};
```

marcbouchenoire: @nimeshnayaju It should always be returned yes, to avoid having to conditionally call it. But we can decide to make it a noop while the query is still loading, or keep the currently loading query and start the next one already?

GuillaumeSalles (approved):

PR #1355, opened 2023-12-15
`useThreads` filter improvements + Set up Inbox notifications cache

nimeshnayaju:
This PR improves the following things about the `useThreads` hook with the new filter options:

1. Simpler revalidation logic during initial revalidation. And, a more unified initial revalidation logic for both normal and suspense `useThreads` hook.
2. Each room has their own revalidation manager (which handles race conditions better during mutation and revalidation).
3. Prepare for inbox notifications cache management.
4. Moved all logic inside `CommentsRoom.tsx` to `factory.tsx` to unify both comment and room related processes. Similar intentions as https://github.com/liveblocks/liveblocks/pull/1351
6. Empty filter options are normalized. Example: `{ }`, `{ query: { } }` and `{ query: { metadata: { } } }` are all normalized to the same filter option.


packages/liveblocks-react/src/index.ts
```diff
@@ -15,3 +15,6 @@ export type {
 // Re-exports from @liveblocks/client, for convenience
 export type { Json, JsonObject } from "@liveblocks/client";
 export { shallow } from "@liveblocks/client";
+
+// Re-exports from @liveblocks/core, for convenience
+export type { ThreadsFilterOptions } from "@liveblocks/core";
```
nimeshnayaju, line 20: We re-export `ThreadsFilterOptions` from `@liveblocks/core`. Want to confirm if this is okay and if we are okay with this name. @marcbouchenoire @GuillaumeSalles

marcbouchenoire: I didn't know you had a follow-up PR, [I renamed it to `ThreadsOptions`](https://github.com/liveblocks/liveblocks/pull/1351/commits/0735a9e6674b2c5768cd5bb97d4a05453827987a) in the `comments-unread` branch since it will include other options over time but it's a private type so it's not super important.

PR #1367, opened 2023-12-29
Add getInboxNotification method to Liveblocks client

FlowFlorent:
### Description

In preparation for email notifications, adding the associated endpoint method that will help process email notification webhook events.

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -118,6 +118,14 @@ export type Schema = {
 
 type SchemaPlain = DateToString<Schema>;
 
+export type InboxNotification = {
+  readAt: string | null;
+  notifiedAt: string;
+  id: string;
+  threadId: string;
+  kind: "thread";
+};
+
 /**
  * Interact with the Liveblocks API from your Node.js backend.
  */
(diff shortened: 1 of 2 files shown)


GuillaumeSalles (approved):
Nice! Can you change the base branch to `comments-unread` to make sure we only release that when @marcbouchenoire works is over?

packages/liveblocks-node/src/client.ts
```diff
@@ -118,6 +118,22 @@ export type Schema = {
 
 type SchemaPlain = DateToString<Schema>;
 
+type InboxNotificationResponse = {
+  readAt: string | null;
+  notifiedAt: string;
+  id: string;
```
nimeshnayaju, line 124: We have a `DateToString<T>` helper type that converts all top level `Date` types to `string` type. @marcbouchenoire recently updated the helper type to also account for `string | null` types in [this PR](https://github.com/liveblocks/liveblocks/pull/1351/files#diff-7b89970372c566b11b3ae169274738ef463e86ec819ad5fbbed6864479911b02), which I think might be helpful in this case.

marcbouchenoire: I also wonder if we should duplicate these types instead of sharing them between packages like we do with `CommentData` and `ThreadData`, both exported from `@liveblocks/core`.

In `comments-unread`, I landed on the names `InboxNotificationData` for the type that contains the thread and `PartialInboxNotificationData` for the type that doesn't, but my confidence level in these is not high so we can still change them. https://github.com/liveblocks/liveblocks/blob/comments-unread/packages/liveblocks-core/src/types/InboxNotificationData.ts

nvie (commented):
Looking good from a "code owner" perspective! 👍

PR #1374, opened 2024-01-04
Refresh threads and notifications cache when receiving comment websocket message

nimeshnayaju:
- [x] Implement `getThread` method that calls the `GET :threadId` API endpoint to retrieve thread with the specified id and associated notifications.
- [x] Handle comment websocket event messages.
- [x] Compare threads using `updatedAt` property before merging new threads into existing ones in cache.
- [x] Add tests to validate the aforementioned features.

### DIFF
--- a/packages/liveblocks-react/src/__tests__/_restMocks.ts
+++ b/packages/liveblocks-react/src/__tests__/_restMocks.ts
@@ -1,7 +1,7 @@
 import type {
-  CommentDataPlain,
-  PartialInboxNotificationDataPlain,
-  ThreadDataPlain,
+  CommentData,
+  PartialInboxNotificationData,
+  ThreadData,
 } from "@liveblocks/core";
 import type { ResponseResolver, RestContext, RestRequest } from "msw";
 import { rest } from "msw";
@@ -11,7 +11,7 @@ export function mockGetThreads(
     RestRequest<never, never>,
     RestContext,
     {
-      data: ThreadDataPlain<any>[];
+      data: ThreadData<any>[];
       inboxNotifications: any[];
     }
   >
@@ -27,8 +27,8 @@ export function mockGetThread(
   resolver: ResponseResolver<
     RestRequest<never, never>,
     RestContext,
-    ThreadDataPlain<any> & {
-      inboxNotification?: PartialInboxNotificationDataPlain;
+    ThreadData<any> & {
+      inboxNotification?: PartialInboxNotificationData;
     }
   >
 ) {
@@ -42,7 +42,7 @@ export function mockCreateThread(
   resolver: ResponseResolver<
     RestRequest<never, never>,
     RestContext,
-    ThreadDataPlain<any>
+    ThreadData<any>
   >
 ) {
   return rest.post(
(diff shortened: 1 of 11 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -365,6 +367,44 @@ export function createRoomContext<
...
+
+        const existingThread = store.get().threads[message.threadId];
+
+        switch (message.type) {
+          case ServerMsgCode.COMMENT_EDITED:
+          case ServerMsgCode.THREAD_METADATA_UPDATED:
+          case ServerMsgCode.COMMENT_REACTION_ADDED:
+          case ServerMsgCode.COMMENT_REACTION_REMOVED:
+            // If the thread doesn't exist in the local cache, we do not update it with the server data as an optimistic update could have deleted the thread locally.
+            if (!existingThread) break;
```
GuillaumeSalles, line 390: How is it possible to delete a thread with an optimistic update? In my mind, it can only happen after getting a 404 or a full refresh of the cache. Never with an optimistic update. What scenario do you have in mind?

nimeshnayaju: Hmmm. I realize the comment description isn't completely accurate. It should say optimistic update followed by successful mutation request.

I thought when we deleted the last comment of a thread (after mutation request is successful, we also deleted the thread from cache, but I checked the implementation of `useDeleteComment`, and that doesn't seem to be the case. Is that intentional or did we forget to account for that case?

https://github.com/liveblocks/liveblocks/assets/35807215/aec86f1e-f98d-4ea3-986c-a8544121bd14

nimeshnayaju: Pushed an update to delete a thread from cache if the deleted comment is the last undeleted comment in the thread. Here's [the commit link](https://github.com/liveblocks/liveblocks/pull/1374/commits/42b4e403944a957ba51a3cb722b86ad9f8aa8655) - let me know what you think!

nvie (commented):
Looking good from a "code owner" perspective! 👍

packages/liveblocks-react/src/__tests__/_restMocks.ts
```diff
@@ -7,7 +11,7 @@ export function mockGetThreads(
     RestRequest<never, never>,
     RestContext,
     {
-      data: ThreadDataPlain<any>[];
+      data: ThreadData<any>[];
```
GuillaumeSalles, line 14: This is wrong. We return plain data from the server so we should not adapt our mocks to make it easier to test.

Created a new issue so we can tackle that later https://github.com/liveblocks/liveblocks.io/issues/1850

GuillaumeSalles (approved):

packages/liveblocks-react/src/__tests__/_restMocks.ts
nimeshnayaju: Thanks for pointing this out. I assumed any `Date` object would be automatically converted to strings in the 8601 format during serialization, so the returned date would indeed be plain. But, I can see how we could miss properly serializing other properties in the future.

PR #1375, opened 2024-01-05
Handle error during initial fetching inside `useThreads` and `useThreadsSuspense`

nimeshnayaju:
Stores any error that occurs during the initial fetch inside `useThreads` and `useThreadsSuspense`.

- [x] `useThreads` should return an error if the initial fetch is failing
- [x] `useThreadsSuspense` should throw an error if the initial fetch is failing
- [x] Integration tests


nimeshnayaju:
@GuillaumeSalles `useThreads` and `useThreadsSuspense` both handle error that are encountered during the initial fetch. Any thoughts on how we could deal with errors thrown during polling and comment Websocket message handling? We've done retries using exponential backoff algorithm on error in our previous implementation.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1085,31 +1085,30 @@ export function createRoomContext<
...
+        ),
+        queryKey
+      );
 
-    poller.start(POLLING_INTERVAL);
+      poller.start(POLLING_INTERVAL);
+    } catch (err) {
+      store.setThreadsQueryState(queryKey, {
+        isLoading: false,
+        error: err as Error,
```
nimeshnayaju, line 1109: **TODO**: If the initial request failed but polling was successful, we want to clear this `error` variable.

nimeshnayaju: Done: https://github.com/liveblocks/liveblocks/pull/1375/commits/7cbfa0ea5ee994ead736f709e21e84190eac875f

marcbouchenoire (approved):
Looking good from my perspective.

PR #1379, opened 2024-01-05
Fix Comments E2E tests

nimeshnayaju:
This fixes the currently skipped E2E tests for comments and introduce new tests.


marcbouchenoire (approved):
🤞

nvie (approved):
Amazing that you got them to work! Did you figure out what was the issue before?

nimeshnayaju:
> Amazing that you got them to work! Did you figure out what was the issue before?

I'm not entirely certain if the tests will pass in the CI environment yet, as I'm encountering persistent failures with other tests like `offline.test.ts` (I've been retrying for a while now, not sure if the new tests have an impact on those). However, I did identify a problem in the previous comments code. Specifically, when a revalidation request was triggered by a comment webhook message, it was incorrectly deduplicated if there were other concurrent requests. We needed to adjust our approach to ensure that these revalidation requests are handled independently, as each one potentially carries unique and important updates that should not be ignored or merged with other requests. 

Also, with the new cache manager, we do not need to worry about request deduplication anymore (thankfully), and I've also made the tests slightly simpler and plan to add more as we see fit.

PR #1383, opened 2024-01-08
Handle invalid response for `getRooms` method

nimeshnayaju:
We missed adding error response handler to the `getRooms` method. This PR updates the method to add that handling logic.


nvie (approved):
Nice, thanks for the quick turnaround! 🙏

PR #1389, opened 2024-01-09
Fix mention suggestions

marcbouchenoire:
Our `Portal` component didn't fallback to `document.body` as expected when not passing a container.


nimeshnayaju (approved):
🚀

PR #1390, opened 2024-01-09
Upgrade examples to 1.9.6

marcbouchenoire:
This PR upgrades all examples to `1.9.6`.


CTNicholas:
A bunch of examples aren't building because of an import we're using in `1.9.5`. Perhaps we need to update [this line](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react/src/comments/CommentsRoom.tsx#L26) to the following? 

```
import { useSyncExternalStore } from "use-sync-external-store/shim/index.js";
```

nimeshnayaju:
> A bunch of examples [aren't building](https://vercel.com/liveblocks/examples-nextjs-live-cursors/EVoHKynVBHmq4zeuojkuTiRq5MKz) because of an import we're using in `1.9.5`. Perhaps we need to update [this line](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react/src/comments/CommentsRoom.tsx#L26) to the following @nimeshnayaju?
> 
> ```
> import { useSyncExternalStore } from "use-sync-external-store/shim/index.js";
> ```

That seems to be the case. Thanks for catching this. I think the fix you suggested should solve the issue. I'll do a quick test and submit a PR to fix it. 

It is a little strange that some of the examples build correctly but some don't.

nimeshnayaju (approved):
Thanks for doing this!

PR #1391, opened 2024-01-09
Update `useSyncExternalStore` import

nimeshnayaju:
Some of our examples were failing to build due to the way `useSyncExternalStore` was imported. This PR updates the import so that it is consistent across both `room.tsx` and `factory.tsx`. Thanks for catching this @CTNicholas 

https://github.com/liveblocks/liveblocks/pull/1390#issuecomment-1883119154

### DIFF
--- a/packages/liveblocks-react/src/comments/CommentsRoom.tsx
+++ b/packages/liveblocks-react/src/comments/CommentsRoom.tsx
@@ -23,9 +23,9 @@ import React, {
   useEffect,
   useMemo,
 } from "react";
-import { useSyncExternalStore } from "use-sync-external-store/shim";
 import { useSyncExternalStoreWithSelector } from "use-sync-external-store/shim/with-selector.js";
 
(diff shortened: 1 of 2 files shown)


packages/liveblocks-react/src/comments/CommentsRoom.tsx
```diff
@@ -23,9 +23,9 @@ import React, {
   useEffect,
   useMemo,
 } from "react";
-import { useSyncExternalStore } from "use-sync-external-store/shim";
 import { useSyncExternalStoreWithSelector } from "use-sync-external-store/shim/with-selector.js";
 
+import { useSyncExternalStore } from "../factory";
```
nimeshnayaju, line 28: I've opted to instead import this from `factory.tsx` which is used in some of our other hooks and doesn't create build issues.

CTNicholas (approved):
I've seen that package cause build issues in other repos too, thanks for making the change to avoid it!

PR #1396, opened 2024-01-10
Move (and improve) `resolveMentionSuggestions`

marcbouchenoire:
Before moving `resolveUsers` to `createClient`, I'm starting with `resolveMentionSuggestions`.

**Context:** We need to move the resolvers out of `createRoomContext` and to the client because we'll need some of them outside the room (for notifications), this one won't be used for notifications but it's better for clarity to have them in the same place and we could leverage it in the future for new use-cases like mentions in text editors.

While I was there, I moved its implementation away from `AsyncCache` to a simpler custom hook (because the mention suggestions use-case doesn't have many edge cases and other issues like `useUser`), and in doing so I was able to improve "performance": we currently always debounce the call to `resolveMentionSuggestions`, even if the current text was already searched (so it's cached) and could be served instantly. Now, we only debounce uncached text.

[Regarding the hidden field on `Client`](https://github.com/liveblocks/liveblocks/pull/1396/commits/03401fce4942792a3f1df7032b653604150c5167): I'm using a Symbol as a "brand" to hide it from autocomplete while still keeping the type information across packages, as opposed to marking a normal field with `@internal` which would strip it from the built types (or marking it with `@private` which would do nothing and would make the field visible in autocomplete when writing `client.`). [I also made it non-enumerable](https://github.com/liveblocks/liveblocks/pull/1396/commits/ebc2ea22ed03c9105ff515745b835bf51c11c558) to keep it hidden in more scenarios.
Edit: [It now uses a single reusable symbol](https://github.com/liveblocks/liveblocks/pull/1396/commits/86ac6638be1fd33803d5cd6a3174ebf9c6ca6711) shared between `Client` and `Room`. (and possibly other places later on)

### Breaking changes

- `resolveMentionSuggestions` moves from `createRoomContext` to `createClient` (and the type `ResolveMentionSuggestionsArgs` moves from `@liveblocks/react` to `@liveblocks/client`)


e2e/next-sandbox/pages/offline.tsx
```diff
@@ -1,5 +1,5 @@
 import { LiveList } from "@liveblocks/client";
-import type { IWebSocketCloseEvent } from "@liveblocks/core";
+import { INTERNAL } from "@liveblocks/core";
```
nvie, line 2: Love the change to symbols for this! I just checked out the `.d.ts` file and I think it's a great solution — glad you found this way 👍 

One small suggestion: when I suggested the `kInternal` naming convention, it was the export/variable that I meant, not the string assigned to the Symbol() function (that's only used for debug logging).

We could define it like so:

```ts
export const kInternal = Symbol('liveblocks internals');
```

or even just

```ts
export const kInternal = Symbol();
```

nvie (approved):
Only looked at the changes from the "symbols" angle, but that looks great. I just left one suggestion re naming, but feel free to ignore.

e2e/next-sandbox/pages/offline.tsx
marcbouchenoire: Updated!

GuillaumeSalles (approved):
:chef-kiss:

Thanks for going the extra mile by improving how we hide internals outside of the scope of comments.

PR #1399, opened 2024-01-11
Implement `useMarkThreadAsRead` hook

nimeshnayaju:
- [x] Move cache store from `room` to `client`. This facilitates other hooks like `useInboxNotifications` as `InboxNotificationData` (and related `ThreadData`) are not fetched via a room. Also makes testing `useMarkThreadAsRead` easier.
- [x] Implement `useMarkThreadAsRead` 
- [x] Unit tests


packages/liveblocks-core/src/client.ts
```diff
@@ -654,3 +655,21 @@ function toURLSearchParams(
...
+
+const CacheStoreMap = new WeakMap<
+  Client,
+  ReturnType<typeof createClientStore<any>>
+>();
+
+export function getCacheStore<TThreadMetadata extends BaseMetadata>(
+  client: Client
+) {
+  let store = CacheStoreMap.get(client);
```
nimeshnayaju, line 667: @GuillaumeSalles I started moving the cache store from `room` to `client` so we could also start working on other hooks like `useInboxNotifications` that fetch data from outside the room. I've gone with the `WeakMap` approach instead of adding a hidden field to `client`. What are your thoughts on this?

marcbouchenoire:
@nimeshnayaju If we move `store.ts` out of `/src/comments`, should we also move `createStore.ts`?

nimeshnayaju:
> @nimeshnayaju If we move `store.ts` out of `/src/comments`, should we also move `createStore.ts`?

I moved `createStore.ts` also to `@liveblocks/core` but forgot to remove it from `@liveblocks/react`. Thanks for reminding!!

marcbouchenoire (approved):

PR #1408, opened 2024-01-16
Move (and improve) `resolveUsers`

marcbouchenoire:
> This PR is a follow-up to https://github.com/liveblocks/liveblocks/pull/1396.

This PR moves `resolveUsers` to `createClient` so that it can be used outside of rooms as well.

While moving it, I also refactored it to use the new batching system which means it now batches calls (and I fully removed `AsyncCache` now that it's not used anywhere anymore).

To be able to share `useUser` hooks between the global context outside the rooms (`LiveblocksContext`) and the room context (`RoomContext`), I created a third (internal-only) context used by both, which can be used to make `useUser` usable and exportable interchangeably between `LiveblocksContext` and `RoomContext`.

Fixes https://github.com/liveblocks/liveblocks.io/issues/1601.
Fixes https://github.com/liveblocks/liveblocks.io/issues/1596.

### Breaking changes

- **`resolveUsers` now only receives `{ userIds: ["xxx"] }` instead of `{ userIds: ["xxx"], roomId: "xxx" }`**
  - **We could technically keep `roomId` as optional but I think it makes things more complicated (having to deal with it being optional in your `resolveUsers` function). It does block use-cases like attaching specific statuses to users based on rooms so maybe it's still necessary in the end.**
- `resolveUsers` moves from `createRoomContext` to `createClient` (and the type `ResolveUsersArgs` moves from `@liveblocks/react` to `@liveblocks/client`)

### Not tackled in this PR

- We need to access the current user's info in Notifications, but we can do that separately: https://github.com/liveblocks/liveblocks.io/issues/1858. Without this, I had to remove the `replaceSelf` prop on the `User` component, we'll need to bring it back once we have a new way to access the current user's info.
- Creating tests for `useUser`: https://github.com/liveblocks/liveblocks.io/issues/1860


GuillaumeSalles (approved):
Way simpler than `AsyncCache`! Great work!

PR #1410, opened 2024-01-17
Better pagination API for rooms

### LINKED ISSUE #1382: Improve pagination API for `client.getRooms()` endpoint in `@liveblocks/node`
If you're using `@liveblocks/node` to iterate over all your rooms, i.e. `client.getRooms(...)`, there is some pagination support, i.e. `client.getRooms({ limit, startingAfter: cursor })`.

However, to fetch the next page, the only way currently to extract the cursor value to use on the next request is by parsing it out of the `nextPage` field manually, which is quite annoying and not a great DX:

```ts
let cursor: string | null = null;

while (true) {
  const resp = await client.getRooms({ startingAfter: cursor });

  doSomethingWithRooms(resp.data);  // current page

  if (resp.nextPage === null) {
    break; // Done paginating
  } else {
    // Ugh :(
    cursor = resp.nextPage.split("startingAfter=")[1].split("&")[0];
  }
}
```

It'd be nice to have a more convenient pagination API.

(Possibly the solution here is not to update the `@liveblocks/node` client, but to make the cursor available in a separate field in the server responses instead so at least we avoid the parsing.)

nvie:
Fixes #1382 and adds the docs, and adds a few fixes for our TypeScript types.

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -327,6 +327,7 @@ export class Liveblocks {
     } = {}
   ): Promise<{
     nextPage: string | null;
+    nextCursor: string | null;
     data: RoomInfo[];
   }> {
     const path = url`/v2/rooms`;
@@ -354,6 +355,7 @@ export class Liveblocks {
 
     const data = (await res.json()) as {
       nextPage: string | null;
+      nextCursor: string | null;
       data: RoomInfoPlain[];
     };
 
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -55,7 +55,7 @@ type Nullable<T> = {
 };
 
 type DateToString<T> = {
-  [P in keyof T]: T[P] extends Date ? string : T[P];
+  [P in keyof T]: Date extends T[P] ? string : T[P];
 };
 
 export type CreateSessionOptions = {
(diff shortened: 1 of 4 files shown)


packages/liveblocks-node/src/client.ts
```diff
@@ -55,7 +55,7 @@ type Nullable<T> = {
 };
 
 type DateToString<T> = {
-  [P in keyof T]: T[P] extends Date ? string : T[P];
+  [P in keyof T]: Date extends T[P] ? string : T[P];
```
nvie, line 58: This wasn't doing what we thought it was, so fixing this while at it.

packages/liveblocks-node/src/client.ts
```diff
@@ -354,6 +355,7 @@ export class Liveblocks {
 
     const data = (await res.json()) as {
       nextPage: string | null;
+      nextCursor: string | null;
```
nvie, line 358: Here is the new field the responses now include.

packages/liveblocks-node/src/client.ts
```diff
@@ -363,12 +365,11 @@ export class Liveblocks {
         ? new Date(room.lastConnectionAt)
         : undefined;
 
-      const createdAt = room.createdAt ? new Date(room.createdAt) : undefined;
-
+      const createdAt = new Date(room.createdAt);
```
nvie, line 368: The `createdAt` field is not optional.

CTNicholas (approved):

PR #1411, opened 2024-01-17
Fix turbo build on Windows

CTNicholas:
`cp` doesn't work on Windows, but [shx](https://www.npmjs.com/package/shx) makes it cross-platform.

### DIFF
--- a/packages/liveblocks-yjs/package.json
+++ b/packages/liveblocks-yjs/package.json
@@ -42,8 +42,7 @@
   "devDependencies": {
     "@liveblocks/eslint-config": "*",
     "@liveblocks/jest-config": "*",
-    "@testing-library/jest-dom": "^5.16.5",
-    "shx": "^0.3.4"
+    "@testing-library/jest-dom": "^5.16.5"
   },
   "sideEffects": false,
   "bugs": {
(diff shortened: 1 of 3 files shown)


packages/liveblocks-yjs/package.json
```diff
@@ -24,7 +24,7 @@
   ],
   "scripts": {
     "dev": "tsup --watch",
-    "build": "tsup && cp dist/index.d.ts dist/index.d.mts",
+    "build": "tsup && shx cp dist/index.d.ts dist/index.d.mts",
```
nvie, line 27: Hmm. This `cp` should no longer be necessary with the latest version of `tsup`. I'll have a look at the configuration and upgrade this.

nvie: Fixed!

nvie (approved):

packages/liveblocks-yjs/package.json
```diff
@@ -24,7 +24,7 @@
   ],
   "scripts": {
     "dev": "tsup --watch",
-    "build": "tsup && cp dist/index.d.ts dist/index.d.mts",
+    "build": "tsup",
```
nvie, line 27: This was necessary in an older version of tsup, so we could simply remove this — sorry about the hassle, and thanks for catching! 👍

CTNicholas: Thanks for making it simpler!

PR #1424, opened 2024-01-23
Use current user id for Notifications

GuillaumeSalles:
- `useCurrentUserId` under `LiveblocksProvider` and `RoomProvider`
- Some tests + error handling


marcbouchenoire (approved):
LGTM!

I'm curious about separating Notifications-related code into its own file: what's your reasoning? I personally wanted to move away from creating silos given that not all parts are separated (why separate Comments or Notifications if other Liveblocks features are all in the same place?). Ideally I wanted to not even have things like `createInboxNotificationsApi` (which are only needed because of generics we're not dealing with at the `Client` level), just one big `client.ts` (and one big `room.tsx` in the React package).

PR #1429, opened 2024-01-25
Synchronizes same lint setting across all repos

nvie:


shared/eslint-config/index.js
```diff
@@ -43,7 +43,11 @@ module.exports = {
     // Customized default rules
     // ------------------------
     eqeqeq: ["error", "always"],
-    quotes: ["error", "double", "avoid-escape"],
```
nvie, line 46: `avoid-escape` is deprecated and will stop working in a future ESLint upgrade.

PR #1440, opened 2024-01-31
Fix incorrect threads ordering

nimeshnayaju:
Fixes https://github.com/liveblocks/liveblocks.io/issues/1887

### DIFF
--- a/packages/liveblocks-react/src/__tests__/useThreads.test.tsx
+++ b/packages/liveblocks-react/src/__tests__/useThreads.test.tsx
@@ -11,9 +11,15 @@ import React, { Suspense } from "react";
 import { ErrorBoundary } from "react-error-boundary";
 
 import { createRoomContext } from "../room";
-import { dummyThreadData } from "./_dummies";
+import { dummyInboxNoficationData, dummyThreadData } from "./_dummies";
 import MockWebSocket, { websocketSimulator } from "./_MockWebSocket";
-import { mockGetThread, mockGetThreads } from "./_restMocks";
+import {
+  mockGetInboxNotifications,
+  mockGetThread,
+  mockGetThreads,
+} from "./_restMocks";
+import { createLiveblocksContext } from "../liveblocks";
+import { wait } from "./_utils";
 
 const server = setupServer();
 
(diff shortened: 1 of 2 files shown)


packages/liveblocks-react/src/__tests__/useThreads.test.tsx
```diff
@@ -492,6 +521,179 @@ describe("useThreads", () => {
...
+    oldThread.createdAt = new Date("2021-01-01T00:00:00Z");
+
+    const newThread = dummyThreadData();
+    newThread.createdAt = new Date("2021-01-02T00:00:00Z");
+
+    server.use(
+      mockGetThreads(async (_req, res, ctx) => {
+        return res(
+          ctx.json({
+            data: [newThread, oldThread], // The order is intentionally reversed to test if the hook sorts the threads by creation date
```
nimeshnayaju, line 536: I understand that the server returns the thread already sorted by creation date, so this test doesn't replicate server behavior correctly but attempts to test the sorting behavior on client side.

GuillaumeSalles (approved):
Sweeeet! I really like the tests! It does not care too much about the internals. Even if we move a lot of code around, we will not have to refactor the tests :chef-kiss:

packages/liveblocks-react/src/__tests__/useThreads.test.tsx
```diff
@@ -492,6 +521,179 @@ describe("useThreads", () => {
...
+        return res(
+          ctx.json({
+            data: [newThread],
+            inboxNotifications: [],
+          })
+        );
+      }),
+      mockGetInboxNotifications(async (_req, res, ctx) => {
+        // Mock a delay in response so that GET THREADS request is resolved before GET NOTIFICATIONS request
+        await wait(100);
```
GuillaumeSalles, line 593: Could you use jest fake timers instead? We're trying to avoid waiting arbitrary values that could create flacky tests on slower machines.

nimeshnayaju: Good catch! Updated to use jest fake timers!

PR #1445, opened 2024-02-02
Add rooms URLs to `resolveRoomsInfo`

marcbouchenoire:
This PR adds support for providing URLs for room IDs in `resolveRoomsInfo` (which currently only asks for room names), and then uses that in `<InboxNotification />` to allow making notifications clickable without having to call your back-end in React-land to generate an `href` value.

IMO, this should be considered the default path, but if you can use `inboxNotification.roomId` to generate URLs and don't need the resolver, setting `href` on `<InboxNotification />` will always take priority and you can just ignore the resolver.


CTNicholas (approved):
Nice!

PR #1446, opened 2024-02-02
Remove unnecessary query params for `getThreads` node method

nimeshnayaju:
We missed to remove an unnecessary query parameter for `getThreads` method under `@liveblocks/node`. Note that the presence of this query parameter doesn't affect the API response because the backend endpoint doesn't look at the query parameter. But, it is still a good idea to remove this query parameter from the method to remove any confusion.


FlowFlorent (approved):

FlowFlorent:
Good catch 🙌

PR #1450, opened 2024-02-07
Improve handling of thread and inbox notification deletion

nimeshnayaju:
### `useThreads`
- [x] Update `getThreads` to accept the new `since` query parameter and deleted information
- [x] Update the logic to merge existing thread with new threads to account for deletion information.
- [x] Update polling logic to call `getThreads({ since })` for each room instead of making a `getThread` call for each filter options.
- [x] Update cache to also include threads marked as deleted. This is necessary to solve potential race conditions. As a result, we will have to update the `DELETE COMMENT` optimistic update logic.
- [x] Add tests
- [x] ~~More tests~~ (Different PR)

### `useInboxNotifications`
- [x] Update `getInboxNotifications` to accept the new `since` query parameter and deleted information.
- [x] Update polling logic to call `getInboxNotifications({ since })`.
- [x] Update notifications merging logic so that a newer notification isn't overridden by an older one
- [x] Add tests 
- [x] ~~More tests~~ (Different PR)

### Optimistic updates
- [x] ~~Update optimistic updates mechanism to also populate updatedAt and deletedAt as appropriate to eliminate race conditions and flickering issues~~ (Different PR)

### Additionally, this PR fixes the following issues:
- [x] Inbox notifications are not always ordered by the `notifiedAt` date.


packages/liveblocks-core/src/room.ts
```diff
@@ -832,6 +850,12 @@ type PrivateRoomApi = {
     explicitClose(event: IWebSocketCloseEvent): void;
     rawSend(data: string): void;
   };
+
+  // Used to store metadata related to comments
+  comments: {
+    lastRequestedAt: Date | null;
+    readonly queries: Set<string>;
```
nimeshnayaju, line 857: @nvie I have added some extra properties related to comments inside the room context and exported as a property of `kInternal`. Wondering if you think it's okay to do so, or if there some things I should be wary about when doing so.

packages/liveblocks-core/src/client.ts
```diff
@@ -614,6 +615,9 @@ export function createClient<TUserMeta extends BaseUserMeta = BaseUserMeta>(
         cacheStore,
         usersStore,
         roomsInfoStore,
+        get rooms() {
+          return Array.from(roomsById.keys());
+        },
```
nimeshnayaju, line 620: @nvie I have exposed a getter that returns the list of room ids. Also curious if it is okay to do so?

packages/liveblocks-core/src/room.ts
```diff
@@ -832,6 +850,12 @@ type PrivateRoomApi = {
     explicitClose(event: IWebSocketCloseEvent): void;
     rawSend(data: string): void;
   };
+
+  // Used to store metadata related to comments
+  comments: {
+    lastRequestedAt: Date | null; // Stores the timestamp when threads and notifications were last requested for the room
+    readonly queries: Set<string>; // Stores queries that have been made in the room
```
nimeshnayaju, line 857: The reason I have stored these information inside a room context is because they are only relevant as long as the room is alive (i.e., not destroyed). For instance, the set `queries` stores query keys for `getThread()` and `getRoomNotificationSettings` queries that have been made. 

I could have stored this as a map inside `room.tsx` where key would be `roomId` and value would be this metadata information. But, I realized there is a new layer of complexity with doing that, especially because a client can have a room with the same id rendered at multiple places (via `RoomProvider`), and these information (about `lastRequestedAt`, `queries`) are expected to be destroyed when the room is destroyed, so we would need to keep track of how many rooms for the same room id is being rendered currently and so on.

packages/liveblocks-react/src/room.tsx
```diff
@@ -996,136 +996,118 @@ export function createRoomContext<
...
-      subscribers: 0,
-    });
+    // If the query has already been recorded in the room, we do not make another fetch request
+    if (queries.has(queryKey)) return;
 
-    store.setQueryState(queryKey, {
-      isLoading: true,
-    });
+    // Add the query to the room's queries set to prevent duplicate fetch requests
+    queries.add(queryKey);
```
nimeshnayaju, line 1073: We currently add a query key to the set of queries before the request starts. There is a possibility that the request might fail, in which case we want to maybe remove this `queryKey` from the set of queries. This could be a part of error handling logic - will have to create a separate PR to think about error handling in general.

nimeshnayaju: Same for `useInboxNotifications`

packages/liveblocks-core/src/room.ts
GuillaumeSalles: Can you clarify why we have to clear this information when the room is destroyed? We're not clearing the cache when the room is destroyed, so I guess we don't have to do it in this case either?

Unrelated to this PR, but even the code in `store.ts` should be moved to `@liveblocks/react` ultimately. It was at some point, but I think it was moved to `@liveblocks/client` when we introduced the hidden symbol field on the client. I want to move all that code to `@liveblocks/react` right after we release the `1.10`

packages/liveblocks-core/src/room.ts
nvie: Nope! I assume these are for unit testing purposes? Nothing special about them. That said, I already removed a couple not too long ago because I would love to keep this list relatively small. But for now, it's fine to ship private APIs like this 👍

packages/liveblocks-core/src/client.ts
nvie: It works, but as a user I think I would prefer getters to be relatively simple "property like" lookups, and not do a computation like this (and return a different reference every time). Would exposing this as a `getRooms()` function not help? I think it better signifies that if you want to read it multiple times, it's better to assign the result to an intermediate variable?

Also… maybe rename this to `getRoomIds()` instead of `getRooms()`?

GuillaumeSalles (commented):
The code is pretty clear and the tests are great! Nice job @nimeshnayaju!

The only thing that is still not clear to me is why we switched from the subscribers approach to keeping some metadata on the room. Are you free later today to discuss about this? I want to make sure we can merge this sooner than later.

packages/liveblocks-core/src/room.ts
nimeshnayaju: These are also internally consumed by `createRoomContext` inside `room.tsx`. The motive is explained here: https://github.com/liveblocks/liveblocks/pull/1450#discussion_r1482163540

I am open to explaining more in detail if it helps.

PR #1451, opened 2024-02-07
Example: Thread notifications + React email

CTNicholas:
This example shows you how to send an email with React Email and Resend whenever an inbox notification webhook is received. React Email 2.0 does not work on Windows yet (but looks like it will soon - https://github.com/resend/react-email/pull/1216), so I had to use version 1.10.

We may be able to show this in our example page:
image

- `npm run dev:email` and http://localhost:3000 for email previews
- `npm run dev` and http://localhost:3001 for comments example

It requires Resend and webhook API keys, and a verified Resend domain. If you're not using Resend, it's easy to replace it with just a few lines.


examples/nextjs-comments-emails/src/app/page.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <main>
+      {threads.map((thread) => (
+        <Thread key={thread.id} thread={thread} className="thread" />
+      ))}
+      <Composer<ThreadMetadata>
+        className="composer"
+        // Add link to the room for email notifications
+        metadata={{ url: roomUrl }}
```
marcbouchenoire, line 27: I don't think we should use metadata for this as it's not realistic (users shouldn't have to do this). You should have access to the room ID on the webhook side?

marcbouchenoire (changes requested):
Thanks for working on this! I left a few comments and one request regarding the use of metadata to pass URLs.
One other possible issue (only if we publish this publicly somehow), it's the same issue as why we don't have a Notifications example: since you use fixed user IDs (e.g. `user-0`), people will see notifications from other rooms. We can ignore this if we don't publish it similarly to how we ignore the problem by not having a Notifications example yet.

examples/nextjs-comments-emails/src/app/page.tsx
FlowFlorent: yes, the entire URL can be constructed from the webhook event. 

For the thread events: `projectId`, `roomId`, `threadId` are available
For comment events: the above and `commentId` are available

marcbouchenoire: @FlowFlorent Could you tweak the example to stop using thread metadata but `roomId` from the webhook events instead to build links within the emails?

FlowFlorent: Yeah on it!

FlowFlorent: Adjusted

examples/nextjs-comments-emails/src/app/api/liveblocks-notifications/threadEmailNotification.tsx
```diff
@@ -53,15 +52,14 @@ export async function threadEmailNotification({
...
   const newCommentsEmail = (
-    <NewComments
-      title={title}
-      href={thread.metadata.url}
-      comments={htmlCommentBodies}
-    />
+    <NewComments title={title} href={roomUrl} comments={htmlCommentBodies} />
   );
-
+  console.log("Sending email", newCommentsEmail);
```
marcbouchenoire, line 62: Is this useful to keep or is it a leftover?

marcbouchenoire (approved):
Thanks @FlowFlorent!

PR #1452, opened 2024-02-07
Make all of our usage of `useSyncExternalStoreWithSelector` referentially stable

GuillaumeSalles:
Solves https://github.com/liveblocks/liveblocks/issues/1448

### DIFF
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -144,6 +144,30 @@ export function createLiveblocksContext<
     return;
   }
 
+  function useInboxNotificationsSelectorCallback(
+    state: CacheState<BaseMetadata>
+  ): InboxNotificationsState {
+    const query = state.queries[INBOX_NOTIFICATIONS_QUERY];
+
+    if (query === undefined || query.isLoading) {
+      return {
+        isLoading: true,
+      };
+    }
+
+    if (query.error !== undefined) {
+      return {
+        error: query.error,
+        isLoading: false,
+      };
+    }
+
+    return {
+      inboxNotifications: selectedInboxNotifications(state),
+      isLoading: false,
+    };
+  }
+
   function useInboxNotifications(): InboxNotificationsState {
     useEffect(() => {
       void fetchInboxNotifications();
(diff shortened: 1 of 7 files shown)


packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -174,7 +174,7 @@ export function createLiveblocksContext<
       incrementInboxNotificationsSubscribers();
 
       return () => decrementInboxNotificationsSubscribers();
-    });
+    }, []);
```
nvie, line 177: Oof, good catch!

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -144,49 +144,57 @@ export function createLiveblocksContext<
     return;
   }
 
+  function useInboxNotificationsSelectorCallback(
+    state: CacheState<BaseMetadata>
+  ): InboxNotificationsState {
```
nimeshnayaju, line 149: @GuillaumeSalles Any particular reason why some selectors are stored as callback and others not? I know it is slightly better for performance to store selectors outside of component  (if possible) because `useCallback` has some performance cost. Curious if there are other reasons too

nvie (approved):
Excellent updates!

In addition to these improvements, you may also want to set the 5th argument to `uSES` to specify a custom `isEqual` function here. Because the selector will be run often, and it will return a new `{ isLoading, inboxNotifications }` object instance every time, then without such `isEqual` function, any component using this hook will still re-render too often, even if there is no new content for this specific component.

nimeshnayaju (approved):

nimeshnayaju:
> In addition to these improvements, you may also want to set the 5th argument to uSES to specify a custom isEqual function here. Because the selector will be run often, and it will return a new { isLoading, inboxNotifications } object instance every time, then without such isEqual function, any component using this hook will still re-render too often, even if there is no new content for this specific component.

@nvie Were you suggesting a deep equal check for each notification returned? Wonder how much of an impact that would have on performance since we could have quite a large list of notifications (same for threads).

nvie:
> @nvie Were you suggesting a deep equal check for each notification returned? Wonder how much of an impact that would have on performance since we could have quite a large list of notifications (same for threads).

Especially if the list is so large, wouldn't constantly re-rendering such large lists in the application (even when nothing changed) create an even bigger performance overhead in the actual components?

Is the list returned by `selectedInboxNotifications()` stable? Or is it re-computed on every call? If it's stable, you can spare the deep equality check here and basically just do a shallow comparison on the `{ isLoading, error, inboxNotifications }` object itself.

packages/liveblocks-react/src/liveblocks.tsx
GuillaumeSalles: I only used `useCallback` if necessary. If the function need to access any local variable, we're using `useCallback`, if not, it's a pure function outside of the hook. It's to be a lean a possible, not for performance.

GuillaumeSalles:
I don't want to use any shallow or deep equality here. We have ways to add some memoization on some sub selectors, but this optimization is not necessary for `1.10`. I just want to make sure there is not infinite re-render is someone is using the antipattern `useEffect(() => { setState(fn(hookResult))}, [hookResult]);`

PR #1453, opened 2024-02-08
Improve auth manager for Comments & Notifications

ofoucherot:
#### Auth Manager

This PR improves the Auth Manager for usage in Comments & Notifications (see [commit](https://github.com/liveblocks/liveblocks/commit/a9e1628763d607c00bf35a97f288f3de9b9e87d3)):
- The Auth manager will verify that the issued Access Token grants enough access for the requested usage. If not, it will throw an error.
- For Access Tokens, if a token without a roomId is requested, the Auth Manager will only accept tokens with a wildcard.
- The scope `comments:read` is enough to get inbox notifications.

#### Liveblocks-node (Session)

- This PR verifies that the resource (roomId or pattern) of an Access Token permissions is a string.
- Allows `*` as a pattern for Access token permissions.

### Related issues

https://github.com/liveblocks/liveblocks.io/issues/1934


packages/liveblocks-node/src/Session.ts
```diff
@@ -43,7 +43,7 @@ const READ_ACCESS = Object.freeze([
  */
 const FULL_ACCESS = Object.freeze(["room:write", "comments:write"] as const);
 
-const roomPatternRegex = /^[^*]{1,128}[*]?$/;
+const roomPatternRegex = /^(?=.+)[^*]{0,128}[*]?$/;
```
nvie, line 46: When possible, I'd try to avoid lookaround patterns in regexes. In some cases they are unavoidable, but in this case, it looks like you're only using it to enforce that the regex isn't the empty string, correct?

Maybe simpler to read and understand if you just add the `*` case explicitly?

```suggestion
const roomPatternRegex = /^([*]|[^*]{1,128}[*]?)$/;
```

packages/liveblocks-node/src/__tests__/Session.test.ts
```diff
@@ -87,6 +88,14 @@ describe("authorization (new API)", () => {
     ).toThrow("Invalid room name or pattern");
   });
 
+  test("throws when room or pattern is undefined", () => {
+    const session = makeSession();
+    expect(() =>
+      // @ts-expect-error - Deliberate incorrect undefined value
+      session.allow(undefined, session.FULL_ACCESS).serializePermissions()
```
nvie, line 95: This will also fail if it's `null`, `false`, a number, etc etc. Maybe change the runtime check to `typeof roomIdOrPattern === 'string'` instead?

packages/liveblocks-node/src/__tests__/Session.test.ts
```diff
@@ -95,6 +104,12 @@ describe("authorization (new API)", () => {
     );
   });
 
+  test("allows asterisk as pattern", () => {
+    expect(makeSession().allow("*", [P1]).serializePermissions()).toEqual({
+      "*": [P1],
+    });
```
nvie, line 110: Can't believe we missed this case! 🤦‍♂️ Totally my bad, thanks for catching and fixing this!

ofoucherot: All good! It wasn't needed before Notifications

packages/liveblocks-node/src/Session.ts
ofoucherot: yes!

CTNicholas (approved):
Thanks for writing the error page!

marcbouchenoire (approved):
Should we cherry-pick the error page into `main` after this PR is merged? Otherwise the link won't be real until `beta` is merged.

docs/pages/errors/liveblocks-client/access-tokens-not-enough-permissions.mdx
```diff
@@ -0,0 +1,26 @@
...
+for the requested usage (Notifications). The token grants access to only one
+room which is not enough when the request is made outside a room.
+
+### Possible ways to fix it
+
+In your auth endpoint, when the request's room ID is undefined, you should
+create a token that grants access to multiple rooms. Example:
+
+```
+ session.allow("orga1*, session.FULL_ACCESS);
```
marcbouchenoire, line 22:
```suggestion
session.allow("orga1*, session.FULL_ACCESS);
```

🙊

ofoucherot:
> Should we cherry-pick the error page into `main` after this PR is merged? Otherwise the link won't be real until `beta` is merged.

@marcbouchenoire Chris is going to redo this error page later, we can cherry pick it into main when it's done.

PR #1455, opened 2024-02-12
Improve unpacking DX slightly

nvie:
This declares that a `user` field will not exist in "reset" events, but you can still unpack it at least this way.

Fixes this type error:

```ts
useOthersListener(({ type, user }) => {
  //                       ^^^^ ❌ (before) Property `user` does not exist
  //                            ✅ (after) No longer
});
```

Also this:

```ts
useErrorListener((err) => {
  err.code
  //  ^^^^ ❌ (before) Property `code` does not exist
  //       ✅ (after) No longer
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Unreleased
+
+### `@liveblocks/react`
+
+- Fix type definitions of `useOthersListener` hook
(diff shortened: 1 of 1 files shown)


CTNicholas (approved):
Thank you!

PR #1456, opened 2024-02-12
Update optimistic updates to include `updatedAt`/`editedAt` field and use for thread/comment comparison.

nimeshnayaju:
This PR updates the mutation hooks to compare threads/comments before merging updates received from server after an optimistic update is made. This is because while a request is in flight, threads/comments might have been updated by another action, so this should prevent some race conditions.

- [x] useEditThreadMetadata
- [x] useCreateComment
- [x] useEditComment
- [x] useDeleteComment
- [x] useAddReaction
- [x] useRemoveReaction


marcbouchenoire (approved):
LGTM! I didn't look _too_ deep into the tests, I'm approving in case missing a review is blocking you.

PR #1458, opened 2024-02-13
Fetch updates since last requested time when room mounts + Set up one poller per room

nimeshnayaju:
This PR aims to fix https://github.com/liveblocks/liveblocks.io/issues/1939 by making the following changes:

- [x] When a `RoomProvider` remounts after being unmounted, we want to retrieve updates since the last time we requested them. We can do so by utilizing the `lastRequestedAt` timestamp, which represents the timestamp when threads and notification updates were last requested. 

Some other improvements that this PR introduces:
- [x] Each room now has their own poller that fetches any thread/notification updates every 5 minutes.

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -1126,6 +1126,8 @@ export function createRoomContext<
     >
   >();
 
+  const roomCountByRoomId = new Map<string, number>();
+
   let isFetchingThreadsUpdates: boolean = false;
 
   async function getThreadsUpdates(roomId: string) {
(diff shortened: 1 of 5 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -1983,10 +2011,6 @@ export function createRoomContext<
     React.useEffect(() => {
       const queryKey = makeNotificationSettingsQueryKey(room.id);
       void getInboxNotificationSettings(room, queryKey);
-
-      incrementQuerySubscribers(queryKey);
```
nimeshnayaju, line 1987: We used to poll room notification settings for the user every 5 minutes (like we did for threads and notifications) but I am curious why we do that? 

@marcbouchenoire In an upcoming PR, I want to increase/decrease polling interval for threads/notifications based on the room connection status, etc, and I am guessing we wouldn't want the same for polling room notification setting if we even do in the first place.

nimeshnayaju: In this PR, I have removed polling room notification settings, but I am happy to introduce it back if you think there is a need for that

marcbouchenoire: I haven't worked on this but I assume it's because there's no WS event for this, so without polling you could have infinitely stale scenarios (e.g. you changed a notification settings dropdown to "none" in one tab and another tab with the same room never gets this change until a reload)

nimeshnayaju:
Closing this in favour of https://github.com/liveblocks/liveblocks/pull/1460. Will create another PR that includes just the poller improvements

PR #1460, opened 2024-02-14
Retrieve threads/notifications updates since last requested at when room remounts

nimeshnayaju:
This PR aims to fix https://github.com/liveblocks/liveblocks.io/issues/1939 by making the following changes:

- [x] When a `RoomProvider` remounts after being unmounted, we want to retrieve updates since the last time we requested them. We can do so by utilizing the `lastRequestedAt` timestamp, which represents the timestamp when threads and notification updates were last requested.

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -1168,10 +1168,14 @@ export function createRoomContext<
 
   const DEFAULT_DEDUPING_INTERVAL = 2000; // 2 seconds
 
-  const lastRequestedAtByRoom = new Map<string, Date>();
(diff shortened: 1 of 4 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -443,7 +443,9 @@ export function createRoomContext<
             >
           }
         >
-          {props.children}
+          <CommentsRoomProvider room={room}>
+            {props.children}
+          </CommentsRoomProvider>
```
marcbouchenoire, line 448: It seems like we don't necessarily need this additional layer and could do `void getThreadsUpdates(room.id);` here instead, next to connecting to rooms, etc.

marcbouchenoire (changes requested):
LGTM, I'm mostly curious about whether we can avoid creating `CommentsRoomProvider`.

PR #1462, opened 2024-02-15
Fix error handling inside inbox notification api implementation inside room

nimeshnayaju:
Fixes https://github.com/liveblocks/liveblocks.io/issues/1955

`onMutationFailure` method throws the error if the error isn't an instance of `CommentsApiError` or `NotificationsApiError`, and the notification related room methods threw `Error` object instead of `NotificationsApiError`, so this wasn't being handled correctly.

```tsx
function onMutationFailure(
  innerError: Error,
  optimisticUpdateId: string,
  createPublicError: (error: Error) => CommentsError<TThreadMetadata>
) {
  store.set((state) => ({
    ...state,
    optimisticUpdates: state.optimisticUpdates.filter(
      (update) => update.id !== optimisticUpdateId
    ),
  }));

  if (innerError instanceof CommentsApiError) {
    const error = handleApiError(innerError);
    commentsErrorEventSource.notify(createPublicError(error));
    return;
  }

  if (innerError instanceof NotificationsApiError) {
    handleApiError(innerError);
    // TODO: Create public error and notify via notificationsErrorEventSource?
    return;
  }

  throw innerError;
}
```

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -1,5 +1,8 @@
 import type { AuthManager, AuthValue } from "./auth-manager";
-import { getAuthBearerHeaderFromAuthValue } from "./client";
+import {
+  getAuthBearerHeaderFromAuthValue,
+  NotificationsApiError,
+} from "./client";
 import type {
   Delegates,
   LegacyConnectionStatus,
@@ -2832,7 +2835,7 @@ export function createRoom<
     fetchClientApi
   );
 
-  async function fetchJson<T>(
+  async function fetchInboxNotificationJson<T>(
     endpoint: string,
     options?: RequestInit
   ): Promise<T> {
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -2832,7 +2835,7 @@ export function createRoom<
     fetchClientApi
   );
 
-  async function fetchJson<T>(
+  async function fetchInboxNotificationJson<T>(
```
nimeshnayaju, line 2838: @marcbouchenoire Calling it `fetchInboxNotificationJson` because it specifically throws `NotificationsApiError`. Do you have an alternative suggestion for the name?

marcbouchenoire: Since it's just `NotificationsApiError` we could call it `fetchNotificationsJson`?

marcbouchenoire (approved):

packages/liveblocks-core/src/room.ts
nimeshnayaju: Yes! I like it over `fetchInboxNotificationJson` and done!

PR #1463, opened 2024-02-15
Fix race condition that could leave zombie websockets open indefinitely

### LINKED ISSUE #1459: Ghost Websocket Connections!
Dear @nvie , first of all I'd like to start with appreciating the level & amount of effort you've put in in implementing the websocket connection system based on a FSM, it truly is very impressive. I was going through your implementation today & may have found a bug/race condition (3 race conditions to trigger the same thing). I am not sure if it not an issue & left out intended since mostly the connection is handled by the library internally (leaving lesser scope for error). 

**Describe the bug**

They way the FSM currently handles the websocket connection & reconnection (even closes), does not accounts for the case where the connection may be accepted/open after the timeout is over, or the machine moves to a different state while the connection is being open. to be more precise I am talking about the `@connecting.busy` state of the machine.  

assume the given scenario: 
  - machine is in `@connecting.busy` state
  - a manual reconnect occurs at the same time
  - machine will now mark the current `@connecting.busy` as stale, and ignore it
  - machine moves to recreating a new connection, due to the manual reconnect
  - uh oh.. but the previous websocket connection attempt is still there leading to an zombie open socket

this above scenario can also happen with if the connection attempt resolves after the `SOCKET_CONNECT_TIMEOUT`. well this scenario is extremely unlikely since the variable is set to 10sec. I'd assume just like reconnect , a manual close can also cause this, haven't tried/tested that case yet but is extremely likely  

**To Reproduce**

there aren't very clear simple steps to reproduce this, but a few that I've tried to make it easier to see in action.

1. (easiest reproducibility)  Manually changing the `SOCKET_CONNECT_TIMEOUT` for the first time to a very low amount like 50, or any amount that is lesser than what it might actually take for the connection promise to resolve. example piece of code I added to lib code itself 
```typescript
      var SLOW_FIRST_TRY = () => {
        if(firstTry) {
          firstTry = false
          return 100
        }

        return SOCKET_CONNECT_TIMEOUT
      } 


      return withTimeout(
        connect$,
        SLOW_FIRST_TRY(),
        "Timed out during websocket connection"
      ).then(
 ```

2. (unpredictable reproducibility)  having a manual reconnect button which `onClick` calls `room.reconnect()`, this has some unpredictable reproducibility but should work. as a matter of fact I was able to generate >=20  ghost connections doing so. effectively reaching the room limit for my free plan. 
dqwdqwdqwd

**Expected behavior**

 Well my examples above may be devoid of actual real world usage of liveblocks, where this case may not even be there, there still exists a possibility of a this bug/race condition happening out there in the wild wild west. so Ideally these ghost connections should be taken care of and cleared/disconnected. 
 
 
 thankx for reading all this if you've made this far, have a great day, and a fantastic year. you people are doing really cool work ^^

### PR DESCRIPTION
**Pull request by nvie**
Fixes #1459.

This PR fixes a race condition in the client that could leave zombie websockets open indefinitely in a small edge case. If this bug manifests, one client would hold multiple connections to the same room, making it appear as if multiple users were connected to the room, and taking up valuable seats in the room. (This incorrect state would only exist temporarily, since the lack of heartbeat would kill those zombie sockets a bit later anyway.)

Can you confirm this fixes the issue, @dev-bahttps://github.com/liveblocks/liveblocks/pull/1463dace?

Thanks so much for reporting and providing these amazingly clear replication steps, @dev-badace! 😍 This made the bug obvious and the fix easy.

### DIFF
--- a/packages/liveblocks-core/src/lib/fsm.ts
+++ b/packages/liveblocks-core/src/lib/fsm.ts
@@ -362,31 +362,37 @@ export class FSM<
 
   public onEnterAsync<T>(
     nameOrPattern: TState | Wildcard<TState>,
-    promiseFn: (context: Readonly<TContext>) => Promise<T>,
+    promiseFn: (context: Readonly<TContext>, signal: AbortSignal) => Promise<T>,
     onOK: Target<TContext, AsyncOKEvent<T>, TState>,
     onError: Target<TContext, AsyncErrorEvent, TState>
   ): this {
     return this.onEnter(nameOrPattern, () => {
-      let cancelled = false;
+      const abortController = new AbortController();
+      const signal = abortController.signal;
 
-      void promiseFn(this.currentContext.current).then(
+      let done = false;
+      void promiseFn(this.currentContext.current, signal).then(
         // On OK
         (data: T) => {
-          if (!cancelled) {
+          if (!signal.aborted) {
+            done = true;
             this.transition({ type: "ASYNC_OK", data }, onOK);
           }
         },
 
         // On Error
         (reason: unknown) => {
-          if (!cancelled) {
+          if (!signal.aborted) {
+            done = true;
             this.transition({ type: "ASYNC_ERROR", reason }, onError);
           }
         }
       );
 
       return () => {
-        cancelled = true;
+        if (!done) {
+          abortController.abort();
+        }
       };
     });
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Unreleased
 
+### `@liveblocks/client`
+
+- Fix race condition in client that could leave zombie websockets open
+  indefinitely in a small edge case (thanks for reporting, @dev-badace)
+
 ### `@liveblocks/react`
 
 - Fix type definitions of `useOthersListener` hook
--- a/packages/liveblocks-core/src/connection.ts
+++ b/packages/liveblocks-core/src/connection.ts
@@ -585,6 +585,7 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
       //
       async (ctx) => {
         let capturedPrematureEvent: IWebSocketEvent | null = null;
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by dev-badace** [comment]
@nvie  I might be wrong, but i think it fixes 1/3 race conditions. the other 2 cases where the promise is resolved but the machine has moved to a different state either via 'reconnect' or 'close' will still leave the zombie connections open no?.  let me get back to you after testing/confirming it

**Comment by nvie** [comment]
Hmm, you may be right! The real issue here may be that `.onEnterAsync()` does not take an abort controller. Exiting a promised-based state should trigger an explicit abort of the promise. Maybe I can add that in. I'll have a play with this.

**Comment by dev-badace** [comment]
another interesting observation is that those zombie connections are staying alive forever (even without ping/pong). and they're not actually contributing to the `useOthers` count to go up, I guess this is by design. having to wait for every other peer's presence message (so i assume the presence is not stored in the server). just an interesting observation ^^

**Comment by dev-badace** [comment]
this is perfect! @nvie , works like a charm ✨  . I am no longer able to reproduce any zombie connections.   👏 👏 🚀  great work 🚀 👏 👏

File: packages/liveblocks-core/src/lib/fsm.ts [hunk 1]
```diff
@@ -362,31 +362,37 @@ export class FSM<
...
   public onEnterAsync<T>(
     nameOrPattern: TState | Wildcard<TState>,
-    promiseFn: (context: Readonly<TContext>) => Promise<T>,
+    promiseFn: (context: Readonly<TContext>, signal: AbortSignal) => Promise<T>,
     onOK: Target<TContext, AsyncOKEvent<T>, TState>,
     onError: Target<TContext, AsyncErrorEvent, TState>
   ): this {
     return this.onEnter(nameOrPattern, () => {
-      let cancelled = false;
+      const abortController = new AbortController();
```
jrowny, line 370: TIL about AbortController, neat!

jrowny (approved):
Nice! Thanks for the report @dev-badace and the fix @nvie

PR #1470, opened 2024-02-16
Make the E2E tests a bit more robust

nvie:
This PR:

- Uses a single `createClient()` helper for all pages, which sets the base URL and configures a lower throttle time.
- Uses `waitUntilFlushed()` helper which you can click after pressing buttons in the UI, to ensure that enough time has passed to flush the ops over the network, before asserting. Oftentimes, the assertions are already waiting, but this makes it more explicit and also silghtly more robust when it comes to edge cases.
- Lowers the default throttle time for our client to 16ms.

I've updated a handful of tests by adding in those `await waitUntilFlushed()` calls, but there are more tests that would benefit from this. The best way to find which ones is by setting the default throttle time to 1000 milliseconds, which increases the change of those tests failing.

### DIFF
--- a/e2e/next-sandbox/test/utils.ts
+++ b/e2e/next-sandbox/test/utils.ts
@@ -4,6 +4,7 @@ import { chromium, expect } from "@playwright/test";
 import _ from "lodash";
 
 import { randomInt } from "../utils";
+import { DEFAULT_THROTTLE } from "../utils/createClient";
 
 export type IDSelector = `#${string}`;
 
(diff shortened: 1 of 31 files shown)


e2e/next-sandbox/test/utils.ts
```diff
@@ -173,6 +174,13 @@ export async function waitForTextContains(
   );
 }
 
+export async function waitUntilFlushed() {
+  // This isn't a fancy implementation. It just waits for <default throttle>
+  // millis, so by the time the test continues, the messages have been sent to
+  // the server ¯\_(ツ)_/¯
```
ofoucherot, line 180: If it works... :D

ofoucherot (approved):
💯

PR #1479, opened 2024-02-22
Add error retrying logic to `useThreads` and `useThreadsSuspense`

nimeshnayaju:
This PR adds error retrying logic to `useThreads` and `useThreadsSuspense`. If an error is encountered while fetching threads/notifications inside `useThreads`(and `useThreadsSuspense`), attempt to fetch is done again in 5 seconds. If the retry attempt fails, we retry again in 10 seconds, 20 seconds, 40 seconds, and so on until we've retried for the maximum number of times (which is set to 8).

- [x] Retry on error while fetching threads/notifications inside `useThreads`.
- [x] Test retrying on error for `useThreads`.
- [x] Test retrying on error for `useThreadsSuspense`.

Next steps (in separate PRs): 
- Retry on error while fetching notification settings inside `useRoomNotificationSettings`
- Retry on error while fetching notification settings inside `useInboxNotifications`

### DIFF
--- a/packages/liveblocks-react/src/__tests__/useThreads.test.tsx
+++ b/packages/liveblocks-react/src/__tests__/useThreads.test.tsx
@@ -17,7 +17,13 @@ import React, { Suspense } from "react";
 import { ErrorBoundary } from "react-error-boundary";
 
 import { createLiveblocksContext } from "../liveblocks";
-import { createRoomContext, generateQueryKey, POLLING_INTERVAL } from "../room";
+import {
+  createRoomContext,
+  ERROR_RETRY_INTERVAL,
+  generateQueryKey,
+  MAX_ERROR_RETRY_COUNT,
+  POLLING_INTERVAL,
+} from "../room";
 import { dummyInboxNoficationData, dummyThreadData } from "./_dummies";
 import MockWebSocket, { websocketSimulator } from "./_MockWebSocket";
 import {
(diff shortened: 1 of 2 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -1006,13 +1006,7 @@ export function createRoomContext<
     throw innerError;
   }
 
-  const requestsCache: Map<
-    string,
-    {
-      promise: Promise<any> | null;
```
nimeshnayaju, line 1012: We store the `getThreads` and `getRoomNotificationSettings` promise for query keys but do not use it anywhere. I do not believe we plan to use it in future either, so removing this to an easier to understand map of `query key` -> `subscribers count`.

packages/liveblocks-react/src/room.tsx
```diff
@@ -124,6 +124,10 @@ const STABLE_EMPTY_LIST = Object.freeze([]);
 
 export const POLLING_INTERVAL = 5 * 60 * 1000; // 5 minutes
 
+export const ERROR_RETRY_INTERVAL = 5000; // 5 seconds
+
+export const MAX_ERROR_RETRY_COUNT = 5;
```
nimeshnayaju, line 129: `ERROR_RETRY_INTERVAL` represents the initial timeout for the error retrying, which is then exponentially incremented for `MAX_ERROR_RETRY_COUNT` number of times. These values are arbitrarily chosen, please feel free to suggest any other values that would make more sense.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1075,66 +1073,85 @@ export function createRoomContext<
...
 
+  /**
+   * Retries an action using the exponential backoff algorithm
+   * @param action The action to retry
+   * @param retryCount The number of times the action has been retried
+   */
+  function handleError(action: () => void, retryCount: number) {
+    if (retryCount >= MAX_ERROR_RETRY_COUNT) return;
+
+    const timeout = Math.pow(2, retryCount) * ERROR_RETRY_INTERVAL;
```
nimeshnayaju, line 1120: This calculates the delay before retrying a failed operation, increasing exponentially with each attempt.

First retry (retryCount = 0):

```ts
timeout = 2^0 * 5000 = 1 * 5000 = 5000ms (5 seconds)
```

Second retry (retryCount = 1):

```ts
timeout = 2^1 * 5000 = 2 * 5000 = 10000ms (10 seconds)
```

and 20000, 40000, 80000 ms and so on.

packages/liveblocks-react/src/__tests__/useThreads.test.tsx
```diff
@@ -1063,6 +1071,193 @@ describe("useThreads", () => {
...
+    await waitFor(() => expect(getThreadsReqCount).toBe(1));
+
+    expect(result.current).toEqual({
+      threads: [],
+      isLoading: false,
+      error: expect.any(Error),
+    });
+
+    // The first retry should be made after ERROR_RETRY_INTERVAL * 2^0
+    jest.advanceTimersByTime(ERROR_RETRY_INTERVAL);
```
nvie, line 1118: Small pet peeve of mine: when writing tests, we should try to avoid importing constants and instead use literals or other static fixture data, to make the test cases easier to read and understand. (For actual code, this would be fine, of course!)

In this case, I'd suggest:

```suggestion
    jest.advanceTimersByTime(5 * SECONDS);
```

If ever we'd change the `ERROR_RETRY_INTERVAL`, I'd want this test case to fail.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1075,66 +1073,85 @@ export function createRoomContext<
...
       poller.stop();
     }
   }
 
+  /**
+   * Retries an action using the exponential backoff algorithm
+   * @param action The action to retry
+   * @param retryCount The number of times the action has been retried
+   */
+  function handleError(action: () => void, retryCount: number) {
```
nvie, line 1117: Maybe rename this to helper to `retry`?

packages/liveblocks-react/src/room.tsx
```diff
@@ -1075,66 +1073,85 @@ export function createRoomContext<
...
+    if (retryCount >= MAX_ERROR_RETRY_COUNT) return;
+
+    const timeout = Math.pow(2, retryCount) * ERROR_RETRY_INTERVAL;
+
+    setTimeout(() => {
+      void action();
+    }, timeout);
```
nvie, line 1124: One thing to be aware of with this is that if you schedule a future event like this, there may be some concurrency bugs down the road. Maybe these don't apply here (I've not completely followed all the implementation details here), but suppose `timeout` is 10 seconds at one point and this retry gets scheduled.

Then, say, after 2 seconds, the room gets switched, or the user disconnects, or the user manually retries and that works. Then the user is happy for about 8 seconds, after which this already-scheduled retry kicks in.

Now again, this may not be an issue if we're lucky, but it could potentially become a source of weird issues like "Huh? I disconnected, and after 8 seconds my data pops up again.", or "Huh? I just switched rooms, and now I'm seeing my old rooms data suddenly again?", etc etc. (Sorry if these examples make no sense in this context.)

I went through this kind of pain in our old client before, which is why we've built the state machine for the 1.1 release, finally putting those issues behind us.

What's missing when setting things up like that is a way to manually cancel such an already-scheduled timeout in hindsight. You could do that the simple way, by tracking a `cancelled` boolean in this closure (or maybe in the `subscribersByQuery` mapping as I think you'll only ever want to run a single fetch or retry per query, right?), and checking it when the timeout fires, just to see if you're still interested in retrying it.

Alternatively, you could use an AbortController, which allows you to explicitly clear any such timeout. I personally really like that API, as it's flexible and allows you to cancel an outstanding promise or timeout at any given time or place.

Let me know if I can help in this area in any way.

packages/liveblocks-react/src/__tests__/useThreads.test.tsx
marcbouchenoire: @nvie For my personal curiosity, about "If ever we'd change the `ERROR_RETRY_INTERVAL`, I'd want this test case to fail.": what would you do in that case, update the test? Isn't that the same result in the sense that this test will always be tied to the value used?

marcbouchenoire: Thanks for the link! I read the test as "it should do something after a specific amount of time, which is a known and specific amount, not sometime in the future" but yeah that perspective could lead to unforeseen consequences.

packages/liveblocks-react/src/room.tsx
GuillaumeSalles: I just realized that we're not keeping an instance of the initial promise anymore and we don't return it for suspense. 

My understanding is that it's a requirement to make sure suspense does not regenerate a promise at every render. And I think this is also recommended when using `use`.

Am I correct? If yes, we might have a subtle suspense bug.

packages/liveblocks-react/src/__tests__/useThreads.test.tsx
nimeshnayaju: Just read the article you shared. Really like this statement from the article. Thanks for sharing.

> Unit tests are meant to repeat values as a way to provide a second point-of-view on the correctness of a given piece of code

packages/liveblocks-react/src/room.tsx
nimeshnayaju: This should not be an issue in our particular situation but you do have me thinking whether it makes sense to cancel the fetch request or timeout if the room is disconnected. Not doing so shouldn't hurt at all, but cancelling the outstanding promises and timeout is more efficient. However, it will increase the code complexity. I want to spend some more time thinking about this and maybe create another PR in future to make these additions. Thanks for offering to help @nvie, and I will definitely be seeking your help when implementing this.

packages/liveblocks-react/src/room.tsx
nvie: > My understanding is that it's a requirement to make sure suspense does not regenerate a promise at every render. And I think this is also recommended when using `use`.

Oh, interesting. Yes, it would make sense to require this from React's perspective, i.e. be able to render a component whenever it needs to for whatever reason, without having to worry about creating a ton of promises as a by-product.

> If yes, we might have a subtle suspense bug.

Do you mean in this new code? Or in our `useSuspendUntil*` functions where we're currently throwing those promises? We are indeed creating a [new promise instance every time](https://github.com/liveblocks/liveblocks/blob/c9deed316115ef76ee8c16aeb337e03bb6d45650/packages/liveblocks-react/src/room.tsx#L863-L865) the function would get executed — even though each such instance subscribes to and waits for the same event, so they'll all resolve at the same moment eventually. (Same [here](https://github.com/liveblocks/liveblocks/blob/c9deed316115ef76ee8c16aeb337e03bb6d45650/packages/liveblocks-react/src/room.tsx#L881-L884).)

That said, I just added in a counter to track how often this actually happens, and it looks like this isn't a pressing issue in practice right now. For each such instance, I found at most 2 outstanding promises, so it looks like React will not re-render that component internally again until the promise resolves. At least in my testing. Maybe there are situations where it would actually keep rendering this component, even though it's already suspended.

Either way, maybe a good idea to be ahead of that and change that code to always return the already-existing promise, using the `use()` polyfill you linked in chat.

GuillaumeSalles (approved):
🚀

nvie (approved):
Lovely work, @nimeshnayaju!

PR #1481, opened 2024-02-22
Fix localized `Timestamp` in `InboxNotification`

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1974.

The `Timestamp` instance in `InboxNotification` wasn't inheriting the locale as it should (it's set to `"en"` by default to fit with the English default localization).


pierrelevaillant (approved):
🌐

PR #1482, opened 2024-02-22
Retrieve threads/notification updates when browser comes back online

nimeshnayaju:
We should retrieve threads/notification updates when browser comes back online

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -426,7 +426,7 @@ export function createRoomContext<
       return () => {
         window.removeEventListener("online", handleIsOnline);
       };
-    }, []);
+    }, [room.id]);
 
     React.useEffect(() => {
       const pair = stableEnterRoom(roomId, frozenProps);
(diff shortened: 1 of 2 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -414,6 +414,20 @@ export function createRoomContext<
...
+    React.useEffect(() => {
+      function handleIsOnline() {
+        void getThreadsUpdates(room.id);
+      }
+
+      window.addEventListener("online", handleIsOnline);
+      return () => {
+        window.removeEventListener("online", handleIsOnline);
+      };
+    }, [room.id]);
```
marcbouchenoire, line 429: Small suggestion: this effect could be combined with the one above, `handleIsOnline` could be called instantly (and renamed to something more generic like `getThreads`) in addition to being tied to the `"online"` event, and the effect above could be removed.

marcbouchenoire (approved):
LGTM!

packages/liveblocks-react/src/room.tsx
nimeshnayaju: I wanted to have a clear separation of what these `useEffect` callbacks do. The first one fetches thread updates when room id changes. The second one (this one) listens to browser going online and performs some tasks (fetching thread updates). In future, we may want to replace this with updating the poller's interval when browser comes back online, so I think the separation might be helpful even though it is a little verbose currently. What do you think?

PR #1484, opened 2024-02-23
remove clientID/actorId hack

jrowny:
This fixes a bug where the first user in a room would not have their yjs awareness (cursors normally) reflected in other rooms.

Previously we set clientID to connectionID to avoid having to map between those values. It turns out this was not very reliable. This PR undoes that and adds clientID to presence and then maps to it.

### DIFF
--- a/packages/liveblocks-yjs/src/awareness.ts
+++ b/packages/liveblocks-yjs/src/awareness.ts
@@ -53,78 +53,38 @@ export class Awareness extends Observable<unknown> {
       [Y_PRESENCE_ID_KEY]: this.doc.clientID,
     });
     this.othersUnsub = this.room.events.others.subscribe((event) => {
+      let updates:
+        | { added: number[]; updated: number[]; removed: number[] }
+        | undefined;
+
+      this.rebuildActorToClientMap(event.others);
       // When others are changed, we emit an event that contains arrays added/updated/removed.
       if (event.type === "leave") {
         const targetClientId = this.actorToClientMap.get(
           event.user.connectionId
         );
-        // REMOVED
-        this.emit("change", [
-          {
-            added: [],
-            updated: [],
-            removed: [targetClientId],
-          },
-          "presence",
-        ]);
-        this.emit("update", [
-          {
-            added: [],
-            updated: [],
-            removed: [targetClientId],
-          },
-          "presence",
-        ]);
+        if (targetClientId !== undefined) {
+          updates = { added: [], updated: [], removed: [targetClientId] };
+        }
         // rebuild after the user leaves so we can get the ID of the user who left
         this.rebuildActorToClientMap(event.others);
       }
-
-      if (event.type === "enter") {
+      if (event.type === "enter" || event.type === "update") {
         this.rebuildActorToClientMap(event.others);
         const targetClientId = this.actorToClientMap.get(
           event.user.connectionId
         );
-        // ADDED
-        this.emit("change", [
-          {
-            added: [targetClientId],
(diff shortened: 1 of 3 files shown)


packages/liveblocks-yjs/src/awareness.ts
```diff
@@ -46,46 +48,93 @@ export class Awareness extends Observable<unknown> {
...
+          },
           "presence",
         ]);
         this.emit("update", [
-          { added: [event.user.connectionId], updated: [], removed: [] },
+          {
+            added: [this.actorToClientMap.get(event.user.connectionId)],
+            updated: [],
+            removed: [],
+          },
```
nvie, line 93: If the user is also using normal Liveblocks presence, and moving their cursor around, generating lots of update events, should we only emit these updates if the `__yjs` key changed here?

```js
{
  "__yjs": { "clientID": 1 },
  "cursor": { "x": 13, "y": 218 }  // 👈 Suppose only this part gets changed a lot
}
```

packages/liveblocks-yjs/src/awareness.ts
```diff
@@ -46,46 +48,93 @@ export class Awareness extends Observable<unknown> {
...
+      this.rebuildActorToClientMap(event.others);
       // When others are changed, we emit an event that contains arrays added/updated/removed.
       if (event.type === "leave") {
         // REMOVED
         this.emit("change", [
-          { added: [], updated: [], removed: [event.user.connectionId] },
+          {
+            added: [],
+            updated: [],
+            removed: [this.actorToClientMap.get(event.user.connectionId)],
```
nvie, line 64: Looks like this `this.actorToClientMap.get(event.user.connectionId)` part is the same everywhere below. Could we put this in a variable? Would also help if these blocks were oneliners again for readability.

packages/liveblocks-yjs/src/awareness.ts
```diff
@@ -30,6 +31,7 @@ export class Awareness extends Observable<unknown> {
   public doc: Y.Doc;
   public clientID: number;
   public states: Map<number, unknown> = new Map();
+  public actorToClientMap: Map<number, number> = new Map();
```
nvie, line 34: Maybe document the purpose of this map in a comment?

nvie (approved):
Looking good. Seeing this fix, it was obvious what the bug was. Thanks! 🙏

packages/liveblocks-yjs/src/awareness.ts
jrowny: the change event doesn't give us that information right? We'd have to actually deep compare or something.

packages/liveblocks-yjs/src/index.ts
```diff
@@ -164,15 +160,6 @@ export default class LiveblocksProvider<
...
 
   private syncDoc = () => {
-    /**
-     * If the connection changes, set the new id, this is used by awareness.
-     * yjs' only requirement for clientID is that it's truly unique and a number.
-     * Liveblock's connectionID satisfies those constraints
-     *  */
-    this.rootDoc.clientID =
-      this.room.getSelf()?.connectionId || this.rootDoc.clientID;
-    this.awareness.clientID = this.rootDoc.clientID; // tell our awareness provider the new ID
```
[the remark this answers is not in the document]
jrowny: it is now important that we don't set it. Overriding the rootDoc.clientID turned out to be a bad idea. In fact, I'm going to push a change removing the clientID property from the awareness class altogether because it should always use doc.clientID and not be able to be set.

PR #1437, opened 2024-01-30
1.10.0: Notifications & Comments improvements

GuillaumeSalles:

PR #1485, opened 2024-02-23
Add more vendor prefixes to default styles

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1975. We were already correctly handling inline elements like mentions but this highlighted a deeper issue:

The issue was not happening in the examples, but it was in the Starter Kit. The reason is that the styles we ship are not fully auto-prefixed when being compiled, so if your setup didn't run them through an auto-prefixer (like the Starter Kit, but unlike the examples), you could be missing some important vendor prefixes (like the one that started this issue: `box-decoration-break` is still only available as `-webkit-box-decoration-break` in the latest Chrome version for example). This PR now makes [our CSS compiler](https://lightningcss.dev/) auto-prefix the files we ship with a pretty balanced target: browsers with at least 1% usage.

### DIFF
--- a/packages/liveblocks-react-comments/src/__tests__/index.test.tsx
+++ b/packages/liveblocks-react-comments/src/__tests__/index.test.tsx
@@ -124,10 +124,10 @@ const server = setupServer(
 
 beforeAll(() => server.listen());
 afterEach(() => {
-  MockWebSocket.reset();
+  MockWebSocket.instances = [];
 });
 beforeEach(() => {
-  MockWebSocket.reset();
+  MockWebSocket.instances = [];
 });
 afterEach(() => server.resetHandlers());
 afterAll(() => server.close());
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
-# v1.10.1
+# v1.10.1 (Unpublished)
 
 ### `@liveblocks/react-comments`
 
 - Fix date localization in `InboxNotification`.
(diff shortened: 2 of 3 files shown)


packages/liveblocks-react-comments/src/__tests__/index.test.tsx
```diff
@@ -124,10 +124,10 @@ const server = setupServer(
 
 beforeAll(() => server.listen());
 afterEach(() => {
-  MockWebSocket.reset();
+  MockWebSocket.instances = [];
 });
 beforeEach(() => {
-  MockWebSocket.reset();
```
nvie, line 130: Was this accidental?

nvie (approved):

packages/liveblocks-react-comments/src/__tests__/index.test.tsx
marcbouchenoire: No, in `beta` when I rebased on `main` I saw your change to this syntax and I used it everywhere but turns out we don't use the same MSW versions across packages and so we had to [revert it for the Redux and Zustand packages](https://github.com/liveblocks/liveblocks/pull/1437/commits/5c963a3bddf60d7d919548972eeca2718de38f5e). This package is also affected but we just didn't catch it because its tests aren't running in CI, so I'm just changing it back now.

nvie: Oh, oof! Sorry I broke stuff for you unknowingly.

PR #1488, opened 2024-02-23
Get base URL from `LIVEBLOCKS_BASE_URL` envvar if not explicitly defined

nvie:
This will make it much easier to test Liveblocks against a locally running backend, i.e. by running an example instead of:

```bash
$ turbo run dev
```

…and instead just run the following command without making any config/code changes:

```bash
$ env NEXT_PUBLIC_LIVEBLOCKS_BASE_URL=http://127.0.0.1:3333 turbo run dev
```

Because this requires both the frontend (in `createClient()`) and the backend (in `new Liveblocks()`) to read the same base URL value, I've stayed pragmatic and allowed it to read the `NEXT_PUBLIC_` value (and `VITE_`) here, since in a Next.js setup it will otherwise not be visible to the frontend for obvious reasons.

For non-Next.js projects, it will also be read from `LIVEBLOCKS_BASE_URL` too.

### DIFF
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -289,7 +289,7 @@ export type ClientOptions<TUserMeta extends BaseUserMeta = BaseUserMeta> = {
 //     | ((room: string) => Promise<{ token: string }>);
 //
 
-function getBaseUrl(baseUrl: string | undefined) {
+function getBaseUrl(baseUrl?: string | undefined): string {
   baseUrl ||=
     process.env.LIVEBLOCKS_BASE_URL ||
     process.env.NEXT_PUBLIC_LIVEBLOCKS_BASE_URL ||
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -293,6 +293,7 @@ function getBaseUrl(baseUrl?: string | undefined): string {
   baseUrl ||=
     process.env.LIVEBLOCKS_BASE_URL ||
     process.env.NEXT_PUBLIC_LIVEBLOCKS_BASE_URL ||
+    process.env.VITE_LIVEBLOCKS_BASE_URL ||
     undefined;
   if (
     typeof baseUrl === "string" &&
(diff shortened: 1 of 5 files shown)


packages/liveblocks-core/src/client.ts
```diff
@@ -289,15 +289,16 @@ export type ClientOptions<TUserMeta extends BaseUserMeta = BaseUserMeta> = {
...
 //
 
-function getBaseUrlFromClientOptions(clientOptions: ClientOptions) {
-  if ("liveblocksServer" in clientOptions) {
-    throw new Error("Client option no longer supported");
-  }
+function getBaseUrl(baseUrl?: string | undefined): string {
+  baseUrl ||=
+    process.env.LIVEBLOCKS_BASE_URL ||
+    process.env.NEXT_PUBLIC_LIVEBLOCKS_BASE_URL ||
```
marcbouchenoire, line 295: We could also add `VITE_LIVEBLOCKS_BASE_URL` to be exhaustive and cover the remaining examples (Vite covers non-Next.js React examples but also Vue, Svelte, etc).

packages/liveblocks-node/src/utils.ts
```diff
@@ -1,6 +1,21 @@
 import type { Brand } from "@liveblocks/core";
 
-export const DEFAULT_BASE_URL = "https://api.liveblocks.io";
+const DEFAULT_BASE_URL = "https://api.liveblocks.io";
+
+export function getBaseUrl(baseUrl?: string | undefined): string {
+  baseUrl ||=
+    process.env.LIVEBLOCKS_BASE_URL ||
+    process.env.NEXT_PUBLIC_LIVEBLOCKS_BASE_URL ||
```
marcbouchenoire, line 8: Ditto, regarding `VITE_LIVEBLOCKS_BASE_URL`.

marcbouchenoire (approved):
Great improvement. 👌

packages/liveblocks-core/src/client.ts
nvie: Great point. Added in ad2e1c5261aabd652056cdad1286cf097886797d 👍

PR #1489, opened 2024-02-23
Retry on error when fetching room notification settings and inbox notifications

nimeshnayaju:
### `useRoomNotificationSettings`

- [x] Retry fetching room notification settings on error
- [x] Test retry on error logic
- [x] Exclude fetching room notifications settings from polling

### `useInboxNotifications`

- [x] Retry fetching inbox notifications on error
- [x] Clear error if polling inbox notifications returns a successful response
- [x] Test retry on error logic

### DIFF
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -245,6 +245,10 @@ export function createLiveblocksContext<
       throw fetchInboxNotifications();
     }
 
+    if (query.error !== undefined) {
+      throw query.error;
+    }
+
     React.useEffect(() => {
       incrementInboxNotificationsSubscribers();
 
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -25,6 +25,7 @@ import { useSyncExternalStore } from "use-sync-external-store/shim/index.js";
 import { useSyncExternalStoreWithSelector } from "use-sync-external-store/shim/with-selector.js";
 
 import { selectedInboxNotifications } from "./comments/lib/selected-inbox-notifications";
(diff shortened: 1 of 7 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -1036,30 +1036,6 @@ export function createRoomContext<
...
 
-      const notificationSettingsQuery = makeNotificationSettingsQueryKey(
-        room.id
-      );
-
-      if (requestsByQuery.has(notificationSettingsQuery)) {
-        requests.push(
-          room[kInternal].notifications
-            .getRoomNotificationSettings()
-            .then((settings) => {
```
nimeshnayaju, line 1047: Polling room notification settings currently only serves the purpose of syncing room notification settings if a user has multiple browsers (or browser tabs) open - but syncing only happens every 5 minutes.

I have removed fetching room notification settings from the polling logic to not overcomplicate the logic for now. Keeping this logic meant making sure the error is cleared if polling returned a successful response when the fetch call from `useRoomNotificationSettings` returned an error and having to account for a few more cases than we normally would.

I am happy to introduce this back in future (with more thought on how we can make this work even more seamlessly across browser tabs) and if customers desire this feature.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -231,6 +245,10 @@ export function createLiveblocksContext<
       throw fetchInboxNotifications();
     }
 
+    if (query.error !== undefined) {
+      throw query.error;
+    }
```
nimeshnayaju, line 250: We had missed this somehow. Figured out when I was writing a test to verify error boundary was triggered.

marcbouchenoire (approved):
It's great to have all these tests 🙌

PR #1494, opened 2024-02-28
Add tests for `useUser` and `useRoomInfo`

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1860. It adds basic tests for `useUser` and `useRoomInfo`, which are the hooks to consume `resolveUsers` and `resolveRoomsInfo` respectively.

Adding the tests made me discover 2 issues:
- **(Fixed)** Because we were doing `{ ...state, user: state.data }` or similar, both hooks were returning their data twice, in the `data` key which shouldn't exist and the `user`/`info` one.
- **(Not fixed yet, open question)** In the types, `resolveUsers` and `resolveRoomsInfo` are allowed to return `undefined` either for a whole batch of user IDs or specific one(s) in the returned array (e.g. the return type of `resolveUsers` is currently `OptionalPromise<(TUserMeta["info"] | undefined)[] | undefined>`). The rationale for this is that sometimes you might not be able to return anything (e.g. a non-existing user, a non-allowed private room, etc). But as proved by the two `test.failing` tests (in https://github.com/liveblocks/liveblocks/pull/1494/commits/f0d440c3034322f2a8b1b9067a880bacb13f77c9#), this isn't actually supported. We can fix that, but that made me realize of a bigger issue: if we allow that, that means the return type of the hooks should reflect that: `user`/`info` in `const { user/info, isLoading, error } = useUser("user-id")` could always be `undefined`, even if `isLoading` is false, even with Suspense, etc. The hooks are not typed like this at the moment, and I feel like it would make using the hooks pretty difficult by always having to check that `user`/`info` exists. Both options below are technically breaking changes, but option 1 is far less hurtful because returning `undefined` isn't really a behavior that we publicly supported.
  - **Option 1:** remove `undefined` from the return type of `resolveUsers` and `resolveRoomsInfo`: you have to either error out or return an array of results.
  - **Option 2:** correctly support `undefined` and update the return type of `useUser` and `useRoomInfo` to reflect that `user`/`info` could always be `undefined`.

I'm curious to hear your thoughts on the second issue. cc: @GuillaumeSalles

---

### Update

For the second issue above, I went ahead with a **third option**: the types of `resolveUsers`/`resolveRoomsInfo` don't change (they still support returning `undefined`) and neither do the types of `useUser`/`useRoomInfo` (they still reflect that if `isLoading` is false, the data (`user`/`info`) will not be `undefined`). To achieve that, returning `undefined` in `resolveUsers`/`resolveRoomsInfo` is now the same as throwing/returning an error. In non-Suspense, the `error` property will be returned by `useUser`/`useRoomInfo`, and in Suspense, the closest error boundary will be triggered.

So, no type breaking changes but a **functional breaking change** for when users use the Suspense version of `useUser`/`useRoomInfo`:
- Non-Suspense version: If you somehow returned `undefined` before, no error was triggered but would still get the data as `undefined` so you had to check for it. Now that an error is returned, it's the same: the data will be `undefined` and you will have to check for it.
- Suspense version: If you somehow returned `undefined` before, you could get away without an error boundary by just checking if the data is `undefined`. This setup will break as it now needs an error boundary.

I'm still curious about your thoughts on this, and if the option I chose seems correct.

Given that there's a functional breaking change, I'd like to wait for `1.11.0` to release this change.


nimeshnayaju:
> To achieve that, returning undefined in resolveUsers/resolveRoomsInfo is now the same as throwing/returning an error. In non-Suspense, the error property will be returned by useUser/useRoomInfo, and in Suspense, the closest error boundary will be triggered.

This makes sense to me 👍

PR #1507, opened 2024-03-05
Fix bug: calling `.clone()` after creating new `LiveObject` could sometimes throw

nvie:
The bug is fixed by skipping over explicitly-`undefined` values when initializing `LiveObject`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,16 @@
+# v1.10.2 (unreleased)
+
+### `@liveblocks/client`
+
+- Fix bug where calling `.clone()` immediately after creating a new `LiveObject`
+  could throw an error
+
 # v1.10.1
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Fix bug where the client’s backoff delay would not be respected correctly in a
   small edge case.
+- Fix bug where calling `.clone()` immediately after creating a new `LiveObject`
```
marcbouchenoire, line 7: `1.10.1` is already released. 😅

marcbouchenoire (approved):
🙌

CHANGELOG.md
nvie: Thanks, will update the changelog 👍

PR #1512, opened 2024-03-06
Add "mark as read" action to `InboxNotification`

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1958.

It adds a `•••` button with a dropdown to `InboxNotification`, which will only include one "mark as read" action for now.

Because we cannot mark notifications as unread for now, the action becomes disabled for read notifications. I wonder if we should hide the `•••` button entirely given that there's only one dropdown item for now.

`InboxNotification` will be used in popovers in many cases, but it's hard to know in advance how many issues related to that there will be, things like pressing "escape" closes the whole popover instead of just the dropdown, etc.

This PR also improves a few other things like some spacings in `InboxNotification` and the actions on hover in `Comment`: they currently always reserve space even when they're not visible, with this PR they only take space when visible.

https://github.com/liveblocks/liveblocks/assets/6959425/70cddc7c-765d-48c8-b1d9-60dec5a62fdd

https://github.com/liveblocks/liveblocks/assets/6959425/38b80af6-625c-496d-bf58-fb3647aa7794

### To do

- [x] More testing
- [x] Update documentation
- [x] Update `CHANGELOG.md`


nimeshnayaju (approved):
🙌

PR #1515, opened 2024-03-06
add modifyStorage api to Node

jrowny:
- Adds a new modifyDocumentStorage method to node client

Requires backend modification 

Use it like this:
```

const liveblocks = new Liveblocks({ secret:  "sk_dev_.....",});

await liveblocks.modifyStorageDocument("roomId", (root) => {
  const obj = root.get("someObject");
  obj.set("99", 12345);
});

```

Root will be a fully hydrated LiveObject.

With types:

```
type MyRoomStorage =  {
    someObject: LiveObject<{
      [key: string]: number | LiveObject<{ a: number }>;
    }>
}
await liveblocks.modifyStorageDocument<MyRoomStorage>("roomId", (root) => {
  const obj = root.get("someObject");
  obj.set("99", 12345);
});
```

### DIFF
--- a/packages/liveblocks-core/src/crdts/LiveObject.ts
+++ b/packages/liveblocks-core/src/crdts/LiveObject.ts
@@ -87,7 +87,6 @@ export class LiveObject<O extends LsonObject> extends AbstractCrdt {
     return [root, parentToChildren];
   }
 
-  /** @internal */
   static _fromItems<O extends LsonObject>(
     items: IdTuple<SerializedCrdt>[],
     pool: ManagedPool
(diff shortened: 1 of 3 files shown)


nvie (commented):
Conceptually, I love this! The flow is so easy to follow:

- Download the full storage document
- Mutate it
- Discard the results
- **Todo:** error if an error happened on the server

See my code review comments on the server side. I think we'll need to figure out how to best define the concept of a "server session". I think the current `"si"` prefix we use to initialize a room from the REST API should become just a special case of whatever server session abstraction we land on.

Amazing work! 🙌

PR #1516, opened 2024-03-06
Scroll into (comment) view when a inbox notification link is clicked

nimeshnayaju:
Fixes: https://github.com/liveblocks/liveblocks.io/issues/1891

This PR updates inbox notification links in the inbox notification lists to append the associated comment id as the hash. When a `useThreads` is initially loaded, it looks at this hash value to see if there is a comment id present and scrolls to the corresponding comment if the comment is a part of the threads returned by the `useThreads`. 

After the initial load, it defaults to browser's hash scrolling browser.

https://github.com/liveblocks/liveblocks/assets/35807215/296a6ae6-9a0e-4c52-bdac-967db8bcd319


packages/liveblocks-react-comments/src/components/InboxNotification.tsx
```diff
@@ -374,63 +372,62 @@ const InboxNotificationThread = forwardRef<
     //       it's not a big deal, the only scenario where it would be superfluous would be if the user provides their own
     //       `href` AND disables room names in the title via `showRoomName={false}`.
     const { info } = useRoomInfo(inboxNotification.roomId);
-    const { unread, date, aside, title, content, threadId, commentId } =
```
nimeshnayaju, line 377: Nothing has been changed in this `useMemo` code snippet. I only removed `threadId` from the return value because we no longer require it.

packages/liveblocks-react-comments/src/utils/url.ts
```diff
@@ -14,6 +15,8 @@ export function setQueryParams(
     }
   }
 
+  urlObject.hash = `#${hash}`;
```
marcbouchenoire, line 18: Idea: I think we should only set the hash if the incoming URL doesn't have one already, that way we respect the users' intention instead of overriding it.

marcbouchenoire (changes requested):
Great work! Really nice to see a first-party solution for this that works out-of-the-box. 🥳

I added a few questions and I have some global ones as well:
- In your video `scrollIntoView` is smooth but in the code it's called without any arguments, is that the default behavior? I think we should mimic the native browser's handling of permalinks and scroll instantly.
- We might not need to support full customization at first (or ever) but I do think we should add an option (`useThreads({ scrollOnLoad: false })` or similar?) to both versions of `useThreads` (in their existing "options" object argument) to disable the behavior: I think it's always useful to provide escape hatches for these default opinionated behaviors. Also, given that we use the `query` option for the cache, we should be careful to see that this new option isn't being used when making cache keys for queries.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1216,12 +1216,40 @@ export function createRoomContext<
...
+        const comments = state.threads.flatMap((thread) => thread.comments);
+        const isCommentInThreads = comments.some(
+          (comment) => comment.id === commentId
+        );
+
+        // If the comment is not in the threads for this hook, we do not scroll to it
+        if (!isCommentInThreads) return;
+
+        const comment = document.getElementById(commentId);
+        if (comment === null) return;
```
marcbouchenoire, line 1244: Should we reverse this logic? Start by `getElementById` **then** make the `cm_`/state checks? I feel like `getElementById` is cheaper than the state check, so if you use primitives (they won't have comment `id`s) or for some reasons have replaced and changed comment `id`s, we would not run the state checks.

Or maybe sandwich it: start with the cheapest `cm_` check, then `getElementById`, and finally the state check.

packages/liveblocks-react/src/types.ts
```diff
@@ -33,6 +33,7 @@ export type UseThreadsOptions<TThreadMetadata extends BaseMetadata> = {
   query?: {
     metadata?: Partial<TThreadMetadata>;
   };
+  scrollOnLoad?: boolean; // (optional) boolean flag to enable or disable automatic scrolling to a comment based on the URL hash when threads are loaded. Defaults to true.
```
marcbouchenoire, line 36: We should use JSDoc (and add a comment to `query` as well) so that it shows this description on hover.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1225,39 +1272,27 @@ export function createRoomContext<
...
-
-        const comment = document.getElementById(commentId);
-        if (comment === null) return;
-
-        comment.scrollIntoView();
+        handleScrollToCommentOnLoad(state.isLoading, scrollOnLoad, state);
       },
       // eslint-disable-next-line react-hooks/exhaustive-deps -- We only want to run this effect once
-      [state.isLoading]
+      [state.isLoading, scrollOnLoad]
```
marcbouchenoire, line 1280: Should we "freeze" `scrollOnLoad` into a ref and not trigger the effect based on it? I don't think we want to support switching from `false` to `true` later in the lifecycle and then triggering the behavior.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1182,9 +1183,55 @@ export function createRoomContext<
     }
   }
 
+  /**
+   *
```
marcbouchenoire, line 1187: Missing comment.

packages/liveblocks-react-comments/src/utils/url.ts
```diff
@@ -3,8 +3,8 @@ const ABSOLUTE_URL_REGEX = /^[a-zA-Z][a-zA-Z\d+\-.]*?:/;
 
 export function generateURL(
   url: string,
-  params: Record<string, string | number | undefined>,
-  hash: string
+  params: Record<string, string | number | undefined> = {},
```
marcbouchenoire, line 6: I was thinking more to keep it `undefined` as a regular optional argument, and to skip the `for...of` loop when it is `undefined`.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1225,39 +1272,27 @@ export function createRoomContext<
 
     React.useEffect(
       () => {
-        if (state.isLoading) return;
+        if (state.isLoading === true) return;
```
marcbouchenoire, line 1275: 🤔

marcbouchenoire (changes requested):

packages/liveblocks-react-comments/src/utils/url.ts
nimeshnayaju: I thought they were effectively the same thing. That being said, I do agree the code would be more readable if I kept it undefined and performed any kind of undefined check inside the function. Updated to the following:

```ts
export function generateURL(
  url: string,
  params?: Record<string, string | number | undefined>,
  hash?: string
) {
  const isAbsolute = ABSOLUTE_URL_REGEX.test(url);
  const urlObject = new URL(url, isAbsolute ? undefined : PLACEHOLDER_BASE_URL);

  if (params !== undefined) {
    for (const [param, value] of Object.entries(params)) {
      if (value) {
        urlObject.searchParams.set(param, String(value));
      }
    }
  }

  // Only add the new hash if the URL does not already have one
  if (!urlObject.hash && hash !== undefined) {
    urlObject.hash = `#${hash}`;
  }

  return isAbsolute
    ? urlObject.href
    : urlObject.href.replace(PLACEHOLDER_BASE_URL, "");
}
```

packages/liveblocks-react/src/room.tsx
nimeshnayaju: 🙈

PR #1520, opened 2024-03-11
Export `EmojiPicker` primitive and update Emojibase

marcbouchenoire:
This PR makes the `EmojiPicker` primitive public (undocumented for now) and updates Emojibase (the emoji database we use as our source of truth).

**Context:**
If you use primitives to build your own `Thread`/`Comment` components, one big missing piece is the emoji picker. We've built our own but kept it private and reserved it for the default components. This choice was made based on the primitive's complexity, but if we never have people trying it we'll never know what works and what doesn't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v1.10.3 (Unpublished)
+# v1.10.3
 
 ### `@liveblocks/react-comments`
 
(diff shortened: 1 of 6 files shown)


GuillaumeSalles (approved):

nimeshnayaju (approved):
🚀

Issue #1521, opened 2024-03-11
Removing thread metadata via `useThreadMetadata` results in metadata being removed followed by a quick flash of the old metadata


nimeshnayaju:
**Describe the bug**

When a property in thread metadata is removed by setting the property to `null`, the thread metadata is successfully removed for a brief period of time followed by the property being set to the old property (before deletion).

PR #1522, opened 2024-03-11
Fix thread metadata not being updated correctly after reading server response

nimeshnayaju:
Fixes https://github.com/liveblocks/liveblocks/issues/1521

See failing test that verifies the bug: https://github.com/liveblocks/liveblocks/blob/a2d9cc0433640c58a63eaa60ae1e43efeb828eca/packages/liveblocks-react/src/__tests__/useEditThreadMetadata.test.tsx#L114-L198

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -1374,7 +1374,7 @@ export function createRoomContext<
         room[kInternal].comments
           .editThreadMetadata({ metadata, threadId })
           .then(
-            (metadata: TThreadMetadata) => {
+            (metadata) => {
               store.set((state) => {
                 const existingThread = state.threads[threadId];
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -1413,10 +1413,9 @@ export function createRoomContext<
                     ...state.threads,
                     [threadId]: {
                       ...existingThread,
-                      metadata: {
-                        ...existingThread.metadata,
-                        ...metadata,
-                      },
+                      metadata: metadata as [TThreadMetadata] extends [never]
+                        ? Record<string, never>
```
nimeshnayaju, line 1417: This is the fix!

The server responds with the total metadata. We used to merge the metadata received from server with the existing metadata, so there was a possibility that stale properties were still stored until a webhook message cleared them.

GuillaumeSalles (approved):

PR #1527, opened 2024-03-13
add update room id method to node client

ofoucherot:
### Description

Add "update room ID" endpoint to liveblocks node client.


nimeshnayaju (approved):
🚀

PR #1529, opened 2024-03-14
Add `guid` params to `sendYjsBinaryUpdate` and `getYjsDocumentAsBinaryUpdate` node methods

nimeshnayaju:
This PR updates the `sendYjsBinaryUpdate` and `getYjsDocumentAsBinaryUpdate` methods inside `@liveblocks/node` to accept an optional `guid` parameter.


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ### `@liveblocks/node`
 
 - Add `updateRoomId` method that lets you update room id of the specified room.
+- Add an optional `guid` parameter to `sendYjsBinaryUpdate` and
+  `getYjsDocumentAsBinaryUpdate` to point to a Yjs subdocument with the
+  specified guid.
```
nvie, line 8: Very much unrelated to this PR, but I noticed the `send` and `get` prefixes here, and wanted to share a thought I had before. Now that the client is getting more and more methods every day, would it be more convenient if we'd start "grouping" our client methods a bit more by product API, to increase discoverability in IDEs a bit more?

Example of what the DX could be like if we'd group them:

https://github.com/liveblocks/liveblocks/assets/83844/b8b3204f-ece3-4124-aad2-26c843afd738

Seems a bit more focused to me, and easier for people to discover methods that could also be of interest, even when they're not looking for them?

nvie (approved):
Nice! 💯

CHANGELOG.md
jrowny: 💯 this would be way better

jrowny (approved):

CHANGELOG.md
nimeshnayaju: I like this too! I know Stripe does something similar with their node APIs. 

```ts

const stripe = require('stripe')('sk_test');

const customer = await stripe.customers.create({
  name: 'Jenny Rosen',
  email: 'jennyrosen@example.com',
});

const balance = await stripe.balance.retrieve();
```

PR #1530, opened 2024-03-14
Correct main

nvie:
Moving these commits into `1.11` soon, but freeing up `main` for patch releases in the mean time.


marcbouchenoire (approved):
🙌

PR #1524, opened 2024-03-12
SvelteKit or Vite issue

CTNicholas:

PR #1531, opened 2024-03-14
Fix regression in Vite builds (introduced in 1.10.1)

### LINKED ISSUE #1523: Updating from v1.10.0 to v1.10.1 throws "process is not defined" error
**Describe the bug**

It seems like, from looking at the diff, [this commit](https://github.com/liveblocks/liveblocks/commit/a848e8b6fbe7deee8dc0d7023a1c9331eb908e31) is breaking our app with the error "process is not defined."

```
@liveblocks_client.js?v=21fbc75a:6692 Uncaught (in promise) ReferenceError: process is not defined
    at getBaseUrl (@liveblocks_client.js?v=21fbc75a:6692:25)
    at createClient (@liveblocks_client.js?v=21fbc75a:6716:19)
```

I know that [the docs](https://liveblocks.io/docs/platform/troubleshooting#process-not-defined) mention this error and say that we should be using a bundler. I am using what I think is a relatively straight-forward setup with SvelteKit/vite, and I still see this. I tried adding `"@liveblocks/client"` to [optimizeDeps](https://vitejs.dev/config/dep-optimization-options#optimizedeps-include) and I still see this error.

At the very least, I would have expected the bump from 1.10.0 to 1.10.1 (a patch version) to mention that this library would no longer work if it wasn't getting bundled.

nvie:
This removes the use of `process.env.WHATEVER` expressions (except `proces.env.NODE_ENV` specifically).

The reason is that Vite does not seem to rewrite any of these in the final bundle, leaving those `process.env.WHATEVER` expressions in the final bundle. Other frameworks/bundlers do rewrite these expressions to `undefined`, which is what the expected behavior was here.

It's still fine to use `process.env.NODE_ENV`, as that specific expression _is_ still recognized by Vite (as well as all other bundlers).

Fixes #1523 and #1524.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v1.10.4 (Unpublished)
+# v1.10.4
 
 - Fix bundling issue in Vite projects, where `process is not defined` could
   happen
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v1.10.4 (Unpublished)
+
+- Fix bundling issue in Vite projects, where `process is not defined` could
+  happen
+
 # v1.10.3
 
 ### `@liveblocks/react-comments`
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v1.10.4 (Unpublished)
```
marcbouchenoire, line 1: Feel free to remove `(Unpublished)` already since we should probably release the fix asap.

marcbouchenoire (approved):
😢

PR #1535, opened 2024-03-18
Yjs subdocument guide and reference

CTNicholas:
I've written up some things:
- API reference for the modified `@liveblocks/node` functions
- Some missing docs Yjs info 
- A guide on using subdocuments

Could @nimeshnayaju or @jrowny take a look at the guide? I won't have a chance to test any of the code snippets, so I'd like some eyes on this.

### DIFF
--- a/docs/pages/api-reference/liveblocks-node.mdx
+++ b/docs/pages/api-reference/liveblocks-node.mdx
@@ -737,9 +737,8 @@ await liveblocks.deleteStorageDocument("my-room-id");
 
 #### Liveblocks.getYjsDocument [#get-rooms-roomId-ydoc]
 
-Returns a JSON representation of a room’s Yjs document. To return a subdocument
-instead of the main document, pass its `guid`. Throws an error if the room isn’t
-found. This is a wrapper around the
+Returns a JSON representation of a room’s Yjs document. Throws an error if the
+room isn’t found. This is a wrapper around the
 [Get Yjs Document API](/docs/api-reference/rest-api-endpoints#get-rooms-roomId-ydoc)
 and returns the same response.
 
@@ -754,9 +753,6 @@ A number of options are available.
 
 ```ts
 const yjsDocument = await liveblocks.getYjsDocument("my-room-id", {
-  // Optional, return a subdocument instead. guid is its unique identifier
-  guid: "c4a755...",
-
   // Optional, if true, `yText` values will return formatting
   formatting: true,
 
--- a/docs/pages/api-reference/liveblocks-node.mdx
+++ b/docs/pages/api-reference/liveblocks-node.mdx
@@ -737,8 +737,9 @@ await liveblocks.deleteStorageDocument("my-room-id");
 
 #### Liveblocks.getYjsDocument [#get-rooms-roomId-ydoc]
 
-Returns a JSON representation of a room’s Yjs document. Throws an error if the
-room isn’t found. This is a wrapper around the
+Returns a JSON representation of a room’s Yjs document. To return a subdocument
+instead of the main document, pass its `guid`. Throws an error if the room isn’t
+found. This is a wrapper around the
 [Get Yjs Document API](/docs/api-reference/rest-api-endpoints#get-rooms-roomId-ydoc)
 and returns the same response.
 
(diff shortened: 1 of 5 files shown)


docs/pages/api-reference/liveblocks-node.mdx
```diff
@@ -753,6 +754,9 @@ A number of options are available.
 
 ```ts
 const yjsDocument = await liveblocks.getYjsDocument("my-room-id", {
+  // Optional, return a subdocument instead. guid is its unique identifier
+  guid: "c4a755...",
+
   // Optional, if true, `yText` values will return formatting
   formatting: true,
```
nimeshnayaju, line 761: We have not yet implemented `guid` support for the `getYjsDocument` node method. There has been discussion about deprecating this API overall especially because the result of this endpoint is not always consistent due to the lack of proper ways to decipher Yjs types on server. Do you reckon we should also add support for this endpoint?

docs/pages/api-reference/liveblocks-yjs.mdx
```diff
@@ -121,6 +121,19 @@ yProvider.on("sync", (sync: boolean) => /* ... */);
 yProvider.on("synced", (sync: boolean) => /* ... */);
 ```
 
+### LiveblocksProvider.on("subdocs") [#LiveblocksProvider.on.subdocs]
+
+Add an event listener for the `subdocs` event. The `subdocs` event is triggered
```
nimeshnayaju, line 126: I know `Y.Doc` has a listener to the `subdocs` event but wasn't aware we also emitted them from our `LiveblocksProvider`. @jrowny Could you confirm if this is true? 

```ts
doc.on('subdocs', ({ added, removed, loaded }) => {
    ...
})
```

docs/pages/api-reference/liveblocks-yjs.mdx
```diff
@@ -151,7 +184,7 @@ fire events when the document has loaded.
 
 ```ts
 // Listen for the sync event only once
-yProvider.on("sync", (isSynced: boolean) => {
+yProvider.once("sync", (isSynced: boolean) => {
```
nimeshnayaju, line 187: Not a comment, just something that crossed my mind - we don't have proper Typescript autocompletion for these event subscription :(
I hope we can improve that soon in the future

guides/pages/how-to-use-yjs-subdocuments.mdx
```diff
@@ -0,0 +1,206 @@
...
+same room, and you wish to lazy-load them individually. Each subdocument works
+similarly to a normal Yjs document, allowing you to use
+[shared types](https://docs.yjs.dev/getting-started/working-with-shared-types),
+[awareness](https://docs.yjs.dev/getting-started/adding-awareness), and more.
+
+### Not necessary for multiple text editors
+
+Please note that **subdocuments are not necessary for displaying multiple text
+editors** on one page. For this use case, it’s often best to create a
+[`Y.Map`](https://docs.yjs.dev/api/shared-types/y.map) in your Yjs document, and
```
nimeshnayaju, line 23: > ### Not necessary for multiple text editors
> 
> Please note that **subdocuments are not necessary for displaying multiple text
editors** on one page. For this use case, it’s often best to create a
[`Y.Map`](https://docs.yjs.dev/api/shared-types/y.map) in your Yjs document

Is this a common misconception around subdocuments, or if this is something our customers enquire a lot about subdocuments?

nimeshnayaju (commented):
Thank you so much for writing this @CTNicholas! It is very well written. I only have a few comments!

docs/pages/api-reference/liveblocks-node.mdx
CTNicholas: My mistake! Nah, I'll remove the documentation 🙂

docs/pages/api-reference/liveblocks-yjs.mdx
CTNicholas: This is an error! I misread the Yjs docs 👍

guides/pages/how-to-use-yjs-subdocuments.mdx
CTNicholas: I've been asked it a lot of times! Customers have thought subdocs is the only way, but they've only needed two editors at once, and this is easier.

PR #1536, opened 2024-03-18
Update examples query parameter

marcbouchenoire:
This PR updates all examples to use the new example query parameter, `exampleId` instead of `roomId`.


CTNicholas (approved):
Looks good!

PR #1537, opened 2024-03-20
Don’t pass `UserMeta` generic to `Client` in `createLiveblocksContext`

marcbouchenoire:
Generics aren't applied everywhere (especially in `@liveblocks/client`) so the current version of `createLiveblocksContext` is hard to use in practice. This PR changes `createLiveblocksContext` to behave like `createRoomContext`.


CTNicholas (approved):
Nice catch!

nvie (approved):
Looks good!

PR #1539, opened 2024-03-20
Add `nextjs-comments-notifications` example

marcbouchenoire:
This PR brings back the `nextjs-comments-notifications` example but rewritten in a simpler way, trying to keep the example/liveblocks.io-specific logic as hidden as possible.

It's quite simple on purpose, like `nextjs-comments`, it's here to be a showcase for the default components without too many additional changes.

In the examples gallery, it will work with 2 users (to be able to try Notifications) side-by-side.


CTNicholas:
Nice, glad we can add this multi-room demo to our examples gallery!

I'm seeing an error and the example doesn't work for me:

By the way, I will be away until next week),

marcbouchenoire:
@CTNicholas I can't reproduce the 2 last issues, we can fix them in follow-up PRs.

pierrelevaillant (approved):

PR #1549, opened 2024-03-28
Add support for `startsWith` metadata queries in `useThreads`

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/1993.

**🚧 WIP 🚧**

### DIFF
--- a/packages/liveblocks-core/src/types/QueryMetadata.ts
+++ b/packages/liveblocks-core/src/types/QueryMetadata.ts
@@ -1,0 +1,20 @@
+import type { BaseMetadata } from "./BaseMetadata";
+
+type QueryMetadataStringValue<T extends string> =
+  | T
+  | {
+      operator: "startsWith";
+      value: T;
+    };
+
+/**
+ * This type can be used to build a metadata query string (compatible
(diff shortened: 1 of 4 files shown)


packages/liveblocks-core/src/types/QueryMetadata.ts
```diff
@@ -0,0 +1,20 @@
...
+      value: T;
+    };
+
+/**
+ * This type can be used to build a metadata query string (compatible
+ * with `@liveblocks/query-parser`) through a type-safe API.
+ *
+ * In addition to exact values (`:` in query string), it adds:
+ * - to strings:
+ *  - `startsWith` (`^` in query string)
```
FlowFlorent, line 16: Is there a way to make this only available if the value is a `string` type?

`boolean` and `number` values can only have an exact match `:`

marcbouchenoire: Yep this type only adds `startsWith` to `string` values, all other values remain exact-only.

FlowFlorent: Great! Sorry I think I misread the comment 😅 

Thanks

PR #1550, opened 2024-03-28
Devtools: Support nested Y.js structures and subdocuments

nimeshnayaju:
This PR updates the devtools to support nested Y.js data structures and subdocuments.

- [x] Support nested Y.js data structures.
- [x] Support subdocuments. Fixes https://github.com/liveblocks/liveblocks.io/issues/1881
- [x] Fix Y.js awareness not displaying in `Awareness` tab. https://github.com/liveblocks/liveblocks.io/issues/2050

Here is a screenshot of the devtools when used with a room containing subdocuments. Notice the top type `folder` with two `subdocuments`: `profile.txt` and `info.txt`, each containing a `Y.Map` called `properties`:

Screenshot 2024-03-29 at 3 17 38 PM


packages/liveblocks-devtools/src/devtools/tabs/yjs/index.tsx
```diff
@@ -228,7 +230,7 @@ function YjsAwareness({ className, ...props }: ComponentProps<"div">) {
           className={cx(className, "absolute inset-0 flex h-full flex-col")}
           {...props}
         >
-          <YjsTree data={presence} />
+          <JsonTree data={awareness} />
```
nimeshnayaju, line 233: We used to pass the `presence` array here which doesn't display Y.js awareness but the presence information. I've updated to pass the awareness array instead and use a `JsonTree` to view the data.

Screenshot 2024-03-29 at 12 09 26 PM

marcbouchenoire: Do we want to keep properties like `__yjs_clientid` visible? I think `YjsTree` was responsible for hiding them.

packages/liveblocks-devtools/src/devtools/components/Tree.tsx
```diff
@@ -326,6 +367,31 @@ function yLogsColor(node: YLogsTreeNode): string {
...
+      return "text-orange-500 dark:text-orange-400";
+
+    case "Y.Doc":
+    case "Y.ContentAny":
+    case "Y.ContentBinary":
+    case "Y.ContentEmbed":
+    case "Y.ContentFormat":
+    case "Y.ContentString":
+      return "text-dark-500 dark:text-light-500";
+  }
```
marcbouchenoire, line 392: We should use the same colors for the same icons, so `Y.XmlFragment` should be the same color as `Y.XmlText` since they both use `<StringIcon />` for example, `Y.ContentString` too, etc. And ideally only `<EllipsisIcon />` should be gray.

packages/liveblocks-devtools/src/devtools/icons/tree.tsx
```diff
@@ -246,3 +246,46 @@ export function TrashIcon(props: ComponentProps<"svg">) {
...
+      />
+      <path
+        d="M14 6H5M16 10H5M12.5 14H5"
+        stroke="currentColor"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+      />
+    </svg>
+  );
+}
```
marcbouchenoire, line 274: We use different icon styles between the dashboard trees and the DevTools ones, I'll update this icon later today to fit the DevTools. (https://www.figma.com/file/WDmXK9fG6aD5OT1AjFVUDr/Liveblocks-%E2%80%93-DevTools?type=design&node-id=0%3A1&mode=design&t=DehwOPjPYvhVbCpt-1)

packages/liveblocks-devtools/src/devtools/icons/tree.tsx
```diff
@@ -246,3 +246,46 @@ export function TrashIcon(props: ComponentProps<"svg">) {
...
+      xmlns="http://www.w3.org/2000/svg"
+      {...props}
+    >
+      <path
+        d="M2 6.117V1.883a.5.5 0 0 1 .757-.429l3.528 2.117a.5.5 0 0 1 0 .858L2.757 6.546A.5.5 0 0 1 2 6.116Z"
+        fill="currentColor"
+      />
+    </svg>
+  );
+}
```
marcbouchenoire, line 291: I see why we should create a component for this but it previously stayed inline because this file is only about the same type of icons and I think we shouldn't mix different types/sizes in the same place. We can keep it a component but I think it should live elsewhere.

packages/liveblocks-devtools/src/devtools/tabs/yjs/to-yjs-tree-node.ts
```diff
@@ -0,0 +1,270 @@
...
+          switch (value.type) {
+            case "Y.Doc":
+              return toTreeYNode(value, key, id);
+            case "Y.Map":
+            case "Y.Array":
+            case "Y.XmlFragment":
+            case "Y.XmlElement":
+            case "Y.Text":
+            case "Y.XmlText":
+              return toTreeYNode(value, key, id);
```
jrowny, line 178: if I'm reading this correctly, all of these switch items result to toTreeYNode, so would it be better to just do like `['Y.Map', etc...].includes(value.type) return toTreeYNode(value, key, id)`

jrowny (approved):

packages/liveblocks-devtools/src/devtools/tabs/yjs/index.tsx
nimeshnayaju: I think it makes sense to at least display the client id because it might be useful with some kind of debugging? 

Values inside `__yjs` is displayed in the `Awareness` tab - so maybe we could also display this client id inside the `Awareness` tab that is purely dedicated to `Yjs`?

packages/liveblocks-devtools/src/devtools/icons/tree.tsx
nimeshnayaju: Thank you @marcbouchenoire ❤️

packages/liveblocks-devtools/src/devtools/tabs/yjs/to-yjs-tree-node.ts
nimeshnayaju: I do not like it either but TypeScript wasn't happy when I did that : (

packages/liveblocks-devtools/src/devtools/components/Tree.tsx
nimeshnayaju: Good point! Will need to do the same for the dashboard colors

packages/liveblocks-devtools/src/devtools/tabs/yjs/index.tsx
marcbouchenoire: Adding it to the `Awareness` tab would make the most sense to me (always keeping all __yjs direct values hidden), displayed differently so it's clear that it's not awareness. Maybe similarly to the presence IDs like `(#216)` next to `Me` in your screenshot?

nimeshnayaju: Going to work on this on a different PR so this PR isn't blocked!

nvie (commented):
Great work, @nimeshnayaju! 🙌

PR #1553, opened 2024-04-04
Avoid clicking on the "clear" button before storage has loaded

nvie:
…as this will trigger the `"This mutation cannot be used until connected to the Liveblocks room"` error.

### DIFF
--- a/e2e/next-sandbox/utils/Button.tsx
+++ b/e2e/next-sandbox/utils/Button.tsx
@@ -18,8 +18,9 @@ export type Props = Omit<
   };
 
 export default function Button(props: Props) {
-  const { enabled, ...rest } = props;
-  const opacity = enabled ?? true ? undefined : 0.7;
+  let { enabled, ...rest } = props;
+  enabled ??= true;
+  const opacity = enabled ? undefined : 0.7;
   return (
     <button
       {...rest}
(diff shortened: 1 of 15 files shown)


CTNicholas (approved):
👍

nvie:
Thanks for the quick review, @CTNicholas! I've just pushed some changes to make this work more generally. No button can be clicked by Playwright until it's explicitly `enabled`. In tests, it means that `await button.click()` will wait until the button becomes enabled, which is exactly what we want in most cases. Only in randomized/fuzz tests, we'll want to force a click, even if the button is currently not enabled. Before I learned about the `{ force: true }` option, before today it was structured like this because I did not know how to avoid the freeze otherwise. This setup is much much nicer! (And simpler, too.)

e2e/next-sandbox/utils/Button.tsx
```diff
@@ -30,6 +31,7 @@ export default function Button(props: Props) {
         opacity,
       }}
       title={`#${props.id}`}
+      disabled={!enabled}
```
marcbouchenoire, line 34: This feels a bit weird from the outside but I can see the readability benefits of doing `enabled={canSet}`.

marcbouchenoire (approved):

PR #1554, opened 2024-04-04
Fix disabled default composer

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks.io/issues/2046.


CTNicholas (approved):
Nice! Thanks for the fix.

PR #1555, opened 2024-04-04
Add filter query to get threads

FlowFlorent:
### Description

Adds query to filter threads by metadata for `GET /rooms/{roomId}/threads`
Ex:
`/rooms/foo/threads?query=metadata['status']:'open' AND metadata['resolved']:false AND metadata['priority']:3`
which can be added using the `options.query` in the node client.

```ts
getThreads(
    { 
      roomId: "foo",
      query: "metadata['status']:'open' AND metadata['resolved']:false AND metadata['priority']:3"
    }
)
```

- [x] REST API doc
- [x] Node client method 
- [x] Node client doc 

In a follow up PR, we will handle typed metadata
```ts
getThreads<ThreadMetadata>(
    { 
      roomId: "foo",
      query: {
          metadata: {
              resolved: true,
              status: "open",
              priority: 3,
              org: {
                 startsWith: "liveblocks"
              }
          }
      }
    }
)
```

### DIFF
--- a/docs/references/v2.openapi.json
+++ b/docs/references/v2.openapi.json
@@ -1369,7 +1369,7 @@
             },
             "in": "query",
             "name": "query",
-            "description": "Query to filter threads. Ex: `metadata[\"status\"]:\"open\" AND metadata[\"resolved\"]:true AND metadata[\"priority\"]:1`"
+            "description": "Query to filter threads. For example, `metadata[\"status\"] AND metadata[\"resolved\"]:true`. Learn more in our guide on [filtering threads with the REST API](/docs/guides/how-to-filter-threads-with-the-rest-api)."
           }
         ],
         "operationId": "get-rooms-roomId-threads",
--- a/docs/references/v2.openapi.json
+++ b/docs/references/v2.openapi.json
@@ -1362,6 +1362,14 @@
               "type": "string"
             },
             "description": "ID of the room"
+          },
+          {
+            "schema": {
+              "type": "string"
+            },
+            "in": "query",
+            "name": "query",
+            "description": "Query to filter threads. Ex: `metadata[\"status\"]:\"open\" AND metadata[\"resolved\"]:true AND metadata[\"priority\"]:1`"
           }
         ],
         "operationId": "get-rooms-roomId-threads",
--- a/guides/pages/how-to-filter-threads-with-the-rest-api.mdx
+++ b/guides/pages/how-to-filter-threads-with-the-rest-api.mdx
@@ -1,0 +1,67 @@
+---
+meta:
+  title: "How to filter threads with the REST API"
+  description:
+    "Learn how to filter for certain threads using their metadata with our REST
+    API"
+---
+
+When using Comments and retrieving threads with our REST API, it’s possible to
+filter for specific threads using
+[custom metadata](/docs/products/comments/metadata) and our custom query
+language. This enables the
+[Get Threads REST API](/docs/api-reference/rest-api-endpoints#get-rooms-roomId-threads)
+to have filtering that works the same as with
+[`useThreads`](/docs/api-reference/liveblocks-react#useThreads) and
+[`liveblocks.getThreads`](/docs/api-reference/liveblocks-node#get-rooms-roomId-threads).
+
+## Query language
+
+You can filter threads by their metadata, allowing you to select for certain
+properties, values, or even for string prefixes. Filters can be combined
+together using `AND` logic.
+
+### Cheatsheet
+
+```js
+// Threads with the `{ status }` metadata
+metadata['status']
+
+// Threads with both `{ status }` and `{ priority }` metadata
+metadata['status'] AND metadata['priority']
+
+// Threads with { status: 'open' } string metadata
+metadata['status']:'open'
+
+// Threads with { priority: 3 } number metadata
+metadata['priority']:3
+
+// Threads with { resolve: true } boolean metadata
+metadata['resolved']:false
+
+// Threads with `{ org }` string metadata that starts with "liveblocks"
+metadata['org']^'liveblocks'
+
+// A mixture of filters
+metadata['status'] AND metadata['status']:'open' AND metadata['org']^'liveblocks'
+```
+
+<Banner>
+
+The `AND` is optional and can actually be omitted, but we’re using it here for
+clarity.
+
+</Banner>
+
+### How to use
+
+To use the query language with the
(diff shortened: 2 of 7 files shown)


FlowFlorent:
Multiple options are possible for the function signature, which do you prefer?

To keep in mind, we will add a `metadata` key at some point to do fully typed queries, in a similar fashion to: https://github.com/liveblocks/liveblocks/pull/1549

```ts
// Option 1, with query in an option object
getThreads(
    { roomId: "foo" }, 
    {
      query: "metadata['status']:'open' AND metadata['resolved']:false AND metadata['priority']:3"
    }
)
// Option 2: query in params directly
getThreads(
    {
      roomId: "foo",
      query: "metadata['status']:'open' AND metadata['resolved']:false AND metadata['priority']:3"
    }
)
```

docs/references/v2.openapi.json
```diff
@@ -1362,6 +1362,14 @@
...
             },
             "description": "ID of the room"
+          },
+          {
+            "schema": {
+              "type": "string"
+            },
+            "in": "query",
+            "name": "query",
+            "description": "Query to filter threads. Ex: `metadata[\"status\"]:\"open\" AND metadata[\"resolved\"]:true AND metadata[\"priority\"]:1`"
```
nvie, line 1372: Should we include a link to some documentation where people can read more about the query language syntax and the possible operators?

nvie (approved):
💯

docs/references/v2.openapi.json
FlowFlorent: That's a very good point, I am wondering where that doc should live? For now I could just put a link to the node doc. But we do need a dedicated page 🤔

CTNicholas: If there's no place for some information, put it in a [guide](https://github.com/liveblocks/liveblocks/tree/main/guides/pages) and link to it! Going to write a quick one now.

CTNicholas (commented):
I've written up a quick guide, please review it and check it's accurate! Also I believe the `@liveblocks/node` still needs to updated with the final solution we come up with.

guides/pages/how-to-filter-threads-with-the-rest-api.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+You can filter threads by their metadata, allowing you to select for certain
+properties, values, or even for string prefixes. Filters can be combined
+together using `AND` logic.
+
+### Cheatsheet
+
+```js
+// Threads with the `{ status }` metadata
+metadata['status']
```
nvie, line 28: This is not valid syntax. "Existence" is not a thing, you always need to use an operator.

guides/pages/how-to-filter-threads-with-the-rest-api.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+// Threads with both `{ status }` and `{ priority }` metadata
+metadata['status'] AND metadata['priority']
```
nvie, line 31: Same here. A valid example would be:

```suggestion
// Threads with both `{ status }` and `{ priority }` metadata
metadata['status']:'open' AND metadata['priority']>=3
```

guides/pages/how-to-filter-threads-with-the-rest-api.mdx
```diff
@@ -0,0 +1,67 @@
...
+metadata['status']:'open'
+
+// Threads with { priority: 3 } number metadata
+metadata['priority']:3
+
+// Threads with { resolve: true } boolean metadata
+metadata['resolved']:false
+
+// Threads with `{ org }` string metadata that starts with "liveblocks"
+metadata['org']^'liveblocks'
```
nvie, line 43: As a good practice, I would always include a separator char with the prefix match, so let's suggest that:

```suggestion
// Threads with `{ org }` string metadata that starts with "liveblocks:"
metadata['org']^'liveblocks:'
```

guides/pages/how-to-filter-threads-with-the-rest-api.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+// A mixture of filters
+metadata['status'] AND metadata['status']:'open' AND metadata['org']^'liveblocks'
```
nvie, line 46:
```suggestion
-metadata['status']:'closed' AND metadata['priority']>=3 AND metadata['org']^'liveblocks:'
```

guides/pages/how-to-filter-threads-with-the-rest-api.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+```js
+// Threads with the `{ status }` metadata
+metadata['status']
+
+// Threads with both `{ status }` and `{ priority }` metadata
+metadata['status'] AND metadata['priority']
+
+// Threads with { status: 'open' } string metadata
+metadata['status']:'open'
```
nvie, line 34: Maybe also show negative matches?

```suggestion
// Threads with { status: 'open' } string metadata
metadata['status']:'open'

// Negative matches
-metadata['status']:'closed'
```

nvie (approved):
Nice work, @CTNicholas! Left a couple suggestions for the guide, but this lgtm!

guides/pages/how-to-filter-threads-with-the-rest-api.mdx
FlowFlorent: For now we do not support `>=` only `:` (equal operator)

nvie: Wait, `metadata` is [declared as mixed](https://github.com/liveblocks/liveblocks-backend/pull/331/files#diff-a616200d88dd4f8e1a8824a81d048bac08d0af283f1dde72d50bd4f0bd872e69R8), right? And mixed [supports](https://github.com/liveblocks/liveblocks-cloudflare/blob/36137bd6d838f64a62d00437b0877e9aa443e6dc/packages/liveblocks-query-parser/README.md?plain=1#L45) `>=` operators when used with numeric values. The language would definitely accept that query. Are you saying the parser will, but then at query time this would be an error?

CTNicholas:
Thanks for the info @nvie, and @FlowFlorent! I've made changes. Do we have an internal document listing all query language info we can use here?

guides/pages/how-to-filter-threads-with-the-rest-api.mdx
FlowFlorent: Oops had missed this comment: the parser accepts the value but I have validation code after which rejects the operator in the backend

PR #1560, opened 2024-04-08
Fix Vite build issue in `@liveblocks/node`, too

### LINKED ISSUE #1557: process is undefined error on Vite project
**Describe the bug**

My project is initialized with Vite and currently using version 1.10.0 for these 4 packages:
"@liveblocks/client": "^1.10.4",
    "@liveblocks/node": "^1.10.4",
    "@liveblocks/react": "^1.10.4",
    "@liveblocks/yjs": "^1.10.4",

But then I download the react-comments package and its latest version is 1.11.0. Then I met a conflict in versions so I upgraded all to 1.11.0 but since then I keep getting the **process is undefined** in the utils.ts file, which is the getBaseUrl function.

I have tried to go back to either 1.10.4 and 1.10.0 but none of them worked. Please help, I'm getting lost on how to solve this though I have look through a similar resolves issue which is solved in 1.10.4 but when I check the function getBaseUrl, it still using the process.env

nvie:
When we fixed this regression in #1531, I forgot to fix this for `@liveblocks/node` too.

Fixes #1557 and https://github.com/liveblocks/liveblocks/issues/1523#issuecomment-2041442328.

### DIFF
--- a/packages/liveblocks-core/src/connection.ts
+++ b/packages/liveblocks-core/src/connection.ts
@@ -306,9 +306,13 @@ function logPrematureErrorOrCloseEvent(e: IWebSocketEvent | Error) {
 }
 
 function logCloseEvent(event: IWebSocketCloseEvent) {
+  const details = [`code: ${event.code}`];
+  if (event.reason) {
+    details.push(`reason: ${event.reason}`);
+  }
   return (ctx: Readonly<Context>) => {
     console.warn(
-      `Connection to Liveblocks websocket server closed (code: ${event.code}). Retrying in ${ctx.backoffDelay}ms.`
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/connection.ts
```diff
@@ -306,9 +306,13 @@ function logPrematureErrorOrCloseEvent(e: IWebSocketEvent | Error) {
 }
 
 function logCloseEvent(event: IWebSocketCloseEvent) {
+  const details = [`code: ${event.code}`];
```
jrowny, line 309: did you mean to have this change as part of this PR?

nvie: I wanted to include this small fix in 1.11.2 too, yes. Should have made that more obvious.

jrowny (approved):

PR #1562, opened 2024-04-09
Fix issue in `link-locally.sh` script

nvie:
This now uses `"workspace:*"` instead of `"*"` to reference local workspace repos only. I've seen other projects use this syntax, and while it does not seem to be documented anywhere officially, it looks like this is working.


marcbouchenoire (approved):

nvie:
Nevermind, this doesn't work after all.

Screen Shot 2024-04-09 at 10 14 29@2x

PR #1568, opened 2024-04-12
Enable typed queries for threads

FlowFlorent:
### Description

**@liveblocks/react**
- `useThreads` now handles typed queries

**@liveblocks/node**
- `getThreads` now handles typed queries

### DIFF
--- a/packages/liveblocks-core/src/lib/objectToQuery.ts
+++ b/packages/liveblocks-core/src/lib/objectToQuery.ts
@@ -1,0 +1,208 @@
+/**
+ * Converts an object to a query string
+ * Example:
+ * ```ts
+ * const query = objectToQuery({
+      resolved: true,
+      metadata: {
+        status: "open",
+        priority: 3,
+        org: {
+          startsWith: "liveblocks:",
+        },
+      },
+});
+
+console.log(query);
+// resolved:true AND metadata["status"]:open AND metadata["priority"]:3 AND metadata["org"]^"liveblocks:"
+
+ * ```
+ *
+ *
+ */
+
+type FilterValue = string | number | boolean | { startsWith: string };
+
+type Filter = {
+  key: string;
+  operator: ":" | "^";
+  value: string | number | boolean;
+};
+
+/**
+ * Converts an object to a query string
+ * @example
+ * ```ts
+ * const query = objectToQuery({
+ *  metadata: {
+ *    status: "open",
+ *    priority: 3,
+ *    org: {
+ *      startsWith: "liveblocks:",
+ *    },
+ *  },
+ * });
+ * console.log(query);
+ * // metadata["status"]:"open" AND metadata["priority"]:3 AND metadata["org"]^"liveblocks:"
+ * ```
+ * @example
+ * ```ts
+ * const query = objectToQuery({
+ *  resolved: true,
+ *  roomId: {
+ *   startsWith: "engineering:",
+ *  },
+ * });
+ * console.log(query);
+ * // resolved:true AND roomId^"engineering:"
+ * ```
+ *
+ */
+export function objectToQuery(obj: {
+  [key: string]: FilterValue | { [key: string]: FilterValue };
+}): string {
+  let filterList: Filter[] = [];
+  const entries = Object.entries(obj);
+
+  /**
+   * Case 1: Simple key value pairs
+   * @example
+   * ```ts
+   * { key: "foo" }
+   * ```
+   */
+  const keyValuePairs = entries.filter(([_, value]) =>
+    isSimpleValue(value)
+  ) as [string, string | number | boolean][];
+
+  filterList = getFiltersFromKeyValuePairs(keyValuePairs);
+
+  /**
+   * Case 2: Key value pairs with operator
+   * @example
+   * ```ts
+   * { key: { startsWith: "fo" } }
+   * ```
+   */
+  const keyValuePairsWithOperator = entries.filter(([_, value]) =>
+    isValueWithOperator(value)
+  ) as [string, Record<"startsWith", string>][];
+
+  filterList = filterList.concat(
+    getFiltersFromKeyValuePairsWithOperator(keyValuePairsWithOperator)
+  );
+
+  /**
+   * Case 3: Indexed key value pairs
+   * @example
+   * ```ts
+   * { key: { nestedKey: "foo" } }
+   * ```
+   * @example
+   * ```ts
+   * { key: { nestedKey: { startsWith: "fo" } } }
+   * ```
+   */
+  const indexedKeys = entries.filter(([_, value]) => {
+    return typeof value === "object" && !("startsWith" in value);
+  });
+
+  indexedKeys.forEach(([key, value]) => {
+    const nestedEntries = Object.entries(value);
+    const nestedKeyValuePairs = nestedEntries
+      .filter(([_, value]) => isSimpleValue(value))
+      .map(([nestedKey, nestedValue]) => [
+        formatFilterKey(key, nestedKey),
+        nestedValue,
+      ]) as [string, string | number | boolean][];
+
+    filterList = filterList.concat(
+      getFiltersFromKeyValuePairs(nestedKeyValuePairs)
+    );
+
+    const nestedKeyValuePairsWithOperator = nestedEntries
+      .filter(([_, value]) => isValueWithOperator(value))
+      .map(([nestedKey, nestedValue]) => [
+        formatFilterKey(key, nestedKey),
+        nestedValue,
+      ]) as [string, { startsWith: string }][];
+
+    filterList = filterList.concat(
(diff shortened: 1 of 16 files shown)


packages/liveblocks-node/src/client.ts
```diff
@@ -896,12 +898,25 @@ export class Liveblocks {
...
+        };
   }): Promise<{ data: ThreadData[] }> {
     const { roomId } = params;
 
+    let query: string | undefined;
+
+    if (typeof params.query === "string") {
+      query = params.query;
+    } else if (typeof params.query === "object") {
+      // @ts-expect-error: FIXME typing
```
FlowFlorent, line 914: struggling to fix this TS error 🤔

```
Argument of type '{ metadata?: Partial<QueryMetadata<TThreadMetadata>> | undefined; }' is not assignable to parameter of type '{ [key: string]: FilterValue | { [key: string]: FilterValue; }; }'.
  Property 'metadata' is incompatible with index signature.
    Type 'Partial<QueryMetadata<TThreadMetadata>>' is not assignable to type 'FilterValue | { [key: string]: FilterValue; }'.
      Type 'Partial<QueryMetadata<TThreadMetadata>>' is not assignable to type '{ [key: string]: FilterValue; }'.
        Type 'QueryMetadata<TThreadMetadata>[P] | undefined' is not assignable to type 'FilterValue'.
          Type 'undefined' is not assignable to type 'FilterValue'.ts(2345)```

File: packages/liveblocks-core/src/room.ts [hunk 2]
```diff
@@ -1114,19 +1116,23 @@ function createCommentsApi(
   async function getThreads<TThreadMetadata extends BaseMetadata = never>(
     options?: GetThreadsOptions<TThreadMetadata>
   ) {
+    let query: string | undefined;
+
+    if (options?.query) {
+      // @ts-expect-error: FIXME typing
```
FlowFlorent, line 1122: same issue as: https://github.com/liveblocks/liveblocks/pull/1568/files#r1566448741

packages/liveblocks-core/src/lib/objectToQuery.ts
```diff
@@ -0,0 +1,222 @@
...
+
+console.log(query);
+// resolved:true AND metadata["status"]:open AND metadata["priority"]:3 AND metadata["org"]^"liveblocks:"
+
+ * ```
+ *
+ *
+ */
+
+type FilterValue = string | number | boolean | { startsWith: string };
```
nvie, line 24: Should we not also support `{ lt: number }`, `{ gte: number }`, etc?

packages/liveblocks-core/src/lib/objectToQuery.ts
```diff
@@ -0,0 +1,222 @@
...
+ *   startsWith: "engineering:",
+ *  },
+ * });
+ * console.log(query);
+ * // resolved:true AND roomId^"engineering:"
+ * ```
+ *
+ */
+export function objectToQuery(obj: {
+  [key: string]: FilterValue | { [key: string]: FilterValue };
```
nvie, line 76: You'll need to account for explicit `undefined`s here too at the TypeScript level. We typically would consider these values to be equal:

```ts
{
  foo: "bar"
}
```

```ts
{
  foo: "bar",
  qux: undefined,
}
```

…but TypeScript knows they're not equal. I think in a function like this, it's fine to accept values containing explicit-`undefined`s and simply ignoring/skip over those keys. Accepting those values makes it much more pragmatic to use this API, as you can use it to conditionally include fields or not, for example:

```ts
{
  foo: myCondition ? "bar" : undefined
}
```

packages/liveblocks-core/src/lib/objectToQuery.ts
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Case 1: Simple key value pairs (equal operator)
+   * @example
+   * ```ts
+   * { key: "foo" }
+   * ```
+   */
+  const keyValuePairs = entries.filter(([_, value]) =>
+    isSimpleValue(value)
+  ) as [string, string | number | boolean][];
```
nvie, line 90: You can avoid doing this type-cast here…

```suggestion
  );
```

packages/liveblocks-core/src/lib/objectToQuery.ts
```diff
@@ -0,0 +1,222 @@
...
+        operator: "^",
+        value: value.startsWith,
+      });
+    }
+  });
+
+  return filters;
+};
+
+const isSimpleValue = (value: unknown) => {
```
nvie, line 188: …if you return a type predicate here:

```suggestion
const isSimpleValue = (value: unknown): value is string | number | boolean => {
```

nvie: In TypeScript 5.5 (yet unreleased), this will get [inferred automatically](https://github.com/microsoft/TypeScript/pull/57465)!

packages/liveblocks-core/src/lib/objectToQuery.ts
```diff
@@ -0,0 +1,222 @@
...
+  return false;
+};
+
+const formatFilter = (key: string, operator: ":" | "^", value: string) => {
+  return `${key}${operator}${value}`;
+};
+
+const formatFilterKey = (key: string, nestedKey?: string) => {
+  if (nestedKey) {
+    return `${key}["${nestedKey}"]`;
```
nvie, line 212: Be careful with trusting user inputted values (which you should assume these inputs are) when formatting. You should ensure you're properly escaping quotation marks to avoid "insertion attacks" here, or validate upfront that certain characters won't exist in the key/nestedKey parts.

Probably best to validate that keys don't include specific chars, and to allow any char in the values, but always escape them.

For example, currently it would be possible to do this at the key level:

```ts
{
  index: { "evil\"]:3 index[\"qux": "bar" },
}
```

Which will produce the query: `index["evil"]:3 index["qux"]:"bar"`.

Similarly, but at the value level:

```ts
{
  "org": "my-org\" OR org:\"another-org"
}
```

…which would produce:

`org:"my-org" OR org:"another-org"`

packages/liveblocks-node/src/client.ts
nvie: I think this will go away once you allow explicit-`undefined`s, see my earlier comment above.

nvie (commented):
I left a few comments. Overall, I think the safest way to write a function that takes a more complex object and produces a valid query string is to use a recursive "prettifier", which takes an AST (in this case the "typed query", which sort of is a simplified AST) and produces a string value, i.e. the reverse operation of a parse.

You can see an example of it here:
https://github.com/liveblocks/liveblocks-cloudflare/blob/main/packages/liveblocks-query-parser/src/prettify/index.ts

With a function like that, it's impossible to generate syntactically invalid outputs, and it will take care of all the proper escaping. (Notice how this function calls itself recursively on sub parts of the AST, until it reaches the "nodes", which it basically just uses `JSON.stringify()` for.)

nvie:
@FlowFlorent Let me know once you're done with this PR. I'm happy to play devils advocate and try to come up with unit tests that won't pass.

packages/liveblocks-node/src/client.ts
FlowFlorent: That was it, thank you!

packages/liveblocks-core/src/lib/objectToQuery.ts
FlowFlorent: Oh nice, about time!

packages/liveblocks-core/src/lib/objectToQuery.ts
FlowFlorent: Would escaping `"` quotes be enough? 

We allow anything but `-` and `+` as the starting characters for metadata keys as seen [here](https://github.com/liveblocks/liveblocks-backend/blob/c9e58d74ef0de67bc785af76f08b355ce9541988/rest-api/src/libs/decoders/index.ts#L37).

FlowFlorent:
@nvie Thanks for all your comments, I've adjusted the PR. 

I am still a bit uncertain about the escaping part though 🤔

packages/liveblocks-core/src/lib/objectToQuery.ts
nvie: Yeah, it would. The easiest way to do that is to use `JSON.stringify()`, which I'm [also using here](https://github.com/liveblocks/liveblocks-cloudflare/blob/main/packages/liveblocks-query-parser/src/prettify/index.ts#L13C38-L13C65).

FlowFlorent: @nvie Thanks for all your feedback, I've adjusted the PR, should be good to go now!

nvie (approved):
Nice work, @FlowFlorent! 🙌

PR #1570, opened 2024-04-15
Fix pointer events in whiteboard-advanced example

nvie:
Small CSS fix that will allow you to use the part of the canvas to the left and right of the toolbar in our whiteboard advanced example. Before this change, those areas were blocked by an invisible box, see video:

https://github.com/liveblocks/liveblocks/assets/83844/bf81aaf1-3c85-41f9-9c52-2ec7c29191de


marcbouchenoire (approved):
I'm surprised it's been like that since the start, it's nice to have it fixed!

PR #1571, opened 2024-04-15
Fix: remove unnecessary check of access token permissions

ofoucherot:
### Fixing a bug

Remove the check of access token permissions before storing it to the local cache.
This validation isn't necessary anymore since we don't use the access token permissions for Notifications.


marcbouchenoire (approved):
Makes sense, as discussed on Friday.

PR #1573, opened 2024-04-15
Fix `nextjs-comments-notifications` example

marcbouchenoire:
The `nextjs-comments-notifications` example is currently broken when the `exampleId` query param isn't set.

The error happens when generating mention suggestions:
- We do `URLSearchParams.get("exampleId")` in `resolveMentionSuggestions` then we pass it to the endpoint responsible for computing the mention suggestions
- But `URLSearchParams.get("exampleId")` returns `null` if it isn't set
- And unlike `undefined`, `null` gets stringified so the endpoint is called with `&exampleId=null`
- Which means that the endpoint receives a literal `"null"` value so it uses it and returns user IDs as `user-0-null` instead of `user-0`

(I also fixed the Next.js errors related to RSC)


CTNicholas (approved):
Nice catch!

PR #1574, opened 2024-04-15
add trigger inboxNotification method to node client

ofoucherot:

### DIFF
--- a/packages/liveblocks-node/src/__tests__/client.test.ts
+++ b/packages/liveblocks-node/src/__tests__/client.test.ts
@@ -969,4 +969,47 @@ describe("client", () => {
       }
     }
   });
+
+  test("should return the created inbox notification when createInboxNotification receives a successful response", async () => {
+    const userId = "user1";
+    const inboxNotificationId = "notification1";
+
+    const notification = {
+      id: inboxNotificationId,
+      kind: "fileUploaded",
+      notifiedAt: new Date().toISOString(),
+      readAt: null,
+      subjectId: "subject1",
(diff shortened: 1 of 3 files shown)


packages/liveblocks-node/src/__tests__/client.test.ts
```diff
@@ -969,4 +969,25 @@ describe("client", () => {
       }
     }
   });
+
+  test("should return the created inbox notification when createInboxNotification receives a successful response", async () => {
```
marcbouchenoire, line 973:
```suggestion
  test("should return the created inbox notification when triggerInboxNotification receives a successful response", async () => {
```

packages/liveblocks-node/src/client.ts
```diff
@@ -1346,6 +1346,21 @@ export class Liveblocks {
         : undefined,
     };
   }
+
+  public async triggerInboxNotification(params: {
+    userId: string;
+    kind: `$${string}`;
+    roomId?: string;
+    subjectId: string;
+    activityData: BaseMetadata;
```
marcbouchenoire, line 1355: They're the same types but for clarity we should use `ActivityData` from `InboxNotificationData.ts` here.

marcbouchenoire (changes requested):

packages/liveblocks-node/src/client.ts
ofoucherot: Fixed!

marcbouchenoire (approved):

PR #1593, opened 2024-04-26
Get rooms query doc

FlowFlorent:
### Description

Adds REST API and @liveblocks/node `query` documentation for `getRooms`


docs/pages/api-reference/liveblocks-node.mdx
```diff
@@ -320,16 +320,31 @@ const {
...
-  metadata: {
-    myRoomType: "whiteboard",
-  },
+  // Optional, use advanced filtering
+  query: {
+    // Optional, filter for rooms with custom metadata in `metadata`
+    metadata: {
+      myRoomType: "whiteboard",
+      shared: true,
+      time: 0,
```
FlowFlorent, line 329:
```suggestion
```
Room metadata only supports `string` as a value so we can only filter with `:` and `string` for now.

PR #1596, opened 2024-04-30
Fix thread notification tests

FlowFlorent:
### Description

Fix tests failing by using thread notifications


packages/liveblocks-core/src/store.ts
```diff
@@ -393,8 +393,9 @@ export function applyOptimisticUpdates<TThreadMetadata extends BaseMetadata>(
 
         const inboxNotification = Object.values(result.inboxNotifications).find(
           (notification) =>
+            // CHECK ALL
```
FlowFlorent, line 396:
```suggestion
```

PR #1601, opened 2024-05-02
Prefactorings to get closer to implementing module augmentation

nvie:
This moves the `getThreadsUpdates()` helper out of the context factory, so it will be easier to reuse outside of it.

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -413,22 +413,22 @@ export function createRoomContext<
 
     React.useEffect(() => {
       // Retrieve threads that have been updated/deleted since the last time the room requested threads updates
-      void getThreadsUpdates(room.id);
-    }, [room.id]);
+      void getThreadsUpdates(client, room.id);
+    }, [client, room.id]);
 
     /**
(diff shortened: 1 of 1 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -2356,6 +2310,52 @@ export function createRoomContext<
   });
 }
 
+const DEFAULT_DEDUPING_INTERVAL = 2000; // 2 seconds
+
+const lastRequestedAtByRoom = new Map<string, Date>(); // A map of room ids to the timestamp when the last request for threads updates was made
+const requestStatusByRoom = new Map<string, boolean>(); // A map of room ids to a boolean indicating whether a request to retrieve threads updates is in progress
```
nvie, line 2316: There are no code changes in this entire block, I just moved it out of the factory body and made `client` a parameter.

The question here now is: is this a safe operation? Especially: will this be safe to do regarding the `lastRequestedAtByRoom` and `requestStatusByRoom` caches? Since those are no longer scoped to the factory instance, but now shared globally.

Will it be important to create a new cache for every `client` instance?

nvie:
Closing, will do a take 2 for this one.

PR #1609, opened 2024-05-13
Also run tests for `@liveblocks/yjs` on CI

nvie:
Somehow we forgot to add this to the CI test runs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,7 @@ jobs:
             "@liveblocks/redux",
             "@liveblocks/zustand",
             "@liveblocks/node",
+            "@liveblocks/yjs",
           ]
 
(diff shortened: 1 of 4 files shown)


nvie:
@jrowny @nimeshnayaju This morning, I noticed we have never been running the `@liveblocks/yjs` unit tests on CI. This PR enables that, but for some reason the client isn't sending the `ROOM_STATE` message (which is expected when using the v7 protocol), or maybe there is something wrong with the mock setup here, because the tests are [failing](https://github.com/liveblocks/liveblocks/actions/runs/9060570982/job/24890537991?pr=1609#step:7:79). (I didn't dig deeper here yet, just opened this PR as a reminder to get this fixed.)

nvie:
Will open a new one for this.

PR #1611, opened 2024-05-14
Prefactorings for module augmentation

nvie:
Prefactorings for the upcoming module augmentation adaption work, which will be noisy.

All changes in this PR **don't change any behavior**, and should be **good to be merged** into `main` right now.

PR #1613, opened 2024-05-14
More prefactorings for module augmentation

nvie:
Follow-up with more prefactorings after #1611. Again, this makes **no functional changes**.


marcbouchenoire (approved):
LGTM 👍

I'm still on the fence about the `use*OrNull` naming convention and I think I would prefer something like `useOptionalClient` but it's purely for stylistic reasons, maybe explicitness should stay the deciding factor.

nvie:
> I'm still on the fence about the `use*OrNull` naming convention and I think I would prefer something like `useOptionalClient` but it's purely for stylistic reasons, maybe explicitness should stay the deciding factor.

I'm still hoping that none of those hooks should have to get exposed publicly, these `OrNull` versions are all private APIs.

PR #1616, opened 2024-05-16
Module augmentation prefactorings

nvie:
This preps the Liveblocks context module by removing all remaining closures from the bundle factory, move all closed-over hooks to the top level and parameterize them all with a client instance. This will pave the way for the addition of the new top-level hooks.

> [!NOTE]
> 💡 Best reviewed commit-by-commit to see each refactoring step!

This PR should be possible to be **merged as-is**, because **no behavioral changes** are expected here.

@marcbouchenoire I'd appreciate if you could manually double-check this work, by running a couple smoke tests manually here. I have run all the unit tests, but this particular part of the code base is not 100% covered.

I'll make a similar changes to the Room context factory tomorrow — but there are a _lot_ more of those there, so I first want to get you on board with this subset.


GuillaumeSalles (approved):
LGTM, no behavioral changes. But the order of notifications is changing in a unit test, but I couldn't figure out why just by looking at the code. 🤔

nvie:
Oh, this is a unit test that I have seen occasionally fail on `main`, even before I started all the refactorings. I think this test is just not deterministic, but I doubt it’s a real issue. I’ll look into it.

marcbouchenoire:
> Do you have an idea what can cause the alternative ordering of those inbox notifications? Are we maybe missing an explicit sort on the backend side?

@nvie Maybe the non-deterministic nature is due to timing? In the failing test I see the dummy notifications' timestamps slightly different, which makes sense since they're created one after another (2024-05-16T22:12:53.561Z vs 2024-05-16T22:12:53.562Z), maybe there are cases/cycles where the timestamps align and the notifications get sorted differently because of that?

packages/liveblocks-react/src/room.tsx
```diff
@@ -2357,14 +2359,19 @@ function makeRoomContextBundle<
...
+
 /**
  * @private
  *
  * This is an internal API, use `createRoomContext` instead.
  */
 export function useRoomContextBundleOrNull() {
   const client = useClientOrNull();
-  return client === null ? null : getOrCreateRoomContextBundle(client);
+  const room = useRoomOrNull();
```
nvie, line 2373: The key line for this fix. `useRoomContextBundleOrNull()` isn't really dependent on having a room in order to create the bundle, but… the way the `useSharedBundle()` function works is that it only expects this function to return a result if a RoomProvider context is active. This was a tricky one! Super glad you discovered this, @marcbouchenoire.

Next week, I'll ensure that the E2E page/test will also capture this edge case before I continue to refactor more.

nvie:
> Maybe the non-deterministic nature is due to timing?

@marcbouchenoire Yeah, maybe. I just added 29bbb87290fd76943c99ceb0ebda6f7ffd6d0d7f which sorts the arrays before comparing them. I really wish there was a native matcher for this in Jest.

marcbouchenoire (approved):
🙌

Issue #1623, opened 2024-05-22
Nested Rooms in React causing error


amcdnl:
## Bug Report

In my application, I have a 'home' room and then session rooms. Currently I have the application structured like:

```
<>
   <RoomProvider id="home">
        ... other stuff ...
       <Suspense fallback={<Loader />}>
         <RoomProvider={sessionId} initialStorage={{ input: new LiveObject({ value: '' }) }}>
             <Suspense fallback={<Loader />}>
               <MyComponent />
             </Suspense>
         </RoomProvider>
       </Suspense>
   </RoomProvider>
</>
```

and in `MyComponent` I would like to set the `input.value` as the user types like:

```
  const updateInput = useMutation(
    ({ storage }, value) => {
        storage.get("input").set("value", value);
    },
    []
  );
```

however, whenever I run the follow I get an error BEFORE `storage.get` is even hit when invoking `updateInput` that reads:

```
This mutation cannot be used until connected to the Liveblocks room
```

however, if I put the following code above the invocation of the `updateInput` function:

```
    const room = useRoom();
    const status = room.getStatus();
    if (status === 'connected') {
       updateInput(inputVal)
    }
```

the code still fails. I'm wondering if it has something to do with the nested rooms though this should totally work in react context.

## Additional Info

```
    "@liveblocks/client": "^1.12.0",
    "@liveblocks/react": "^1.12.0",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
```

nvie:
Hi @amcdnl – thanks for opening a support ticket. The nesting of RoomProviders should not be an issue indeed—it's supported since [v1.5.0](https://github.com/liveblocks/liveblocks/releases/tag/v1.5.0).

The `useMutation` callback can only be called after the initial storage document has been loaded by your application. Waiting for `status === 'connected'` is not enough for this. That is only the WebSocket status. After the WebSocket is connected, the Storage loading phase begins. You can listen for the `storageStatus` status instead. If it's `synchronized` or `synchronizing`, it means it has been loaded. You can find the [relevant docs](https://liveblocks.io/docs/api-reference/liveblocks-client#Room.getStorageStatus) here.

Hope that helps to resolve it for you!

nvie:
I realize that the error message may have put you on the wrong track. I'll get that fixed 👍

amcdnl:
@nvie - Sadly, it never moves out of `not-connected`. Any clues why this might be?

[...]

PR #1624, opened 2024-05-23
Small tweak in error message

nvie:
Small update to the error message we throw in `useMutation` when called too soon. This should make it clearer that unpacking `storage` can only happen here after storage has been loaded.

See https://github.com/liveblocks/liveblocks/issues/1623#issue-2311341430 which brought this to my attention.


CTNicholas (approved):
Good change!

PR #1625, opened 2024-05-23
Text mention inbox notification

FlowFlorent:
#### Description

Adds a basic text mention inbox notification

Screenshot 2024-05-28 at 3 08 56 PM


packages/liveblocks-react-comments/src/components/InboxNotification.tsx
```diff
@@ -489,6 +499,45 @@ const InboxNotificationThread = forwardRef<
...
+        <Room roomId={inboxNotification.roomId} />
+      )}
+      date={inboxNotification.notifiedAt}
+      unread={unread}
+      overrides={overrides}
+      showActions={showActions}
+      {...props}
+      ref={ref}
+    />
+    // </InboxNotificationLayout>
```
FlowFlorent, line 537:
```suggestion
```

FlowFlorent:
Hello @marcbouchenoire!

A question for you, should we add "content" for this new component?

For now it's not possible to get the text editor content (without a lot of hoops), so I figured just using the `title` property would be enough.

That said, should we allow for a content override?

marcbouchenoire:
Hey @FlowFlorent, hmm it does look odd without a description below, I wonder how other products handle this? (if any have the same requirement as us and can't show the text) 

And regarding overriding, I think for now the mention notification should allow the same customization as the thread one, via the `kinds` prop. For example, by default thread notifications show room names in the title (like the mention one you have), but this can be tweaked:

```
kinds={{
  thread: (props) => {
    return <InboxNotification.Thread {...props} showRoomName={false} />;
  }
}}
```

If they want more granularity like changing just the description/content they can go lower with `InboxNotification.Custom` or fully custom.

PR #1627, opened 2024-05-24
Fix two minor regressions

nvie:
This PR fixes:

- Two places where I introduced a regression in the public API surface
- Upgrades `@types/react` for all comments examples
- Fixes two misconfigurations of TypeScript which future TS versions will nag about when we upgrade

### DIFF
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -136,7 +136,7 @@ function selectorFor_useUnreadInboxNotificationsCountSuspense(
 function getOrCreateContextBundle<
   U extends BaseUserMeta,
   M extends BaseMetadata,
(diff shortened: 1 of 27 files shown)


packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -631,6 +631,6 @@ export function useLiveblocksContextBundle() {
 export function createLiveblocksContext<
   U extends BaseUserMeta = DU,
   M extends BaseMetadata = never, // TODO Change this to DM for 2.0
->(client: Client<U>): LiveblocksContextBundle<U, M> {
+>(client: OpaqueClient): LiveblocksContextBundle<U, M> {
```
nvie, line 634: These are the two regressions, which degrade the DX when working with the factories.

marcbouchenoire (approved):
🧹

PR #1629, opened 2024-05-24
Widen `BaseMetadata` type

nvie:
At the heart of this PR is the following change (it's the first commit):

```diff
 export type BaseMetadata = Record<
   string,
-  string | boolean | number
+  string | boolean | number | undefined
 >;
```

This would better reflect reality, as accessing `metadata.foo` at runtime is always possible, but potentially `undefined`. For a base type, this is a requirement. For _writing_ metadata, we'll also want to allow explicitly-undefined in the base type, for pragmatic reasons.

Since this is a widening, this is a non-breaking change.
(The rest of the PR only contains trivial code moves and name changes.)

Fixes LB-607.


GuillaumeSalles:
IIRC, we tried this at some point and reverted back to not allow undefined. But I don't remember why. Does it ring a bell @marcbouchenoire ? 🤔 

Was it because because writing does not work because of `undefined` serialization? Could it be technically solved if we improve our create/update thread endpoints?

nvie:
I already talked to @marcbouchenoire about this last week, and it seemed indeed that the argument for not including it was the write API. However, I think including it in the write API would _also_ provide better DX. This point is currently moot, because our `PartialNullable<>` wrapper is already doing that.

We're entering subtle TS territory here, but since `undefined` is currently not part of the value type, it means that using optional fields in your custom metadata type should problably be a type error.

1. [Playground example 1](https://www.typescriptlang.org/play/?ts=4.7.4#code/C4TwDgpgBAQghgZwgWQsOATO6oF4oBKEAxgPYBOGAPAFBRQLDkCWAdgOYA0dDTb7UAD5QARqVIAbCHFZCorAK4BbERHI0AfAG4aNUJCgANPFADeUHgDNxAfgBcvFhx0BfXZYWtiwZqVmWqZCgIAA9gCFYMBFhEFDRMbDgNAAolB2QASjM3GgDDFPNrUnTsAAsAOnIZDFIlZKyNKAAGKBsoAHJS5naoB08MCEs2CAwoFwydIA) (surprisingly working, we may be relying on this 🤔)
1. [Playground example 2](https://www.typescriptlang.org/play/?ts=4.7.4#code/C4TwDgpgBAQghgZwgWQsOATO6oF4oBKEAxgPYBOGAPAFBRQLDkCWAdgOYA0dDTb7UAD5QARqVIAbCHFZCorAK4BbERHI0AfAG4aNUJCgANPFADeUHgDNxALl4sOchawwRLbCBh0BfXZefEwMykspZUyFAQAB7AEC4IsIgoaJjYcBoAFEp2yACUZr40YYaZ5takOdgAFgB05DIYpEoZ+RpQAAxQAPxQAORVzL1Qds6u7qyeUN65OkA) (TS error, as expected)
1. [Playground example 3](https://www.typescriptlang.org/play/?ts=4.7.4#code/C4TwDgpgBAQghgZwgWQsOATO6oF4oBKEAxgPYBOGAPAFBRQLDkCWAdgOYA0dDTb7UAD5QARqVIAbCHFZCorAK4BbERHI0AfAG4aNNsDUAzOMWgANKBAAeB1hgSxEKNJmxwoAbx6HxAfgBcvCwcOgC+uoYKrMTAzKSyhlTIljYQdg7wSKjoWOgaABRKAVDIAJSe4TSJZgWlOlVUNfkeUD6kgcjYABYAdOQyGKRK+eUaUAAMUL5QAORdzDNQgVEYEIZsEBhQoXU0QA) (TS error, as expected)

Surprisingly, TS doesn't reject playground example (1) at the moment, which is likely why we were thinking this is working, but honestly we may rely on a narrow TS implementation edge case here, because you can clearly see that examples (2) and (3) _are_ broken (as expected).

Adding `| undefined` to the values of the base type is the solution for all of these cases. (You can try it out in the playgrounds.)

I don't see any negative side effects to adding `undefined` to the values here, but please do let me know if you know any.

marcbouchenoire (approved):

marcbouchenoire:
> Was it because because writing does not work because of undefined serialization? Could it be technically solved if we improve our create/update thread endpoints?

@GuillaumeSalles I think it was mostly because I wanted explicit `undefined` to be disallowed (given that it disappears during serialization so it amounts to doing nothing), but given that those objects are partial, we can't really disallow `undefined` at all, it doesn't make sense.

nvie:
To summarize, there are mainly three reasons to do this:

1. Need. To add to this, allowing `undefined` is needed to support data migrations too, i.e. allow users to over time add a new optional properties to the metadata.
2. Pragmatism. Being able to write `{ a: 123, b: some?.optional }` is much nicer DX than having to write `some?.optional ? { a: 123, b: some.optional } : { a: 123 }`. Fortunately this is not an issue with editing our metadata, because we are already using `PartialNullable` there. But it would be nice to be able to use this construct when creating initial metadata as well.
3. The fact that `undefined` values "disappear" during serialization is consistent by me. Many APIs use this pattern. In other words, if you write `{ a: 123, b: undefined }`, then what comes out is:
   ```
   metadata.a  // 123
   metadata.b  // undefined
   ```
   Which is consistent.

PR #1639, opened 2024-05-29
Release 2.0

nvie:
This PR is the merge target for everything that will land in the upcoming v2.0.0.


docs/pages/get-started/lexical-react.mdx
```diff
@@ -0,0 +1,232 @@
...
+        LiveblocksPlugin,
+        FloatingComposer,
+        ThreadPanel,
+      } from "@liveblocks/react-lexical";
+
+      export default Editor() {
+        const room = useRoom();
+
+        // Wrap your initial config with `liveblocksConfig`
+        const initialConfig = liveblocksConfig({
```
nimeshnayaju, line 146: Should we also highlight the fact that users are expected to wrap their lexical configuration within `liveblocksConfig`? This function ensures `ThreadMarkNode` and `MentionNode` are correctly added to the editor config.

docs/pages/api-reference/liveblocks-react-lexical.mdx
```diff
@@ -0,0 +1,223 @@
...
+## useEditorStatus
+
+Returns the current editor status.
+
+```ts
+import { useEditorStatus } from "@liveblocks/react-lexical";
+
+const status = useEditorStatus();
+```
+
```
nimeshnayaju, line 118: `<FloatingComposer />` does not currently need to be placed inside `LiveblocksPlugin` component. I wonder if we should still encourage this just to be safe.

docs/pages/api-reference/liveblocks-react-lexical.mdx
```diff
@@ -0,0 +1,223 @@
...
+  <LiveblocksPlugin>
+    <FloatingComposer />
+  </LiveblocksPlugin>
+</LexicalComposer>
+```
+
+{/* TODO: Image */}
+
+Submitting a comment will attach an annotation thread at the current selection.
+Should be nested inside [`LiveblocksPlugin`][].
```
nimeshnayaju, line 142: `<FloatingComposer />` does not currently need to be placed inside `LiveblocksPlugin` component. I wonder if we should still encourage this just to be safe.

marcbouchenoire (approved):
🔜

PR #1640, opened 2024-05-29
Liveblocks 2.0 docs: deprecations

CTNicholas:
https://linear.app/liveblocks/issue/LB-596/document-simpler-dx


nvie (approved):
Nice! Glad we're finally removing some of these APIs! 🧹

Should we however maybe keep the documentation for these removed endpoints, so that existing links to them would still point to useful info, e.g. for users still on 0.x or 1.x?

One way we might be able to solve this is like how we do it on our REST API page, where we moved all deprecated APIs to the bottom, under the "Deprecated" section. Links to these sections would still work. And this way, people can still read the upgrade instructions.

Screen Shot 2024-06-03 at 21 53 46@2x

CTNicholas:
> Nice! Glad we're finally removing some of these APIs! 🧹
> 
> Should we however maybe keep the documentation for these removed endpoints, so that existing links to them would still point to useful info, e.g. for users still on 0.x or 1.x?
> 
> One way we might be able to solve this is like how we do it on our REST API page, where we moved all deprecated APIs to the bottom, under the "Deprecated" section. Links to these sections would still work. And this way, people can still read the upgrade instructions.
> 
> Screen Shot 2024-06-03 at 21 53 46@2x

Good idea, I've done this for the more commonly used and most recent deprecations.

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -3842,12 +3773,53 @@ const user = room.getSelf();
...
-    true if they can only read but not mutate it.
-  </PropertiesListItem>
 </PropertiesList>
 
+## Deprecated
+
+### Client.enter [@badge=Deprecated]
+
+<Banner title="Deprecated" type="error">
+  This is no longer supported. We recommend using [`client.enterRoom`][]
```
nvie, line 3783: @CTNicholas Should we maybe say "This API was removed in 2.0" instead of saying it's not supported?

PR #1641, opened 2024-05-29
Add `@liveblocks/react/suspense` subpath export

nvie:
This PR adds the subpath export `@liveblocks/react/suspense`, and exports all the hooks at those top-levels. Think of `@liveblocks/react/suspense` as having exactly the same exports as `@liveblocks/react`, _except_ for the few hooks that have a Suspense version.

This video shows how it will _eventually_ work. (I'm cheating a bit here, because this integration is not yet implemented—that will happen in one of the next PRs.)

https://github.com/liveblocks/liveblocks/assets/83844/037b7b68-d6b8-4a48-a516-a45caca81a96

Fixes LB-557.
Fixes LB-586.
Fixes LB-587.
Fixes LB-588.
Fixes LB-670.
Fixes LB-671.
Fixes LB-675.


packages/liveblocks-react/package.json
```diff
@@ -17,10 +17,22 @@
         "module": "./dist/index.mjs",
         "default": "./dist/index.js"
       }
+    },
+    "./suspense": {
```
GuillaumeSalles, line 21: I remember this failing on some version of typescript, or maybe the ecosystem was not fully ready. Do you know if it's still the case?

GuillaumeSalles (approved):

packages/liveblocks-react/package.json
nvie: I think [this is the issue](https://github.com/liveblocks/liveblocks/pull/1641/commits/90eff09db6ecc011e6c41dbbfc8fb78670c6f6c4) you may remember. It was indeed a problem, but I found a solution for it yesterday! See the commit. Full context is in the link from the commit message.

PR #1643, opened 2024-05-29
Text mention notification doc

FlowFlorent:
#### @liveblocks/node
Add notification types

#### Docs
Add notification types for node client
Add notification types for webhook

### DIFF
--- a/docs/pages/platform/webhooks.mdx
+++ b/docs/pages/platform/webhooks.mdx
@@ -851,6 +851,10 @@ const threadMetadataUpdatedEvent = {
 
 #### NotificationEvent
 
+There are three notification kinds: "thread", "textMention", and custom.
+
+##### Thread notification
+
 An event is triggered 30 minutes after a user has been mentioned or replied to
 in a thread, and has not seen the thread. It will also be triggered if the user
 has subscribed to the thread and has not seen the thread. The event won’t be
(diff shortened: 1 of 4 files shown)


docs/pages/platform/webhooks.mdx
```diff
@@ -874,18 +885,93 @@ type NotificationEvent = {
...
+  },
+};
+```
+
+##### Custom notification
+
+An event is triggered 30 minutes after the user has been notified of a custom
+event and has not seen the notification. All custom notification `kinds` are
+prefixed with `$` and are manually by you on the server. Learn more about
+[triggering custom notifications](/docs/api-reference/rest-api-endpoints#post-inbox-notifications-trigger).
```
CTNicholas, line 946: I've changed the link from this:
```md
Learn more about triggering custom notifications [here](/docs/...).
```
To this:
```md
Learn more about [triggering custom notifications](/docs/...).
```

**Interesting fact**
If you describe where a link goes inside the `<a>` tag, instead of using a word like "here", it makes links more accessible and understandable (e.g. for those with screen readers). It's also better for SEO, as then Google can understand the meaning of our links better!

CTNicholas (approved):
Looks good, thanks @FlowFlorent! I've added more of an explanation, please check I've got it right.

docs/pages/platform/webhooks.mdx
FlowFlorent: TIL, thank you!

PR #1645, opened 2024-05-30
Custom notifications example

CTNicholas:
How to use custom notifications. This example loads the same user every time, as it refreshes after a notification, and is confusing otherwise.

[pwJPeHWmwN.webm](https://github.com/liveblocks/liveblocks/assets/33033422/76e06e68-92e2-4954-acd7-5d7d697f8dd5)


marcbouchenoire (approved):
Nice! Some optional feedback/questions:

- We should be able to remove everything related to Comments: things like `.thread` in `globals.css` and the mentions resolver (+ endpoint)
- Some notifications could be links (they're already links but adding an `href`, even "#"), for example to replace the `Details` button
- Instead of setting `documentTitle` in the notification directly, we could use `resolveRoomsInfo` with `useRoomInfo` maybe, since that's what is used in thread/text editor notifications to show the room/document's name

CTNicholas:
Good feedback, thanks! Implemented it. Need to come back to this after we have improved types, `useRoomInfo` is not ideal currently.

CTNicholas:
Upgraded properly to 2.0 now 👍

marcbouchenoire:
Nice! Also, we have to document it somewhere (and it to various places like config generation/snippets, etc) but we'll also be able to type custom notifications with `ActivitiesData`: https://github.com/liveblocks/liveblocks/pull/1669

PR #1648, opened 2024-05-30
Rename Monthly anonymous connection to Monthly public key connections to match emails

adigau:
What I did:
- Rename Monthly anonymous connections to Monthly public key connectiosn to match what @stevenfabre corrected in the usage progress emails
CleanShot 2024-05-30 at 15 52 32@2x


stevenfabre (approved):
Thank you!

PR #1650, opened 2024-05-31
Refactor some of the `useCurrentUserId` complexity

nvie:
> [!NOTE]
> This PR is stacked on top of #1649.

In order to get rid of the `useXxxBundle()` private/internal hooks which we'll no longer need, I first needed to get rid of the `useCurrentUserId` API. This caused me some headaches, because there are two instances internally, and depending on which context they're used in one or the other function will execute, but things were written in a way where you could not tell from the call site which one would get executed.

By moving the code around like this, hopefully these implementation detials will be easier to clean up a bit later.

For right now, it's important to check that I did not introduce a regression here — would love a confirmation the `react-comments` package is still working as-expected. Unit tests and E2E tests are obviously passing, but would like a smoke test! 🙏

### DIFF
--- a/packages/liveblocks-react-comments/src/components/InboxNotification.tsx
+++ b/packages/liveblocks-react-comments/src/components/InboxNotification.tsx
@@ -360,10 +360,10 @@ const InboxNotificationThread = forwardRef<
     const {
       useRoomInfo,
       useInboxNotificationThread,
-      [kInternal]: { useCurrentUserId },
+      [kInternal]: { useCurrentUserIdFromClient },
     } = useLiveblocksContextBundle();
     const thread = useInboxNotificationThread(inboxNotification.id);
-    const currentUserId = useCurrentUserId();
+    const currentUserId = useCurrentUserIdFromClient();
     // TODO: If you provide `href` (or plan to), we shouldn't run this hook. We should find a way to conditionally run it.
     //       Because of batching and the fact that the same hook will be called within <Room /> in the notification's title,
     //       it's not a big deal, the only scenario where it would be superfluous would be if the user provides their own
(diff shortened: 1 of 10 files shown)


packages/liveblocks-react-comments/src/components/InboxNotification.tsx
```diff
@@ -360,10 +360,10 @@ const InboxNotificationThread = forwardRef<
     const {
       useRoomInfo,
       useInboxNotificationThread,
-      [kInternal]: { useCurrentUserId },
+      [kInternal]: { useCurrentUserIdFromClient },
     } = useLiveblocksContextBundle();
     const thread = useInboxNotificationThread(inboxNotification.id);
-    const currentUserId = useCurrentUserId();
+    const currentUserId = useCurrentUserIdFromClient();
```
nvie, line 366: @marcbouchenoire @GuillaumeSalles I've not changed this one, but notice how my pure refactoring here shows that this was/is always calling the useCurrentUserId from the Liveblocks context here. Even if you would have a <RoomProvider> in there. I think that's potentially confusing, right? I think we'll want to simply use useCurrentUserId() here, which will call the appropriate method on either the client or the room, depending on the context.

marcbouchenoire: I don't know if it's confusing since their difference is an implementation detail and they return the same information

marcbouchenoire (approved):

packages/liveblocks-react-comments/src/components/InboxNotification.tsx
nvie: I would like to and can clean this up even later so we don't even have to have those two different implementations at all, but that's a post-2.0 thing, an 100% an implementation detail. So no rush with that.

nvie: OK, I just made the decision to do it: 551f5d460b46d93c2618830e298b299a9b7954df.

@GuillaumeSalles Lmk if I'm missing something here, as you were the original author for this part.

PR #1651, opened 2024-05-31
`@liveblocks/react-comments` → `@liveblocks/react-ui`

marcbouchenoire:
I had some time to help so I took a stab at this.

Fixes LB-377

The codemod package part could be the foundation for other codemods like the config changes ([LB-418](https://linear.app/liveblocks/issue/LB-418/codemods-for-20-package-rename-factory-etc))

### Changes

- `@liveblocks/react-comments` → `@liveblocks/react-ui`
- `<CommentsConfig />` → `<LiveblocksUIConfig />`
   - Feedback appreciated on this one, [docs](https://liveblocks.io/docs/api-reference/liveblocks-react-comments#CommentsConfig) if you're not familiar with what it does, it follows the naming convention laid out by `<SWRConfig />` and (Framer's) `<MotionConfig />`
 - Adds a new `@liveblocks/codemod` CLI package based on `@next/codemod` (which is itself based on `react-codemod`)
   - For now it contains a basic transform that will handle both `@liveblocks/react-comments` → `@liveblocks/react-ui` and `<CommentsConfig />` → `<LiveblocksUIConfig />`

### To-do

- [ ] I assume just changing the package name will mean that we will publish the package under the new name starting at 2.0, but we'll still have to [manually deprecate](https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions) `@liveblocks/react-comments`
- [ ] Upgrade all examples/starter kit etc, but I guess that's a bigger task of upgrading them to 2.0 entirely, not just this rename change


nvie (approved):
Lovely—thanks for lending a hand with this! ❤️

PR #1653, opened 2024-06-02
Improve generics overrides DX 

GuillaumeSalles:
- Use a `Liveblocks` interface instead of a namespace to support discriminated unions for `RoomEvent`
- Use the slate.js extended types trick with a third argument to restrict base type
- Reuse "bundles" types to get JSDoc and remove type expansion (LB-674)
- Fixes lose types test
- Migrate `nextjs-whiteboard-advanced` to new API

### DIFF
--- a/packages/liveblocks-react/test-d/augmentation.test-d.ts
+++ b/packages/liveblocks-react/test-d/augmentation.test-d.ts
@@ -1,4 +1,3 @@
-import type { Json } from "@liveblocks/client";
 import * as classic from "@liveblocks/react";
 import * as suspense from "@liveblocks/react/suspense";
 import { expectAssignable, expectError, expectType } from "tsd";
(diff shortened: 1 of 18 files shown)


packages/liveblocks-react/test-d/augmentation.test-d.ts
```diff
@@ -183,20 +172,20 @@ declare global {
   const { user, error, isLoading } = classic.useUser("user-id");
   expectType<boolean>(isLoading);
   expectType<string | undefined>(user?.name);
-  expectType<string | undefined>(user?.avatar);
-  expectType<Json | undefined>(user?.age);
-  expectType<Json | undefined>(user?.anyOtherProp);
+  expectError(user?.avatar);
+  expectType<number | undefined>(user?.age);
+  expectError(user?.anyOtherProp);
```
nvie, line 177: 👌

CTNicholas (approved):
Nice, looks good! 

Quick question, do we still need `createClient`? We could have this instead:

```tsx
<LiveblocksProvider options={{ throttle: 16, apiEndpoint: "..." }}>
```

It would make it easier to document if this is the only way to create a client in React.

nvie (approved):
Amazing improvement, solving all of our remaining blockers. Going to merge this in right now, and continue polishing it in the next PRs. Thanks for finding a path to unblock these final issues! 😍

PR #1656, opened 2024-06-03
Update `@beta` tags

marcbouchenoire:
This PR updates `@beta` tags in JSDoc comments to reflect that Comments is going to be out of beta, but not Notifications.


nvie (approved):
🙏

PR #1658, opened 2024-06-03
Remove bundle hooks

nvie:
Removes the internal `useLiveblocksContextBundle` and `useRoomContextBundle` internal helper hooks. The concept of a "bundle" is now 100% an implementation detail, no longer visible on the outer API surface area.

As an aside, I made the `useCurrentUserId` helper hook a 100% implementation detail inside `@liveblocks/react-ui`, and was now able to completely remove the `[kInternal]` API fields from the "bundles". This should make the implementation much easier to understand, as there are way fewer layers to be aware of and change when we want to make a change here.

Here is the final, complete, implementation of the [`useCurrentUserId()` hook](https://github.com/liveblocks/liveblocks/blob/8de44ae3d44d4b6e03f0160d8f9a0e461054f53d/packages/liveblocks-react-ui/src/shared.ts#L103-L128). No support code needed inside `@liveblocks/react`, no private APIs, etc. So `@liveblocks/react` doesn't even have to know about this.

### DIFF
--- a/packages/liveblocks-react-ui/src/shared.ts
+++ b/packages/liveblocks-react-ui/src/shared.ts
@@ -5,6 +5,7 @@ import {
   useLiveblocksContextBundleOrNull__,
   useRoom,
   useRoomContextBundleOrNull__,
+  useSelf,
 } from "@liveblocks/react";
 import React from "react";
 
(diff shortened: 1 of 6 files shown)


marcbouchenoire (approved):
🧹

PR #1661, opened 2024-06-04
Fix `InboxNotification` props types

marcbouchenoire:
We were trying too hard to get away with one props type for each kind but it doesn't really scale, the props we give to the users in their `kinds` components shouldn't always match the component (e.g. `InboxNotification.Custom`) props.

Example: We don't give users `title` and `children` in custom kinds props, but we want these props to be required on `<InboxNotification.Custom />`.

Fixes LB-687


CTNicholas (approved):
Great!

PR #1662, opened 2024-06-04
Add 5th `M` param to all internal types that didn't get it yet

nvie:
This PR changes:

```
Room<P, S, U, E>     // ❌
Room<P, S, U, E, M>  // ✅
```

As-is, this is a breaking change for people that have used the `Room` type in their application code. But I plan to set these Room type arguments to the `DP, DS, DU, DE, DM` defaults, so that in practice it will not be a breaking change for people. In fact, most people will be able to migrate by changing `Room<P, S, U, E>` to simply `Room`, and it will pick up those changes from their global types.

Please note that this PR does NOT change this just yet. The reason is that I want to ensure that we're never relying on these default type args ourselves. Default type args should are purely an end-user convenience feature. I will make this change a bit later, after I know for sure it works everywhere.

I've also DRYed up the `OpaqueRoom` definition here (an internal type), which is a little shorter to write APIs that take _any_ Room, whether it's coming from a global type, or set from a specific factory. Note that this is different from the `Room` type (without type args): `Room` means a room using all the global types, whereas `OpaqueRoom` means "any room, wherever the types are coming from".

Fixes LB-654


packages/liveblocks-core/src/custom-types.ts
```diff
@@ -1,10 +1,16 @@
...
+import type { BaseMetadata } from "./protocol/Comments";
+
+declare global {
+  /**
+   * Namespace for user-defined Liveblocks types.
+   */
+  export interface Liveblocks {
+    [key: string]: unknown;
+  }
+}
```
marcbouchenoire, line 13: 👀 👀 👀

marcbouchenoire (approved):

PR #1663, opened 2024-06-04
Use augmentation for `RoomInfo` type

marcbouchenoire:
This PR replaces the static `RoomInfo` type by an augmentable version.

Fixes LB-717


nvie (approved):
Beautiful! There is one linter error here, but otherwise this looks great! Thanks for adding it 😍

PR #1668, opened 2024-06-05
Rename `@liveblocks/node`'s `RoomInfo` to `RoomData`

marcbouchenoire:
To avoid a naming conflict, this PR renames `@liveblocks/node`'s `RoomInfo` type to `RoomData` (which also harmonizes it with our other `*Data` named types).

**Slack thread:** https://liveblocks.slack.com/archives/C02PZL7QAAW/p1717498677779919


packages/liveblocks-node/src/index.ts
```diff
@@ -7,7 +7,7 @@ detectDupes(PKG_NAME, PKG_VERSION, PKG_FORMAT);
 export type {
   LiveblocksOptions,
   RoomAccesses,
-  RoomInfo,
+  RoomData,
```
nvie, line 10: Should we add this to the end of `src/index.ts` here?

```ts
/**
 * @deprecated RoomInfo was renamed to RoomData, to avoid
 * confusion with the globally augmentable RoomInfo type. This
 * alias will be removed in a future version.
 */
export type RoomInfo = RoomData;
```

This may ease the transition if people upgrade to 2.0. We could remove the alias in 2.1.

nvie (approved):

packages/liveblocks-node/src/index.ts
marcbouchenoire: Good idea yeah, will do 👍

PR #1669, opened 2024-06-05
Allow typing custom notifications' `activities` via augmentation

marcbouchenoire:
This PR explores how we could allow typing custom notifications' `activities` via augmentation.

Since only activities' `data` is customizable and nothing else inside `InboxNotificationCustomData`, I created an intermediate type (`ActivitiesData`) that we use internally to do the matching between kinds and their activities' `data` type.

**Without augmentation, same as today:**

**With augmentation:**

This PR doesn't contain `tsd` tests yet, I wasn't really sure how I should go about it.

Fixes LB-724


packages/liveblocks-core/src/protocol/BaseActivitiesData.ts
```diff
@@ -0,0 +1,5 @@
+import type { ActivityData } from "./InboxNotifications";
+
+export type BaseActivitiesData = {
+  [key: `$${string}`]: ActivityData;
```
nvie, line 4: For the same reason as with our other base types, we should use
```suggestion
  [key: `$${string}`]: ActivityData | undefined;
```
here I think.

nvie (approved):
Looks good to me. What's the DX like? Is it what you hoped to get out of this?

nvie (commented):
Would love to see some type-level tests for this DX before merging.

packages/liveblocks-core/src/protocol/BaseActivitiesData.ts
marcbouchenoire: I'm not sure because it's an intermediate type: it doesn't represent any real data (unlike `ActivityData`) and it's never used by users, only augmented. It's only a way for users to give us a mapping "a kind" ↔ "its activity data type", and I don't think we want them to be allowed to give us `undefined` for a kind.

nvie: Oh, I see. That's actually a good point. Discard my comment 👍

marcbouchenoire:
@nvie Thanks again for the help with `tsd`, I think I'm happy with the set of tests I added (only covering this PR's changes for now).

PR #1672, opened 2024-06-05
Adds beautiful and helpful error messages for when the user-provided augmented types don't match their requirements

nvie:
This is what the erorr message look like.

In your favorite IDE:
In Vim

In the console:
In the console

Fixes LB-736.


marcbouchenoire (approved):
:chef-kiss:

PR #1675, opened 2024-06-05
Upgrade all examples to 2.0

nvie:
Fixes LB-658.


marcbouchenoire (approved):
We might have follow-ups to do for new examples (`nextjs-lexical` and the custom notifications one), after this one has landed.

PR #1679, opened 2024-06-05
Docs: 2.0 API reference updates

CTNicholas:
Documenting DX changes in the API reference pages.

LB-756


docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -10,21 +10,1134 @@ alwaysShowAllNavigationLevels: false
...
+}
+
+```
+
+### createLiveblocksContext
+
+<Banner>
+
+This used to be the default way to start your app, but now it’s recommend for
+advanced usage only. We generally recommend using ... TODO ... unless you need
```
nvie, line 904: @CTNicholas Just a small reminder to update this section. Looks unfinished.

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -10,21 +10,1134 @@ alwaysShowAllNavigationLevels: false
...
+  // authEndpoint: "/api/liveblocks-auth",
+  // throttle: 100,
+});
+
+// ...
+
+export const { RoomProvider } = createRoomContext(client);
+
+export const {
+  LiveblocksProvider,
```
nvie, line 933: Not sure where we should mention it, but _this_ `LiveblocksProvider`, in contrast to a "normal" imported `LiveblocksProvider`, does not take any props. Those are already bound from the `client`.

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -10,21 +10,1134 @@ alwaysShowAllNavigationLevels: false
...
+        colors: ["red", "purple"],
+        selection: {
+          id: 72426,
+        }}
+    >
+      {/* children */}
+    </RoomProvider>
+  );
+}
+```
```
nvie, line 1014: Not sure why the markdown seems to be flipping out here.

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -10,21 +10,1134 @@ alwaysShowAllNavigationLevels: false
...
+
 ### createRoomContext
 
-Creates a [`RoomProvider`][] and a set of typed hooks. We recommend using it in
-`liveblocks.config.ts` and re-exporting your typed hooks like below.
+<Banner>
 
-```tsx file="liveblocks.config.ts"
+This used to be the default way to start your app, but now it’s recommend for
+advanced usage only. We generally recommend using ... TODO ... unless you need
```
nvie, line 1060: @CTNicholas Similar reminder to update.

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -10,21 +10,1134 @@ alwaysShowAllNavigationLevels: false
...
+});
+
+type Presence = {};
+type Storage = {};
+type UserMeta = {};
+type RoomEvent = {};
+type ThreadMetadata = {};
+
+export const {
+  RoomProvider,
```
nvie, line 1086: Not sure where we should mention it, but the `RoomProvider` exported here does not need to be wrapped in a `LiveblocksProvider`, like it normally would have to. This is true only for the factory-exported one, since it's already bound to the passed-in `client` instance.

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -10,21 +10,1134 @@ alwaysShowAllNavigationLevels: false
...
+    </RoomProvider>
+  );
+}
+```
+
+Any
+[conflict-free live structures](/docs/api-reference/liveblocks-client#Storage)
+and JSON-serializable objects are allowed (the `LsonObject` type).
+
 ### createRoomContext
```
nvie, line 1055: Maybe we could make a separate page in the docs on the topic of "looks like I can only set one global room type, but my app needs multiple room types", and link there?

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -1241,6 +2408,174 @@ To use `useRoomInfo`, you should provide a resolver function to the
...
+```
+
+### Typing with createRoomContext
+
+Before Liveblocks 2.0, it was recommend to create your hooks using
+[`createRoomContext`][], and manually pass your types to this function. This is
+no longer [the recommended method](#Typing-your-data) for setting up Liveblocks,
+but it can still be helpful, for example you can use `createRoomContext`
+multiple times to create different room types, each with their own correctly
+typed hooks.
```
nvie, line 2517: Would it be helpful to link to the migration guide here?

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -1467,139 +2802,15 @@ And the following timeline:
...
-
 [`createclient`]: /docs/api-reference/liveblocks-client#createClient
+[`createRoomContext`]: /docs/api-reference/liveblocks-react#createRoomContext
 [`livelist`]: /docs/api-reference/liveblocks-client#LiveList
 [`livemap`]: /docs/api-reference/liveblocks-client#LiveMap
 [`liveobject`]: /docs/api-reference/liveblocks-client#LiveObject
 [`lostConnectionTimeout`]:
   /docs/api-reference/liveblocks-client#createClientLostConnectionTimeout
 [`room.history`]: /docs/api-reference/liveblocks-client#Room.history
+[`cl`]: /docs/api-reference/liveblocks-react#RoomProvider
```
nvie, line 2813: `cl`? Typo?

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -10,21 +10,1134 @@ alwaysShowAllNavigationLevels: false
...
+  useInboxNotifications,
+  /* ...all the other hooks you’re using... */
+} = createLiveblocksContext(client);
+```
+
 ## Room
 
+### RoomProvider
+
+Makes a [`Room`] available in the component hierarchy below. Joins the room when
```
nvie, line 943: Should this be ``[`Room`][]`` instead?

docs/pages/api-reference/liveblocks-react.mdx
CTNicholas: Good idea! It's on my to-do list for after the launch LB-852.

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -3653,14 +3669,114 @@ await stringifyCommentBody(comment.body, {
...
+a global `Liveblocks` interface. We recommend doing this in a
+`liveblocks.config.ts` file in the root of your app, so it’s easy to keep track
+of your types. Each type (`Presence`, `Storage`, etc.), is optional, but it’s
+recommended to make use of them.
+
+```ts file="liveblocks.config.ts"
+declare global {
+  interface Liveblocks {
+    // Each user's Presence
+    Presence: {};
```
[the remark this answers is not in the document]
CTNicholas: Hmm, yeah! I was trying to show people `Presence` is an object here. Do you recommend I show something else? Or maybe I just cut out this code snippet, and only use the one below. Going to merge everything now, but we can still change this.

CTNicholas:
Implemented everything, thank you for the thorough review @nvie!

docs/pages/api-reference/liveblocks-client.mdx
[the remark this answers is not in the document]
nvie: Actually, I just checked, and I think the DX is not as bad as I thought it would be. I think we can keep this after all 👍

PR #1680, opened 2024-06-06
Remove ComposerFocusCallbackProvider

GuillaumeSalles:


jrowny (approved):

packages/liveblocks-react-lexical/src/comments/floating-composer.tsx
```diff
@@ -114,21 +116,33 @@ export const FloatingComposer = forwardRef<
...
+        alignOffset={0}
+        collisionPadding={10}
+      >
+        <Composer
+          autoFocus
+          {...composerProps}
+          onKeyDown={handleKeyDown}
+          onComposerSubmit={handleComposerSubmit}
+          ref={forwardedRef}
+          onFocus={() => setShowActiveSelection(true)}
```
nimeshnayaju, line 143: Because we still accept an `onFocus` prop in this component, we should still execute the `onFocus` callback passed as prop before/after executing our callback. We do this for `onComposerSubmit` and `onKeyDown`.

PR #1684, opened 2024-06-06
Fix `react-comments-to-react-ui` codemod

marcbouchenoire:
Fixes LB-778

### DIFF
--- a/docs/pages/platform/upgrading/2.0.mdx
+++ b/docs/pages/platform/upgrading/2.0.mdx
@@ -23,27 +23,55 @@ Breaking changes:
 
 ## Step-wise migration guide
 
-I found the following order of mechanical changes to be most ideal path to upgrade. You can make a Git commit after every one of these changes.
+I found the following order of mechanical changes to be most ideal path to
+upgrade. You can make a Git commit after every one of these changes.
 
-1. Start with a clean slate, run `tsc` to make sure your project contains no TypeScript errors to begin with.
+1. Start with a clean slate, run `tsc` to make sure your project contains no
+   TypeScript errors to begin with.
 1. Replace `@liveblocks/react-comments` by `@liveblocks/react-ui`:
-   1. Run the codemod: `npx @liveblocks/codemod@latest react-comments-to-react-ui`
+   1. Run the codemod:
+      `npx @liveblocks/codemod@latest react-comments-to-react-ui`
    1. Run `npm uninstall @liveblocks/react-comments`
    1. Run `npm i @liveblocks/react-ui@latest`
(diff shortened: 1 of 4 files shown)


docs/pages/platform/upgrading/2.0.mdx
```diff
@@ -23,27 +23,55 @@ Breaking changes:
...
+   TypeScript errors to begin with.
 1. Replace `@liveblocks/react-comments` by `@liveblocks/react-ui`:
-   1. Run the codemod: `npx @liveblocks/codemod@latest react-comments-to-react-ui`
+   1. Run the codemod:
+      `npx @liveblocks/codemod@latest react-comments-to-react-ui`
    1. Run `npm uninstall @liveblocks/react-comments`
    1. Run `npm i @liveblocks/react-ui@latest`
+   1. Do a search for `@liveblocks/react-comments` to find any remaining
+      references and rename them manually, for example if you import the default
+      styles from within your own CSS files using `@import`.
```
nvie, line 38: Is this still necessary after you fixed the codemod? Or won't the codemod change any css files?

packages/liveblocks-codemod/src/transforms/react-comments-to-react-ui.ts
nvie, line 1: Not sure if possible, but could we match the source by regex somehow?

If we could change `.find({ source: { value: "@liveblocks/react-comments" } })` to something like `.find({ source: { value: /^@liveblocks\/react-comments\b/ } })` instead, that'd be great, because we wouldn't have to treat each subpath separately and it would also work for CSS imports automatically.

nvie (approved):
Nice, thanks for the fix! I love the input/output test setup too. 💯

docs/pages/platform/upgrading/2.0.mdx
marcbouchenoire: Yes, the codemods are JS/TS-only unfortunately 😕 But I think we could do a simpler find-and-replace in all CSS files for this use-case outside of the codemod itself, I'll merge this for now but tackle this tomorrow in a separate PR.

packages/liveblocks-codemod/src/transforms/react-comments-to-react-ui.ts
marcbouchenoire: The special `value` field can't be used with regexes but it's possible to just find all imports and filter them:
```
root
    .find(j.ImportDeclaration)
    .filter((path) => {
      return (
        path.value.source.type === "StringLiteral" &&
        CONFIG_IMPORT_REGEX.test(path.value.source.value)
      );
    })
```

I think in this context I like the repetition, to be safer than sorry.

PR #1685, opened 2024-06-06
Write error pages for types, and improve the error messages a bit further

nvie:
This PR writes specific error detail pages for each of our augmentable types, including examples of commonly possible errors in each case. I've tried my best to come up with good examples of code that actual users may have written when they hit these errors.

### Checklist

- [X] Presence
- [X] Storage
- [X] RoomEvent
- [X] ThreadMetadata
- [X] UserMeta
- [x] RoomInfo
- [ ] ActivitiesData

Fixes LB-761.

### DIFF
--- a/docs/pages/errors/ActivitiesData.mdx
+++ b/docs/pages/errors/ActivitiesData.mdx
@@ -1,0 +1,15 @@
+---
+meta:
+  title:
+    "The type you provided for ActivitiesData does not match its requirements"
+  parentTitle: "Error"
+  description: "Your ActivitiesData type is incorrect and needs to be updated"
+---
+
+## Why this error occurred
+
+...
+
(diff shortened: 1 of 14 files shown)


GuillaumeSalles:
The pages look amazing! :chef-kiss: They're pretty clear.

Maybe we can anticipate the most common errors when we except JSON? A small sentence along the line of `For example, Date, function, Map and classes are not valid JSON`.

GuillaumeSalles (approved):

nvie:
> Maybe we can anticipate the most common errors when we except JSON? A small sentence along the line of `For example, Date, function, Map and classes are not valid JSON`.

Great points, added in b3a900110e8c01e62d4b7c53ef04d901b4752891 👍

PR #1688, opened 2024-06-07
Enrich typings for `@liveblocks/node`

nvie:
This really improves the DX of using `@liveblocks/node`.

There is technically a breaking change here, but it's super simple to address:

```ts
const client = new Liveblocks()
client.getThread<MyMetadata>(...)
//              ^^^^^^^^^^^^ ❌ Will be removed, so breaking change
```

Instead do this:

```ts
// Just add the annotation globally
declare global {
  interface Liveblocks {
    ThreadMetadata: MyMetadata
  }
}

const client = new Liveblocks()
client.getThread(...)  // ✅ No annotation needed anymore
```

Fixes LB-828.


packages/liveblocks-node/src/client.ts
```diff
@@ -130,7 +157,13 @@ type SchemaPlain = DateToString<Schema>;
 /**
  * Interact with the Liveblocks API from your Node.js backend.
  */
-export class Liveblocks {
+export class Liveblocks<
+  _P extends JsonObject = DP,
```
nvie, line 161: @GuillaumeSalles @marcbouchenoire @nimeshnayaju I would like to hear your professional opinion here.

The most pragmatic approach here for now may be to remove the unused `P` param here. Or do we anticipate our Node client ever going to read or set Presence somehow? I don't think it makes much sense.

However, if we remove it here now, and it turns out later that we need to add it in anyway, then we must add it at the end and live with the inconsistent param order. (Not a big issue apart from the aestetics/consistency.)

I'm for removing it here now, and just having this client take 4 out of our 5 params instead. Thoughts?

packages/liveblocks-core/src/types/Patchable.ts
```diff
@@ -0,0 +1,9 @@
+type OptionalKeys<T> = {
+  [K in keyof T]-?: {} extends Pick<T, K> ? K : never;
+}[keyof T];
+
+type MakeOptionalFieldsNullable<T> = {
+  [K in keyof T]: K extends OptionalKeys<T> ? T[K] | null : T[K];
+};
+
+export type Patchable<T> = Partial<MakeOptionalFieldsNullable<T>>;
```
nvie, line 9: @marcbouchenoire I've replaced `PartialNullable<T>` with `Patchable<T>` across the board here. This new type is subtly closer to what we want when patching metadata.

Suppose you have:

```ts
type Meta = {
  color: "red" | "blue";
  resolved?: boolean;
}
```

Then:

```ts
PartialNullable<Meta>
// {
//   color?: "red" | "blue" | null;
//   resolved?: boolean | null;
// }
```

Whereas:

```ts
PartialNullable<Meta>
// {
//   color?: "red" | "blue";
//   resolved?: boolean | null;
// }
```

This ensures you cannot accidentally wipe mandatory fields when patching.

packages/liveblocks-node/src/client.ts
GuillaumeSalles: I would remove all the generic types on the `Liveblocks` class and only use module augmentation for the launch. Let's wait until we get enough users asking for this level of control. Generics are easy to add later on.

GuillaumeSalles (approved):
That's amazing! The last piece of the puzzle :chef-kiss:

Pre approving but would like to remove the generics on the `Liveblocks` class.

packages/liveblocks-node/src/client.ts
nvie: Ah, yes. That might actually be nice. They are harder to remove than to add. Perfect. Goes without saying that in that case you can only use the client for a single room type. If your app requires multiple room types, it’s not going to be possible.

I’m on board with that until someone needs it 👍

Will remove them before meeting.

nvie: @GuillaumeSalles While making the change, I realized that one downside is that unit tests [like this one](https://github.com/liveblocks/liveblocks/blob/736a08abbdc37362b77730aca28a57ac04b84ac1/packages/liveblocks-node/src/__tests__/client.test.ts#L581-L585) could no longer be used to test one-off edge cases. If the only way to control the types is the `global`, then all tests will always use the same types, and it's also impossible to test them for a user without augmentation 🤔

I think to get past this limitation, I'll export a private class that still has these type params, but bind them to `DS`, `DU`, etc in our public export. That way, users will not have the type params, but we can still use them in unit tests?

GuillaumeSalles: Mhh, I don't think tests should influence external API and internal complexity that much.

I believe it's okay to adapt the tests by adding default metadata all the time or cast to `any` if this is the only choice. It just feels simpler to maintain and more pragmatic, no?

nvie: A'ight, did that in 2bfa19799a573845bb9b3b23500fcc457bbaa2d8 now 👍  I think this is ready for merge.

PR #1689, opened 2024-06-08
`create-liveblocks-app`: 2.0 config generation

CTNicholas:
The new config file is [here](https://github.com/liveblocks/liveblocks/blob/8a1e5d6ed83f4d58f955502850107aef171fa9cd/packages/create-liveblocks-app/bin/templates/init/config-file.ts). I've removed some flags that are no longer needed.


marcbouchenoire (approved):
👌

nvie (commented):
👨‍🍳🤌

PR #1691, opened 2024-06-09
Fix `EnsureJson` helper

nvie:
I noticed a bug in `EnsureJson` helper after upgrading our `zustand-flowchart` example. We should recursively remove all functions (like in this case `toString() => string`) from all nested types instead of retaining them as-is. Maybe this is an issue that came to light only on higher TS versions.


CTNicholas (approved):
Looks good!

PR #1694, opened 2024-06-10
Add upgrade command to `create-liveblocks-app`

marcbouchenoire:
It's not the best place for it but in the interest of time, I think it's still beneficial so that we can have a simple command for that everywhere.

Ideally it would be more fleshed out and would live in `@liveblocks/upgrade` maybe (like [Astro](https://x.com/cpenned/status/1798819246136524873)), or in a future `liveblocks` general CLI.

@CTNicholas There were a few formatting changes so I'll wait for https://github.com/liveblocks/liveblocks/pull/1689 to land before merging this one.

### Usage

```
npx create-liveblocks-app@latest --upgrade
```

### DIFF
--- a/packages/create-liveblocks-app/scripts/demo.js
+++ b/packages/create-liveblocks-app/scripts/demo.js
@@ -18,7 +18,6 @@ const pkgJson = `{
     "@liveblocks/client": "^1.11.0",
     "@liveblocks/node": "^1.11.0",
     "@liveblocks/react": "^1.11.0",
-    "@liveblocks/react-comments": "^1.11.0",
     "next": "^13.4.16",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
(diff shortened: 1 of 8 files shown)


packages/create-liveblocks-app/scripts/demo.js
```diff
@@ -4,6 +4,34 @@ import { esbuildOptions } from "./esbuild.js";
...
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start"
+  },
+  "dependencies": {
+    "@liveblocks/client": "^1.11.0",
+    "@liveblocks/node": "^1.11.0",
+    "@liveblocks/react": "^1.11.0",
+    "@liveblocks/react-comments": "^1.11.0",
```
nvie, line 21: Should this be `react-ui` or is that the entire goal here? To illustrate how it will get upgraded?

nvie (approved):

packages/create-liveblocks-app/scripts/demo.js
marcbouchenoire: It's a naive test so it doesn't matter too much, I'll remove it altogether to avoid confusion.

CTNicholas:
Nice! Was going to add this if I had time 👍

PR #1695, opened 2024-06-10
Remove need for thunking in `<ClientSideSuspense>`

nvie:
The goal is to no longer require:

```tsx
<ClientSideSuspense fallback={<Loading />}>
  {() => <MyComponent />
</ClientSideSuspense>
```

But instead allow this:

```tsx
<ClientSideSuspense fallback={<Loading />}>
  <MyComponent />
</ClientSideSuspense>
```

### Roll-out plan

I plan to remove the ability to pass in a thunk here first, release that as 2.0.0-alpha2 today, use that to force a rewrite of all examples by removing that, and then restore the ability to allow passing a thunk again for backward-compatibility. We're not going to force people to upgrade, but they can.

We should also update the documentation.

### DIFF
--- a/packages/liveblocks-react/src/ClientSideSuspense.tsx
+++ b/packages/liveblocks-react/src/ClientSideSuspense.tsx
@@ -1,9 +1,11 @@
-import type { ReactElement, ReactNode } from "react";
+import type { ReactNode } from "react";
 import * as React from "react";
 
 type Props = {
-  fallback: NonNullable<ReactNode> | null;
-  children: () => ReactNode | undefined;
+  fallback: ReactNode;
+  children: // TODO Restore this again later
+  // (() => ReactNode | undefined) |
+  ReactNode | undefined;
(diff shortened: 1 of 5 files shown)


packages/liveblocks-react/src/ClientSideSuspense.tsx
```diff
@@ -1,9 +1,11 @@
...
+import type { ReactNode } from "react";
 import * as React from "react";
 
 type Props = {
-  fallback: NonNullable<ReactNode> | null;
-  children: () => ReactNode | undefined;
+  fallback: ReactNode;
+  children: // TODO Restore this again later
+  // (() => ReactNode | undefined) |
+  ReactNode | undefined;
```
nvie, line 8: I will restore this ability again after the 2.0.0-alpha2 release.

marcbouchenoire (approved):
🙌

PR #1700, opened 2024-06-10
Text collaboration tweaks

marcbouchenoire:
A few text collaborations (the only ones I'll have time for unfortunately), mostly around `ThreadPanel` and styles.


nimeshnayaju (approved):
🚀

PR #1703, opened 2024-06-10
Updated README for 2.0

stevenfabre:


GuillaumeSalles (approved):

GuillaumeSalles:
I would just add a link to get started and/or signup somewhere, but it's looking good!

stevenfabre:
> I would just add a link to get started and/or signup somewhere, but it's looking good!

Good call! Just added that.

PR #1706, opened 2024-06-12
New codemods

marcbouchenoire:


packages/liveblocks-codemod/src/transforms/liveblocks-yjs-provider.ts
nvie, line 1: Could we rename this codemod to `remove-yjs-default-export`?

nvie (approved):
Lovely! Just one naming suggestion.

PR #1708, opened 2024-06-12
Small typo on id-tokens.mdx

davenewza:
Minor typo fix, [here](https://liveblocks.io/docs/authentication/id-token#permissions-levels-groups-accesses-example), on the id-token page.  

Typo:

 > After calling this, ever user in the “engineering” group will have read-only access. 

Fix:

> After calling this, every user in the “engineering” group will have read-only access.


nvie (approved):
Nice catch, thanks, @davenewza!

PR #1713, opened 2024-06-12
Allow upgrading to specific version using `create-liveblocks-app --upgrade`

nvie:
To upgrade all Liveblocks packages to the latest version:

    $ npx create-liveblocks-app@latest --upgrade

To upgrade all Liveblocks packages to the `beta` version:

    $ npx create-liveblocks-app@latest --upgrade beta

To upgrade all Liveblocks packages to a specific version:

    $ npx create-liveblocks-app@latest --upgrade 1.12.0

### DIFF
--- a/packages/create-liveblocks-app/scripts/demo.js
+++ b/packages/create-liveblocks-app/scripts/demo.js
@@ -31,8 +31,6 @@ const pkgJson = `{
   }
 }`;
 
-const flags = "";
-
 /**
  * Use `npm run demo` and check in .demo folder
  */
@@ -45,7 +43,7 @@ fs.rmSync(testDir, { recursive: true, force: true });
 fs.mkdirSync(testDir);
 fs.writeFileSync(path.join(testDir, "package.json"), pkgJson);
 
-execSync(`node ${esbuildOptions.outfile} ${flags}`, {
+execSync(`node ${esbuildOptions.outfile} ${process.argv.slice(2)}`, {
   cwd: testDir,
   stdio: "inherit",
 });
--- a/packages/create-liveblocks-app/README.MD
+++ b/packages/create-liveblocks-app/README.MD
@@ -93,8 +93,9 @@ Options:
   --open
   Open browser without asking permission, when deploying to Vercel or getting API key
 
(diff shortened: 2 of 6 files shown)


marcbouchenoire (commented):
I think stylistically I would prefer `--upgrade` (latest) and `--upgrade=1.12.0`/`--upgrade="beta"` given that it's already a flag and not a command, would it be weird?

nvie:
> I think stylistically I would prefer `--upgrade` (latest) and `--upgrade=1.12.0`/`--upgrade="beta"` given that it's already a flag and not a command, would it be weird?

@marcbouchenoire I tried, but I could not get it to work. A flag is either a boolean, or it's a string, in which case it takes an argument, but I couldn't get it to work with the cmdline arg parser to do this hybrid form, so I settled on this optional `tag` flag, which I didn't mind too much. If you have an idea of how to implement the hybrid flag, I'm still open to it.

packages/create-liveblocks-app/scripts/demo.js
```diff
@@ -45,7 +43,7 @@ fs.rmSync(testDir, { recursive: true, force: true });
 fs.mkdirSync(testDir);
 fs.writeFileSync(path.join(testDir, "package.json"), pkgJson);
 
-execSync(`node ${esbuildOptions.outfile} ${flags}`, {
+execSync(`node ${esbuildOptions.outfile} ${process.argv.slice(2)}`, {
```
nvie, line 46: What's happening here? This feels a bit brittle.

marcbouchenoire: This is just a demo script to test the built CLI locally, `npm run demo` etc. This change forwards all flags (the two first args are `node` and the file path) so we can do `npm run demo --upgrade` with it.

CTNicholas (approved):
Hey guys, this had a few bugs, so I've just fixed them. You can use these commands:

```bash
# latest
--upgrade

## 2.0.0
--upgrade 2.0.0

# Also works
--upgrade @2.0.0
```

If you use the UI it will always upgrade to the latest version.

PR #1718, opened 2024-06-13
Fix type signatures in `@liveblocks/node` APIs

nvie:
This PR fixes the type signature of `client.identifyUser()` and `client.prepareSession()` in our `@liveblocks/node` package, to require `userInfo` if it's mandatory according to your global `UserMeta` definition.


CTNicholas (approved):
Good improvement!

PR #1719, opened 2024-06-13
feat: Add support for deleting threads

FlowFlorent:
### Description

**@liveblocks/react**

Room & thread cache now handle thread deletion by listening to THREAD_DELETED event

**@liveblocks/node**
- Adds `threadDeleted` webhook event
- Adds `deleteThread` method to client


packages/liveblocks-core/src/protocol/ServerMsg.ts
```diff
@@ -26,6 +26,7 @@ export enum ServerMsgCode {
   COMMENT_DELETED = 404,
   COMMENT_REACTION_ADDED = 405,
   COMMENT_REACTION_REMOVED = 406,
+  THREAD_DELETED = 407,
```
nvie, line 29: We can safely move this a couple lines higher, together with the other `THREAD_` enum members. You don't have to worry about not noticing the code if we add new members later. We already have a lint rule that will fail if there are duplicate members.

nvie (approved):
💯

packages/liveblocks-core/src/protocol/ServerMsg.ts
FlowFlorent: nice, adjusting :)

packages/liveblocks-node/src/client.ts
```diff
@@ -1224,6 +1224,25 @@ export class Liveblocks {
...
+   * @param params.threadId The thread ID to delete.
+   */
+  public async deleteThread(params: {
+    roomId: string;
+    threadId: string;
+  }): Promise<void> {
+    const { roomId, threadId } = params;
+
+    const res = await this.delete(url`/v2/rooms/${roomId}/threads/${threadId}`);
+    console.log("deleteThread res", res);
```
FlowFlorent, line 1239:
```suggestion
```

Issue #1717, opened 2024-06-13
Bug: Comment Splits into Two When Typing in Vietnamese and Pressing Enter


lamtuanamc:
**Describe the bug**

When using the comment feature in Liveblocks and typing in Vietnamese, if there are still underlined words (indicating that they are in the process of being composed), pressing Enter splits the comment into two. However, clicking the send button does not cause this issue.

**To Reproduce**

1. Open the comment feature in Liveblocks.
2. Start typing a comment in Vietnamese.
3. Ensure that there are still underlined words (indicating that the words are being composed).
4. Press Enter to submit the comment.

**Expected behavior**

The comment should be posted as a single comment regardless of whether the Enter key or the send button is used.

**Actual Behavior:**
When pressing Enter, the comment is split into two separate comments. When clicking the send button, the comment is posted as a single comment as expected.

**Illustrations**

https://github.com/liveblocks/liveblocks/assets/37234239/51ac76ab-c52e-4902-8697-3884b9415377

marcbouchenoire:
Hi @lamtuanamc, thanks for the detailed report!

We'll investigate, and in the meantime could you try `Enter`ing on https://www.slatejs.org/examples/richtext and tell us if there are any unexpected behaviors? This is the text editor used in the Comments composer.

PR #1723, opened 2024-06-14
Comments: Composer fixes/improvements

marcbouchenoire:
This contains a few fixes and improvements for the Comments composer:

- Don't clear the composer if it's not handled (via `onComposerSubmit`), fixes LB-874
- Prevent the composer from splitting text being composed, fixes https://github.com/liveblocks/liveblocks/issues/1717
- Count whitespace as empty to prevent posting empty comments
- Handle parentheses around and within (if they are balanced, e.g. `https://google.com/hello(world))` will not contain the last parenthesis) auto-links, fixes LB-875

### To-do

- [ ] Follow-up on reports
- [x] More testing around the fix for https://github.com/liveblocks/liveblocks/issues/1717

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,10 @@
 
 ### `@liveblocks/react-ui`
 
-- Improve the composer if it's not handled. (via `onComposerSubmit`)
 - Prevent the composer from splitting text being composed.
-- Count whitespace as empty to prevent posting empty comments.
 - Handle parentheses around and within auto-links.
+- Count whitespace as empty to prevent posting empty comments.
+- Prevent clearing the composer if it's not handled. (via `onComposerSubmit`)
 
 ## v2.0.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,17 @@
-## v2.0.1
+## v2.0.1 (not released yet)
 
 ### `@liveblocks/node`
 
 - Fix type signatures of `client.identifyUser()` and `client.prepareSession()`
   to require `userInfo` if it's mandatory according to your global `UserMeta`
   type definition.
+
+### `@liveblocks/react-ui`
+
+- Don't clear the composer if it's not handled. (via `onComposerSubmit`)
+- Prevent the composer from splitting text being composed.
+- Count whitespace as empty to prevent posting empty comments.
+- Handle parentheses around and within auto-links.
 
 ## v2.0.0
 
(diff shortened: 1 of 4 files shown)


nimeshnayaju (approved):
Thank you for working on this @marcbouchenoire <3

CHANGELOG.md
```diff
@@ -1,11 +1,18 @@
-## v2.0.1
+## v2.0.1 (not released yet)
```
nvie, line 1: 2.0.2 was released last Friday, so we'll have to move this to 2.0.3

```suggestion
## v2.0.3 (not released yet)
```

packages/liveblocks-react-ui/src/slate/plugins/auto-links.ts
```diff
@@ -137,6 +139,38 @@ function endsWithPeriodOrQuestionMark(textContent: string): boolean {
   );
 }
 
+/**
+ * Helper function to get the "logical length" of a URL, taking into account things like opening/closing parentheses
+ * @param url The URL to check
+ * @returns The "logical length" of the URL
```
nvie, line 145: Out of curiosity, what kind of URLs are being analysed here?

packages/liveblocks-react-ui/src/slate/utils/is-empty.ts
```diff
@@ -1,10 +1,39 @@
...
-  return (
-    children.length <= 1 &&
-    !SlateText.isText(children[0]) &&
-    SlateEditor.isEmpty(editor, children[0])
-  );
+  // No children
+  if (children.length === 0) {
+    return true;
+  }
+
```
nvie, line 11: You could remove this case, because it's already covered by the loop that won't do any looping.

nvie (approved):
Nice that you got this fixed! 👍

packages/liveblocks-react-ui/src/slate/plugins/auto-links.ts
marcbouchenoire: Any URL being composed that matched `URL_REGEX` above, so we already know they are valid URLs when this helper is invoked.

packages/liveblocks-react-ui/src/slate/utils/is-empty.ts
marcbouchenoire: 😅 Thanks for the catch!

packages/liveblocks-react-ui/src/slate/plugins/auto-links.ts
nvie: I was specifically curious about the need for doing the balanced parentheses counting. What's that needed for?

marcbouchenoire: It allows you to type "Hello ([https://google.com/hello(world)](https://google.com/hello(world)))". Otherwise, it would either be:
- parentheses are not handled in links
- once you start typing a link, it can't be followed by a closing parenthesis because it would still be counted as part of the link

nvie: Ohh, that's quite nice, and seems to be what GitHub is also doing! 💯

PR #1725, opened 2024-06-17
Make `initialPresence` and `initialStorage` optional…

nvie:
…but only if they truly are optional in your type definition.

Some examples:

```tsx
declare global {
  interface Liveblocks {
    // Not using Presence here
  }
}

// ✅ No longer needed if you're not using/interested in Presence
<RoomProvider initialPresence={{}}>...</RoomProvider>
<RoomProvider>...</RoomProvider>
```

```tsx
declare global {
  interface Liveblocks {
    Presence: {
      cursor: { x: number; y: number },
      foo?: string,
    }
  }
}

// ❌ Now a TS error: missing mandatory `initialPresence` prop
<RoomProvider>...</RoomProvider>

// ❌ Now a TS error: missing field `cursor`
<RoomProvider initialPresence={{}}>...</RoomProvider>
```

```tsx
declare global {
  interface Liveblocks {
    Presence: {
      cursor?: { x: number; y: number },  // 👈 Only changed this line
      foo?: string,
    }
  }
}

// ✅ Now perfectly fine!
<RoomProvider>...</RoomProvider>

// ✅ Now perfectly fine!
<RoomProvider initialPresence={{}}>...</RoomProvider>
```

Same logic applies to the `initialStorage` prop.

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -21,6 +21,8 @@ import type {
   Json,
   JsonObject,
   LsonObject,
+  OptionalTupleUnless,
+  PartialUnless,
   Patchable,
   PlainLsonObject,
   QueryMetadata,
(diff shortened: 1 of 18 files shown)


marcbouchenoire (approved):
Beautiful ❤️

packages/liveblocks-node/src/client.ts
```diff
@@ -101,7 +101,7 @@ export type CreateThreadOptions<M extends BaseMetadata> = {
   roomId: string;
   data: {
     comment: { userId: string; createdAt?: Date; body: CommentBody };
-  } & (Record<string, never> extends M ? { metadata?: M } : { metadata: M });
+  } & PartialUnless<M, { metadata: M }>;
```
nvie, line 104: @marcbouchenoire The central way I found here to DRY up these definitions, is using this `PartialUnless<X, T>` helper, which will be `Partial<{ metadata: M }>` (= `{ metadata?: M }`) if the empty object can be assigned to type `X`, but `{ metadata: M }` otherwise.

PR #1731, opened 2024-06-17
Avoid legacy APIs in examples

nvie:
I found these two examples were still using legacy APIs. Specifically `room.subscribe("connection")` no longer exists. At first, I switched these over to `room.subscribe("status")` (which replaces the legacy socket connection API). However, when I did I noticed a small regression. When the `"connected"` event is emitted, the `.getSelf()` will still return `null`. The value of `.getSelf()` is only updated _after_ that connected event happens.

This may be a regression. I'd have to trace back to where it has been introduced.

However, we also have our event hub API, which is much nicer to use than the `.subscribe("...")` API, and it has a `room.events.self.subscribe()` which is triggered whenever "self" gets updated. This is a much nicer event to listen to anyway (than relying on `"status"` updates), so I've fixed that in these examples.

@CTNicholas Do you want to give these examples a quick test to validate I updated them correctly?

### DIFF
--- a/examples/nuxtjs-live-avatars/app.vue
+++ b/examples/nuxtjs-live-avatars/app.vue
@@ -84,7 +84,7 @@ export default {
     this._room = room;
     this._leave = leave;
     this._unsubscribeOthers = room.events.others.subscribe(this.onOthersChange);
-    this._unsubscribeSelf = room.events.self.subscribe(this.onStatusChange);
+    this._unsubscribeSelf = room.events.self.subscribe(this.onSelfChange);
   },
   destroyed() {
     this._unsubscribeOthers();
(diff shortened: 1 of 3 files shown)


GuillaumeSalles:
@nvie: I don't think we've ever documented `room.events`. If we start using `room.events` in the examples, we should document it first IMO

GuillaumeSalles (approved):

nvie:
> @nvie: I don't think we've ever documented `room.events`. If we start using `room.events` in the examples, we should document it first IMO

Agreed. The only reason I'm residing to `room.events` here is because we don't have a `room.subscribe("self")` here, and the example is otherwise broken. I'll get those new event APIs documented tomorrow 👍

CTNicholas (approved):
Yep, both working! Thanks

nvie:
Closing all PRs that have been opened for more than a year.

PR #1732, opened 2024-06-17
Prevent `LiveblocksProvider` nesting

nvie:
It makes no sense to nest `LiveblocksProvider`s, and we should prevent it with an error if it would ever happen.

Currently this is possible, but leads to unpredictable behavior.

```tsx
<LiveblocksProvider /* options */>
  <LiveblocksProvider /* other options */>
    <MyApp />
  </LiveblocksProvider>
</LiveblocksProvider>
```

Things would get even weirder here, so all in all better to prevent it entirely:

```tsx
<LiveblocksProvider /* options */>
  <RoomProvider id="my-room">
    <LiveblocksProvider /* other options */>
      <MyApp />
    </LiveblocksProvider>
  </RoomProvider>
</LiveblocksProvider>
```

Fixes LB-873.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 - In `<RoomProvider>`, the props `initialPresence` and `initialStorage` are now
   only mandatory if your custom type requires them to be.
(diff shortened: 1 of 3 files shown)


marcbouchenoire (approved):
👮

PR #1736, opened 2024-06-18
Improve `.initializeStorageDocument()` API

nvie:
This now takes a `LiveObject` instance directly, so that users no longer have to bother with knowing about or calling `toPlainLson()` manually.

Creating a new room from the backend is now as easy as:

```ts
await client.initializeStorageDocument("my-room", new LiveObject({
  layers: new LiveMap(),
  layerIds: new LiveList([]),
}));
```

Fixes LB-884.


packages/liveblocks-node/src/client.ts
```diff
@@ -705,9 +702,11 @@ export class Liveblocks {
    */
   public async initializeStorageDocument(
     roomId: string,
-    document: PlainLsonObject
+    document: LiveObject<S> | PlainLsonObject
   ): Promise<PlainLsonObject> {
```
nvie, line 706: Actual question: I just noticed we're returning the document in plain LSON format here. This format is of course technically needed to get the Live objects over the wire, but I don't think in actual apps it's a nice format to work with. Should we not lift the level of abstraction a bit higher here, and either return:

- `void`
- `LiveObject` instance
- Simplified JSON (i.e. the equivalent of calling `storage.toImmutable()`)

All of those three seem better than returning a `PlainLsonObject`.

@CTNicholas @nimeshnayaju What do you think?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## v2.0.4 (not released yet)
+
+### `@liveblocks/node`
+
+- Fix `client.initializeStorageDocument()` which can now take a `LiveObject`
```
marcbouchenoire, line 5:
```suggestion
- Improve `client.initializeStorageDocument()` which can now take a `LiveObject`
```

marcbouchenoire (approved):

CTNicholas:
Nice! This would definitely be easier to use. I'm assuming `LiveMap`, `LiveList`, etc. work in all environments (Browser, Node.js, Workers, etc.).

Why not just have a plain object at the root of the argument? Since it's always a `LiveObject` anyway, or at least, it always is when you use `initialStorage`.

```ts
await client.initializeStorageDocument("my-room", {
  layers: new LiveMap(),
  layerIds: new LiveList([]),
});
```

packages/liveblocks-node/src/client.ts
CTNicholas: Right! I think `LiveObject` is not particularly useful on the back end, so I'd probably skip that. I think `void` or JSON makes the most sense to me. However, bear in mind that `getStorageDocument` returns LSON by default, and we want our APIs to be consistent.

nvie:
> Nice! This would definitely be easier to use. I'm assuming `LiveMap`, `LiveList`, etc. work in all environments (Browser, Node.js, Workers, etc.).

Yeah, they do, until you start "using" them, i.e. start calling `myObj.set("foo", "bar")`. This will want to start emitting messages to the server, which requires a WebSocket polyfill. But as long as you're only using them to build a document like in this case, that should work.

> Why not just have a plain object at the root of the argument? Since it's always a `LiveObject` anyway, or at least, it always is when you use `initialStorage`.

We could do that too, can update 👍

packages/liveblocks-node/src/client.ts
nvie: Agreed, so the consistency argument rules out option 3, too. So… should we make this `void` instead? That, however, would be a breaking change, so maybe not change anything after all?

CTNicholas: Vincent's written up more ideas here: LB-891

nvie:
Closing all PRs that have been opened for more than a year.

PR #1737, opened 2024-06-18
Don't try to write missing `initialStorage` keys if current user has no write access

nvie:
This PR changes the client so that it won’t attempt to write missing `initialStorage` keys if the current user has no write access to storage. This will no longer throw, but issue a warning message in the console. This came up in [this support question](https://discord.com/channels/913109211746009108/1251201132538363924/1251201132538363924).

Fixes LB-888.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## v2.0.4 (not released yet)
+
+### `@liveblocks/client`
+
+- Don’t attempt to write missing initialStorage keys if the current user has no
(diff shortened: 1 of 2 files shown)


CTNicholas (approved):
Nice! Should we also add something like... "consider initializing your Storage on the server: https://liveblocks.io/docs/api-reference/liveblocks-node#post-rooms-roomId-storage"?

GuillaumeSalles (approved):

PR #1739, opened 2024-06-18
Add missing type export for `CommentReaction`

nvie:


packages/liveblocks-client/src/index.ts
```diff
@@ -24,6 +24,7 @@ export type {
   CommentBodyText,
   CommentBodyTextElementArgs,
   CommentData,
+  CommentReaction,
```
nvie, line 27: This was only exported from `@liveblocks/core`, and not yet from `@liveblocks/client`.

marcbouchenoire (approved):
👀

PR #1740, opened 2024-06-19
Fix DX issue around visibility of custom `UserMeta` type errors

nvie:
Fixes a type error visiblity issue, see this video for an explainer:

https://github.com/liveblocks/liveblocks/assets/83844/ca835aa1-8cd5-4c20-9e4f-49ca47ff731f

Fixes LB-904.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## v2.0.4 (not released yet)
 
+### All packages
+
+- Improve TS error message location if `UserMeta` type is not compatible with
+  base requirement
+
 ### `@liveblocks/client`
 
 - Add missing type export for `CommentReaction`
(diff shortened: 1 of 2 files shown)


marcbouchenoire (approved):
Nice find!

Like you said I think we should be able to do the same for `ActivitiesData`.
The only place where the augmented/provided `ActivitiesData` type is used is [the following](https://github.com/liveblocks/liveblocks/blob/168ce0f804b47c5a46b9fcf850912eb3291c46df/packages/liveblocks-core/src/protocol/InboxNotifications.ts#L31-L35):
```
type InboxNotificationActivity<K extends `$${string}` = `$${string}`> = {
  id: string;
  createdAt: Date;
  data: K extends keyof DAD ? DAD[K] : ActivityData;
};
```

So if `DAD`'s error would be ``Record<`$${string}`, MakeErrorString<"ActivitiesData">`` maybe it would trickle down as expected?

nvie:
@marcbouchenoire I'll take a look at the ActivitiesData DX and open another PR for that if I can find a solution.

PR #1741, opened 2024-06-19
Upgrade GA actions to their latest versions

nvie:
[We need to upgrade.](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) The main change in these versions is that they're using Node 20, instead of Node 16 (which is EoL).

Fixes LB-886.

### DIFF
--- a/.github/workflows/publish-devtools.yml
+++ b/.github/workflows/publish-devtools.yml
@@ -13,6 +13,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
+          node-version: "18.x"
           cache: "npm"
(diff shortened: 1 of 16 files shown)


marcbouchenoire (commented):
Should we change `node-version: [18.x]` to `node-version: [20.x]` in actions where it's specified?

nvie:
> Should we change `node-version: [18.x]` to `node-version: [20.x]` in actions where it's specified?

I'm sticking with 18.x for now, just because that's the version of Node we're all using. Happy to bump it to 20.x soon, though, but we should do it all together and in all repos, ideally so saving that for a separate change.

marcbouchenoire (approved):

PR #1743, opened 2024-06-19
Improve error messages and error locations for `ActivitiesData` errors

nvie:
Suppose you make a mistake in your `ActivitiesData` type definition _somewhere_:

## Before

Types would break in unhelpful ways:

Screen Shot 2024-06-19 at 14 01 14@2x

## After

Types now show the error, pointing to our URL with a better explanation and ways to fix:

Screen Shot 2024-06-19 at 14 02 51@2x


packages/liveblocks-react-ui/src/components/InboxNotification.tsx
```diff
@@ -59,11 +59,8 @@ type ComponentTypeWithRef<
   P,
 > = ComponentType<P & Pick<ComponentProps<T>, "ref">>;
 
-type InboxNotificationKinds = {
-  [K in keyof DAD]: ComponentTypeWithRef<
-    "a",
-    InboxNotificationCustomKindProps<K>
-  >;
+type InboxNotificationKinds<KS extends KDAD = KDAD> = {
```
marcbouchenoire, line 62: Could you add a comment here to clarify why we need to do that?

packages/liveblocks-core/src/globals/augmentation.ts
```diff
@@ -76,4 +76,15 @@ export type DE = GetOverride<"RoomEvent", Json, "is not a valid JSON value">;
...
+  BaseActivitiesData,
+  {
+    [K in keyof Liveblocks["ActivitiesData"]]: `At least one of the custom notification kinds you provided for 'ActivitiesData' does not match its requirements. To learn how to fix this, see https://liveblocks.io/docs/errors/ActivitiesData`;
+  }
+>;
+
+export type KDAD = keyof DAD extends `$${string}`
+  ? keyof DAD
+  : `Your custom 'ActivitiesData' type can only contain kind names that start with '$'. To learn how to fix this, see https://liveblocks.io/docs/errors/ActivitiesData`;
+// XXX Change error message to include link
```
marcbouchenoire, line 90: To be removed.

marcbouchenoire (approved):
Thanks for improving this! I left a few comments but I'm approving already.

packages/liveblocks-react-ui/src/components/InboxNotification.tsx
nvie: Added in 5ef77741f60c4c204c512e30901a96df6a9e5ed2 👍

packages/liveblocks-core/src/globals/augmentation.ts
nvie: Thx, removed.

PR #1745, opened 2024-06-19
Fix token info in docs

CTNicholas:
Should not be a string


nvie (approved):
Oh, I thought it had to be. My bad!

PR #1746, opened 2024-06-19
useDeleteThread

FlowFlorent:
### Description

Adds a `useDeleteThread` hook. Only the thread owner can delete the thread. The thread owner is the user of the first comment in the thread.

```tsx
function Example() {
  const { threads } = useThreads();
  const self = useSelf();

  const deleteThread = useDeleteThread();

  return (
    <main>
      {threads.map((thread) => (
        <>
          {self.id === thread.comments[0].userId ? (
            <button onClick={() => deleteThread(thread.id)}>
              Delete thread below
            </button>
          ) : null}
          <Thread key={thread.id} thread={thread} className="thread" />
        </>
      ))}
      <Composer className="composer" />
    </main>
  );
}
```

Screenshot 2024-06-19 at 3 46 42 PM
Screenshot 2024-06-19 at 3 46 30 PM

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -66,6 +66,7 @@ import {
   MarkInboxNotificationAsReadError,
   RemoveReactionError,
   UpdateNotificationSettingsError,
+  DeleteThreadError,
 } from "./comments/errors";
 import { createCommentId, createThreadId } from "./comments/lib/createIds";
 import { selectNotificationSettings } from "./comments/lib/select-notification-settings";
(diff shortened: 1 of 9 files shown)


GuillaumeSalles (approved):
Looking good @FlowFlorent! Just missing the docs but the rest looks clean!

packages/liveblocks-react/src/room.tsx
```diff
@@ -538,6 +539,7 @@ function makeRoomContextBundle<
     // context consistent internally.
     return (
       <LiveblocksProviderWithClient client={client} allowNesting>
+        {/* Why type error? */}
```
nvie, line 542: @FlowFlorent What's this? There should not be any type error, so lmk if this is causing issues for you.

nvie (approved):
💯

ofoucherot (approved):

packages/liveblocks-react/src/room.tsx
FlowFlorent, line 542:
```suggestion
```

FlowFlorent: FYI: I removed the comment to tidy up this PR

PR #1747, opened 2024-06-20
Cleans up all remaining empty `initialPresence={{}}`

nvie:
...also notices a few places were we're still using `<ClientSideSuspense>` with a function as children. All no longer needed, so a nice simplification of our examples. ✨🧹🧽✨


marcbouchenoire (approved):

CTNicholas (approved):
Great, thank you!

PR #1749, opened 2024-06-20
Re-enable CI for codemod test suite

nvie:
Re-enable CI tests for `liveblocks-codemod`. I found a lil' hack/workaround. Still not sure why those tests are failing on CI but not locally, but this seems to work.

Fixes LB-897.


marcbouchenoire (approved):
🧙🪄

PR #1753, opened 2024-06-20
useDeleteThread throws client side if user not creator

FlowFlorent:
### Description

useDeleteThread throws client side if user not creator

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -1405,9 +1405,7 @@ function useDeleteThread(): (threadId: string) => void {
 
       const userId = getCurrentUserId(room);
 
-      if (
-        thread?.comments?.[0]?.userId !== userId
-      ) {
+      if (thread?.comments?.[0]?.userId !== userId) {
         throw new Error("Only the thread creator can delete the thread");
       }
 
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -1400,6 +1400,20 @@ function useDeleteThread(): (threadId: string) => void {
       const optimisticUpdateId = nanoid();
 
       const { store, onMutationFailure } = getExtrasForClient(client);
+
+      const thread = store.get().threads[threadId];
+
+      const userId = getCurrentUserId(room);
+
+      if (
+        thread &&
+        thread.comments &&
+        thread.comments[0] &&
+        thread.comments[0].userId !== userId
+      ) {
+        throw new Error("Only the thread creator can delete the thread");
+      }
+
       store.pushOptimisticUpdate({
         type: "delete-thread",
         id: optimisticUpdateId,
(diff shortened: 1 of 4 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -1400,6 +1400,20 @@ function useDeleteThread(): (threadId: string) => void {
...
+
+      const thread = store.get().threads[threadId];
+
+      const userId = getCurrentUserId(room);
+
+      if (
+        thread &&
+        thread.comments &&
+        thread.comments[0] &&
+        thread.comments[0].userId !== userId
```
nvie, line 1412:
```suggestion
        thread?.comments?.[0]?.userId !== userId
```

nvie (approved):
Nice, good DX change. Better to not even make the optimistic update if we know pessimism will win here.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1400,6 +1400,20 @@ function useDeleteThread(): (threadId: string) => void {
...
+      ) {
+        throw new Error("Only the thread creator can delete the thread");
```
nimeshnayaju, line 1414: I am trying to recall if the public API endpoint also returns deleted comments to ensure that the first comment under a thread would always come from the thread owner.

nvie: I think it isn’t if user A creates the thread, then user B leaves a comment, and user A delete their initial thread‘s comment. Now user B’s comment will be the new thread‘s “root”, so I guess it makes sense to declare user B the new thread owner, right?

nvie: Owned by inheritance.

nimeshnayaju: As long as the backend is also consistent with what we do in the frontend. Otherwise, we will be preventing thread deletion too eagerly.

nimeshnayaju: It looks like deleted comments are returned in the API response. So, in your example @nvie, User A is still the owner of the thread. Backend also seems to think the same.

nimeshnayaju: @FlowFlorent We also need to make sure when a comment is deleted, the optimistic update only soft deletes the comment just to be sure? The thread  should, nonetheless be replaced very quickly with the most recent value after receiving a websocket event.

nvie: Agreed, being consistent with the backend is a must.

FlowFlorent: We soft delete comments through the client API so the owner of the thread will always be the original thread creator.

There's one area where this changes, the REST API endpoint to delete a comment hard deletes the comment, meaning that the owner could change by inheritance. Wonder if we should set it as a soft delete instead 🤔

FlowFlorent: The optimistic update indeed soft deletes the comment, so we should be good! Thanks for raising the point, I forgot to check that!
```ts
  const updatedComments = thread.comments.map((comment) =>
    comment.id === commentId
      ? {
          ...comment,
          deletedAt,
          body: undefined,
        }
      : comment
  );
```

PR #1760, opened 2024-06-24
Expose `isStorageReady` on mutation context

nvie:
This fixes a DX issue with `useMutation`.

## Problem

Currently, it's annoying to unpack `{ storage }` in a `useMutation(({ storage }) => ..., [])` call because it will throw immediately upon unpacking when Storage hasn't loaded at that point in time yet, leading to:

    Error: This mutation cannot be used until Storage has loaded

This could be happening if there is a button in the UI that is clicked before Storage has loaded, or if the mutation function gets called in a `useEffect(() => {}, [])` handler, for example.

The problem is that it's currently annoying and unclear to users how to properly avoid this error. For example, given this mutation…

```ts
const myMutation = useMutation(({ storage }) => {
  storage.doSomething();
}, []);
```

…really the only safe way around this is quite annoying:

```ts
// Ugh 😒
const room = useRoom();
const myMutation = useMutation((context) => {
  if (room.getStorageSnapshot() === null) {
    // Storage not ready yet, ignore
    return;
  }
  // Only unpack `storage` after it's loaded
  const { storage } = context;
  storage.doSomething();
}, [room]);
```

## ~Proposed Solution~ _We haven't decided yet!_

This PR makes two (backward-compatible) changes to this API, so the example above can now safely be written as:

```ts
// ✅ After this PR
const myMutation = useMutation(({ storage, isStorageReady }) => {
  // If storage not ready yet, ignore
  if (!isStorageReady) return;

  storage.doSomething();
}, []);
```

1. It no longer throws upon _unpacking_ `storage` (so we can keep all our examples the same), but yet it will throw when _accessing_ methods on the unpacked `storage`. Until storage has actually loaded, this object will be a stub that will throw errors when used.
1. It adds a `isStorageReady` boolean to the mutation context. You can check this boolean to determine if Storage has been loaded, which you can use to exit early. In the use cases of the early button click or `useEffect`, you can now conditionally allow it, rather than getting the error.

### DIFF
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -2058,14 +2058,14 @@ Both are equally fine, just a matter of preference.
 If Storage is modified inside a mutation before it has been initially loaded, an
 error will occur. It’s recommended to not call your mutation functions until
 Storage has loaded. If that is impractical, you can detect this situation using
-the `isStorageLoaded` boolean.
+the `isStorageReady` boolean.
 
 ```tsx
 const deleteCircle = useMutation(
   // Mutation context is passed as the first argument
-  ({ storage, isStorageLoaded }) => {
+  ({ storage, isStorageReady }) => {
     // No-op if storage has already loaded
-    if (!isStorageLoaded) return;
+    if (!isStorageReady) return;
 
     // Here, `storage` is guaranteed to be available
     storage.get("shapes").get("circle1").delete();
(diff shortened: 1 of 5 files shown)


CTNicholas (approved):
Great improvement, love it!

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -2053,6 +2053,27 @@ options:
 
 Both are equally fine, just a matter of preference.
 
+#### Waiting for Storage to load
```
nvie, line 2056: @CTNicholas Do you prefer to move this subsection down? If so, where do you prefer it?

CTNicholas: Need to add props tables for all the React hooks when I have a chance.

nvie:
Closing this PR for now. We haven't found a solution that generally works and that everyone liked yet for this. We will keep debating this change.

PR #1765, opened 2024-06-24
Re-mention new hook in 2.1 changelog

nvie:
Last week, 2.0.4 was released which introduced a new hook. New features should land in a minor version, instead of in patch-level releases. We'll leave it be for now, but let's at least re-mention the addition of this hook in the 2.1 changelog, so people that skip reading all the patch level changelog entries will still see this one.


marcbouchenoire (approved):
👌

PR #1768, opened 2024-06-24
Improve Suspense internals

nvie:
Adds a polyfill for React 19's `use`, and uses this construct now to implement our suspending internals. Behaviorally this is the same as what we're doing currently, but it will make our implementation more recognizable as a pattern and ultimately simplify the code a bit, because the throwing-a-promise behavior is better abstracted. By adopting the `use()` abstraction, we will end up with code that's simpler and easier to reason about.

Currently done for:

- `useSuspendUntilPresenceReady()`
- `useSuspendUntilStorageReady()`

Fixes LB-903.


nvie:
As discussed on our call yesterday, we're merging this incremental improvement in now, but will come back to rewriting/fixing these cases in a later version:

001e90c1-a9ee-46d5-aa15-79fe45a905de

marcbouchenoire (approved):

PR #1770, opened 2024-06-25
Add script to lint all package exports

nvie:
Checks the following things in `@liveblocks/react`:

Implemented:

- [x] That all exports from `./index` have a matching `./suspense`
- [x] That all exports have a doc string
- [x] Allow `@private` exports _not_ to have a twin export

There are more checks I'd like to add later, but I'm going to track those separately in LB-945.

Fixes LB-669.


packages/liveblocks-react/scripts/check-exports.ts
```diff
@@ -0,0 +1,278 @@
...
+import { relative } from "path";
+import { Project, Node, SyntaxKind } from "ts-morph";
+import { sorted } from "itertools";
+
+// Configuration
+const ALLOW_NO_JSDOCS = ["MutationContext", "UseThreadsOptions"];
+
+// Add any hooks here that are allowed to have a different doc string between
+// their classic and suspense versions! Because... they're actually different!
+// TODO Expand to still enforce a certain similarity percentage maybe later?
```
marcbouchenoire, line 12: 😄

marcbouchenoire (approved):
Nice!

PR #1771, opened 2024-06-25
Use `actions/upload-artifact@v4`

marcbouchenoire:
We currently use `actions/upload-artifact@v2` but `v1`, `v2`, and `v3` are all [deprecated](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/).

Given our simple usage of it, we're not affected by the breaking changes of `v3` and `v4`.


nvie (approved):
Nice, thanks for noticing / upgrading! 🙏

PR #1772, opened 2024-06-25
Add missing JSDocs

nvie:
Add back all the missing JSDocs that got lost in the 2.0 refactoring.

This PR copies them all back in. Note that currently they are duplicated, because there is no way to "share" them in TypeScript (for good reason). However, to avoid a massive maintenance burden, I'm working on a script to check that all of these exports (and their JSDoc strings) are in sync in all their various forms, so that CI will fail as a reminder if we update one but forget to publish a suspense version.

In fact, I've used that WIP script to find all of these, and ensure they're correct. However, I want to get this fix out for 2.1 before publishing the script, because that may take a bit longer to finish completely.

Fixes LB-730.

### DIFF
--- a/packages/liveblocks-react/src/comments/errors.ts
+++ b/packages/liveblocks-react/src/comments/errors.ts
@@ -1,5 +1,8 @@
 import type { BaseMetadata, CommentBody, Patchable } from "@liveblocks/core";
 
+/**
+ * @private Internal API, do not rely on it.
+ */
 export class CreateThreadError<M extends BaseMetadata> extends Error {
   constructor(
     public cause: Error,
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -2710,8 +2710,13 @@ const _useEditThreadMetadata: TypedBundle["useEditThreadMetadata"] =
  * useEventListener is a React hook that allows you to respond to events broadcast
  * by other users in the room.
  *
+ * The `user` argument will indicate which `User` instance sent the message.
+ * This will be equal to one of the others in the room, but it can be `null`
+ * in case this event was broadcasted from the server.
+ *
  * @example
- * useEventListener(({ connectionId, event }) => {
+ * useEventListener(({ event, user, connectionId }) => {
+ * //                         ^^^^ Will be Client A
  *   if (event.type === "CUSTOM_EVENT") {
  *     // Do something
  *   }
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -2815,6 +2815,7 @@ const _useThreadsSuspense: TypedBundle["suspense"]["useThreads"] =
  */
 const _useOther: TypedBundle["useOther"] = useOther;
 
+// TODO This one is tricky, as it has overloads
 const _useOthers: TypedBundle["useOthers"] = useOthers;
 
 /**
@@ -2828,6 +2829,7 @@ const _useOthers: TypedBundle["useOthers"] = useOthers;
  */
 const _useOtherSuspense: TypedBundle["suspense"]["useOther"] = useOtherSuspense;
 
+// TODO This one is tricky, as it has overloads
 const _useOthersSuspense: TypedBundle["suspense"]["useOthers"] =
   useOthersSuspense;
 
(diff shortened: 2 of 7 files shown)


packages/liveblocks-react/src/comments/errors.ts
```diff
@@ -1,5 +1,8 @@
 import type { BaseMetadata, CommentBody, Patchable } from "@liveblocks/core";
 
+/**
+ * @private Internal API, do not rely on it.
+ */
```
marcbouchenoire, line 5: Is this really private? Can't it be used with `instanceof` and `useErrorListener`? (I'm not sure where we left off with error handling for Comments/Notifications)

CTNicholas (approved):
Looks good! Thanks for fixing 🙌

packages/liveblocks-react/src/room.tsx
```diff
@@ -2402,48 +2583,326 @@ export function generateQueryKey(
...
+ */
+const _useAddReaction: TypedBundle["useAddReaction"] = useAddReaction;
+
+/**
+ * Create a callback function that lets you mutate Liveblocks state.
+ *
+ * The first argument that gets passed into your callback will be
+ * a "mutation context", which exposes the following:
+ *
+ *   - `root` - The mutable Storage root.
```
CTNicholas, line 2640: I think this is `storage` not `root`. It's mentioned a couple of other times in this JSDoc comment too.

packages/liveblocks-react/src/room.tsx
```diff
@@ -2402,48 +2583,326 @@ export function generateQueryKey(
...
-const _useOthersMapped: DefaultRoomContextBundle["useOthersMapped"] =
-  useOthersMapped;
-const _useOthersMappedSuspense: DefaultRoomContextBundle["suspense"]["useOthersMapped"] =
+
+/**
+ * useEventListener is a React hook that allows you to respond to events broadcast
+ * by other users in the room.
+ *
+ * @example
+ * useEventListener(({ connectionId, event }) => {
```
CTNicholas, line 2715: Should we mention `user` too? 

Also it's `null` when broadcast from server, we could make a note: https://liveblocks.io/docs/api-reference/liveblocks-react#useEventListener

packages/liveblocks-react/src/comments/errors.ts
nvie: Maybe it's not intended to be private, but in that case, we should publish and document all of these:

https://github.com/liveblocks/liveblocks/blob/9029869472aba1e325584afb41ec6d86599fb2f4/packages/liveblocks-react/src/comments/errors.ts#L147-L154

Currently, only `CreateThreadError` is exported here, and given that it's in a separate re-export together with `useCommentsErrorListener`, which is also private, I figured it's at least supposed to be private _for now_.

packages/liveblocks-react/src/room.tsx
nvie: Ah, good catch. These comments are all blindly copy-pasted from their equals, so I didn't read them through during the process. Fixed in ba88ab8bdfddf48080af0186ac00325ce76762fe 👍

packages/liveblocks-react/src/room.tsx
nvie: Good point, fixed in a94d52e4f0611ea8f87293bd683d132ad520af2e 👍

packages/liveblocks-react/src/comments/errors.ts
marcbouchenoire: Ah yep, looks like it's a case of needing it in `react-lexical` again so `@private` indeed for now.

marcbouchenoire (approved):

packages/liveblocks-react/src/room.tsx
```diff
@@ -2402,48 +2583,326 @@ export function generateQueryKey(
...
+ * component re-render if properties for this particular user change.
+ *
+ * @example
+ * // Returns only the selected values re-renders whenever that selection changes)
+ * const { x, y } = useOther(2, user => user.presence.cursor);
+ */
+const _useOther: TypedBundle["useOther"] = useOther;
+
+// TODO This one is tricky, as it has overloads
+const _useOthers: TypedBundle["useOthers"] = useOthers;
```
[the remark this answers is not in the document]
nvie: I'll keep digging and try to add these in a follow-up, but thus far I tried many options, but haven't found a satisfactory solution yet. Maybe we'll have to eventually de-DRY this entire function and accept the duplication.

PR #1774, opened 2024-06-26
Fix `useSyncExternalStore` usage in `@liveblocks/react-ui`

marcbouchenoire:
This PR fixes a case where `useSyncExternalStore` was imported directly from `react` instead of its shim.
While I was there I also improved two things:
- improve how our Rollup configs list external imports (it currently does exact matching so it complains about subpaths)
- create a set of ESLint errors for React >=18 hooks whenever we have an alternative

### DIFF
--- a/packages/liveblocks-react-ui/.eslintrc.js
+++ b/packages/liveblocks-react-ui/.eslintrc.js
@@ -31,13 +31,13 @@ module.exports = {
         selector:
           "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='useId']",
         message:
-          "useId is only available on React >=18. Import it from 'src/utils/use-id' instead.",
+          "useId is only available on React >=18. Import it from '/src/utils/use-id' instead.",
       },
       {
         selector:
           "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='useTransition']",
         message:
-          "useTransition is only available on React >=18. Import it from 'src/utils/use-transition' instead.",
+          "useTransition is only available on React >=18. Import it from '/src/utils/use-transition' instead.",
       },
     ],
 
--- a/packages/liveblocks-react-ui/.eslintrc.js
+++ b/packages/liveblocks-react-ui/.eslintrc.js
@@ -23,6 +23,11 @@ module.exports = {
       ...commonRestrictedSyntax,
       {
         selector:
+          "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='use']",
+        message: "use is only available on React >=19.",
+      },
+      {
+        selector:
           "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='useSyncExternalStore']",
         message:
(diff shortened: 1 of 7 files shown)


packages/liveblocks-react-ui/.eslintrc.js
```diff
@@ -18,7 +18,28 @@ module.exports = {
...
+          "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='useId']",
+        message:
+          "useId is only available on React >=18. Import it from '/src/utils/use-id' instead.",
+      },
+      {
+        selector:
+          "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='useTransition']",
+        message:
+          "useTransition is only available on React >=18. Import it from '/src/utils/use-transition' instead.",
+      },
```
nvie, line 41: Shall we add `use` here too, while at it?

packages/liveblocks-react/.eslintrc.js
```diff
@@ -8,7 +8,23 @@ module.exports = {
...
+      //     "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='use']",
+      //   message:
+      //     "use is only available on React >=19. Import it from '/src/lib/use-polyfill' instead.",
+      // },
+      {
+        selector:
+          "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='useSyncExternalStore']",
+        message:
+          "useSyncExternalStore is only available on React >=18. Import it from 'use-sync-external-store/shim/index.js' instead.",
+      },
```
nvie, line 26: Should these be the same as in the other `.eslintrc` file?

packages/liveblocks-react-ui/.eslintrc.js
marcbouchenoire: Since we don't have a polyfill usable in `react-ui` for it yet I wouldn't know what to recommend to fix the error?

packages/liveblocks-react/.eslintrc.js
marcbouchenoire: For `useTransition` and `useId` we have polyfills only in `react-ui` not here because we don't use them, I guess ultimately it's the same question as above with `use`, should we error out even if there's no alternative way (yet)? Maybe that's wiser, since it would prevent all cases, not just ones that have been dealt with before and led to a polyfill.

packages/liveblocks-react-ui/.eslintrc.js
marcbouchenoire: Updated in https://github.com/liveblocks/liveblocks/pull/1774/commits/e99a0e3e2e5738c8bb570d491282c436582233d7.

packages/liveblocks-react/.eslintrc.js
marcbouchenoire: Updated in https://github.com/liveblocks/liveblocks/pull/1774/commits/e99a0e3e2e5738c8bb570d491282c436582233d7.

PR #1775, opened 2024-06-26
Add missing JSDocs for `useOthers` and `useSelf`

nvie:
Unfortunately, I had to de-DRY (duplicate) the function signatures to get this to work with the overloading, but so be it. The DX is more important than our maintenance burden.

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -2820,8 +2820,8 @@ const _useThreadsSuspense: TypedBundle["suspense"]["useThreads"] =
 const _useOther: TypedBundle["useOther"] = useOther;
 
 /**
- * Returns an object that lets you get information about all the users
- * currently connected in the room.
+ * Returns an array with information about all the users currently connected in
+ * the room (except yourself).
  *
  * @example
  * const others = useOthers();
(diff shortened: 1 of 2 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -2819,8 +2819,54 @@ const _useThreadsSuspense: TypedBundle["suspense"]["useThreads"] =
...
+ * const cursors = useOthers(users => users.map(u => u.presence.cursor), shallow);
+ * const someoneIsTyping = useOthers(users => users.some(u => u.presence.isTyping));
+ *
+ */
+function _useOthers<T>(
+  selector: (others: readonly User<DP, DU>[]) => T,
+  isEqual?: (prev: T, curr: T) => boolean
+): T;
+function _useOthers(...args: any[]) {
+  return useOthers(...(args as []));
```
nvie, line 2868: These `any`s won't show up in the final `*.d.ts` outputs. There just here so we don't have to keep this signature in sync too.

packages/liveblocks-react/src/room.tsx
```diff
@@ -2819,8 +2819,54 @@ const _useThreadsSuspense: TypedBundle["suspense"]["useThreads"] =
  */
 const _useOther: TypedBundle["useOther"] = useOther;
 
-// TODO This one is tricky, as it has overloads
-const _useOthers: TypedBundle["useOthers"] = useOthers;
+/**
+ * Returns an object that lets you get information about all the users
```
marcbouchenoire, line 2823: Was it written like this before? I find it a complicated way of describing an array.

nvie: Ha, great point! Fixed in 203f5a584192934227c78862bc764f0a99ff7148. For context, Back In The Days™, this used to be an object that "proxied" an array (was array-like). But in 0.18 we changed it to an actual array.

CTNicholas (approved):
Looks good!

PR #1778, opened 2024-06-27
Decouple publishing packages from publishing CLIs

nvie:
Splits publishing `create-liveblocks-app` and `@liveblocks/codemod` off from the rest of our public packages. Our public packages need to be strictly versioned together, but making a patch to our codemod script is not something that requires a bump to the rest of the public packages.

This PR splits those up. I've moved these workspaces into a new top-level `tools/*` dir, a sibling to `packages/*`. This will hopefully make it clearer they are on a separate versioning scheme. They are not dependencies that users will pin as dependencies, but just run as CLI commands.

You always want to be running the latest versions of these CLI tools (i.e. `npx create-liveblocks-app@latest`), so I've settled for a versioning scheme of `2.<YYYYMMDD>.0` (the last patch version can be incremented if you need to publish more than once on a single day). There is no semantic versioning here. But if ever we really must, we can by bumping the 2 to 3. I'm starting with `2` as a major here, because `2.1.0` is already out there, and `2.20240627.0` is "higher".

Fixes LB-869.


marcbouchenoire (approved):

nvie:
Nice, thanks for all the additions, @marcbouchenoire 🙌 !

PR #1780, opened 2024-06-27
Add support for custom links to Comments composer

marcbouchenoire:
This PR adds support for custom links to the Comments composer, we currently only support "auto" links (turning any valid URL into a link).

As a V1, there's no edition UI (e.g. a floating toolbar that lets you update a link and its destination) yet, but there are two ways to create them:
- select some plain text (no rich text, no mentions, etc) and paste a valid URL
- paste any text from somewhere else, links within the pasted contents will be kept accordingly (as "auto" or custom links depending on how they're constructed)

Working on the changes above also led to more general improvements:
- links cannot have rich text in them given how they're serialized in the `CommentBody` format, and yet it was allowed in the composer (and you would see something different when posting your comment). This is now prevented so the composer will always match the output after posting.
- to support pasting text with links in it, we had to support pasting any HTML. So we now support pasting rich text generally, be it keeping links from a Slack message or the code blocks from a GitHub description.

Fixes LB-905

https://github.com/liveblocks/liveblocks/assets/6959425/b6a12df6-ef9e-40bf-99f2-ffec30edb396


packages/liveblocks-react-ui/src/primitives/Comment/index.tsx
```diff
@@ -111,7 +111,7 @@ const CommentBody = forwardRef<HTMLDivElement, CommentBodyProps>(
 
                       return (
                         <Link href={href} key={index}>
-                          {inline.url}
+                          {inline.text ?? inline.url}
```
nvie, line 114: Can `inline.text` ever be `""`? If so, maybe in this case, we should use `||` instead of `??` (which I'd normally always prefer).

nvie (approved):
Nice, this will be so nice to have!

One thing I noticed is the similarity between these data types:

https://github.com/liveblocks/liveblocks/blob/65cd3a832fdf29d9c8f6ad4b9733fa74ba795cd0/packages/liveblocks-react-ui/src/types.ts#L32-L42

Should the data type here not be "link", since both are links? The only difference between them is how they are created by the end user: either by typing or by explicitly creating them (by pasting or whatever UI we support for it later). But should we distinguish them at the data structure level?

packages/liveblocks-react-ui/src/primitives/Comment/index.tsx
marcbouchenoire: No that would mean a 0-characters link and Slate's normalization process prevents empty nodes from existing (they either get merged with compatible neighbors or get removed).

packages/liveblocks-react-ui/src/primitives/Composer/utils.ts
```diff
@@ -54,45 +66,60 @@ export function commentBodyMentionToComposerBodyMention(
...
+              return composerBodyCustomLinkToCommentBodyLink(inline);
+            }
 
-      return {
-        ...block,
-        children,
-      };
-    }),
+            if (isText(inline)) {
+              return inline as CommentBodyText;
```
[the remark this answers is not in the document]
marcbouchenoire: Yep! We also need to remove the unused import so I'll commit directly.

marcbouchenoire:
> But should we distinguish them at the data structure level?

@nvie This (`ComposerBody`) is the client-side data structure (used by Slate.js), not the one we send, store, and (de)serialize (`CommentBody`). In `CommentBody` we do use a single `"link"` type with an optional `text` property to distinguish them, but in the editor they need to be 2 different nodes because they have different behaviors (e.g. auto links should be created/broken based on how you type or "untype" them, not custom ones)

PR #1783, opened 2024-06-27
Implement check that matches factory exports from top-level exports

nvie:
Adds check that ensures that any top-level export we add, must also…

✔️ exist in `./suspense` subpath export  
✔️ exist in factory's return value  
✔️ exist in factory's `{ suspense }` key

For example, just adding this to `src/index.ts`:

Will lead to these errors:

Screen Shot 2024-06-28 at 00 41 41@2x

### DIFF
--- a/packages/liveblocks-react/scripts/check-exports.ts
+++ b/packages/liveblocks-react/scripts/check-exports.ts
@@ -19,6 +19,19 @@ const ALLOW_DIFFERENT_JSDOCS = [
   "useUser",
 ];
 
+// These exports may exist at the top-level without a factory equivalent
+const ALLOW_NO_FACTORY = [
+  "ClientSideSuspense",
+  "createLiveblocksContext", // the factories themselves, obviously
+  "createRoomContext", // the factories themselves, obviously
+  "Json",
+  "JsonObject",
+  "MutationContext",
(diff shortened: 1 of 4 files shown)


packages/liveblocks-react/scripts/check-exports.ts
```diff
@@ -19,6 +19,19 @@ const ALLOW_DIFFERENT_JSDOCS = [
...
+const ALLOW_NO_FACTORY = [
+  "ClientSideSuspense",
+  "createLiveblocksContext", // the factories themselves, obviously
+  "createRoomContext", // the factories themselves, obviously
+  "Json",
+  "JsonObject",
+  "MutationContext",
+  "shallow",
+  "UseThreadsOptions",
+  "useClient", // XXX Should we export this from the factory too, for consistency? 🤔
```
nvie, line 32: Maybe this is actually a valid complaint, will look into it tomorrow. Maybe it's nice to also export this new hook from the factory!

nvie: Fixed in 9073bffd2493cff90bde1c903f98ac8a1a10498d.

marcbouchenoire (approved):
🤝

PR #1784, opened 2024-06-28
LB-899 Update useEditorStatus

jrowny:
Store pending updates/clocks for a given client so we can provide an ack/synchronize status. 
Update useEditorStatus to use the new status event and getStatus function


packages/liveblocks-yjs/src/index.ts
```diff
@@ -24,6 +25,13 @@ const DefaultOptions: ProviderOptions = {
   autoloadSubdocs: false,
 };
 
+enum SyncStatus {
+  Loading = "loading",
+  NotLoaded = "not-loaded",
```
nvie, line 30: Minor nit, but I prefer to flip these to make them match the order of state transitions over time:

```suggestion
  NotLoaded = "not-loaded",
  Loading = "loading",
```

Note that everywhere else we use `not-loaded` as the initial state, think NotYetEvenStartedToLoad.

packages/liveblocks-yjs/src/index.ts
```diff
@@ -173,6 +214,18 @@ export class LiveblocksYjsProvider<
...
+      return SyncStatus.NotLoaded;
+    }
+    if (!this.synced) {
+      return SyncStatus.Loading;
+    }
+    return this.pending.length === 0
+      ? SyncStatus.Synchronized
+      : SyncStatus.Synchronizing;
+  }
+
```
nvie, line 228: Let's make sure we follow the same convention as we do with `getStorageStatus()` here.

This is the goal behavior as a state diagram:

```mermaid
flowchart LR
    Not-Loaded --> Loading --> Synchronizing --> Synchronized --> Synchronizing
```

Not 100% clear on how this works internally, but I assume this can be in `Not-Loaded` state only during the constructor, but since at the end of the constructor the synchronization seems to start due to [the `this.syncDoc()` call](https://github.com/liveblocks/liveblocks/blob/c93c119cc19627d900b1bd964bcb39076e67e360/packages/liveblocks-yjs/src/index.ts#L133) that should already immediately put this status into the `Loading` state, so `Not-Loaded` should not even be observable for this Yjs provider I think. Therefore I think you can safely remove the `roomConnected` boolean.

So how to decide between `Loading` (initial sync needs to happen) and `Synchronizing`? Could we use something like this?

```suggestion
  public getStatus(): SyncStatus {
    if (this.synced) {
      return SyncStatus.Synchronized;
    }
    
    if (hasEverBeenSynched) {
      return SyncStatus.Synchronizing;
    } else {
      return SyncStatus.Loading;
    }
  }

```

I'm even wondering if you need to do the explicit `pending` tracking here. But maybe I'm missing something here! It looks to me like you could simply set `hasEverBeenSynched = true` when the initial (any, really) sync event happens.

jrowny: We do need to track updates because we currently have no acknowledgement from the server that an update was applied to the source of truth. Most real world issues are pretty well masked by the fact that a full sync happens on reconnect, but if you send off some updates while your internet is farting and close your browser, then those updates will be lost.

nvie: Oh I see. Fair point. Well, in that case, I think you just need to change the condition of this example to use the `pending` array, I think?

jrowny: What if the room gets disconnected or subdoc is loading? Should that be considered syncronizing?

nvie: Subdocs I'm not sure about, but when the room gets disconnected and yet the client would have pending changes, the status should still be "synchronizing". To me "synchronizing" means that the client and server are no longer in sync at the moment. The client has changes that the server must be made aware of.

nvie: This is also how Storage does it.

nvie: Regarding subdocs, my instinct is to try to treat this the same way. So if the main doc, or _any_ of its subdocs have known pending changes, then the status would be synchronizing. Only if the main doc, and all subdocs, got a synched event, then the entire status should be synchronized. So that the meaning of synchronized means that the client has no changes that need to be submitted.

Another way to think of it is that it's safe to close your browser tab without losing data.

packages/liveblocks-yjs/src/index.ts
jrowny: okay, I removed not-loaded from yjs provider because it will never be in that state if the provider exists, then the room exists and it will be loading. The hooks that implement this, i.e. getEditorStatus do implement not-loaded.

packages/liveblocks-yjs/src/index.ts
jrowny: I removed roomConnected and not-loaded, but I want to keep loading intact for when the 2-stage sync happens because some editors (like lexical) rely on waiting for sync to complete to show the editor. If we used "synchronizing" for that, then the editor would constantly be flickering in/out, so I think having an explicit loading during initial sync makes more sense. It's a common use case for Yjs editors to wait until sync to allow editing because even though it's technically "okay" to edit during that stage, it looks/feels weird as content pops in.

nvie: > I think having an explicit loading during initial sync makes more sense

Absolutely – agreed!

packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
```diff
@@ -70,10 +70,12 @@ export function useEditorStatus(): EditorStatus {
...
   const subscribe = useCallback(
     (onStoreChange: () => void) => {
-      if (provider === undefined) return () => {};
+      if (provider === undefined) return () => { };
+      provider.on("status", onStoreChange);
       provider.on("sync", onStoreChange);
       return () => {
         provider.off("sync", onStoreChange);
+        provider.off("status", onStoreChange);
       };
```
nimeshnayaju, line 79: Is it okay to remove the subscription to `provider.on("sync", onStorageChange)` now that `provider.on("status", onStorageChange)` does the same thing and more?

jrowny: yea, probably

packages/liveblocks-yjs/src/index.ts
```diff
@@ -133,10 +161,19 @@ export class LiveblocksYjsProvider<
     }
   };
 
-  private updateDoc = (update: string, guid?: string) => {
+  private getUniqueUpdateId = (update: Uint8Array) => {
+    const clock =
+      Y.parseUpdateMeta(update).to.get(this.rootDoc.clientID) ?? "-1";
+    return this.rootDoc.clientID + ":" + clock;
+  };
```
nimeshnayaju, line 168: Is the idea behind this to extract the `clockId` behind the update and appending it to the Y.Doc's client ID? Not suggesting but just curious if the appending to the client ID necessary?

nimeshnayaju (approved):
Looks very good! This will bring more accurate editor synchronization statuses to our Y.js provider.

Do you know how much effort it would be to write minimal test to verify editor status is 
correctly determined after this change?

packages/liveblocks-yjs/src/index.ts
jrowny: yes, it's just a way to identify the update. Otherwise we would need to generate a UUID or nanoid with each update and return it from the server.

PR #1786, opened 2024-06-28
Fix and improve `release.sh`

marcbouchenoire:
This PR:
- bumps DevTools to 1.12.0 manually
- makes the DevTools publishing workflow push its commit
- improves bump commits (the packages remain as "Bump to 2.2.0" but all other bumps now specify their source: e.g. "Bump DevTools to 1.12.0")

### DIFF
--- a/.github/scripts/release.sh
+++ b/.github/scripts/release.sh
@@ -8,19 +8,22 @@ err () {
 }
 
 usage () {
-    err "usage: release.sh [-h] [-V <version>] <pkgdir> [<pkgdir>...]"
+    err "usage: release.sh [-h] [-V <version>] [-p] <pkgdir> [<pkgdir>...]"
     err
     err "Prepare a release by updating files in this repo."
-    err "Run this prior to publishing to NPM."
+    err "Run this prior to publishing to NPM (or elsewhere, e.g. DevTools)."
     err ""
     err "    -V   the new NPM version"
+    err "    -p   push the bump commit (no need if publish.sh is called afterwards)"
     err ""
 }
 
 VERSION=
-while getopts V:h flag; do
+PUSH_COMMIT="false"
+while getopts V:p:h flag; do
     case "$flag" in
         V) VERSION=$OPTARG;;
+        p) PUSH_COMMIT="true";;
         *) usage; exit 2;;
     esac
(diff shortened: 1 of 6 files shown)


.github/scripts/release.sh
```diff
@@ -8,19 +8,25 @@ err () {
...
+    err "    -p   push the bump commit (no need if publish.sh is called afterwards)"
+    err "    -m   improve the commit message by specifying what was bumped"
     err ""
 }
 
 VERSION=
-while getopts V:h flag; do
+PUSH_COMMIT="false"
+COMMIT_MESSAGE="Bump to "
+while getopts V:p:h flag; do
```
nvie, line 25: You'll have to add `m:` here. `m` being the flag, and `:` meaning that the flag takes an argument, i.e. take the next command line argument and put it in `$OPTARG`.

```suggestion
while getopts V:p:m:h flag; do
```

Without this, you'll get a `unknown flag -m` like error.

.github/workflows/publish-cli-tools.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         env:
           VERSION: ${{steps.version.outputs.value}}
         run:
-          ./.github/scripts/release.sh -V "$VERSION"
+          ./.github/scripts/release.sh -V "$VERSION" -m "Bump tools to "
```
nvie, line 79:
```suggestion
          ./.github/scripts/release.sh -V "$VERSION" -m "Bump CLI tools to "
```

.github/workflows/publish-devtools.yml
```diff
@@ -74,8 +74,8 @@ jobs:
         env:
           STRIPPED_VERSION: ${{steps.version.outputs.value}}
         run:
-          ./.github/scripts/release.sh -V "$STRIPPED_VERSION"
-          "tools/liveblocks-devtools"
+          ./.github/scripts/release.sh -V "$STRIPPED_VERSION" -p -m "Bump
+          DevTools to " "tools/liveblocks-devtools"
```
nvie, line 78: Nice, thanks for fixing this!

nvie (approved):
Nice! 💯

PR #1791, opened 2024-06-28
Support/use resolved first-class property in React packages

marcbouchenoire:
→ LB-939

This PR handles the core and React parts of the new first-class `resolved` property:
- Update types
- Add 2 new hooks (`useMarkThreadAsResolved` and `useMarkThreadAsUnresolved`)
   - And all the necessary changes (new core methods, updates to the cache, etc)
- Switch the `Thread` default component to use and update the first-class property instead of metadata

The docs aren't fully done but I'd prefer to finish them in https://github.com/liveblocks/liveblocks/tree/resolved-first-party, as we now 3 different branches (including this one) with docs changes.


docs/references/v2.openapi.json
```diff
@@ -1368,7 +1368,7 @@
             },
             "in": "query",
             "name": "query",
-            "description": "Query to filter threads. You can filter by `metadata`, for example, `metadata[\"status\"]:\"open\" AND metadata[\"resolved\"]:true`. Learn more about [filtering threads with query language](/docs/guides/how-to-filter-threads-using-query-language)."
+            "description": "Query to filter threads. You can filter by `metadata` and `resolved`, for example, `metadata[\"status\"]:\"open\" AND metadata[\"resolved\"]:true AND resolved:true`. Learn more about [filtering threads with query language](/docs/guides/how-to-filter-threads-using-query-language)."
```
nvie, line 1371: I think this example is now a bit confusing, let's use another field as an example, or remove it:

```suggestion
            "description": "Query to filter threads. You can filter by `metadata` and `resolved`, for example, `metadata[\"status\"]:\"open\" AND metadata[\"color\"]:"red" AND resolved:true`. Learn more about [filtering threads with query language](/docs/guides/how-to-filter-threads-using-query-language)."
```

nvie (approved):
Lovely! 💯

docs/references/v2.openapi.json
marcbouchenoire: Definitely, I tried to do this everywhere but this one slipped through the cracks.

PR #1792, opened 2024-06-28
add mark thread as resolved/unresolved methods

ofoucherot:
### Description

Add methods to liveblocks-node to call endpoints to mark a thread as resolved/unresolved.

#### Related issue(s)

https://linear.app/liveblocks/issue/LB-940/client-node-add-new-methods-use-new-endpoint

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -1250,10 +1250,10 @@ export class Liveblocks {
 
   /**
    * Mark a thread as resolved.
-   * @param params.roomId The room ID to update the thread in.
-   * @param params.threadId The thread ID to update.
-   * @param params.data.userId The user ID of the user who updated the thread.
-   * @returns The updated thread.
+   * @param params.roomId the room ID of the thread.
+   * @param params.threadId The thread ID to mark as resolved.
+   * @param params.data.userId The user ID of the user who marked the thread as resolved.
+   * @returns The thread marked as resolved.
    */
   public async markThreadAsResolved(params: {
     roomId: string;
@@ -1261,7 +1261,7 @@ export class Liveblocks {
     data: {
       userId: string;
     };
-  }): Promise<M> {
+  }): Promise<ThreadData<M>> {
     const { roomId, threadId } = params;
 
     const res = await this.post(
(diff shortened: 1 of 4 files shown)


packages/liveblocks-node/src/client.ts
```diff
@@ -1246,6 +1248,64 @@ export class Liveblocks {
     }
   }
 
+  /**
+   * Mark a thread as resolved.
+   * @param params.roomId The room ID to update the thread in.
+   * @param params.threadId The thread ID to update.
+   * @param params.data.userId The user ID of the user who updated the thread.
+   * @returns The updated thread.
```
marcbouchenoire, line 1256:
```suggestion
   * @param params.roomId The room ID to mark the thread as resolved in.
   * @param params.threadId The thread ID to mark as resolved.
   * @param params.data.userId The user ID of the user who marked the thread as resolved.
   * @returns The thread marked as resolved.
```

packages/liveblocks-node/src/client.ts
```diff
@@ -1246,6 +1248,64 @@ export class Liveblocks {
...
+
+    return (await res.json()) as M;
+  }
+
+  /**
+   * Mark a thread as unresolved.
+   * @param params.roomId The room ID to update the thread in.
+   * @param params.threadId The thread ID to update.
+   * @param params.data.userId The user ID of the user who updated the thread.
+   * @returns The updated thread.
```
marcbouchenoire, line 1285:
```suggestion
   * @param params.roomId The room ID to mark the thread as unresolved in.
   * @param params.threadId The thread ID to mark as unresolved.
   * @param params.data.userId The user ID of the user who marked the thread as unresolved.
   * @returns The thread marked as unresolved.
```

marcbouchenoire (approved):
👍 I added a couple of suggestions about the JSDoc comments but feel free to disregard them if you prefer the current ones.

packages/liveblocks-node/src/client.ts
```diff
@@ -1246,6 +1248,64 @@ export class Liveblocks {
...
+      url`/v2/rooms/${roomId}/threads/${threadId}/mark-as-resolved`,
+      {}
+    );
+
+    if (!res.ok) {
+      const text = await res.text();
+      throw new LiveblocksError(res.status, text);
+    }
+
+    return (await res.json()) as M;
```
marcbouchenoire, line 1277: Oh I missed this, the endpoints don't return the thread's metadata no?

nvie: Oh yeah, that's indeed a bug, good catch 👍

nvie: Let's add type-level tests for this while we're fixing this, to avoid regressions. Here are the relevant tests for `editThreadMetadata`:

- When [using augmentation](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-node/test-d/augmentation.test-d.ts#L256-L312)
- When [not using augmentation](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-node/test-d/no-augmentation.test-d.ts#L207-L245)

@ofoucherot For context, in case you have never updated these type-level tests: the files inside `test-d/*` are static tests only. The code you write in there isn't actually (going to be) executed. However, you can add `expectType<Foo>(someExpression)` assertions in there, to ensure that calling `someExpression` will return a value of a specific type (in this case `Foo`). To run these, run `turbo run test:types` in any package.

ofoucherot: Edit thread metadata only returns the metadata. I'm fixing the two new endpoints' signature.

PR #1795, opened 2024-07-01
Add smooth option to `useStorageStatus` hook

nvie:
This PR adds a `{ smooth: true }` option to the `useStorageStatus()` hook, making it easier to build calm UIs. It delays the `"synchronized"` status update, but keeps all other status updates immediate. Exact timing need to be determined.

### TODO

@CTNicholas We'll need to decide on these questions.

- [x] We settled on alternative (1)
- [x] We settled on 1000ms
- [x] We settled on `smooth` for the final option name
- [x] We settled on not allowing to provide custom timing/control for now
- [ ] Write the docs
- [x] ~Do the same for `useEditorStatus` API (see LB-899)~ (moving to that ticket)

Fixes LB-947.


packages/liveblocks-react/src/room.tsx
```diff
@@ -100,10 +100,13 @@ import type {
   ThreadsState,
   ThreadsStateSuccess,
   ThreadSubscription,
+  UseStorageStatusOptions,
   UseThreadsOptions,
 } from "./types";
 import { useScrollToCommentOnLoadEffect } from "./use-scroll-to-comment-on-load-effect";
 
+const SMOOTH_DELAY = 1000;
```
nvie, line 108: Currently, the `"synchronized"` event will be delivered after 1 second. If you want to see it live, run the `e2e/next-sandbox` app (`turbo run dev`) and open [this page](http://localhost:3007/storage/list?room=e2e-storage-list).


https://github.com/liveblocks/liveblocks/assets/83844/7d770d5d-7843-40a7-a3c5-4722d4940c9f

e2e/next-sandbox/pages/storage/list.tsx
```diff
@@ -52,6 +53,9 @@ function Sandbox() {
   const items = useStorage((root) => root.items);
   const me = useSelf();
   const status = useStatus();
+  const storageStatus = useStorageStatus();
+  const smoothStorageStatus1 = useStorageStatus({ smooth: 1 });
+  const smoothStorageStatus2 = useStorageStatus({ smooth: 2 });
```
nvie, line 58: @CTNicholas Hey! Rather than having a single `{ smooth: true }`, I've added two different implementations that could both work. Would you mind giving these a try and decide which one is preferable?

nvie: https://github.com/liveblocks/liveblocks/assets/83844/59d0abbc-0ab9-406d-96f6-01f3a6db8702

packages/liveblocks-react/src/types.ts
```diff
@@ -31,6 +31,16 @@ import type {
...
 } from "@liveblocks/core";
 
+export type UseStorageStatusOptions = {
+  /**
+   * When setting smooth, the hook will not update immediately as status
+   * changes. This is because in typical applications, these states can change
+   * quickly between synchronizing and synchronized. If you use this hook to
+   * build a "Saving changes..." style UI, prefer setting `smooth: true`.
+   */
+  smooth?: 0 | 1 | 2;
```
nvie, line 41: To be clear: this is not the final API. Just adding these here for now to test two alternative possible implementatoins.

e2e/next-sandbox/pages/storage/list.tsx
CTNicholas: I definitely prefer option 2, for a couple of reasons.

- Smoother. We din't want to create a highly visible flicker.
- Liveblocks synchs so quickly it actually makes me mildly suspicious. The artificial delay is nice.

nvie: Agreed, I think I'd also prefer (1). Wanted to make sure we try them both out "in the wild" first, before committing to it.

Here is the plan:
- I will remove alternative (2) from this PR
- @CTNicholas We need to decide on the final duration before we hardcode it. It's currently hard-coded to 1 second. Should we keep it, or make it faster? Not slower I think?
- @CTNicholas Do you think allowing configuration of this duration makes sense? I.e. allow both `{ smooth: true }` (1 second), or `{ smooth: 600 }` if you want full control? I don't think I like it that much and would prefer picking a good opinionated default value—at least not until we get a request for it. Do you agree?

nvie: Removed alternative 2 in c52d99fd9254d2508d20db2dd320e6529873d750 👍

CTNicholas:
I like 1000ms as a good default. Part of me is tempted to go lower, but it's important it's not distracting, and Google Docs has it lots slower than this even.

I'm happy not offering an option until someone needs it, I think 1000ms will work for all scenarios I can think of.

I prefer `smooth` over `calm`.

CTNicholas (approved):

PR #1796, opened 2024-07-01
Start using `corepack` and upgrade to Turbo v2

nvie:
This PR switches to Turbo v2, following [this migration guide](https://turbo.build/blog/turbo-2-0#breaking-changes), and enables the top-level `packageManager` field, which also enables the use of [corepack](https://www.totaltypescript.com/how-to-use-corepack).

After this is merged, everyone should run `corepack enable && corepack enable npm` once on their machines.

Fixes LB-713.


marcbouchenoire (approved):
👮‍♂️

PR #1797, opened 2024-07-01
Fix logic bug in incremental cache updates

nvie:
Refactoring the internals (for LB-944), I spotted a small bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - The `useClient()` hook is now also available for users of
   `createRoomContext()` and/or `createLiveblocksContext()`
-- Fix bug where all subsequent fetches of inbox notification were not
+- Fix bug where all subsequent fetches of inbox notifications were not
   incremental.
 
 ## v2.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - The `useClient()` hook is now also available for users of
   `createRoomContext()` and/or `createLiveblocksContext()`
+- Fix bug where all subsequent fetches of inbox notification were not
+  incremental.
 
 ## v2.1.0
 
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -253,7 +253,7 @@ function makeExtrasForClient<U extends BaseUserMeta, M extends BaseMetadata>(
        */
       if (
         lastRequestedAt === undefined ||
-        lastRequestedAt > result.meta.requestedAt
+        lastRequestedAt < result.meta.requestedAt
       ) {
         lastRequestedAt = result.meta.requestedAt;
       }


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - The `useClient()` hook is now also available for users of
   `createRoomContext()` and/or `createLiveblocksContext()`
+- Fix bug where all subsequent fetches of inbox notification were not
```
marcbouchenoire, line 7:
```suggestion
- Fix bug where all subsequent fetches of inbox notifications were not
```

marcbouchenoire (approved):

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -253,7 +253,7 @@ function makeExtrasForClient<U extends BaseUserMeta, M extends BaseMetadata>(
        */
       if (
         lastRequestedAt === undefined ||
-        lastRequestedAt > result.meta.requestedAt
+        lastRequestedAt < result.meta.requestedAt
       ) {
         lastRequestedAt = result.meta.requestedAt;
       }
```
nimeshnayaju, line 259: I think the current code is fine? The comment, I believe may be incorrect 🤔. The `lastRequestedAt` value represents the timestamp we last requested information from the server. This code is meant to ensure that we take the smaller of the two `requestedAt` value so that we do not miss any updates in between.

nimeshnayaju: Do you think it would be better if we had a test to verify this?

nvie: Because this timestamp is also used as the “since” param for the next request, I assume it’s supposed to “move forward” over time. In the old logic, it would get set once, initially, and then never updated beyond that in subsequent requests. Basically making all future/ subsequent fetches use the same “since” param, not?

Maybe the purpose of this isn’t what I assumed?

nvie: > Do you think it would be better if we had a test to verify this?

Being able to change logic like this without breaking a test is indeed not ideal! Curious what that test would look like in this case.

nimeshnayaju: > Because this timestamp is also used as the “since” param for the next request, I assume it’s supposed to “move forward” over time. In the old logic, it would get set once, initially, and then never updated beyond that in subsequent requests. Basically making all future/ subsequent fetches use the same “since” param, not?

https://github.com/liveblocks/liveblocks/blob/70f31dada35da015fb51f50f375e8c42d58655a6/packages/liveblocks-react/src/liveblocks.tsx#L205-L208

It does get updated after each successful polling request.

nvie: Ohh, I did not notice that! Will re-read this part today with this new insight.

nvie:
Closing this PR for now. Will make this change part of #1809's holistic refactoring.

PR #1800, opened 2024-07-01
Fix broken `useTransition` fallback

marcbouchenoire:
We had a fallback in place for `useTransition` (like `useId`), but for some reason it wasn't the same signature as the official version so it was never going to work. 🤔

### DIFF
--- a/packages/liveblocks-react-ui/src/utils/use-transition.ts
+++ b/packages/liveblocks-react-ui/src/utils/use-transition.ts
@@ -1,4 +1,5 @@
-import React from "react";
+import type { TransitionFunction } from "react";
+import React, { useCallback } from "react";
 
 // Prevent bundlers from importing `useTransition` directly
 // See https://github.com/radix-ui/primitives/pull/1028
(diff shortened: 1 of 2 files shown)


packages/liveblocks-react-ui/src/utils/use-transition.ts
```diff
@@ -8,8 +9,13 @@ const useReactTransition: typeof React.useTransition = (React as any)[
...
 
-function useTransitionFallback(transition: () => void) {
-  return [false, transition] as ReturnType<typeof useReactTransition>;
+function useTransitionFallback() {
+  const startTransition = useCallback(
+    (callback: TransitionFunction) => callback(),
+    []
+  );
+
+  return [false, startTransition] as ReturnType<typeof useReactTransition>;
```
nvie, line 18: Prefer an explicit return type annotation over `as`-casting (which can hide bugs and cause headaches later):

```suggestion
function useTransitionFallback(): ReturnType<typeof useReactTransition> {
  const startTransition = useCallback(
    (callback: TransitionFunction) => callback(),
    []
  );

  return [false, startTransition];
```

nvie (approved):
Noooice! 🙌

PR #1809, opened 2024-07-02
Refactor some Suspense internals

nvie:
This refactors the Suspense implementation internals for `useInboxNotifications()`. The internals are supposed to be somewhat simpler to follow / reason about, following the design layed out in LB-948.

Fixes LB-948.


packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -384,10 +361,16 @@ function makeLiveblocksContextBundle<
...
-  const { store, useSubscribeToInboxNotificationsEffect } =
+  const { loadInboxNotifications, store, useEnableInboxNotificationsPolling } =
     getExtrasForClient(client);
 
-  useSubscribeToInboxNotificationsEffect();
+  // Trigger initial loading of inbox notifications if it hasn't started
+  // already, but don't await its promise.
+  useEffect(() => {
+    loadInboxNotifications();
+  }, [loadInboxNotifications]);
```
nimeshnayaju, line 371: Since, this hook `useInboxNotifications_withClient` will also be called from `useInboxNotificationsSuspense_withClient` hook, this will mean suspense hook will perform data fetching twice? - first during render and second during this `useEffect` callback.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -398,37 +381,30 @@ function useInboxNotifications_withClient(client: OpaqueClient) {
...
-  const {
-    store,
-    fetchInboxNotifications,
-    useSubscribeToInboxNotificationsEffect,
-  } = getExtrasForClient(client);
+  const { waitUntilInboxNotificationsLoaded } = getExtrasForClient(client);
 
-  const query = store.get().queries[INBOX_NOTIFICATIONS_QUERY];
+  // Suspend until there are at least some inbox notifications
+  use(waitUntilInboxNotificationsLoaded());
```
nimeshnayaju, line 387: Beautiful 😍

packages/liveblocks-core/src/lib/utils.ts
```diff
@@ -208,7 +209,11 @@ export function memoizeOnSuccess<T>(
   return () => {
     if (cached === null) {
       cached = factoryFn().catch((err) => {
-        cached = null;
+        // Keep returning the failed promise for the next 5 seconds, after
+        // which we'll trigger a fresh promise again.
+        setTimeout(() => {
+          cached = null;
+        }, 5_000);
```
nimeshnayaju, line 216: We will probably (and hopefully) not get to this point (because we auto retry for [5 times](https://github.com/liveblocks/liveblocks/blob/0dbcecfb6eea4ea389e8f4228a753a8ebaaead17/packages/liveblocks-react/src/liveblocks.tsx#L236-L240) before throwing an error), but I would like to hear your thought on this @GuillaumeSalles . 

The implication of doing this, from my understanding, is that if user placed an error boundary above a component that has errored out, resetting the error boundary (until this timeout has completed) will not do anything.

nvie: > The implication of doing this, from my understanding, is that if user placed an error boundary above a component that has errored out, resetting the error boundary (until this timeout has completed) will not do anything.

This is correct! 👍 

It was either this timeout, or we would have to provide an explicit API to reset this memorized value, which I have tried various forms of, but haven’t found a satisfying solution for that won’t over-complicate things thus far.

I don’t like this timeout, but in practice it works pretty nicely.

packages/liveblocks-react/src/liveblocks.tsx
nvie: It will indeed _trigger_ twice, but since the value of the promise is memorized after the first attempt, it will not actually fetch twice.

But you’re right: the moment of first trigger is different between the Suspense and non-Suspense versions:
- Suspense version will first trigger during render (because of the `use(…)`)
- Non-Suspense will first trigger in a `useEffect` (so after mounting)

nimeshnayaju (approved):

packages/liveblocks-core/src/lib/utils.ts
nvie: Explained this a bit more in a code comment for posterity in 7cf046d3c9289aa7ce12ea47b79c95e9b6821961.

PR #1810, opened 2024-07-02
Upgrade `lexical` peer dependency to `^0.16.1`

nimeshnayaju:
This pull request upgrades the `lexical` peer dependency to version `^0.16.1` in `@liveblocks/react-lexical` and `@liveblocks/node-lexical`. This fixes compatibility issues with Next.js versions `14.2.0` and above.


nvie (approved):
Nice, finally! 🥹

Do you mind also updating the CHANGELOG?

packages/liveblocks-node-lexical/package.json
```diff
@@ -39,11 +39,11 @@
...
-    "@lexical/headless": "0.16.0",
-    "@lexical/selection": "0.16.0",
-    "@lexical/yjs": "0.16.0",
-    "@lexical/markdown": "0.16.0",
-    "lexical": "0.16.0"
+    "@lexical/headless": "0.16.1",
+    "@lexical/selection": "0.16.1",
+    "@lexical/yjs": "0.16.1",
+    "@lexical/markdown": "0.16.1",
+    "lexical": "0.16.1"
```
nimeshnayaju, line 46: @jrowny Updated the version of these packages on `@liveblocks/node-lexical` to `0.16.1` too. Do you see any issues with it?

nvie: Should these be using `^0.16.1` now, instead of `0.16.1` like we do in `@liveblocks/react-lexical`?

nimeshnayaju: @nvie I noticed they were pinned to the exact `0.16.0` version. I do not recall why we did that. @jrowny Do you know why that was done and if this change might create issues or if it is okay to set it to `^0.16.1`?

jrowny: should be good, for node specifically, it needs to be locked, no ~ because of the fact that Lexical stores globals and we can't have different versions between what we build and what the user has.

PR #1813, opened 2024-07-03
Fix code example in liveblocks-react.mdx

Teddarific:
## For Maintainers
Discovered an issue in the code example in the `Nesting Data Structures` code example. It was missing a pair of brackets in the `LiveMap` type — took me a minute to figure out what was going on after copy/pasting this example into my own code base to try to understand how `LiveMap` worked.

### Description
tldr: Fixing one of the code examples. The `LiveMap` API expects an array of key-value pairs to be passed in, not just a single array.


nvie:
Thanks for spotting and fixing this, @Teddarific! ❤️

PR #1816, opened 2024-07-03
Only mark threads as read if the window is focused

marcbouchenoire:
We mark threads as read (in the default `Thread` component) once their last comment becomes visible using the `IntersectionObserver` API. But @nvie noticed that this logic was triggered even if the window wasn't focused.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 - Add support for custom links to the composer. (either by pasting URLs with
   plain text selected or by pasting existing links)
 - Preserve whitespace and empty lines in comments.
+- Take into account the window's focus state when marking threads as read once
+  they are visible.
 - Fix improper `useTransition` fallback which would break on React versions
   lower than 18.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 - Add support for custom links to the composer. (either by pasting URLs with
   plain text selected or by pasting existing links)
 - Preserve whitespace and empty lines in comments.
+- Take into account the window's focus state when marking threads as read once
+  they are visible.
```
nvie, line 20: What do you think of describing it like this?

```suggestion
- Mark threads as read when visible (like before), but only if the window is focused
```

packages/liveblocks-react-ui/src/components/Comment.tsx
```diff
@@ -481,6 +500,12 @@ export const Comment = forwardRef<HTMLDivElement, CommentProps>(
 
     return (
       <TooltipProvider>
+        {markThreadAsReadWhenVisible && (
+          <MarkThreadAsReadWhenVisibleHandler
+            commentRef={ref}
+            threadId={markThreadAsReadWhenVisible}
```
nvie, line 506: Should we rename the `markThreadAsReadWhenVisible` prop to `autoMarkReadThreadId` maybe?

nvie (approved):
Lovely! 😍

packages/liveblocks-react-ui/src/components/Comment.tsx
marcbouchenoire: Yep that's probably easier to understand.

CTNicholas (approved):

PR #1817, opened 2024-07-04
Storage status docs

CTNicholas:

### DIFF
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -1270,10 +1270,8 @@ The possible value are: `initial`, `connecting`, `connected`, `reconnecting`, or
 ### useStorageStatus [@badge=RoomProvider]
 
 Returns the current storage status of the room, and will re-render your
-component whenever it changes. Can be helpful for
-[preventing users from leaving your app before changes are saved](#prevent-users-losing-unsaved-changes).
-A `{ smooth: true }` option is also available, which prevents quick changes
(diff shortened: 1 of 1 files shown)


docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -1284,10 +1288,60 @@ const status = useStatus();
...
+      }
+      e.preventDefault();
+    }
+
+    window.addEventListener("beforeunload", beforeUnload);
+
+    return () => {
+      window.removeEventListener("beforeunload", beforeUnload);
+    };
+  }, [storageStatus]);
```
nvie, line 1322: Not sure if I would prefer this implementation. I'd recommend the following implementation:

```suggestion
  const room = useRoom();

  useEffect(() => {
    function beforeUnload(e: BeforeUnloadEvent) {
      if (room.getStorageStatus() === "synchronized") {
        return;
      }
      e.preventDefault();
    }

    window.addEventListener("beforeunload", beforeUnload);

    return () => {
      window.removeEventListener("beforeunload", beforeUnload);
    };
  }, [room]);
```

This is because it would save (a lot of) re-renders for a component like this. It may actually be good to call out in the docs that if you use `useStorageStatus()` you should expect lots and lots of rerenders for any component that uses it. This is less true for the `{ smooth: true }` version, of course.

nvie (approved):
LGTM, but one suggestion for the pattern we recommend.

nvie (approved):

PR #1821, opened 2024-07-05
Infer readonly-ness in Yjs provider, rather than having the user set it

nvie:
This is a quick follow-up after #1820, where I think it's a bit nicer DX-wise if we automatically infer this property from the session rather than to ask the user to set it manually.


CTNicholas (approved):
Docs 👍

jrowny (approved):

PR #1822, opened 2024-07-05
Improve unit tests

marcbouchenoire:
This PR attempts to fix the flaky `useDeleteThread` tests. In the process, I improved dummy data creation and forced each test to be assigned a random room ID.


packages/liveblocks-react/src/__tests__/_dummies.ts
```diff
@@ -12,50 +12,65 @@ import {
   createThreadId,
 } from "./_createIds";
 
-export function dummyThreadData(
-  overrides?: Partial<ThreadData<BaseMetadata>>
-): ThreadData<BaseMetadata> {
+type AtLeast<T, K extends keyof T> = Partial<T> & Pick<T, K>;
```
nvie, line 15: 💯

nvie (approved):
Thanks for improving this test setup, @marcbouchenoire! 🙏

marcbouchenoire:
@nvie I ran the tests 3 times successfully (and 10+ times locally), so the delay is sufficient for now until the correct fix.

PR #1831, opened 2024-07-08
Implement `AnchoredThreads` and `FloatingThreads` component

nimeshnayaju:
This pull request adds the `ThreadsPanel` and `FloatingThreads` component to the project. 

- [x] Implement `ThreadsPanel` component.
- [x] Implement `FloatingThreads` component.
- [x] Update `nextjs-lexical` example to use the new components.
- [x] Update `CHANGELOG.md`.
- [x] Implement styles for `ThreadsPanel` component.
  - [x] Ensure class names match the name of the component. For e.g. if we decide to use `AsideThreads`, we want to make sure classnames are labelled as `aside-threads-thread`, etc.
  - [x] Disable thread animation for media query `prefers-reduced-motion`.
- [x] Implement styles for `FloatingThreads` component.
  - [x] Set thread `width` to `~350px`.
- [x] Update `react-lexical` documentation.
  - [x] `AnchoredThreads`
  - [x] `FloatingThreads`
  - [x] Include images if possible (similar to `react-ui`)


packages/liveblocks-react-lexical/src/comments/threads-panel.tsx
```diff
@@ -0,0 +1,429 @@
...
+    return () => onItemRemove(thread.id);
+  }, [thread.id, onItemAdd, onItemRemove]);
+
+  return (
+    <div ref={divRef} {...divProps}>
+      <Thread
+        thread={thread}
+        data-state={isActive ? "active" : "inactive"}
+        onClick={handleThreadClick}
+        className={"lb-lexical-threads-panel-thread lb-elevation"}
```
nimeshnayaju, line 406: @marcbouchenoire I have added `lb-elevation` to threads inside the `ThreadsPanel` component (name not confirmed yet). This is what it looks like:

Screenshot 2024-07-08 at 9 18 59 PM

Screenshot 2024-07-08 at 9 19 31 PM

Do you think we should use a different elevation style? I am not sure the intention behind `lb-elevation` but was it for components that would float on top of something? These threads don't necessarily feel like floating elements. Would love your help figuring this out.

packages/liveblocks-react-lexical/src/comments/threads-panel.tsx
```diff
@@ -0,0 +1,429 @@
...
+            thread={thread}
+            onItemAdd={onItemAdd}
+            onItemRemove={onItemRemove}
+            style={{
+              position: "absolute",
+              transform: `translateX(${isActive ? -12 : 0}px) translateY(${top}px)`,
+              left: 0,
+              width: "100%",
+              paddingBottom: gap,
+              transition: "transform 0.2s",
```
nimeshnayaju, line 322: @marcbouchenoire This will likely need to be removed but if we choose to control this animations via JavaScript in future, should we make this default for now so that users aren't surprised when they see threads being animated? 🤔

packages/liveblocks-react-lexical/src/comments/threads-panel.tsx
```diff
@@ -0,0 +1,429 @@
...
+        return (
+          <ThreadWrapper
+            key={thread.id}
+            Thread={Thread}
+            thread={thread}
+            onItemAdd={onItemAdd}
+            onItemRemove={onItemRemove}
+            style={{
+              position: "absolute",
+              transform: `translateX(${isActive ? -12 : 0}px) translateY(${top}px)`,
```
nimeshnayaju, line 318: @marcbouchenoire Threads that are active will also be translated left by `-12px`. Any suggestion on this number. I think it is okay to hard code this for now and in future, if users would like a different value, we could allow a prop to override this.

packages/liveblocks-react-ui/src/components/Composer.tsx
```diff
@@ -325,9 +325,7 @@ const ComposerWithContext = forwardRef<
     }, []);
 
     const handleEditorClick = useCallback(
-      (event: MouseEvent<HTMLDivElement>) => {
-        event.stopPropagation();
```
marcbouchenoire, line 329: Did you see any issues after this change?

packages/liveblocks-react-lexical/src/comments/threads-panel.tsx
marcbouchenoire: I'll create a softer shadow and yep I don't think we should use `lb-elevation` here, only default to a shadow.

packages/liveblocks-react-lexical/src/comments/threads-panel.tsx
marcbouchenoire: I think we can make it a CSS variable defined for this component only.

I also noticed that you don't actually use the `gap` value in your layout calculations so I'm tempted to also make it a CSS variable.

packages/liveblocks-react-lexical/src/comments/threads-panel.tsx
marcbouchenoire: Regardless of this I think it belongs in the external styles if we want to keep it by default

marcbouchenoire:
I noticed that sorting threads on the same line vs different lines doesn't give the same result. 

https://github.com/liveblocks/liveblocks/assets/6959425/a9c4b1f5-863e-44bc-b381-29b5c1a4d3c9

packages/liveblocks-react-lexical/src/comments/threads-panel.tsx
```diff
@@ -0,0 +1,429 @@
...
+            if (!(composer instanceof HTMLElement)) return;
+            composer.focus();
+          }
+          // If the composer is not found (and this can happen if `showComposer` is set to `false`), wait for the next render and try again to focus the composer.
+          else {
+            setTimeout(() => {
+              const composer = container.querySelector(".lb-composer-editor");
+              if (composer === null) return;
+              if (!(composer instanceof HTMLElement)) return;
+              composer.focus();
```
marcbouchenoire, line 384: After trying it a bit, I feel like this has too many cases where I would prefer to not focus the composer on active, just opening it.

packages/liveblocks-react-lexical/src/comments/threads-panel.tsx
marcbouchenoire: I personally generally prefer CSS variables because you can set them in React and in CSS (supporting responsive values, etc), compared to props where you're stuck in React only (it might be fine for many use-cases but it removes possibilities).

Maybe we could have a mixed API with a prop that sets the default value of a CSS variable, best of both worlds?

packages/liveblocks-react-ui/src/components/Composer.tsx
nimeshnayaju: Nothing strange yet. But, will revert this change since we won't be focusing the composer on click.

nimeshnayaju: Nothing strange yet. I could not think of how this would affect other things but I am ready to be surprised 😬

marcbouchenoire: Since we're hiding the composers by default now, this change is not needed anymore, should we keep it as it was before until we need it?

nimeshnayaju: We still want to select the associated mark node (so that the selected thread is active), so this will still be needed. Otherwise, clicking on the composer will not result in the thread being active.

examples/nextjs-lexical/app/lexical/editor.tsx
```diff
@@ -87,23 +80,26 @@ export default function Editor() {
...
-        return <ThreadWrapper key={thread.id} thread={thread} />;
-      })}
-    </div>
+  const isMobile = useIsMobile();
+
+  return isMobile ? (
+    <FloatingThreads threads={threads} className="w-[350px]" />
+  ) : (
+    <AnchoredThreads threads={threads} className="w-[350px]" />
   );
```
nimeshnayaju, line 89: @marcbouchenoire Updated to use a `useIsMobile` hook to determine if the user is on mobile depending on screen size. The `useIsMobile` hook uses `useSyncExternalStore` under the hood. Curious if you think this will make understanding this easier or more difficult?

packages/liveblocks-react-lexical/src/comments/floating-threads.tsx
```diff
@@ -0,0 +1,321 @@
...
+  return (
+    <Thread
+      thread={thread}
+      onKeyDown={handleKeyDown}
+      style={{
+        maxHeight: "var(--lb-lexical-floating-threads-available-height)",
+        maxWidth: "var(--lb-lexical-floating-threads-available-width)",
+        overflow: "auto",
+        ...style,
+      }}
```
nimeshnayaju, line 275: Added the `maxHeight` and `maxWidth` style here on the `Thread` component within `ThreadWrapper` component.

packages/liveblocks-react-lexical/src/comments/floating-threads.tsx
```diff
@@ -0,0 +1,321 @@
...
+        position: strategy,
+        top: 0,
+        left: 0,
+        transform: `translate3d(${Math.round(x)}px, ${Math.round(y)}px, 0)`,
+        minWidth: "max-content",
+      }}
+      className={classNames(
+        "lb-root lb-portal lb-elevation lb-lexical-floating-threads",
+        className
+      )}
```
nimeshnayaju, line 232: The container that holds floating threads is labelled with classname `lb-lexical-floating-threads`.

packages/liveblocks-react-lexical/src/comments/floating-threads.tsx
```diff
@@ -0,0 +1,321 @@
...
+      {...divProps}
+    >
+      {range.threads.map((thread) => (
+        <ThreadWrapper
+          key={thread.id}
+          thread={thread}
+          Thread={Thread}
+          onEscapeKeydown={handleEscapeKeydown}
+          className="lb-lexical-floating-threads-thread"
+        />
```
nimeshnayaju, line 156: Each `Thread` in the `FloatingThreads` component is labelled with classname `lb-lexical-floating-threads-thread`. And, because this is passed directly from the `FloatingThreads` component, users are not able to remove this classname but only add to the list of classnames. I noticed we were doing this with the old `ThreadPanel` component. Is that intentional @marcbouchenoire ?

packages/liveblocks-react-lexical/src/comments/floating-threads.tsx
```diff
@@ -0,0 +1,321 @@
...
+  ...threadProps
+}: ThreadWrapperProps) {
+  const handleKeyDown = useCallback(
+    (event: KeyboardEvent<HTMLDivElement>) => {
+      onKeyDown?.(event);
+
+      // TODO: Add ability to preventDefault on keydown to override the default behavior, e.g. to show an alert dialog
+      if (event.key === "Escape") {
+        onEscapeKeydown();
+      }
```
nimeshnayaju, line 261: We want users to be able to override this behaviour so that they are able to for e.g. display an alert dialog when they press <kbd>Esc</kbd> key.

examples/nextjs-lexical/app/lexical/editor.tsx
marcbouchenoire: I think it's better 👍 I did [move it to its own file](https://github.com/liveblocks/liveblocks/pull/1831/commits/0f9f90ab00612419eb228f5fc23abaf3c0df7f11) given how much an implementation detail it is.

packages/liveblocks-react-lexical/src/comments/floating-threads.tsx
marcbouchenoire: Yes we do this for all our "render props", it does prevent you from removing it (I think that's fine) but otherwise you'd have to hardcode it yourself to keep the default styles working, I don't think we want our users to manually set our class names on their end.

packages/liveblocks-react-lexical/src/comments/floating-threads.tsx
marcbouchenoire: What happens in `FloatingComposer` at the moment?

packages/liveblocks-react-ui/src/components/Composer.tsx
```diff
@@ -326,8 +326,7 @@ const ComposerWithContext = forwardRef<
 
     const handleEditorClick = useCallback(
       (event: MouseEvent<HTMLDivElement>) => {
-        event.stopPropagation();
-
+        event.preventDefault();
```
marcbouchenoire, line 329: 👀

packages/liveblocks-react-lexical/src/comments/floating-threads.tsx
nimeshnayaju: Same behavior unfortunately. The `onKeyDown` event handler on `Composer` is passed to the `form` (and not the Slate.js contenteditable `div`), so the event we get here is a propagated event from the contenteditable div, which is already default prevented. See:

https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react-ui/src/primitives/Composer/index.tsx#L750-L754

Happy to discuss further about this.

CTNicholas (commented):
Hey, just trying out the package and I've got this

CleanShot 2024-07-11 at 14 28 09@2x

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## v2.3.0 (Not released)
+
+### `@liveblocks/react-lexical`
+
+- New default components: `AnchoredThreads` and `FlatingThreads` to display
+  threads that are tied to a specific part of the document, similar to Notion,
+  Linear.
```
nimeshnayaju, line 7: @CTNicholas Would love an extra set of eyes to help make this better.

CTNicholas: I'll add more to this after we've figured out the docs!

CTNicholas (approved):

marcbouchenoire (approved):

PR #1832, opened 2024-07-09
Expose comments and notifications API on `@liveblocks/client`

GuillaumeSalles:
Expose the following internal APIs to let our users build comments and notifications components if they're not using React.

I had to cut some corners on the documentation for lack of time. The priority is to ship this to unblock customers and polish it right after it if possible.

**Review commit by commit recommended.** 

### APIs

- `Client.getInboxNotifications` and `Client.getInboxNotificationsSince`: 
I split`getInboxNotifications` into two functions to simplify the signature. `getInboxNotifications` is used for the initial call and `getInboxNotificationsSince` for subsequent polling. `getInboxNotificationSince` returns the deleted inbox notification and threads. 
I don't really like that `getInboxNotifications` returns inbox notifications and threads but we don't have the choice because these entities are intrinsically related. I considered renaming it `getInboxNotificationsWithThreads` but it felt too long.
- `Client.getUnreadInboxNotificationsCount`
- `Client.markAllInboxNotificationsAsRead`
- `Client.markInboxNotificationAsRead`
- `Client.deleteAllInboxNotifications`
- `Client.deleteInboxNotification`

- `Room.getThreads` and `Room.getThreadsSince`: I split `getThreads` into two functions for the same reasons that I split `getInboxNotifications`.

- `Room.getThread`: I simplified the signature to allow restructuring, even if endpoint returns `404`
- `Room.createThread`: `threadId` and `commentId` are now optional for users that do not want to build an optimistic cache.
- `Room.deleteThread`: I simplified the signature to only take an id instead of an object to match `useDeleteThread`.
- `Room.editThreadMetadata`
- `Room.markThreadAsResolved`: I simplified the signature to only take an id instead of an object to match `useMarkThreadAsResolved`.
- `Room.markThreadAsUnresolved`: I simplified the signature to only take an id instead of an object to match `useMarkThreadAsUnresolved`.
- `Room.createComment`: `commentId` is now optional for users that do not want to build an optimistic cache.
- `Room.editComment`
- `Room.deleteComment`
- `Room.addReaction`
- `Room.removeReaction`
- `Room.getNotificationSettings`. Renamed from `getRoomNotificationSettings`
- `Room.updateNotificationSettings`. Renamed from `updateRoomNotificationSettings`
- `Room.markInboxNotificationAsRead`: Exposed but marked as internal because I believe we should have `markThreadAsRead` instead.

### Other

I had to move our id creation utilities to `core`. After discussing with @nvie, we decided to not include `nanoid` as a dependency to avoid potential `CJS/ESM` issues. `nanoid@3.3.7` as been copy pasted inside the source and removed from our dependencies.


packages/liveblocks-core/src/room.ts
```diff
@@ -486,8 +489,8 @@ type CommentsApi<M extends BaseMetadata> = {
     | undefined
   >;
   createThread(options: {
-    threadId: string;
-    commentId: string;
+    threadId?: string;
```
GuillaumeSalles, line 492: We're setting `threadId` and `commentId` from `@liveblocks/react` to be able to reconcile optimistic updates. I made them optional to simplify the default usage. How do you feel about this @nvie?

nvie: Are all of these valid/meaningful calls?

1. `createThread()`
2. `createThread({ threadId })`
3. `createThread({ threadId, commentId })`
4. `createThread({ commentId })`

I don't think (2) and (4) should ever be used, right? I don't think we need to solve/protect against that at the type level, though. But I would like it if we'd add a runtime check there. Either both are undefined, or both are specified.

GuillaumeSalles: I can change that! Thank you!

packages/liveblocks-core/package.json
```diff
@@ -46,6 +46,9 @@
     "showdeps:high-level": "depcruise src --include-only '^src' --exclude='(^src/index.ts|shallow.ts|__tests__)' --collapse='^src/(refs|lib|compat|types|crdts|protocol)' --output-type dot | dot -T svg > /tmp/dependency-graph.svg && open /tmp/dependency-graph.svg"
   },
   "license": "Apache-2.0",
+  "dependencies": {
+    "nanoid": "^3"
+  },
```
nvie, line 51: As I mentioned to you already, I'd prefer to vendor in `nanoid` at this point if the license allows it, or implement a similar function (it's really small) ourselves, rather than to introduce this as a dependency. There is lots of pain with the dependencies that it's worth it, like the ESM/CJS stance/issues that v4 and v5 are causing which is causing us to keep it pinned at v3, the workarounds for the jest config etc etc. It's just a function that shouldn't cause all these issues.

GuillaumeSalles: I inlined nanoid v3.3.7 and also removed the dependency from our other packages.

CTNicholas (approved):
I've improved the docs a little, and I'm happy with them now! Thanks for writing half of them @GuillaumeSalles 🙌 please check I've understood everything.

packages/liveblocks-core/src/room.ts
```diff
@@ -1122,6 +1262,60 @@ function createCommentsApi<M extends BaseMetadata>(
     return body;
   }
 
+  async function getThreadsSince(options: { since: Date }) {
```
FlowFlorent, line 1265: What would be the use case of using this versus adding the `since` parameter on `getThreads`?

I could see getting resolved threads or threads having a certain metadata since a certain date as a possible call and this would make it confusing: should I use getThreads or getThreadsSince

packages/liveblocks-core/src/room.ts
FlowFlorent: Agreed with @nvie here, it doesn't make sense to pass just one, good catch!

packages/liveblocks-core/src/room.ts
FlowFlorent: You can actually disregard my above comment, I understand the logic better now, it provides the changes in the return model!

I like it :)

packages/liveblocks-core/src/room.ts
```diff
@@ -457,65 +459,190 @@ type SubscribeFn<
...
+   * Returns the modified and deleted threads and their associated inbox notifications since the requested date.
+   *
+   * @example
+   * const result = await room.getThreads();
+   * // ... //
+   * await room.getThreadsSince({ since: result.requestedAt });
+   */
+  getThreadsSince(options: { since: Date }): Promise<{
+    threads: {
+      modified: ThreadData<M>[];
```
FlowFlorent, line 501: should we use `updated` instead of `modified`? To match `updatedAt`, we don't have `modifedAt`

FlowFlorent: this would also hold for `getInboxNotificationsSince`

FlowFlorent (approved):

marcbouchenoire (approved):

nimeshnayaju (approved):
Looks really good. Thanks for working on this!!

PR #1834, opened 2024-07-09
Various `textMention` notification fixes

marcbouchenoire:
Someone reported that the `showRoomName` prop didn't work on `textMention` notifications, which led me to fix a few other issues.

- We weren't using the provided (via `kinds`) components for `textMention` notifications, that's why the `showRoomName` prop didn't work
- We were using an empty icon but we have the mention's creator ID so we can show their avatar like thread notifications
- A type issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### `@liveblocks/react-ui`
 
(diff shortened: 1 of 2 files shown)


nvie (approved):

CTNicholas (approved):
Good fixes/improvements!

PR #1835, opened 2024-07-09
Refactor internals of Batch APIs to take only a single arg as input

nvie:
The "multi-arg" batch API was a premature optimization we never used or needed, and this will make maintaining it going forward a bit easier. Look at the changed unit tests to see the difference of (internal) behavior. This only changes internal behavior. No public APIs are affected by this change.

Targeted to be included in #1809.


marcbouchenoire (approved):
🧹

PR #1843, opened 2024-07-12
Fix resolved threads not being filtered correctly

nimeshnayaju:
The selectedThreads function was not correctly filtering threads based on the 'resolved' query parameter. This pull request fixes the bug by adding a check for the 'resolved' value in the query and excluding threads that do not match. Additionally, a failing test was added to demonstrate the bug and then removed after the fix was implemented.

### DIFF
--- a/packages/liveblocks-react/src/comments/lib/selected-threads.ts
+++ b/packages/liveblocks-react/src/comments/lib/selected-threads.ts
@@ -30,6 +30,11 @@ export function selectedThreads<M extends BaseMetadata>(
       const query = options.query;
       if (!query) return true;
 
+      // If the query includes 'resolved' filter and the thread's 'resolved' value does not match the query's 'resolved' value, exclude the thread
+      if (query.resolved !== undefined && thread.resolved !== query.resolved) {
+        return false;
+      }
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react/src/comments/lib/selected-threads.ts
```diff
@@ -30,6 +30,11 @@ export function selectedThreads<M extends BaseMetadata>(
       const query = options.query;
       if (!query) return true;
 
+      // If the query includes 'resolved' filter and the thread's 'resolved' value does not match the query's 'resolved' value, exclude the thread
+      if (query.resolved !== undefined && thread.resolved !== query.resolved) {
+        return false;
+      }
```
nimeshnayaju, line 36: @nvie I noticed we do not check for the resolved property inside `query` when filtering for threads. They may have been missed during the change to make `resolved` as first class citizen.

ofoucherot (approved):

nvie (approved):
> _“Additionally, a failing test was added to demonstrate the bug and then removed after the fix was implemented.”_

😍

PR #1844, opened 2024-07-12
Use event source to dispatch provider status inside `useEditorStatus`

nimeshnayaju:
This PR updates `useEditorStatus` to use event source to notify when the provider status updated. The current implementation fails when `LexicalPlugin` is rendered conditionally because the hook expects `provider` to be available after the first render. This fixes [LB-1011](https://linear.app/liveblocks/issue/LB-1011/useeditorstatus-always-returns-not-loaded-if-liveblocksplugin-is).

This PR also fixes issue with [remote cursor names not displaying](https://linear.app/liveblocks/issue/LB-878/user-names-on-remote-cursors-are-not-displayed) by rendering `CollaborationPlugin` only after `useSelf` returns a defined value, which is not ideal as we have to wait till `useSelf` returns a defined value to render `CollaborationPlugin`. But, hopefully, we can invest more time to think deeper about the issues around awareness and see if forking `CollaborationPlugin` from `@lexical/react` is necessary.

### DIFF
--- a/packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
+++ b/packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
@@ -244,7 +244,7 @@ export const LiveblocksPlugin = ({
         }}
       />
 
-      {self && (
+      {info && (
         <CollaborationPlugin
           // Setting the key allows us to reset the internal Y.doc used by useYjsCollaboration
           // without implementing `reload` event
--- a/packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
+++ b/packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
@@ -1,15 +1,13 @@
 import { autoUpdate, useFloating } from "@floating-ui/react-dom";
-import { CollaborationContext } from "@lexical/react/LexicalCollaborationContext";
 import { CollaborationPlugin } from "@lexical/react/LexicalCollaborationPlugin";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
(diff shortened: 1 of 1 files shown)


packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
```diff
@@ -205,6 +197,13 @@ export const LiveblocksPlugin = ({
         const provider = new LiveblocksYjsProvider(room, doc);
         yjsDocMap.set(id, doc);
         providersMap.set(id, provider);
+
+        provider.on("sync", () =>
+          providerEventSource.notify({
+            id: room.id,
+            status: getEditorStatus(provider),
+          })
+        );
```
nvie, line 206: You may have to call `provider.off("sync")` to unsub, if I remember correctly. Not sure if "returning" the unsub function works with the Yjs Provider API.

packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
```diff
@@ -62,29 +60,24 @@ function getEditorStatus(
...
-
-    const cb = () => setStatus(getEditorStatus(provider));
-
-    provider.on("sync", cb);
-
-    return () => provider.off("sync", cb);
+    return providerEventSource.subscribe((event) => {
+      if (event.id !== room.id) return;
+      setStatus(event.status);
+    });
```
nvie, line 69: I think this is a useful pattern to turn _any_ event source into a stateful hook easily. You could rewrite this into a custom hook, i.e. so you can use `const status = useEventSource(providerEventSource)` here.

nvie (approved):
I left a couple suggestions, but this is looking good already! 👍

PR #1846, opened 2024-07-15
Improve emoji picker keyboard navigation

marcbouchenoire:
While looking at LB-960, I remembered LB-227, and I discovered that it was also resulting in a bigger issue I missed before: the fact that it would be stuck to the bottom edge is just a nitpick but for the top edge it wasn't taking the sticky headers into account so pressing `↑` would do nothing visually half the time because the highlighted emoji was behind one of the sticky headers.

Fixes LB-227.

### Before

https://github.com/user-attachments/assets/ff514c9b-20c5-4ad2-ada3-d99a9d6d7ca8

Notice how navigating up only "works" half of the time.

### After

https://github.com/user-attachments/assets/b556a51e-3697-43df-bd16-48e949b0a62b

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 - Fix `resolved` query not being applied when filtering threads inside
   `useThreads` hook.
 
+### `@liveblocks/react-ui`
+
+- Improve keyboard navigation within emoji pickers.
+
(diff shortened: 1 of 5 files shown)


CTNicholas (approved):
Good spot (and fix)!

PR #1861, opened 2024-07-26
docs: Update how-to-modify-liveblocks-storage-from-the-server.mdx

Teddarific:
Updating the code example in the modifying liveblocks from the server snippet.

Recently implemented this, and somethings were out of date:

- (Style change, nit) Removed the outer Promise since felt unnecessary + probably should just let any errors boil up
- Found through trial and error had to access the `room` property from an object that is returned from the `enterRoom` versus `enterRoom` doesn't seem to return just the `room` object anymore


CTNicholas:
Thank you, this guide was out of date! Applying your fix here: - https://github.com/liveblocks/liveblocks/pull/1867

PR #1862, opened 2024-07-26
`lexical-react`: Allow adding a provider to room

nimeshnayaju:
This pull request updates the `LiveblocksYjsProvider` and `Room` API to allow adding a provider to a room. 

It also includes a fix for `useEditorStatus` not displaying the correct provider status if `LexicalPlugin` is rendered conditionally.

We also now render `CollaborationPlugin` only after `self` is available. This fixes an issue with remote cursor names not displaying.


packages/liveblocks-core/src/room.ts
```diff
@@ -795,6 +801,12 @@ export type PrivateRoomApi<M extends BaseMetadata> = {
   undoStack: readonly (readonly Readonly<HistoryOp<JsonObject>>[])[];
   nodeCount: number;
 
+  // For usage in Y.js provider
+  getProvider(): Provider | undefined;
+  setProvider(provider: Provider | undefined): void;
+
+  onProviderUpdate: Observable<void>;
```
nimeshnayaju, line 808: We now add three internal methods inside a `Room` object. 

1. `getProvider`: Returns the currently attached (Y.js) provider in a room.
2. `setProvider`: Attaches the specified provider in the room. Could be set to `undefined` (currently no usage).
3. `onProviderUpdate`: Subscription to changes in the provider attached to a room.

packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
```diff
@@ -42,14 +40,21 @@ export type EditorStatus =
...
-  }
+function useProvider() {
+  const room = useRoom();
+
+  const subscribe = useCallback(
+    (onStoreChange: () => void) => {
+      return room[kInternal].onProviderUpdate.subscribe(onStoreChange);
+    },
+    [room]
+  );
```
nimeshnayaju, line 51: This `useProvider` hook returns the currently attached provider in a room. In order to do so, we subscribe to any provider changes in the room and always return the most recently attached provider.

packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
```diff
@@ -249,17 +246,19 @@ export const LiveblocksPlugin = ({
...
-        key={room.id}
-        id={room.id}
-        providerFactory={providerFactory}
-        username={username}
-        cursorColor={cursorcolor}
-        cursorsContainerRef={containerRef}
-        shouldBootstrap={true}
-      />
+      {self && (
+        <CollaborationPlugin
```
nimeshnayaju, line 250: `CollaborationPlugin` is only rendered once `self` is available. This is to work around Lexical's `CollaborationPlugin` issues around inability to update `username` and `cursor` after being set once 😭

jrowny (approved):
looks good!

PR #1863, opened 2024-07-30
Prototype comments without websocket & with toggle on polling

FlowFlorent:
#### Description

This aims to setup a prototype for an application which renders a list of many rooms in a single page (up to 500). Only a few rooms are in view at a time (roughly 25).
Each room must have the comments functionality available. It is acceptable for comments to not be live and be polled at an interval.

To achieve that, two things must be done:
- autoConnect is set to false on the RoomProvider, that's already the case out of the box
- hiddenEnablePolling api should be built to poll for comments when enable, and not when false

#### What's needed for `autoConnect` to work well:
- [ ] allow Composer to be focused & editable while room is not connected
- [ ] setup polling based on room connection status -> "initial" or "disconnected" or introduce another status

#### What's needed for `hiddenEnablePolling`:
- [ ] Setup an api that reacts on the change of the value of `hiddenEnablePolling`

### DIFF
--- a/examples/nextjs-comments/src/app/page.tsx
+++ b/examples/nextjs-comments/src/app/page.tsx
@@ -21,16 +21,33 @@ function Example() {
       {threads.map((thread) => (
         <Thread key={thread.id} thread={thread} className="thread" />
       ))}
-      <Composer className="composer" />
+      <Composer className="composer"/>
     </main>
   );
 }
 
 export default function Page() {
-  const roomId = useExampleRoomId("liveblocks:examples:nextjs-comments");
+  // 10 rooms are created for the example
+  const exampleIds = Array.from({ length: 1 }, (_, i) => i + 1);
 
   return (
-    <RoomProvider id={roomId}>
+    <div>
+      {exampleIds.map((exampleId) => (
+        <Room key={exampleId} exampleId={exampleId.toString()} />
+      ))}
+    </div>
+  );
+}
+
+function Room(props: { roomId?: string; exampleId: string }) {
+  // const roomId = useExampleRoomId("liveblocks:examples:nextjs-comments");
+  const roomId =
+    props.roomId ?? props.exampleId
+      ? `liveblocks:examples:nextjs-comments:${props.exampleId}`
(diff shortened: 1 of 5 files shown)


packages/liveblocks-react/src/room.tsx
```diff
@@ -1415,12 +1441,13 @@ function useThreads<M extends BaseMetadata>(
     [room, options]
   );
 
+  // check how polling works here
```
FlowFlorent, line 1444:
```suggestion
```

examples/nextjs-comments/src/app/page.tsx
```diff
@@ -13,31 +13,79 @@ import { ErrorBoundary } from "react-error-boundary";
...
   const { threads } = useThreads();
 
   return (
-    <main>
+    <main ref={props.ref}>
       {threads.map((thread) => (
         <Thread key={thread.id} thread={thread} className="thread" />
       ))}
-      <Composer className="composer" />
+      <Composer className="composer" disabled={false}/>
```
FlowFlorent, line 24: does not work, explore why

examples/nextjs-comments/src/app/page.tsx
```diff
@@ -58,3 +106,24 @@ function useExampleRoomId(roomId: string) {
 
   return exampleRoomId;
 }
+
+
+export function useIsVisible(ref: React.RefObject<Element>): boolean {
```
FlowFlorent, line 111: does not work yet

packages/liveblocks-react-ui/src/components/Composer.tsx
```diff
@@ -396,13 +398,13 @@ const ComposerWithContext = forwardRef<
               {hasResolveMentionSuggestions && (
                 <ComposerInsertMentionEditorAction
                   label={$.COMPOSER_INSERT_MENTION}
-                  disabled={isDisabled}
+                  disabled={false}
```
FlowFlorent, line 401: dirty enable logic, understand why composer doesn't allow focus

PR #1864, opened 2024-07-30
Various Comments fixes

marcbouchenoire:


FlowFlorent:
To answer your [comment on slack](https://liveblocks.slack.com/archives/C02PZL7QAAW/p1722338426890649?thread_ts=1722282822.560749&cid=C02PZL7QAAW). The auth call is made even if the room is not connected, so we will already have access to the userId using the token

GuillaumeSalles (approved):

PR #1873, opened 2024-08-08
`useIsInsideRoom` hook

CTNicholas:
`useIsInsideRoom`: Returns `true` if the hook is inside `RoomProvider`, `false` otherwise.

```tsx
import { useIsInsideRoom } from "@liveblocks/react";

function Header() {
  const isInsideRoom = useIsInsideRoom();
  return isInsideRoom ? <LiveAvatars /> : <JustMyAvatar />;
}
```

This is not currently possible with `useRoom`, because it throws an error outside of `RoomProvider`.

LB-1123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## 2.4.1 (Not yet published)
 
+### `@liveblocks/react`
+
+- Add
+  [`useIsInsideRoom`](https://liveblocks.io/docs/api-reference/liveblocks-react#useIsInsideRoom)
+  hook, useful for rendering different components inside and outside of
+  [`RoomProvider`](https://liveblocks.io/docs/api-reference/liveblocks-react#RoomProvider).
+
 ### `@liveblocks/react-lexical`
 
 - Fix a bug in `useEditorStatus` which prevented the hook from returning correct
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -1220,6 +1220,71 @@ import { useRoom } from "@liveblocks/react/suspense";
 const room = useRoom();
 ```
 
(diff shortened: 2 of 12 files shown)


nimeshnayaju (approved):
@CTNicholas Looks really good, and thank you for working on this!

I am trying to compare the names `useInsideRoom` and `useIsInsideRoom`. The latter is longer to write and is difficult to enunciate provides a clearer idea of the return type of the hook. Curious if you had certain thoughts in mind when you chose `useInsideRoom` for the hook name.

CTNicholas:
> @CTNicholas Looks really good, and thank you for working on this!
> 
> I am trying to compare the names `useInsideRoom` and `useIsInsideRoom`. The latter is longer to write and is difficult to enunciate provides a clearer idea of the return type of the hook. Curious if you had certain thoughts in mind when you chose `useInsideRoom` for the hook name.

I think you're right, `useIsInsideRoom` is a better name! Will change it soon, if no one disagrees.

marcbouchenoire:
I wonder if the API for this would be too specific? You can't call `useRoom` conditionally even if you checked that you're in a room so you're forced to use multiple components like in your example, maybe it's fine if this is the only supported use-case but I'm wondering 🤔

CTNicholas:
@marcbouchenoire I like this as a `boolean` personally, since I never actually need the `room` object! 

I often come across times where it would be easiest to render the same UI inside and outside of a room, especially since Notifications, but can't because our hooks need `RoomProvider`, so I have to create two entire layouts, instead of just having this switch.

The only time users commonly need a `room` object is when creating a Yjs app (e.g. [Step 5](https://liveblocks.io/docs/get-started/yjs-slate-react)), but I think `useIsInsideRoom` in the component above would be easier than `useOptionalRoom` there.

nvie:
💯

PR #1879, opened 2024-08-14
Fix method name in update room node documentation

SugarDarius:
Hi @CTNicholas 👋🏻 

I found this one in the doc. I suppose we should display `updateRoom` and not `createRoom` right?

Thanks 🙏🏻


CTNicholas (approved):
Good catch, thanks!

PR #1881, opened 2024-08-14
`removed` array not emitted in Yjs awareness event

CTNicholas:
The `removed` array in the `update` event was always empty because `actorToClientMap` was updated before the removed user could be retrieved.

```tsx
yProvider.awareness.on("update", ({ added, updated, removed } => {
  // always an empty array `[]`
  console.log(removed);
});
```


CTNicholas (commented):
This line was never called!

nvie (approved):
🙏

PR #1882, opened 2024-08-15
Merge internal `CommentsApi`

marcbouchenoire:
As part of https://github.com/liveblocks/liveblocks/tree/attachments, I made this change to simplify exporting Comments-related internal values (since `[kInternal]` is defined in `createRoom`). Since https://github.com/liveblocks/liveblocks/tree/attachments might be a long-lived branch, I'm making this PR to avoid diverging too much from `main` over time.

The diffs are awful, so just to make sure: this is just moving things around, there aren't any actual changes.

I didn't do the same for `createNotificationsApi` in `client.ts` since I suspect it might be a bit trickier but we could tackle that in a second PR.

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -470,186 +470,6 @@ export type GetThreadsOptions<M extends BaseMetadata> = {
   };
 };
 
-type CommentsApi<M extends BaseMetadata> = {
-  /**
-   * Returns the threads within the current room and their associated inbox notifications.
-   * It also returns the request date that can be used for subsequent polling.
-   *
-   * @example
-   * const {
-   *   threads,
-   *   inboxNotifications,
-   *   requestedAt
-   * } = await room.getThreads({ query: { resolved: false }});
-   */
-  getThreads(options?: GetThreadsOptions<M>): Promise<{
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -3057,98 +2655,461 @@ export function createRoom<
...
+    params?: QueryParams,
+    options?: RequestInit
+  ): Promise<Response> {
+    // TODO: Use the right scope
+    const authValue = await delegates.authenticate();
+
+    return fetchClientApi(config.roomId, endpoint, authValue, options, params);
+  }
+
+  async function fetchCommentsJson<T>(
```
nvie, line 2756: Looks like this is a new function, right? Did you intend to include it in this PR? I'm fine with it, but just double-checking since you mentioned it was only moving code around.

nvie (approved):
✨🧽✨

packages/liveblocks-core/src/room.ts
marcbouchenoire: 🦅👁️ `createCommentsApi` had it as `fetchJson` so I renamed it to fit in `createRoom`, but its implementation hasn't changed.

PR #1883, opened 2024-08-15
feat: add getInboxNotifications method to Liveblocks node client

FlowFlorent:
#### Description

Adds `getInboxNotifications` with an `unread` query parameter.

```ts
  /**
   * Returns the inbox notifications for a user.
   * @param params.userId The user ID to get the inbox notifications from.
   * @param params.query The query to filter inbox notifications by. It is based on our query language and can filter by unread.
   */
  public async getInboxNotifications(params: {
    userId: string;
    /**
     * The query to filter inbox notifications by. It is based on our query language.
     *
     * @example
     * ```
     * {
     *  query: "unread:true"
     * }
     * ```
     *
     * @example
     * ```
     * {
     *   query: {
     *     unread: true
     *   }
     * }
     * ```
     *
     */
    query?: string | { unread: boolean };
  })
  ```

### DIFF
--- a/packages/liveblocks-node/src/__tests__/client.test.ts
+++ b/packages/liveblocks-node/src/__tests__/client.test.ts
@@ -918,7 +918,7 @@ describe("client", () => {
     });
   });
 
-  test("getInboxNotifications works with a query", () => {
+  test("getInboxNotifications works with a query", async () => {
     const userId = "user1";
     const notifications = [
       {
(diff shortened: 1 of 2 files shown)


nvie (approved):
LGTM, but I also requested a review from @nimeshnayaju to do a double-check for consistency with the rest of the Node client's methods, which he implemented.

nimeshnayaju (approved):
🚀🚀🚀

PR #1885, opened 2024-08-16
Prevent codemod false positives

marcbouchenoire:
There are some cases where the `text → AST → text` process will result in different code even if the AST wasn't changed. Inspired by [this list of codemods](https://github.com/es-tooling/module-replacements-codemods/blob/main/codemods/has/index.js), this PR adds a dirty flag to all our codemods and uses it to return the unchanged input if the codemod didn't change the AST. 

**False positive example with the current version of `@liveblocks/codemod`:**
false-positive


nvie (approved):
💯

PR #1889, opened 2024-08-16
Remove recently-introduced test delays

marcbouchenoire:
I found [this trick](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react/src/__tests__/useEditThreadMetadata.test.tsx#L92) for the `editThreadMetadata` tests and I remembered that for basically the same problem I went with 1s delays last month or so 🥲, this PR replaces the delays by the same trick. (until we have an API (internal or not) to check for pending optimistic updates)


nvie (approved):
Nice! While still not the ideal expression of this, it's better than waiting an arbitrary number of seconds. I agree that all usage of `waitFor` is a hack and indicates we're missing a proper synchronization event.

nimeshnayaju (approved):

PR #1890, opened 2024-08-16
Get user inbox notifications doc

FlowFlorent:
Screenshot 2024-08-20 at 8 26 19 AM
Screenshot 2024-08-20 at 8 26 02 AM

### DIFF
--- a/docs/references/v2.openapi.json
+++ b/docs/references/v2.openapi.json
@@ -2777,7 +2777,7 @@
           },
           "403": {
             "description": "Forbidden"
-          },
+          }
         }
       }
     },
--- a/docs/references/v2.openapi.json
+++ b/docs/references/v2.openapi.json
@@ -2778,9 +2778,6 @@
           "403": {
             "description": "Forbidden"
           },
-          "404": {
-            "description": "Not Found"
-          }
         }
       }
     },
--- a/docs/references/v2.openapi.json
+++ b/docs/references/v2.openapi.json
@@ -2726,6 +2726,62 @@
             "description": "Not Found"
           }
         }
+      },
+      "get": {
+        "summary": "Get all inbox notifications",
+        "tags": ["Notifications"],
+        "parameters": [
+          {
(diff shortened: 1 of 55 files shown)


docs/references/v2.openapi.json
```diff
@@ -2726,6 +2726,62 @@
...
+          },
+          "401": {
+            "description": "Unauthorized"
+          },
+          "403": {
+            "description": "Forbidden"
+          },
+          "404": {
+            "description": "Not Found"
+          }
```
FlowFlorent, line 2783:
```suggestion
```

docs/references/v2.openapi.json
```diff
@@ -2726,6 +2726,59 @@
...
+                  "items": {
+                    "oneOf": [
+                      {
+                        "$ref": "#/components/schemas/InboxNotificationThreadData"
+                      },
+                      {
+                        "$ref": "#/components/schemas/InboxNotificationCustomData"
+                      }
+                    ]
+                  }
```
CTNicholas, line 2770:
```suggestion
                  "oneOf": [
                    {
                      "$ref": "#/components/schemas/InboxNotificationThreadData"
                    },
                    {
                      "$ref": "#/components/schemas/InboxNotificationCustomData"
                    }
                  ]
```
I think it will work like this!

CTNicholas (commented):
Great, thanks for doing this! I've just noticed we still have no example bodies, we should [look at that soon](https://linear.app/liveblocks/issue/LB-1180/add-example-requests-to-notifications-rest-api-docs).

CTNicholas (approved):

docs/references/v2.openapi.json
FlowFlorent: Unfortunately it does not, it instead provides that the response will be one model or the other 🤔

FlowFlorent:
Thank you @CTNicholas !

Doing the example in a follow up PR

PR #1914, opened 2024-09-06
Fix mention suggestions not scrolling

marcbouchenoire:
This PR fixes LB-1235 by using the same `autoUpdate` option of using animation frames with `@floating-ui/react-dom` that we use everywhere else.


CTNicholas (approved):
Nice!

PR #1916, opened 2024-09-09
Refactor cache store

nvie:
This PR starts refactoring the "cache store". Last week, we learned that the thing called the "cache store" is actually a store for multiple Liveblocks-specific concepts, like a cache for threads, inbox notifications, etc, as well as a query state cache, and a store for optimistic updates.

I'm starting the refactoring efforts small with some non-behavioral changes:

- `CacheState` → `UmbrellaStoreState` (the purpose is to not expose this directly soon)
- `CacheStore` → `UmbrellaStore` (the purpose is to make this a black box)
- Make `UmbrellaStore` a class
- Refactor all "setters" on the `UmbrellaStore` into a collection of public methods, so that call sites aren't responsible for knowing about, and updating the internal data structures in a integrity-safe way. How the internal cache data is maintained should be (and remain) a private implementation detail, which we can go refactor further later.

### About the `UmbrellaStore` ☂️🏪

For lack of a better name, Marc and I talked about a few alternatives last week and found "umbrella store" to be the least worst of the alternatives, mainly because it's not as generic as "cache store", and "umbrella" should not associate to any existing concepts, and at the same time associate roughly to "multiple things under one roof". There is, and will be, only one umbrella store, so we can also talk about _the_ umbrella store.

If a better name pops up, we can change it again later.

### Almost all pure refactorings…

These are almost all safe/pure refactorings, not changing any behavior. However, there is one small behavioral change introduced here, which is this commit: https://github.com/liveblocks/liveblocks/pull/1916/commits/4029f69e8d8ac136365dce3113dbcdbdd3fb3b4f (we made hard and soft deletions consistent for threads).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.6.2 (not released yet)
+
+### `@liveblocks/core`
+
+- Refactor caching internals to prepare for upcoming features
+
 ## 2.6.1
 
 ### `@liveblocks/react-ui`
--- a/e2e/next-sandbox/pages/inbox-notifications/index.tsx
+++ b/e2e/next-sandbox/pages/inbox-notifications/index.tsx
@@ -181,7 +181,13 @@ function usePendingUpdatesCount() {
     () => store.get().optimisticUpdates.length,
     [store]
   );
-  return React.useSyncExternalStore(store.subscribe, getter, getter);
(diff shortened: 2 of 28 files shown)


marcbouchenoire (approved):
☔🏪

nimeshnayaju (approved):

PR #1920, opened 2024-09-10
Refactor cache store internals (day 2)

nvie:
Results of the next refactoring steps, where I ended on day 2. The code is best reviewed commit-by-commit.

This PR introduces a `.batch()` method on the abstract `Store` class (a new sibling to `.get()`, `.set()`, and `.subscribe()`). This allows us to perform multiple successive `.set()` calls on the store without triggering multiple notifications to subscribers. Separating `.set()` calls from notifying allows us to better split and reuse individual cache mutations and enables more DRYing up of code.

More tomorrow, where I'll be focusing on moving the UmbrellaCache into `@liveblocks/react` (out of `@liveblocks/core`).

**One behavioral change:** Please note that the only behavioral change introduces here is in [this commit](https://github.com/liveblocks/liveblocks/commit/ecf7a0df6e7451a1a2cb9022878b8d76ca790953) (which we [discussed here](https://liveblocks.slack.com/archives/C02PZL7QAAW/p1725980167039189)).

### DIFF
--- a/packages/liveblocks-core/src/lib/create-store.ts
+++ b/packages/liveblocks-core/src/lib/create-store.ts
@@ -5,12 +5,15 @@ export type Store<T> = {
   get: () => Readonly<T>;
   set: (callback: (currentState: Readonly<T>) => Readonly<T>) => void;
   subscribe: (callback: (state: Readonly<T>) => void) => () => void;
+  batch: (callback: () => void) => void;
 };
 
 /**
  * Create a store for an immutable state. Close to Zustand's vanilla store conceptually but with less features.
  */
 export function createStore<T>(initialState: T): Store<T> {
+  let notifyImmediately = true;
+  let dirty = false;
   let state = initialState;
   const subscribers = new Set<(state: T) => void>();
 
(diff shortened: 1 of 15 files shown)


packages/liveblocks-core/src/lib/create-store.ts
```diff
@@ -25,18 +28,50 @@ export function createStore<T>(initialState: T): Store<T> {
...
+  /**
+   * While the callback is running, do not notify any subscribers. Keep
+   * collecting (potentially multiple) .set() updates to the store, and only
+   * notify subscribers at the end of the callback.
+   */
+  function batch(cb: () => void): void {
+    if (notifyImmediately === false) {
+      // Already in a batch, make this inner batch a no-op
+      return cb();
+    }
```
nimeshnayaju, line 64: Is this to guard against nested batch calls and unwanted asynchronous operations in the callback?

nimeshnayaju (approved):
> [Start decoupling UmbrellaStore from core, with the goal of moving it into react](https://github.com/liveblocks/liveblocks/pull/1920/commits/cfdc44ae53eef23d51284d21c27fd8436470d089)

If I recall correctly, one of the reasons why we kept the store in `core` was to potentially reuse various logic related to store subscription, notifying, applying optimistic updates, etc. on other frameworks too or if our customers wanted to build their own integrations. In retrospective, it feels we probably were trying to optimize for some unknown thing in the future. So, I am all for this change! Early abstractions are not a great idea.

packages/liveblocks-core/src/lib/create-store.ts
nvie: Yep! If you'd ever do:

```tsx
store.batch(() => {
  store.set();
  store.batch(() => {
    store.set();
    store.set();
  });
  store.set();
});
```

Then only the outermost `.batch()` will trigger one notify/rerender.

Async functions should never be used here. This only supports synchronous updates.

nvie: You can think of "batching" as "wait to notify subscribers until I'm done with all these state changes first". The benefit is that you can turn any statements inside the `.batch()` into a helper function to mutate state, without having to worry about accidentally triggering subscription notifications. The behavior is very much like how our public [`Room.batch`](https://liveblocks.io/docs/api-reference/liveblocks-client#Room.batch) API works.

nvie:
> Early abstractions are not a great idea.

Exactly! I talked about this a bit with Guillaume a couple of times, and all frontend frameworks have their own strong opinions on how to do state mutations. React, Vue, Svelte, Solid, etc all take a very different stance on it. While it may be true that we would need to duplicate the optimistic update knowledge if we wanted to make a Vue package, for now, we're moving these abstractions out of core, so they don't get shipped to those that don't use/need React.

But if it turns out that later it may be beneficial to move this back into core, that's still an open road AFAIAC. It's just that if that happens, we'll make a more informed decision about it, and _until then_ it's easier to move this code into the React package.

PR #1925, opened 2024-09-11
Refactor cache store (day 3)

nvie:
Full refactoring history:
- #1916
- #1920
- #1925 👈 This PR!

---

This PR moves UmbrellaStore from `core` → `react`.

Fixes LB-1254.

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -75,6 +75,7 @@ import { useLatest } from "./lib/use-latest";
 import { use } from "./lib/use-polyfill";
 import {
   createSharedContext,
+  getUmbrellaStoreForClient,
   LiveblocksProviderWithClient,
   useClient,
   useClientOrNull,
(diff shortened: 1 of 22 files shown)


nimeshnayaju (approved):

packages/liveblocks-react/src/room.tsx
```diff
@@ -108,11 +103,18 @@ import type {
...
+  deleteComment,
+  removeReaction,
+  upsertComment,
+} from "./umbrella-store";
 import { useScrollToCommentOnLoadEffect } from "./use-scroll-to-comment-on-load-effect";
 
 const SMOOTH_DELAY = 1000;
 
-const noop = () => { };
+const noop = () => {};
```
nvie, line 117: @jrowny I noticed this as well on your PR, but our linters somehow keep fighting over these changes. I think this only happens if we have different versions of Prettier (or ESLint maybe) installed. Can you make sure your VSCode is configured to use the ESLint and Prettier versions from this repo (the version in `node_modules`), and not maybe some globally installed different version of ESLint or Prettier?



The way to verify is to run `npm ci` in the root, and then run `turbo run format` on this package and there should not be any changes.

PR #1926, opened 2024-09-12
Fix: version history documentation

SugarDarius:
Small fix into the version history documentation to fix the `liveblocks.io` build


nvie (approved):
Great spot! 🦅👁️

PR #1935, opened 2024-09-16
Fix shallow checks to reduce unnecessary re-rendering

nvie:
This PR adds a `shallow2` helper which is like `shallow`, except that it will perform a 2-level deep shallow test instead.

This is needed to avoid unnecessary re-renders of components where the selector returns values like this:

```tsx
return {
  isLoading: false,
  threads: selectThreads(...),
}
```

Those will never pass the `shallow()` test, because the `selectThreads()` helper is basically a filter that will thus never return a stable value, because it's a new array every time.

This is only needed for the hooks that currently do dynamic filtering in the selector, because most other hooks by now will already just return a stable result. This didn't use to be the case, but after last week's refactorings all pre-processing, filtering, and ordering is now happening on writes to the store, no longer on reads.

### DIFF
--- a/packages/liveblocks-core/src/lib/shallow.ts
+++ b/packages/liveblocks-core/src/lib/shallow.ts
@@ -1,3 +1,5 @@
+import { isPlainObject } from "./utils";
+
 function shallowArray(xs: unknown[], ys: unknown[]): boolean {
   if (xs.length !== ys.length) {
     return false;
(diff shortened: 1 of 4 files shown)


packages/liveblocks-core/src/lib/shallow.ts
nvie, line 1: No functional changes made here, these are just cleanups.

packages/liveblocks-react/src/lib/__tests__/shallow2.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    const y = { isLoading: false, myData: THINGS };
+    expect(shallow(x, y)).toBe(true);
+    expect(shallow2(x, y)).toBe(true);
+
+    // But here, there is the difference
+    const xx = { isLoading: false, myData: THINGS.slice() };
+    const yy = { isLoading: false, myData: THINGS.slice() };
+    expect(shallow(xx, yy)).toBe(false);
+    expect(shallow2(xx, yy)).toBe(true);
+  });
```
nvie, line 31: This test should make it obvious what the two-level deep check of `shallow2` is doing, compared to a "normal" `shallow`.

marcbouchenoire (approved):
_`shallow2`: Electric Boogaloo_

PR #1941, opened 2024-09-17
Refactor cache store

nvie:
Continuation of #1939. Now also adopting this pattern for `useHistoryVersions()`.

Stop disabling the TSConfig setting `noUncheckedIndexedAccess`, making TS stricter, which is a good idea when building a library.

Also fixes three bugs:
- Fix a bug where under some conditions threads could end up without comments
- Fix a bug where notifications associated to deleted threads would not be deleted
- Fix a bug where subsequent optimistic updates to the same inbox notification could sometimes not get applied correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### `@liveblocks/react`
 
+- Fix a bug where under some conditions threads could end up without comments.
 - Fix a bug where notifications associated to deleted threads would not be
   deleted.
 - Fix a bug where subsequent optimistic updates to the same inbox notification
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### `@liveblocks/react`
 
+- Fix a bug where notifications associated to deleted threads would not be
+  deleted.
 - Fix a bug where subsequent optimistic updates to the same inbox notification
   could sometimes not get applied correctly.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.7.2 (not released yet)
+## 2.7.2
 
 ### `@liveblocks/react`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 2.7.2 (not released yet)
+
+### `@liveblocks/react`
+
+- Fix a bug where subsequent optimistic updates to the same inbox notification
+  could sometimes not get applied correctly.
+
 ## 2.7.1
 
 ### `@liveblocks/react-lexical`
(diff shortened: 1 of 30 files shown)


packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -639,16 +635,14 @@ function useInboxNotificationsSuspense_withClient(client: OpaqueClient) {
...
+  const { store, loadInboxNotifications, startPolling } =
     getExtrasForClient(client);
 
   // Trigger initial loading of inbox notifications if it hasn't started
   // already, but don't await its promise.
-  useEffect(() => {
-    loadInboxNotifications();
-  }, [loadInboxNotifications]);
+  useEffect(loadInboxNotifications, []); // eslint-disable-line react-hooks/exhaustive-deps
+  useEffect(startPolling, []); // eslint-disable-line react-hooks/exhaustive-deps
```
nimeshnayaju, line 644: Is there a particular reason why we do not add `startPolling` to the dependency here? Same for `loadInboxNotifications` above this

nvie: We could, but I wanted to clarify those should be stable handlers that should not get executed, and the expectation is that they only execute on mount and unmount.

nvie: But I'm also not a fan of shushing the linter, so I can add them. Technically, if the `client` instance would change you _would_ want to execute them. So maybe better anyway.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1081,36 +1105,56 @@ function internalToExternalState<M extends BaseMetadata>(
...
       case "update-notification-settings": {
-        output.notificationSettings[optimisticUpdate.roomId] = {
-          ...output.notificationSettings[optimisticUpdate.roomId],
+        const settings =
+          computed.notificationSettingsByRoomId[optimisticUpdate.roomId];
+
+        // If the inbox notification doesn't exist in the cache, we do not apply the update
+        if (settings === undefined) {
+          break;
+        }
```
nimeshnayaju, line 1154: Was one of the bugs caused by us not checking if an inbox notification existed before applying any optimistic update to it?

nvie: Correct! Since we weren't catching the `foo[0]` undefined index access here, it was possible (in theory, not sure if it's an actual practical problem), that we'd spread `{ ...undefined, updatedAt: new Date() }` in there, effectively creating an incorrect resource with only an `updatedAt` field and injecting that in the store state.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 2.7.2 (not released yet)
```
marcbouchenoire, line 1: So we can release `2.7.2` immediately after this lands:

```suggestion
## 2.7.2
```

marcbouchenoire (approved):
🙌

PR #1950, opened 2024-09-19
Make URL paths static, so they're better searchable

nvie:
Updates the client, making it more obvious what backend URLs are called exactly from the client, and requiring the use of URL-safe URL builder, making it unnecessary (and impossible to forget) to call `encodeURIComponent()` on all path arguments.

The other benefit is that all endpoint URL paths are now statically searchable, making it easier to see which paths are used where, and easier to find the backend endpoint implementation in the router.

### DIFF
--- a/packages/liveblocks-core/src/notifications.ts
+++ b/packages/liveblocks-core/src/notifications.ts
@@ -60,7 +60,7 @@ export function createNotificationsApi<M extends BaseMetadata>({
   }>;
 } {
   async function fetchJson<T>(
-    endpoint: string,
+    endpoint: `/v2/c/${string}`,
     options?: RequestInit,
     params?: QueryParams
   ): Promise<T> {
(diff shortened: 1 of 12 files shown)


packages/liveblocks-core/src/notifications.ts
```diff
@@ -60,10 +62,14 @@ export function createNotificationsApi<M extends BaseMetadata>({
   }>;
 } {
   async function fetchJson<T>(
-    endpoint: string,
+    endpoint: URLSafeString,
```
nimeshnayaju, line 65: Really like this! Do you know if it's possible to enforce this argument to have a `/v2/c/` prefix on a type level?

nimeshnayaju (approved):
Thank you for making this change. It is much easier to track which URL paths a method relates to now.

PR #1952, opened 2024-09-19
Retry attachment upload

FlowFlorent:


packages/liveblocks-core/src/room.ts
```diff
@@ -3244,56 +3273,98 @@ export function createRoom<
...
+          {
+            method: "PUT",
+            body: attachment.file,
+            signal: abortSignal,
+          },
+          {
+            fileSize: attachment.size,
+          }
+        );
+        // Remove the upload from the map
```
FlowFlorent, line 3325:
```suggestion
```

packages/liveblocks-core/src/room.ts
```diff
@@ -3244,56 +3273,97 @@ export function createRoom<
     if (abortSignal?.aborted) {
       throw abortError;
     }
+    setupFileUpload(attachment, options);
```
FlowFlorent, line 3276:
```suggestion
    setupFileUpload(attachment, options);
```

make this noop so `uploadAttachment` is the only thing that's exposed, it would also be called for uploadAttachment

PR #1954, opened 2024-09-20
DRY up AsyncResult types

nvie:
This is in preparation of upcoming pagination implementation, which will have to change these types. This PR makes no functional changes. It just changes the type aliases, but the types are exactly identical, just simpler to read.

### DIFF
--- a/packages/liveblocks-react/src/types/index.ts
+++ b/packages/liveblocks-react/src/types/index.ts
@@ -163,22 +163,7 @@ export type RoomNotificationSettingsStateError = AsyncError<"settings">;
 export type RoomNotificationSettingsStateSuccess = AsyncSuccess<RoomNotificationSettings, "settings">; // prettier-ignore
 export type RoomNotificationSettingsState = AsyncResult<RoomNotificationSettings, "settings">; // prettier-ignore
 
-export type HistoryVersionDataStateLoading = AsyncLoading;
-export type HistoryVersionDataStateError = AsyncError;
-
-// XXX Refactor this type away! We should NOT have data & error state simultaneously!
-// XXX Re-express this as AsyncSuccess<Uint8Array>
-export type HistoryVersionDataStateResolved = {
-  isLoading: false;
-  data: Uint8Array;
-  error?: Error;
-};
-
-export type HistoryVersionDataState =
-  | HistoryVersionDataStateLoading
-  // XXX Refactor this member away! We should NOT have threads & error state simultaneously!
-  | HistoryVersionDataStateResolved
-  | HistoryVersionDataStateError;
+export type HistoryVersionDataState = AsyncResult<Uint8Array>;
 
 export type HistoryVersionsStateLoading = AsyncLoading<"versions">;
 export type HistoryVersionsStateError = AsyncError<"versions">;
--- a/packages/liveblocks-react/src/types/index.ts
+++ b/packages/liveblocks-react/src/types/index.ts
@@ -130,22 +130,22 @@ export type CommentReactionOptions = {
   emoji: string;
 };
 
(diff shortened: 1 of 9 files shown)


packages/liveblocks-react/src/types/index.ts
```diff
@@ -136,142 +132,68 @@ export type CommentReactionOptions = {
...
-export type ThreadsStateLoading = {
-  isLoading: true;
-  threads?: never;
-  error?: never;
-};
+export type ThreadsStateLoading = AsyncLoading<"threads">;
+export type ThreadsStateSuccess<M extends BaseMetadata> = AsyncSuccess<ThreadData<M>[], "threads">; // prettier-ignore
 
+// XXX Refactor this type away! We should NOT have threads & error state simultaneously!
+// XXX Re-express this as AsyncSuccess<ThreadData<M>[], 'threads'>
```
nvie, line 139: @nimeshnayaju These XXX comments are what's currently annoying/inconsistent between various hooks. I think building a consistent pagination implementation will benefit from these inconsistencies to be cleaned up first, so all hooks work the same, before we build new features on top.

This is related to the comment I left [here in the tech design](https://www.notion.so/liveblocks/Pagination-support-to-React-hooks-10682084c812800ab19bdf8f2a962cc0?pvs=4#10782084c812805399caedb1f8dd853a).

packages/liveblocks-react/src/types/index.ts
```diff
@@ -136,142 +132,53 @@ export type CommentReactionOptions = {
...
+export type RoomNotificationSettingsStateError = AsyncError<"settings">;
+export type RoomNotificationSettingsStateSuccess = AsyncSuccess<RoomNotificationSettings, "settings">; // prettier-ignore
+export type RoomNotificationSettingsState = AsyncResult<RoomNotificationSettings, "settings">; // prettier-ignore
 
-export type UnreadInboxNotificationsCountStateSuccess = {
-  isLoading: false;
-  count: number;
-  error?: never;
-};
+export type HistoryVersionDataState = AsyncResult<Uint8Array>;
```
nvie, line 166: See commit dbc6d8428edcb090bd8a70071b19dce587aa1130 that changed this. It was fine here, because the implementation would never return dual data/error values, so we could safely simplify this type here already, without implementation changes.

packages/liveblocks-react/src/types/index.ts
nvie: I just went over the XXXs and most of them were trivial: the types were simply wrong, but the implementation was fine (no dual error/data). The only offender remaining is the `useThreads` hook!

nimeshnayaju: That sounds correct to me too. The types were wrong (on the `useThreads` and `useUserThreads`) but the implementation ensured that once initial fetching was completed, data and error wouldn't exist at the same time. 

Your recent update to `useInboxNotifications` hook ensured that both types and implementation were consistent.

nvie: Correct. Now we need to make the similar change to the `useThreads()` and `useUserThreads()` hooks as well to make them behave consistently. To summarize:

- No dual error/data states (this PR takes care of that)
- Restructure the internals of `useThreads()` so that we have a stable `waitUntilThreads()` (stable!) promise which we can use with `use()`
- This promise will: try to download initial 5 times, with incremental backoff delay, and only throw (reject) after the 5th attempt. No errors observable unti that happens.
- When that happens, the promise is cleared after 5 seconds, so the next render will trigger it again from the start

nvie: @nimeshnayaju Is going to work on that implementation today. We can merge this PR already, so you can make your changes on a different PR.

nimeshnayaju (approved):
Thank you for making return behaviour consistent across all our hooks!

PR #1958, opened 2024-09-23
Refactor: Split store queries into individual query based on resource type

nimeshnayaju:
This is a temporary refactoring artifact from Vincent and Nimesh. Each query corresponds to a resource which should eventually have its own type. This is why we split it for now.


nvie (approved):
💯💨

PR #1961, opened 2024-09-24
Refactor `useThreads` and `useThreadsSuspense` to use the `use` api/polyfill

nimeshnayaju:
This PR refactors the `useThreadsSuspense` hook to use the `use` API.

- [ ] `useThreadsSuspense`
- [ ] `useThreads`
- [ ] Fix failing test (add new tests if relevant)

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -269,7 +269,7 @@ function getExtrasForClient<M extends BaseMetadata>(client: OpaqueClient) {
 }
 
 function makeExtrasForClient<M extends BaseMetadata>(client: OpaqueClient) {
-  const store = getUmbrellaStoreForClient(client);
+  const store = getUmbrellaStoreForClient<M>(client);
 
   const DEFAULT_DEDUPING_INTERVAL = 2000; // 2 seconds
 
@@ -408,6 +408,48 @@ function makeExtrasForClient<M extends BaseMetadata>(client: OpaqueClient) {
     return;
   }
 
+  const loadRoomThreads = (
+    room: Room<JsonObject, LsonObject, BaseUserMeta, Json, M>,
+    queryKey: string,
+    options: UseThreadsOptions<M>
+  ): Promise<ThreadData<M>[]> => {
+    const existingPromise = store.getLoadThreadsPromise(queryKey);
+    if (existingPromise !== undefined) return existingPromise;
+
+    const getThreadsAndInboxNotifications = async () => {
+      const result = await room.getThreads(options);
+
+      const lastRequestedAt = lastRequestedAtByRoom.get(room.id);
+
+      /**
+       * We set the `lastRequestedAt` value for the room to the timestamp returned by the current request if:
+       * 1. The `lastRequestedAt` value for the room has not been set
+       * OR
+       * 2. The `lastRequestedAt` value for the room is older than the timestamp returned by the current request
+       */
+      if (
+        lastRequestedAt === undefined ||
+        lastRequestedAt > result.requestedAt
+      ) {
+        lastRequestedAtByRoom.set(room.id, result.requestedAt);
+      }
+
+      poller.start(POLLING_INTERVAL);
+
+      return {
+        threads: result.threads,
+        inboxNotifications: result.inboxNotifications,
+      };
+    };
+
+    const loadThreadsPromise = store.setLoadThreadsPromise(
+      queryKey,
+      getThreadsAndInboxNotifications()
+    );
+
+    return loadThreadsPromise;
+  };
+
   async function getThreadsAndInboxNotifications(
     room: OpaqueRoom,
     queryKey: string,
@@ -536,6 +578,7 @@ function makeExtrasForClient<M extends BaseMetadata>(client: OpaqueClient) {
     getInboxNotificationSettings,
     getRoomVersions,
     onMutationFailure,
+    loadRoomThreads,
   };
 }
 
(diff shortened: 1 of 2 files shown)


packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -182,6 +183,10 @@ type InternalState<M extends BaseMetadata> = Readonly<{
   queries3: Record<string, QueryAsyncResult>; // Notification settings
   queries4: Record<string, QueryAsyncResult>; // Versions
 
+  threads: {
+    loadThreadsPromises: Record<string, Promise<ThreadData<M>[]>>;
```
nimeshnayaju, line 187: We may also need to create a new separate field to track the promise status because we for the suspense `useThreads` hook, we can expect the `use` polyfill to add the `status` property on the promise but the non-suspense hook doesn't use the `use` API.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1025,6 +1033,57 @@ export class UmbrellaStore<M extends BaseMetadata> {
     this.setQuery2State(queryKey, { isLoading: false, error });
   }
 
+  public setLoadThreadsPromise(
+    queryKey: string,
+    getThreadsAndNotificationsPromise: Promise<{
+      threads: ThreadData<M>[];
+      inboxNotifications: InboxNotificationData[];
+    }>
+  ): Promise<ThreadData<M>[]> {
```
nimeshnayaju, line 1042: The goal of this was to delegate as much work as I could to the store, including auto retrying. There was still some work that I just couldn't pull out of the `extras` unless I spent a lot of time refactoring a lot of the code.

packages/liveblocks-react/src/room.tsx
```diff
@@ -2500,45 +2544,45 @@ function useThreadsSuspense<M extends BaseMetadata>(
...
+
       return {
+        isLoading: false,
         threads: selectThreads(state, {
           roomId: room.id,
           query: options.query,
           orderBy: "age",
         }),
-        isLoading: false,
       };
```
nimeshnayaju, line 2569: Here, we expect this selector to run only if the promise resolved correctly.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1025,6 +1033,57 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
+            return result.threads;
+          })
+          .catch((err) => {
+            // Throw the error as we want `autoRetry` function to handle the error
+            throw err;
+          });
+      },
+      5,
+      [5000, 5000, 10000, 15000]
+    ).catch((err) => {
```
nimeshnayaju, line 1064: We may have to set the promise status here (in the then and catch clauses) of autoRetry and after the request has been kicked off instead of doing it inside the use polyfill.

packages/liveblocks-react/src/umbrella-store.ts
nvie: Actually I had a thought about this this morning! Could we create our own promise wrapper, that would add the `use()`-compatible `status` fields?

```tsx
const coolerPromise = pinkyPromise(originalPromise);
coolerPromise.status // 'pending'
await coolerPromise;
coolerPromise.status // 'resolved'
```

This way we can synchronously read the promise's status (which is useful for us anyway), and it will additionally be compatible with `use()` already, so `use()` will never have to patch anything.

nvie: In places like this record, we could simply type the values part as a `PinkyPromise`:

```tsx
    loadThreadsPromises: Record<string, PinkyPromise<ThreadData<M>[]>>;
```

nvie: The implementation would be almost the same as `use()`, except it would not _throw_.

packages/liveblocks-react/src/umbrella-store.ts
nvie: Yess! Exactly my thinking, see comments above.

packages/liveblocks-react/src/room.tsx
nvie: Yep! After `use(loadThreadsPromise)`, I'd indeed expect the promise to be in `fulfilled` state. That's exactly what `use()` guarantees, so this makes sense to me. AFAIAC, you can remove this if-check and just leave a comment that at this point in time, the promise will always be resolved.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1438,11 +1481,12 @@ function useThreads<M extends BaseMetadata>(
     [room, options]
   );
 
-  const { store, getThreadsAndInboxNotifications, incrementQuerySubscribers } =
+  const { store, loadRoomThreads, incrementQuerySubscribers } =
     getExtrasForClient<M>(client);
 
   React.useEffect(() => {
-    void getThreadsAndInboxNotifications(room, queryKey, options);
+    // void getThreadsAndInboxNotifications(room, queryKey, options);
```
nvie, line 1488:
```suggestion
```

packages/liveblocks-react/src/room.tsx
```diff
@@ -408,6 +408,48 @@ function makeExtrasForClient<M extends BaseMetadata>(client: OpaqueClient) {
     return;
   }
 
+  const loadRoomThreads = (
```
nvie, line 411: This is now similar to `waitUntilInboxNotificationsLoaded`, right? Should we unify the naming of these helpers if so? Maybe `waitUntilRoomThreadsLoaded(roomId, queryKey)`?

packages/liveblocks-react/src/room.tsx
```diff
@@ -408,6 +408,48 @@ function makeExtrasForClient<M extends BaseMetadata>(client: OpaqueClient) {
...
+       * 2. The `lastRequestedAt` value for the room is older than the timestamp returned by the current request
+       */
+      if (
+        lastRequestedAt === undefined ||
+        lastRequestedAt > result.requestedAt
+      ) {
+        lastRequestedAtByRoom.set(room.id, result.requestedAt);
+      }
+
+      poller.start(POLLING_INTERVAL);
```
nvie, line 437: This is one thing I'm not sure about. Did we change behavior here? In the inbox notifications case, we only start polling once a component using `useInboxNotifications` is mounted, and we stop polling when the last component using it unmounts. Here we're always starting a poller, but never stopping it?

And slight aside: from just this PR diff, it's unclear to me _which_ poller this is starting? Do we have a poller per query for threads? Or do we have a single poller for _all_ threads?

PR #1962, opened 2024-09-24
Pagination

nvie:
This PR adds pagination support to the following hooks:

- `useInboxNotifications()`
- `useThreads()`
- `useUserThreads_experimental()`

Example:

```tsx
import { useInboxNotifications } from "@liveblocks/react";

const {
  inboxNotifications,
  isLoading,
  error,

  // ✨ New APIs
  fetchMore,
  isFetchingMore,
  hasFetchedAll,
  fetchMoreError,
} = useInboxNotifications();
```

The PR is relatively large, because we also did a lot of internal cleanups alongside the actual hook changes to make this implementation easier to maintain going forward.

### DIFF
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -141,12 +141,15 @@ export type PrivateClientApi<U extends BaseUserMeta, M extends BaseMetadata> = {
   readonly usersStore: BatchStore<U["info"] | undefined, string>;
   readonly roomsInfoStore: BatchStore<DRI | undefined, string>;
   readonly getRoomIds: () => string[];
-  readonly getThreads: (options: GetThreadsOptions<M>) => Promise<{
+  readonly getUserThreads_experimental: (
+    options: GetThreadsOptions<M>
+  ) => Promise<{
     threads: ThreadData<M>[];
     inboxNotifications: InboxNotificationData[];
+    nextCursor: string | null;
     requestedAt: Date;
   }>;
-  readonly getThreadsSince: (
+  readonly getUserThreadsSince_experimental: (
     options: { since: Date } & GetThreadsOptions<M>
   ) => Promise<{
     inboxNotifications: {
@@ -668,8 +671,9 @@ export function createClient<U extends BaseUserMeta = DU>(
         },
 
         // "All" threads (= "user" threads)
-        getThreads: httpClientLike.getThreads,
-        getThreadsSince: httpClientLike.getThreadsSince,
+        getUserThreads_experimental: httpClientLike.getUserThreads_experimental,
+        getUserThreadsSince_experimental:
+          httpClientLike.getUserThreadsSince_experimental,
       },
     },
     kInternal,
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -655,7 +655,7 @@ export function createClient<U extends BaseUserMeta = DU>(
 
       logout,
 
-      // XXX Eventually, once this is actually using the HttpClient class,
+      // XXXX Eventually, once this is actually using the HttpClient class,
       // "just" expose a reference to it here, instead of spreading all of its
       // methods
       ...httpClientLike,
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -655,9 +655,6 @@ export function createClient<U extends BaseUserMeta = DU>(
 
       logout,
 
-      // XXXX Eventually, once this is actually using the HttpClient class,
-      // "just" expose a reference to it here, instead of spreading all of its
-      // methods
       ...httpClientLike,
 
       // Internal
--- a/packages/liveblocks-core/src/lib/autoRetry.ts
+++ b/packages/liveblocks-core/src/lib/autoRetry.ts
@@ -44,6 +44,7 @@ export async function autoRetry<T>(
 
     // Do another retry
     const delay = backoff[attempt - 1] ?? fallbackBackoff;
+    // XXX - Add a console warn to notify the failure and delay before retrying
     await wait(delay);
   }
 }
--- a/packages/liveblocks-core/src/lib/autoRetry.ts
+++ b/packages/liveblocks-core/src/lib/autoRetry.ts
@@ -1,3 +1,4 @@
+import * as console from "./fancy-console";
 import { wait } from "./utils";
 
 /**
@@ -44,7 +45,10 @@ export async function autoRetry<T>(
 
     // Do another retry
     const delay = backoff[attempt - 1] ?? fallbackBackoff;
-    // XXX - Add a console warn to notify the failure and delay before retrying
+
+    console.warn(
+      `Attempt ${attempt} was unsuccessful. Retrying in ${delay} milliseconds.`
+    );
     await wait(delay);
   }
 }
--- a/packages/liveblocks-core/src/lib/autoRetry.ts
+++ b/packages/liveblocks-core/src/lib/autoRetry.ts
@@ -49,6 +49,8 @@ export async function autoRetry<T>(
   }
 }
 
+// XXXX Either DRY this up with the equivalent class in @liveblocks/core, or
+// find a better solution for this!
 export class StopRetrying extends Error {
   constructor(reason: string) {
     super(reason);
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -2960,7 +2960,7 @@ export function createRoom<
       query = objectToQuery(options.query);
     }
 
-    const PAGE_SIZE = 2; // XXX TODO: Bump this to a reasonable number
+    const PAGE_SIZE = 50;
 
     const response = await fetchCommentsApi(
       url`/v2/c/rooms/${config.roomId}/threads`,
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -472,10 +472,15 @@ type SubscribeFn<
 };
 
 export type GetThreadsOptions<M extends BaseMetadata> = {
+  cursor?: string;
   query?: {
     resolved?: boolean;
     metadata?: Partial<QueryMetadata<M>>;
   };
+};
+
+export type GetThreadsSinceOptions = {
+  since: Date;
 };
 
 export type UploadAttachmentOptions = {
@@ -765,7 +770,7 @@ export type Room<
    * // ... //
    * await room.getThreadsSince({ since: result.requestedAt });
    */
-  getThreadsSince(options: { since: Date }): Promise<{
+  getThreadsSince(options: GetThreadsSinceOptions): Promise<{
     threads: {
       updated: ThreadData<M>[];
       deleted: ThreadDeleteInfo[];
@@ -2898,12 +2903,10 @@ export function createRoom<
     return body;
   }
 
-  async function getThreadsSince(options: { since: Date }) {
+  async function getThreadsSince(options: GetThreadsSinceOptions) {
     const response = await fetchCommentsApi(
       url`/v2/c/rooms/${config.roomId}/threads`,
-      {
-        since: options?.since?.toISOString(),
-      },
+      { since: options?.since?.toISOString() },
       {
         headers: {
           "Content-Type": "application/json",
@@ -2961,7 +2964,10 @@ export function createRoom<
 
     const response = await fetchCommentsApi(
       url`/v2/c/rooms/${config.roomId}/threads`,
-      { query },
+      {
+        cursor: options?.cursor,
+        query,
+      },
       { headers: { "Content-Type": "application/json" } }
     );
 
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -1294,9 +1294,6 @@ export class CommentsApiError extends Error {
   }
 }
 
-const MARK_INBOX_NOTIFICATIONS_AS_READ_BATCH_DELAY = 50;
-const ROOM_THREADS_PAGE_SIZE = 2; // XXX TODO: Bump this to a reasonable number
-
 /**
  * @internal
  * Initializes a new Room, and returns its public API.
@@ -2963,12 +2960,14 @@ export function createRoom<
       query = objectToQuery(options.query);
     }
 
+    const PAGE_SIZE = 2; // XXX TODO: Bump this to a reasonable number
+
     const response = await fetchCommentsApi(
       url`/v2/c/rooms/${config.roomId}/threads`,
       {
         cursor: options?.cursor,
         query,
-        limit: ROOM_THREADS_PAGE_SIZE,
+        limit: PAGE_SIZE,
       },
       { headers: { "Content-Type": "application/json" } }
     );
@@ -3523,7 +3522,7 @@ export function createRoom<
 
       return inboxNotificationIds;
     },
-    { delay: MARK_INBOX_NOTIFICATIONS_AS_READ_BATCH_DELAY }
+    { delay: 50 }
   );
 
   async function markInboxNotificationAsRead(inboxNotificationId: string) {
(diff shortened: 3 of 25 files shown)


packages/liveblocks-core/src/client.ts
```diff
@@ -600,11 +623,136 @@ export function createClient<U extends BaseUserMeta = DU>(
...
+
+    const url = urljoin(baseUrl, endpoint, params);
+    return await fetcher(url, {
+      ...options,
+      headers: {
+        ...options?.headers,
+        Authorization: `Bearer ${getAuthBearerHeaderFromAuthValue(authValue)}`,
+      },
+    });
+  }
```
nvie, line 655: @nimeshnayaju This code looks really familiar. Where was this copied from?

packages/liveblocks-core/src/client.ts
```diff
@@ -600,11 +623,136 @@ export function createClient<U extends BaseUserMeta = DU>(
...
+        inboxNotifications: {
+          updated: [],
+          deleted: [],
+        },
+        requestedAt: new Date(),
+      };
+    } else {
+      throw new Error("There was an error while getting threads.");
+    }
+  }
```
nvie, line 754: @nimeshnayaju For inbox notifications, I just DRYed these up, and only have a single getter, where `?since=` is now optional. The only difference is that if `since` is undefined, it will do an initial fetch (i.e. page 1), and otherwise it will do delta updates (used by the poller). Also, if `since` is present, there can be "deleted" threads. If `since` isn't set, it will always return an empty array of deleted threads. (This is at least how it works for inbox notifications in the backend.)

packages/liveblocks-core/src/client.ts
```diff
@@ -666,8 +813,14 @@ export function createClient<U extends BaseUserMeta = DU>(
...
           return Array.from(roomsById.keys());
         },
+
+        // "All" threads (= "user" threads)
         getThreads,
         getThreadsSince,
+
+        // "Room" threads (= "normal" threads)
+        getRoomThreads,
+        getRoomThreadsSince,
```
nvie, line 823: I don't like all these variants here. Could we DRY them up into one?

packages/liveblocks-core/src/client.ts
nimeshnayaju: Yes! It was copied from `room.ts` file. Ideally, I'd like to remove all the fetch related logic from `room.ts` and have them call methods from client instead. Currently, I've only implemented `get(Room)Threads` and `get(Room)ThreadsSince` but haven't updated the respective methods on `room.ts` to use the client method instead.

https://github.com/liveblocks/liveblocks/blob/aca360f891ecbf5d2657e64240cef36b2df23c14/packages/liveblocks-core/src/room.ts#L1520-L1538

packages/liveblocks-core/src/client.ts
nimeshnayaju: Agreed. One seems to be coming from `notifications` related export (not sure if it is a fitting place). I will look into the efforts required to consolidate the two. Good point!

packages/liveblocks-core/src/client.ts
nimeshnayaju: > Also, if since is present, there can be "deleted" threads. If since isn't set, it will always return an empty array of deleted threads. (This is at least how it works for inbox notifications in the backend.)

I checked the backend for the `GET: threads` endpoint and it seems that the endpoint **CAN** return deleted threads even without `since` parameter. So, the batch of `2000` threads is deleted threads + non-deleted threads 🤔.

nvie: It seems to me that that is not correct and should get fixed. Maybe I'm overlooking something, but there should be 2 use cases only I think:

1. The first use case is a "delta update", aka a `?since=` query. It should return _all_ (not capped to 2000, but unlimited) threads created since the given date, including all threads deleted since that moment.
2. The second use case is the initial fetch (= page 1) and all subsequent page fetches. Those will use `?cursor&limit` params (and never `?since=`) and should _not_ return any deleted threads, ever.

This is my current understanding.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1022,6 +1129,181 @@ export class UmbrellaStore<M extends BaseMetadata> {
   public setQuery4Error(queryKey: string, error: Error): void {
     this.setQuery4State(queryKey, { isLoading: false, error });
   }
+
+  public loadThreadsAndNotifications(
+    roomId: string,
+    options: GetThreadsOptions<BaseMetadata>,
+    queryKey: string
+  ): Promise<ThreadData[]> {
```
nimeshnayaju, line 1137: Maybe, in future, we could not require the `queryKey` so that `useThreads` and others to only pass the `roomId` and `options` as that is enough to generate the query key.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -243,21 +279,26 @@ export type UmbrellaStoreState<M extends BaseMetadata> = {
 };
 
 export class UmbrellaStore<M extends BaseMetadata> {
+  private _client?: OpaqueClient;
```
nimeshnayaju, line 282: Can we make `client` required? Is there a particular reason why it is not?

nvie: Only didn't do so because we would have to create a client in all unit tests, which won't be used by most of them at the moment.

packages/liveblocks-core/src/client.ts
nvie: Perfect, as long as we don't forget to remove the original source and accidentally introduce the duplication here 👍

packages/liveblocks-core/src/room.ts
```diff
@@ -2825,6 +2825,7 @@ export function createRoom<
     return body;
   }
 
+  // XXX Use the `getRoomThreadsSince` method in the client
   async function getThreadsSince(options: { since: Date }) {
     const response = await fetchCommentsApi(
       url`/v2/c/rooms/${config.roomId}/threads`,
```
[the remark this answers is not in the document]
nvie: Delegates (callbacks) can get complicated to maintain over time when over-used: you end up in callback hell, where anything can call anything at any time and the overall structure is lost. So I wouldn't recommend the callback pattern here (or anywhere in most circumstances for that matter).

In this case, I'm actually not sure why the Room wouldn't be able to have access to (its) Client through some kind of back-reference? Maybe it's obvious once you try to add it, but right now I don't have it clear why it's set up like that.

It's like a 1-to-many relationship: one client instance can create multiple rooms, and each room can know which client it is managed by. All we have to do is add a way to obtain it from a Room, or am I missing something? As long as we don't create a cyclical dependency (in module imports, but also in avoiding a memory leak).

packages/liveblocks-core/src/lib/autoRetry.ts
```diff
@@ -46,3 +47,9 @@ export async function autoRetry<T>(
     await wait(delay);
   }
 }
+
+export class StopRetrying extends Error {
+  constructor(reason: string) {
+    super(reason);
+  }
+}
```
nvie, line 55: Maybe we should DRY this up with the existing:

https://github.com/liveblocks/liveblocks/blob/699db03ac6d9fb684c439a9e1e087eef80e2b088/packages/liveblocks-core/src/connection.ts#L192-L200

…and just import that one from `core` now?

packages/liveblocks-react/src/__tests__/useThreads.test.tsx
```diff
@@ -1490,9 +1490,6 @@ describe("useThreads: error", () => {
 
     expect(result.current).toEqual({ isLoading: true });
 
-    // Unmount so polling doesn't interfere with the test
-    unmount();
```
nvie: Can you explain why this was necessary?

packages/liveblocks-react/src/room.tsx
```diff
@@ -287,10 +287,15 @@ function makeExtrasForClient<M extends BaseMetadata>(client: OpaqueClient) {
...
+      requests.push(store.fetchThreadsDeltaUpdate(room.id));
     });
 
-    await Promise.allSettled(requests);
+    try {
+      await Promise.allSettled(requests);
+    } catch (err) {
+      // When polling, we don't want to throw errors, ever
+      console.warn(`Polling new threads failed: ${String(err)}`);
+    }
```
nvie, line 298: This try/catch won't be needed, because the promise returned by `Promise.allSettled()` will never be rejected and will always succeed. (Different from `Promise.all()`.)

packages/liveblocks-react/src/__tests__/useThreads.test.tsx
nimeshnayaju: This unit test is checks if we retry with a backoff if there was an error while fetching threads. The fetcher looks like the following:

https://github.com/liveblocks/liveblocks/blob/734f5b6a8a40b2bbf2fc9b79699e3cc64e22c467/packages/liveblocks-react/src/umbrella-store.ts#L1336-L1377

If we unmount, the room is left and removed from the client in which case `client.getRoom(roomId)` fails and throws a `StopRetrying` error which is not retried unlike a fetching error, which interferes with the goal of the test.

packages/liveblocks-core/src/lib/autoRetry.ts
nimeshnayaju: Good idea! Not ideal to have it defined in multiple places even though we may plan to remove it in future.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -990,29 +989,26 @@ function useUserThreads_experimental<M extends BaseMetadata>(
...
   const { store, incrementUserThreadsQuerySubscribers } =
     getExtrasForClient<M>(client);
 
   useEffect(() => {
-    store.waitUntilUserThreadsLoaded(options, queryKey).catch(() => {
+    store.waitUntilUserThreadsLoaded(options.query).catch(() => {
       // Deliberately catch and ignore any errors here
     });
-  }, [queryKey, store]); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [store, options.query]);
```
nimeshnayaju, line 1039: Only possible issue I can think of with this change is that if the user passes in an inline object to the hook like the following:

```tsx
useUserThreads({ query: { resolved: false } });
```

The code inside `useEffect` will run on each render as the `query` gets recreated on each render.

```tsx
useEffect(() => {
  store.waitUntilUserThreadsLoaded(options.query).catch(() => {
    // Deliberately catch and ignore any errors here
  });
}, [store, options.query]);
```

Because we used `queryKey` (which is a string) instead of `options.query` in the past as dependency (and ignored the dependency rule for `useEffect`), this was not an issue although still a hack.

That being said, I think this is fine because the `store.waitUntilUserThreadsLoaded` function will only make a fetch request once.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -362,6 +360,7 @@ function makeExtrasForClient(client: OpaqueClient) {
   // TODO Hmm. All of this is stuff that should be managed by the cache. Now we have caches in different places.
   const userThreadsSubscribersByQuery = new Map<string, number>();
 
+  // XXX Stop using a queryKey here!
   function incrementUserThreadsQuerySubscribers(queryKey: string) {
```
nimeshnayaju: This would be great if we can do this!!

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1002,9 +1002,10 @@ function useUserThreads_experimental<M extends BaseMetadata>(
 
   const queryKey = makeUserThreadsQueryKey(options.query);
 
-  useEffect(() => {
-    incrementUserThreadsQuerySubscribers(queryKey);
-  }, [incrementUserThreadsQuerySubscribers, queryKey]);
+  useEffect(
+    () => incrementUserThreadsQuerySubscribers(queryKey),
```
nimeshnayaju: Nice catch 😮

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -350,6 +350,17 @@ function makeLiveblocksExtrasForClient(client: OpaqueClient) {
...
     try {
+      //
+      // XXX Think about this! To remain symmetric with the solid inbox
+      // XXX notifications implementation, we should first call something like:
+      //       await store.waitUntilUserThreadsLoaded(query);
+      //                                              ^^^^^
+      // XXX                         However... which query should we use here?
+      //
+      // XXX Maybe this warrents an API like? 🤔
+      //       await store.waitUntilAnyUserThreadsLoaded();
```
nimeshnayaju, line 315: Hmm. Here's my understanding of how this currently works:

A poller starts as soon as a `useUserThreads` hook mounts and the poller runs, say every 2 hours. Let's say the first time it runs, user threads for none of the query key has been successfully loaded yet. In that case, the poller does nothing because the poller doesn't know when to fetch user threads delta since, i.e. it doesn't have the `lastRequestedAt` date value. So, we hope the next time the poller run (in 2 hours), user threads for at least a query key has been successfully loaded and the poller can then start fetching updates with the some value of `lastRequestedAt`.

With the change you're proposing (to also keep the logic consistent with inbox notifications), each time the poller runs (every 2 hours), we'll wait until until user threads for at least one query is loaded (this will be instant if the condition is true), and only then do we request delta updates. With this behavior, what if no user threads have been loaded both the first and the second time the poller runs and say after a few seconds the user threads are loaded, then both requests that were waiting for user threads to be available for at least a query is complete will kick off.

Do we instead want to only start the poller only after user threads are loaded for at least a query key? Happy to discuss this in the call.

packages/liveblocks-react/src/room.tsx
```diff
@@ -264,7 +264,14 @@ function makeDeltaPoller_RoomThreads(client: OpaqueClient) {
...
 
-  // XXX Stop using a queryKey here!
+  //
+  // XXX DISCUSSION
+  // XXX ----------
+  // XXX I _think_ we can remove the queryKey as a param here entirely and
+  // XXX "just" keep a flat count, like we do for our inbox notifications
+  // XXX poller. It would make the internals a lot simpler.
+  // XXX The query key isn't really used anyway.
+  //
```
nimeshnayaju, line 274: I do not recall why we did not keep a flat count to start with. But, in the current settings, I think you're right - we can keep a flat count instead and get rid of queryKey generation inside of the hook for good!!!

nvie: Thanks for fixing this already 👍

packages/liveblocks-react/src/liveblocks.tsx
nvie: Updating for posterity. This was discussed in yesterday's call, and we decided that we'll keep the current no-op behavior if the `lastRequestedAt` date is still unknown, but next week we'll want to give it a shot to clean this behavior up and not even make the poller start and make it fully dependent on the promise that needs to be fulfilled first (aka the moment the `lastRequestedAt` becomes set). Only _then_ we'll want to enable the poller.

Also we observed that two poll executions from the same poller can currently overlap, which is undesirable because of the possible race conditions. This isn't likely to happen in practice because the fetch requests are typically faster than the `POLLING_INTERVAL`, but there is a race condition there.

Ideally we'll fix this by making the following changes:
1. Set a max timeout for a single poll cycle (proposal POLLING_INTERVAL + some buffer).
2. Ensure we'll only schedule the next poll `POLLING_INTERVAL` after the polling function has been settled.

Current (possible) timeline:
```
|<- POLL_INTERVAL ->|
|-------------------|------------o------|-------------------|-------------------|-------------------|...
^ Mount             ^ Poll 1     ^      ^ Poll 2            ^ Poll 3    ^       ^ Poll 4   ^          ^
                      (noop)     |                                      |                  |          |
                            First page loaded                       Poll 2 ends      Poll 3 ends   Poll 4 ends
```

Proposed new behavior (researching/implementing next week):
```
                                 |<- POLL_INTERVAL ->|          |<- POLL_INTERVAL ->|
|--------------------------------o-------------------|----------|-------------------|----------|...
^ Mount                          ^                   |< Poll 1 >|                   |< Poll 2 >|
                                 |
                            First page loaded
```

packages/liveblocks-react/src/liveblocks.tsx
nvie: We discussed this yesterday, and this is indeed a bug I introduced with this refactoring! Great catch, @nimeshnayaju! 🙏

nimeshnayaju (approved):

PR #1967, opened 2024-09-26
Rename `EventSource`s `.clear()` to `._forceClear()` and mark it as internal

nvie:
This API only really exists for unit testing purposes, and should never be relied on in real world scenarios, as it will lead to gotchas and surprises.

### DIFF
--- a/packages/liveblocks-react-ui/src/primitives/Composer/utils.ts
+++ b/packages/liveblocks-react-ui/src/primitives/Composer/utils.ts
@@ -459,20 +459,12 @@ function createComposerAttachmentsManager(
     notifySubscribers();
   }
 
-  function dispose() {
-    clear();
-    // TODO Avoid having to rely on this API
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react-ui/src/primitives/Composer/utils.ts
```diff
@@ -461,7 +461,9 @@ function createComposerAttachmentsManager(
 
   function dispose() {
     clear();
-    eventSource.clear();
+    // TODO Avoid having to rely on this API
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access
+    (eventSource as any)._forceClear();
```
nvie, line 466: @marcbouchenoire Deliberately made this a bit ugly temporarily, to make it more obvious we shouldn't use this API here.

marcbouchenoire (approved):
Thanks for changing this!

packages/liveblocks-react-ui/src/primitives/Composer/utils.ts
nvie: Thank you! 🙏

PR #1968, opened 2024-09-26
Fix broken references in Open API spec

### LINKED ISSUE #1720: OpenAPI spec is invalid
**Describe the bug**

The [OpenAPI spec](https://github.com/liveblocks/liveblocks/blob/main/docs/references/v2.openapi.json) includes several `$ref`s to `'#/components/schemas/Schema'`, but that schema is not defined in the file.

**Expected behavior**

Schema file should be valid so it will work with codegen tools

nvie:
Clean-up prior to our hackathon 😇

Fixes #1720

### DIFF
--- a/docs/references/v2.openapi.json
+++ b/docs/references/v2.openapi.json
@@ -398,7 +398,7 @@
             "schema": {
               "type": "string"
             },
-            "name": "newRoomId",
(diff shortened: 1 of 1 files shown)


CTNicholas (approved):
Thanks!

PR #1969, opened 2024-09-26
[Hackathon] Open API spec cleanups

nvie:
Some cleanups prior to generating/fixing the contents:

- [x] Prettier
- [x] Rename param names to match our internal routes (so they won't be seen as a different route soon)
- [x] Reorder fields in a more natural / human-readable / consistent fashion
- [x] Move all "path params" into a central definition that's shared between all endpoints
- [x] Make all route path start with `/v2`
- [x] ~Can we move all "tags" to the path, not endpoint level?~ Nope, this is not valid syntax.
- [x] Make all route params work with $ref, too

### DIFF
--- a/docs/references/v2.openapi.json
+++ b/docs/references/v2.openapi.json
@@ -798,7 +798,7 @@
         }
       }
     },
-    "/schemas/{id}": {
+    "/schemas/{schemaId}": {
       "get": {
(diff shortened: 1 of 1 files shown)


nvie:
Closing this old hackathon PR, to remove it from my open PR lists.

PR #1974, opened 2024-10-01
Update examples deployment workflow

marcbouchenoire:
The examples deployment workflow is only up-to-date on the https://github.com/liveblocks/liveblocks/tree/examples branch, that's where it's used so nothing is broken but I think we should try to always work on the examples in the `main` → `examples` direction.


nvie (approved):

nvie:
You almost want to think about `examples` label not as a Git _branch_, but as a _moving tag_. We should therefore not commit to the `examples` branch directly and instead only add commits to `main`. Instead, in the ideal workflow, we can "just point the `examples` tag" to an arbitrary commit, and then those versions would get deployed. Maybe we can define an explicit GitHub action that would do just that?

marcbouchenoire:
@nvie Maybe yeah 🤔

PR #1976, opened 2024-10-01
Send client version in header to client-side API requests

nvie:
This will help identify older API clients over time, so we can maintain backward compatible API responses and detect when we can sunset legacy logic in our backend over time.


nimeshnayaju (approved):
🚀 🚀 🚀

PR #1979, opened 2024-10-03
Add E2E tests for Comments' `<Composer />`

marcbouchenoire:
This PR adds E2E tests for the `<Composer />` component. All of its text editing behaviors with Slate.js are currently not tested in any way and are easy to break as we add new nodes/behaviors.

(The tests don't actually interact with Liveblocks)

I got stuck testing keyboard shortcuts and copy-pasting, if anyone has ideas or experience about this I'd love to hear about it!

https://github.com/user-attachments/assets/1f0bff3f-6dbb-4697-aeba-b34d110e991c

Fixes LB-1325


SugarDarius:
Quick question @marcbouchenoire after reading your description here.
Did you try to test and use `KeyboardEvent` to emulate keyboard shortcuts?
```ts
const keyboardEvent = new KeyboardEvent('keydown', { key: 'Enter', metaKey: true, cancellable: true, bubbles: true })
document.dispatchEvent(keyboardEvent)
```

I do not know `playwright` well enough but regarding his capacities I suppose it should be possible.

marcbouchenoire:
@SugarDarius I haven't! I did try manually dispatching `ClipboardEvent` for copy-pasting and that didn't work but it could be due to other issues like permissions, I'll try `KeyboardEvent`. 👍

SugarDarius:
Arf :/ 
So I suppose things like `await page.keyboard.press('Ctrl+Enter');` do not work neither?

marcbouchenoire:
`await page.keyboard.press("Enter")` and `await editor.press("Enter")` work for submitting, same with new lines with `"Shift+Enter"`. But if I select a word and press `"Control+B"`/`"Meta+B"`/`"ControlOrMeta+B"`, it doesn't turn bold as expected ([source](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react-ui/src/primitives/Composer/index.tsx#L799)), but if I just press `"B"` then the selected word does get replaced by "B" this time.

nvie (approved):
Thanks for adding these tests, and for upgrading Playwright! 🙏

PR #1985, opened 2024-10-07
Various `Composer` improvements

marcbouchenoire:
I was debugging event propagation in our components and noticed a couple of low-hanging `Composer` improvements.

Here's a video of the non-disabled-actions-while-the-composer-is-disabled issue:

https://github.com/user-attachments/assets/28daa5a3-20aa-44bb-823f-8dbed25d6d69


packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
```diff
@@ -1198,9 +1219,13 @@ const ComposerForm = forwardRef<HTMLFormElement, ComposerFormProps>(
           onSubmitEnd();
         }
       },
-      [editor, attachments, onComposerSubmit, onSubmit, onSubmitEnd]
+      [disabled, editor, attachments, onComposerSubmit, onSubmit, onSubmitEnd]
     );
 
+    const stopPropagation = useCallback((event: SyntheticEvent) => {
+      event.stopPropagation();
+    }, []);
```
nimeshnayaju, line 1227: Is this is the hidden input used for attachment upload? Was this previously causing an issue?

nimeshnayaju (approved):

nvie (approved):

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
marcbouchenoire: Yes that's the one, we didn't get any reports but we stop propagation for all `onClick` events so we should do it here as well because it can be triggered by clicking on the "Add attachments" button.

PR #1986, opened 2024-10-07
Pagination cleanup

nimeshnayaju:
This PR follows up on the initial pagination work for useThreads and useInboxNotifications ([PR](https://github.com/liveblocks/liveblocks/pull/1962)). Our aim was to create a consistent approach for using the internal API, particularly the PaginatedResource introduced in the aforementioned PR. We also identified bugs and inconsistencies in poller usage across user threads, room threads, notifications, and history versions. The goal of this PR is to leverage the foundation built in the pagination work and apply it internally across all our existing hooks, ensuring consistent polling and fetching behavior. Additionally, we have refactored a few other internal APIs to be simpler and more robust.


nvie (approved):
🚀!

PR #1987, opened 2024-10-08
Update `ThreadData` type to always have an `updatedAt` date

nvie:
Updates the `ThreadData` type definition to make `updatedAt` a non-optional field. All threads should have an `updatedAt` time at this point, so we can remove a lot of uncertainty in the code around this.


nimeshnayaju (approved):
Code looks much simpler thanks to this! Please feel free to merge once backend confirmation is received!

ofoucherot (approved):

PR #1989, opened 2024-10-08
Refactor `useHistoryVersions` and `useRoomNotificationSettings`

nimeshnayaju:


packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -299,6 +299,72 @@ const noop = Promise.resolve();
 
 const ASYNC_LOADING = Object.freeze({ isLoading: true });
 
+class SingleResource {
+  public readonly observable: Observable<void>;
+  private _eventSource: EventSource<void>;
+  private _fetchPage: () => Promise<void>;
+  private _cachedPromise: UsablePromise<void> | null = null;
```
nimeshnayaju, line 306: I've duplicated the implementation from `PaginatedResource` for now but we most likely need an abstraction that both single and paginated resources can extends.

nvie: Yep! Duplication isn't bad here. Premature DRYing up or using subclasses would be much worse. Until the abstraction becomes apparent, we can keep the duplication. If they're still somewhat symmetric, this is fine.

nvie (approved):
Looking good! 👌

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -299,6 +299,72 @@ const noop = Promise.resolve();
 
 const ASYNC_LOADING = Object.freeze({ isLoading: true });
 
+class SingleResource {
```
nvie, line 302: Maybe `SinglePageResource`?

nvie:
@nimeshnayaju I just back-merged `pagination-cleanups` into this PR, and resolved [one conflict](https://github.com/liveblocks/liveblocks/pull/1989/commits/10c7a3692319f3da1961eccba52ddcc6bf0019b4) caused by #1995 🙏

packages/liveblocks-core/src/room.ts
```diff
@@ -1646,7 +1650,32 @@ export function createRoom<
...
+        }
+      );
+
+      return {
+        versions,
+      };
+    } else {
+      throw new Error(
+        `There was an error while getting text versions for room ${config.roomId}`
+      );
```
nvie, line 1677: @nimeshnayaju I think we can now replace this handling with a "normal" call to `.fetchJson()` here, and just run a "convert to dates" helper function on the success result. It would turn this `Error` into an `HttpError`. Agreed?

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1550,6 +1619,52 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
+    const queryKey = makeVersionsQueryKey(roomId);
+    let resource = this._roomVersions.get(queryKey);
+    if (resource === undefined) {
+      const roomVersionsFetcher = async () => {
+        if (this._client === undefined) {
+          // TODO: Think about other ways to structure this. Throwing a StopRetrying only
+          // makes sense only if we can easily know if the fetcher is going to be wrapped inside the autoRetry function
+          throw new StopRetrying(
+            "Client is required in order to load threads for the room"
+          );
```
nvie, line 1633: This should now also be more natural to handle, using the `HttpError` check.

packages/liveblocks-react/src/config.ts
```diff
@@ -11,4 +11,7 @@ export const config = {
 
   USER_THREADS_POLL_INTERVAL: 1 * MINUTES,
   USER_THREADS_MAX_STALE_TIME: 5 * SECONDS,
+
+  HISTORY_VERSIONS_POLL_INTERVAL: 1 * MINUTES,
+  HISTORY_VERSIONS_MAX_STALE_TIME: 3 * SECONDS,
```
nvie, line 16: Any reason to not use 5 seconds here?

packages/liveblocks-react/src/room.tsx
```diff
@@ -2115,23 +2099,42 @@ function useHistoryVersions(): HistoryVersionsAsyncResult {
...
+    //    *next* render after that, a *new* fetch/promise will get created.
+  );
 
   const state = useSyncExternalStoreWithSelector(
     store.subscribe,
     getter,
     getter,
     identity,
-    shallow
+    shallow2
```
nvie, line 2137: Good catch!

packages/liveblocks-react/src/room.tsx
```diff
@@ -2146,30 +2149,14 @@ function useHistoryVersions(): HistoryVersionsAsyncResult {
...
-    throw getRoomVersions(room);
-  } else if (state.error) {
-    throw state.error;
-  }
-
-  return state;
+  const result = useHistoryVersions();
+  assert(!result.error, "Did not expect error");
+  assert(!result.isLoading, "Did not expect loading");
+  return result;
```
nvie, line 2159: Love seeing the symmetry here! 😍

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -534,7 +595,7 @@ export type UmbrellaStoreState<M extends BaseMetadata> = {
    * Versions by roomId
    * e.g. { 'room-abc': {versions: "all versions"}}
    */
-  versionsByRoomId: Record<string, HistoryVersion[]>;
+  versionsByRoomId: Record<string, Record<string, HistoryVersion>>;
```
nvie, line 598: Why did you change this data structure here internally? When we fetch these from the backend, it's a list, and when we expose it from the hook it's also a list. In between we now have to convert the list to a Map, and back to an array again when we read it.

Is this done to remove "duplicate" version IDs? Ideally there would not be such duplicates in the first place, right?

nvie (approved):
Lovely changes! Great work, @nimeshnayaju! 😍

packages/liveblocks-react/src/config.ts
nimeshnayaju: This is supposed to be 5 seconds 😅 . It slipped from my example testing. Thanks for pointing out.

PR #1990, opened 2024-10-09
Use delta endpoints for `getThreadsSince` and `getUserThreadsSince` methods

nimeshnayaju:
We already do this for `getInboxNotificationsSince` method but missed doing the same for `getThreadsSince` and `getUserThreadsSince` methods.


nvie (approved):
Oof! How have I missed this? Thanks for catching this! 😇

PR #1992, opened 2024-10-09
Resolvers invalidation

marcbouchenoire:
Add advanced APIs to manually invalidate the cached results for each resolver.

```tsx
// Invalidate all users
client.resolvers.invalidateUsers();
// Invalidate specific users
client.resolvers.invalidateUsers(["user-1", "user-2"]);

// Invalidate all rooms
client.resolvers.invalidateRoomsInfo();
// Invalidate specific rooms
client.resolvers.invalidateRoomsInfo(["room-1", "room-2"]);

// Invalidate all mention suggestions
client.resolvers.invalidateMentionSuggestions()
```


packages/liveblocks-core/src/client.ts
```diff
@@ -314,6 +315,26 @@ export type Client<U extends BaseUserMeta = DU, M extends BaseMetadata = DM> = {
...
   logout(): void;
 
+  /**
+   * Advanced APIs related to the resolvers.
+   */
+  resolvers: {
+    /**
+     * Invalidate one or all users that were previously cached by `resolveUsers`.
+     */
+    invalidateUsers(userId?: string): void;
```
nvie, line 325: Maybe instead of `userId?: string`, you could accept an array by default:

```suggestion
    invalidateUsers(userIds: string[]): void;
```

Perhaps even also let it take a predicate function, so they can implement the invalidation very precisely according to whatever logic makes sense to them:

```suggestion
    // All of these overloads would make sense
    invalidateUsers(): void;
    invalidateUsers(userIds: string[]): void;
    invalidateUsers(predicate: (user: User) => boolean): void;
```

This would allow these calls:

```ts
client.resolves.invalidateUsers();  // Invalidate all
client.resolves.invalidateUsers(["user1"]);  // Invalidate a single user
client.resolves.invalidateUsers(["user1", "user2"]);  // Invalidate multiple users
client.resolves.invalidateUsers((user) => user.info.tags.includes("Engineering"));
```

What do you think?

nvie (commented):
The proposal looks great! I left one comment about a possible API that might be nice to expose here. We use something similar in the backend when we invalidate sessions and it's pretty nice to use. Just my $0.02.

packages/liveblocks-core/src/client.ts
marcbouchenoire: Yes, love it! I'm going to implement that on the plane ✈️

marcbouchenoire:
@nvie As (very briefly) discussed last week, I removed the predicate APIs because they feel too fragile given that the user/room objects don't contain the user/room IDs. We can think about them later based on feedback/requests.

packages/liveblocks-core/src/client.ts
```diff
@@ -657,9 +712,17 @@ export function createClient<U extends BaseUserMeta = DU>(
...
+      resolvers: {
+        invalidateUsers: invalidateResolvedUsers,
+        invalidateRoomsInfo: invalidateResolvedRoomsInfo,
+        invalidateMentionSuggestions: invalidateResolvedMentionSuggestions,
+      },
+
       // Internal
       [kInternal]: {
         currentUserIdStore,
+        mentionSuggestionsCache,
```
nvie, line 725: Is this only exported to access it in unit tests?

packages/liveblocks-react/src/index.ts
```diff
@@ -90,6 +90,11 @@ export {
...
 
-// Private APIs (for use in react-lexical only)
+// Private APIs
+
+// Used in @liveblocks/react-lexical
 export { getUmbrellaStoreForClient } from "./liveblocks";
 export { selectThreads } from "./umbrella-store";
+
+// Used in @liveblocks/react-ui and @liveblocks/react-lexical
+export { useMentionSuggestions } from "./use-mention-suggestions";
```
nvie, line 100: Should we export this with an uglier name, to make it clear users should not rely on it? Just adding a `@private` field will still make it pop up in autocompletions I think.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -666,24 +666,29 @@ function useUser_withClient<U extends BaseUserMeta>(
...
+
+  // Trigger a fetch if we don't have any data yet (whether initially or after an invalidation)
+  useEffect(() => {
+    if (!result.user && result.isLoading) {
+      // NOTE: .get() will trigger any actual fetches, whereas .getState() will not
+      void usersStore.get(userId);
+    }
+  }, [usersStore, userId, result]);
+
+  return result;
```
nvie, line 691: Not entirely sure why this change is needed? My expectation would be that `void usersStore.get(userId)` would not do anything if it's already loading or successfully loaded. I.e. this condition should get checked by the `usersStore.get()` method, not here, not?

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -743,17 +748,23 @@ function useRoomInfo_withClient(
...
     shallow
   );
+
+  // Trigger a fetch if we don't have any data yet (whether initially or after an invalidation)
+  useEffect(() => {
+    if (!result.info && result.isLoading) {
+      // NOTE: .get() will trigger any actual fetches, whereas .getState() will not
+      void roomsInfoStore.get(roomId);
+    }
+  }, [roomsInfoStore, roomId, result]);
```
nvie, line 765: Ditto.

nvie (approved):
Left a couple questions/suggestions, but overall this is looking good! 👍

packages/liveblocks-core/src/client.ts
marcbouchenoire: No it's used by `useMentionSuggestions` in `liveblocks-react` but it needs to live here at the client level in order to be accessible by `client.resolvers.invalidateMentionSuggestions`.

packages/liveblocks-react/src/index.ts
marcbouchenoire: I was on the fence about doing too many changes in this area because of the TipTap work being done at the same time but there are already changes anyway, I'll go with the `@liveblocks/react/_private` convention 👍

packages/liveblocks-react/src/liveblocks.tsx
marcbouchenoire: The invalidate methods don't trigger a revalidation on their own, it's up to the hooks (or any other consumer) to trigger one (to make sure we only revalidate if needed, e.g. at least one consumer is rendered and needs revalidated data). This change here makes the effect re-trigger if `result` goes back to an initial state, it's not related to deduping but I can see why the `if` statement would make it look that way, I'll remove it because yeah it's technically unnecessary.

packages/liveblocks-react/src/index.ts
marcbouchenoire: Done in https://github.com/liveblocks/liveblocks/pull/1992/commits/2ca30ffeda6232ee72079e957283db4b89bba940.

packages/liveblocks-react/src/liveblocks.tsx
marcbouchenoire: Improved in https://github.com/liveblocks/liveblocks/pull/1992/commits/25abd0b3eb9d59c81cddade5f48de59522abe419.

PR #1994, opened 2024-10-10
Throw `StopRetry` error in event of `403 Forbidden` while fetching inbox notifications, room threads or user threads.

nimeshnayaju:
We currently keep retrying on any kinds of error that happened while fetching inbox notifications, room threads or user threads. We do not want to retry if the response is of status`4xx`.


packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -610,20 +612,32 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
+        }
+
+        const nextCursor = result.nextCursor;
+        return nextCursor;
+      } catch (err) {
+        // If the error is a 403 Forbidden error, we do not want to keep retrying
+        if (err instanceof NotificationsApiError && err.status === 403) {
+          throw new StopRetrying(
+            "403 Forbidden: Please check your permissions and try again."
+          );
```
nimeshnayaju, line 635: This error object will be thrown to the nearest `ErrorBoundary` on Suspense and will be used to populate the `error` object in non-suspense, so the message will say the following:

```
403 Forbidden: Please check your permissions and try again.
```

@nvie Please feel free to suggest alternative words that may be easier to understand for users.

nimeshnayaju: We could also instead do a simple:

```ts
throw new StopRetrying(err.message);
```

And, the message is normally:

```
Resource forbidden
```

packages/liveblocks-core/src/room.ts
```diff
@@ -2962,48 +2962,36 @@ export function createRoom<
 
     const PAGE_SIZE = 50;
 
-    const response = await fetchCommentsApi(
+    const result = await fetchCommentsJson<{
```
nimeshnayaju, line 2965: Noticed we have a helper method `fetchCommentsJson` that checks the status in case of !response.ok and throws `CommentsApiError` with response status and message, so instead used it so I could check for response status from the store.

https://github.com/liveblocks/liveblocks/blob/7e6c8cb485f9a827523c5d0402bc009b8430ea5b/packages/liveblocks-core/src/room.ts#L2866-L2902

packages/liveblocks-react/src/umbrella-store.ts
nimeshnayaju: Hmm. I also realized this will expose `StopRetrying` error publicly. Does that mean we would need to also export `StopRetrying`? Maybe, time to think about a better error to expose 🤔

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -610,20 +612,32 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
+        // We initialize the `_lastRequestedNotificationsAt` date using the server timestamp after we've loaded the first page of inbox notifications.
+        if (this._notificationsLastRequestedAt === null) {
+          this._notificationsLastRequestedAt = result.requestedAt;
+        }
+
+        const nextCursor = result.nextCursor;
+        return nextCursor;
+      } catch (err) {
+        // If the error is a 403 Forbidden error, we do not want to keep retrying
+        if (err instanceof NotificationsApiError && err.status === 403) {
```
nvie, line 632: I think we'll want to extend this to any error in the 4xx range, not just the 403. What 4xx means is: you had no access and without changing the request you just made retrying will not have an effect.

4xx: Request refused for a known reason, you need to make a different request
5xx: Request refused for an unknown/unexpected reason, please retry

packages/liveblocks-react/src/umbrella-store.ts
nvie: I don't prefer catching and re-wrapping any existing errors like this, because it might become hard to trace where an error is coming from, what the real error/stack trace is, and where things are handled. Ideally, this function wouldn't have to be changed at all. Maybe a more solid approach is to start using the `throwError?: (err: any) => boolean` param from the `autoRetry` function, and somehow just define this logic in there.

For now, I'm okay hardcoding this `err instanceof NotificationsApiError && err.status === 403` check [where we set up the autoRetry call](https://github.com/liveblocks/liveblocks/blob/7e6c8cb485f9a827523c5d0402bc009b8430ea5b/packages/liveblocks-react/src/umbrella-store.ts#L478-L482), until we figure out a more flexible way of doing this, and do it similarly for `CommentsApiError` / `LiveblocksApiError`, etc. At least we maintain this ugly handling in a single place that way, and we don't have to spread it out across the code everywhere.

nvie (changes requested):
Starting to look good!

I just would like to settle on the best way of dealing with these errors before we merge this, so I'll leave this out of 2.9.1 for right now, but maybe we can make a couple changes and release it as 2.9.2 later this week.

packages/liveblocks-react/src/umbrella-store.ts
nvie: > _"Please feel free to suggest alternative words that may be easier to understand for users."_

One of the things about our server error responses, is that they should absolutely contain the reason for the rejection in the HTTP response body, so ideally we would simply use that as the message. Ideally the various `*ApiError`s would expose those as `err.message` already, and we should simply reuse those.

nimeshnayaju: Good idea! I also do not like that `PaginatedResource` is now aware of `HttpError` but it is okay for now until we come up with better abstractions! Thanks for the pointer.

```ts
const initialFetcher = autoRetry(
  () => this._fetchPage(undefined),
  5,
  [5000, 5000, 10000, 15000],
  (err) => {
    // We do not want to retry if a 4xx HTTP error is received
    if (err instanceof HttpError && err.status >= 400 && err.status < 500) {
      return true;
    }
    return false;
  }
```

nvie: I quite like this! Maybe we should make this PR part of `pagination-cleanups`, just to avoid any further merge conflicts there. We can just publish this fix a little bit later.

nvie: Oh, nevermind, I just saw you already did that! Perfect! 💯

packages/liveblocks-core/src/room.ts
```diff
@@ -2848,48 +2848,29 @@ export function createRoom<
...
-        requestedAt: new Date(json.meta.requestedAt),
-      };
-    } else if (response.status === 404) {
-      return {
-        threads: [],
-        inboxNotifications: [],
-        deletedThreads: [],
-        deletedInboxNotifications: [],
-        nextCursor: null,
-        requestedAt: new Date(),
```
nvie, line 2888: This code we had here looks a little bit weird to me so maybe indeed it's fine to handle it, but I just want to confirm there isn't an edge case we're overlooking with this and changing in an unexpected way. If this 404s, it most likely means the room does not exist, which would happen either if the room actually does not exist, or it was deleted.

In the new code, this edge case would throw an HttpError. Maybe this is fine, but a bit hard to tell without understanding the reason this code existed. Either way, if the code is still relevant, I would really love to see a unit test for this, just to capture this required behavior deliberately, which is currently not the case.

nvie (approved):
Pretty clean now! I left one comment about the handling of 404, which we should double-check, but otherwise this looks good to me! 👍

PR #1995, opened 2024-10-10
Refactorings towards a unified API client

nvie:
This DRYs up all the various HTTP fetch calls our client is making into a single "http client" abstraction, responsible for constructing, making, authorizing the request, and parsing/throwing the response. There used to be many various and subtly different ways we did this before, mostly copy-pasted and slightly adjusted. This PR mostly DRYs them all up. I've very carefully refactored things here, so it should not have introduced a breaking change.

I don't consider this fully done yet. This is step 1 of 2.

- **Step 1** (this PR): Have a single `HttpClient.fetch()` and `HttpClient.fetchJson()` method as the foundational building blocks for making HTTP calls from the client.
- **Step 2**: Move all possible API calls on there as methods, fully type-safe, and, make the `fetch` and `fetchJson` private methods.

The only "public" change here should be the removal of the `NotificationsApiError` and `CommentsApiError` classes, which are now both `HttpError` (which is more closely to what it actually is).

### DIFF
--- a/packages/liveblocks-core/src/http-client.ts
+++ b/packages/liveblocks-core/src/http-client.ts
@@ -5,7 +5,7 @@ import { urljoin } from "./lib/url";
 import { raise } from "./lib/utils";
 import { PKG_VERSION } from "./version";
 
-export class CommentsApiError extends Error {
+export class HttpError extends Error {
   constructor(
     public message: string,
     public status: number,
@@ -15,15 +15,9 @@ export class CommentsApiError extends Error {
   }
 }
 
-export class NotificationsApiError extends Error {
-  constructor(
-    public message: string,
-    public status: number,
-    public details?: JsonObject
-  ) {
-    super(message);
-  }
-}
+// TODO Deprecate these error aliases
+export const CommentsApiError = HttpError;
+export const NotificationsApiError = HttpError;
 
 export function getBearerTokenFromAuthValue(authValue: AuthValue): string {
   if (authValue.type === "public") {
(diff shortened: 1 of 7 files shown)


packages/liveblocks-core/src/http-client.ts
```diff
@@ -7,3 +22,117 @@ export function getBearerTokenFromAuthValue(authValue: AuthValue): string {
...
+      }
+    }
+
+    let body;
+    try {
+      body = (await response.json()) as T;
+    } catch {
+      // XXX This looks wrong 🤔 !
+      // XXX We should be throwing this error if something fails to parse.
+      body = {} as T;
```
nimeshnayaju, line 134: I'd also like an error thrown in this case.

packages/liveblocks-core/src/http-client.ts
```diff
@@ -7,3 +22,117 @@ export function getBearerTokenFromAuthValue(authValue: AuthValue): string {
...
+   *   4. Parse response body as Json
+   *   5. ...but silently return `{}` if that parsing fails
+   *   6. Throw HttpError if response is an error
+   */
+  // XXX Ultimately, might be nice to make this a private method? It might be much nicer if this method would _always_ deal with the JSON parsing.
+  public async fetch(
+    endpoint: URLSafeString,
+    options?: RequestInit,
+    params?: QueryParams
+  ): Promise<Response> {
```
nimeshnayaju, line 72: Like you mentioned, it'd be ideal for all methods to be using the `fetchJson()` method and `fetch` to be an implementation detail, but there can be some breaking changes if we do that change in a rush.

nimeshnayaju (approved):

packages/liveblocks-core/src/http-client.ts
nvie: Maybe this code exists because in some endpoints in the backends, we historically used to return an empty body (`''`), would would fail to parse, but I think the issue in that case is that we should not be doing that. I'll double-check this in the backend before removing this.

PR #1996, opened 2024-10-11
Refactor polling

nimeshnayaju:


packages/liveblocks-core/src/__tests__/Poller.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    const callback = jest.fn();
+    const poller = makePoller(callback, 1000); // 1000ms interval
+
+    poller.enable(true);
+
+    // Fast-forward until all timers have been executed
+    await jest.advanceTimersByTimeAsync(1000);
+
+    // Expect the callback to be called after the interval
+    expect(callback).toHaveBeenCalledTimes(1);
```
nvie, line 25: Setting up the poller is great, but should we ideally structure all tests in this way?

```tsx
const poller = makePoller(callback, 1000);
try {
  // Run all tests
} finally {
  poller.enable(false);
}
```

Just to ensure that polling in the background actually stops as well?

I really wish we could use `using` here already!

```
using poller = makePoller(callback, 1000);
// Just run all tests 💆‍♂️
```

nvie, line 25: Might be nice to add an assertion that after 500ms, the callback has not been invoked yet.

packages/liveblocks-core/src/lib/Poller.ts
```diff
@@ -23,32 +32,35 @@ export function makePoller(
 
   function poll() {
     if (context.state === "running") {
-      //
-      // XXX Should we make this contingent on the callback's execution, i.e. do
-      // XXX try { await callback() } finally { schedule() } instead, so that
-      // XXX two polling functions will never "overlap"?
+      // XXX Set a max timeout for the `callback()` (make `callback` take
+      // a signal, and protect each call with AbortSignal.timeout)
```
nvie, line 36: Yes! This is a must-have before we merge this change in, and we should unit test it properly.

packages/liveblocks-core/src/lib/Poller.ts
```diff
@@ -79,7 +91,31 @@ export function makePoller(
...
+    }
+
+    // If a poll is already in progress, do nothing
+    if (context.pollPromise !== null) {
+      return;
+    }
+
+    const lastPolledAt = context.lastPolledAt ?? 0;
+
+    if (performance.now() - lastPolledAt > maxStaleTimeMs) {
```
nvie, line 106: To me the conditional would be slightly more readable if we expressed it as:

```suggestion
    if (lastPolledAt + maxStaleTimeMs < performance.now()) {
```

nvie (approved):
Looking great! I love this new behavior, it makes so much more sense! Thank you for writing these excellent unit tests too! I just did a pass over them, read and played with them carefully and tried to break it without looking at the implementation first. I've also added some diagrams to make the tests a bit more readable at a glance.

It's all working as-expected, and I could not break it! 😇

Ensuring that a single poll cannot take longer than a certain max timeout is important to avoid the possibility of "freezing" the poller, but I don't think that's going to be much more work.

Excellent stuff!

packages/liveblocks-core/src/lib/Poller.ts
nvie: Actually, now that I'm looking at the implementation, I think there is a tiny edge case bug here. If there is no initial data, you _always_ want to poll, so you cannot use the `?? 0` construct here. I'll push a fix while at it.

nvie: Bug illustrated in cbc98aa.
Bug fixed in 5bb8a24.

nvie:
Closing in favor of #2002.

PR #2002, opened 2024-10-22
Refactor poller

nimeshnayaju:


packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -223,31 +223,43 @@ export function getLiveblocksExtrasForClient<M extends BaseMetadata>(
...
+        );
+      });
 
     return () => {
       pollerSubscribers--;
 
-      // NIMESH - When stopping the poller, we should also ideally abort its
+      // XXX - When stopping the poller, we should also ideally abort its
       // poller function, maybe using an AbortController? This functionality
       // should be automatic and handled by the Poller abstraction, not here!
```
nvie, line 264: Having second thoughts about this comment. In the current implementation, when the machine is in `@polling` state, it won't respond to `STOP` events — by design. I think it makes sense to not abort the currently-ongoing poll and just handle its response, even if there are no subscribers anymore by the time the response arrives. This means the current implementation is already correct, and I think we can safely remove this comment.

Do you agree, @nimeshnayaju?

packages/liveblocks-core/src/http-client.ts
```diff
@@ -113,16 +104,14 @@ export class HttpClient {
...
-        }
-        throw error;
+      let error: HttpError;
+      try {
+        const errorBody = (await response.json()) as { message: string };
+        error = new HttpError(errorBody.message, response.status, errorBody);
+      } catch {
+        error = new HttpError(response.statusText, response.status);
       }
+      throw error;
```
nvie, line 114: @nimeshnayaju I just removed this `if`-condition here. I don't think we have to check the status code at all. If the response is not OK, always turn it into an error.

packages/liveblocks-react/src/__tests__/useInboxNotifications.test.tsx
```diff
@@ -368,10 +368,15 @@ describe("useInboxNotifications: error", () => {
...
     await waitFor(() => expect(getInboxNotificationsReqCount).toBe(4));
+    expect(result.current).toEqual({ isLoading: true });
 
     // The fourth retry should be made after 10s
     await jest.advanceTimersByTimeAsync(15_000);
     await waitFor(() => expect(getInboxNotificationsReqCount).toBe(5));
+    expect(result.current).toEqual({
+      isLoading: false,
+      error: expect.any(Error),
+    });
```
nvie, line 379: @nimeshnayaju Thanks for fixing the unit tests! 🙏 I added in this extra assertion and noticed it doesn't pass. Maybe we should discuss on a call what we think the behaviour here should be, because I would expect this to transition to an error state at this point. Instead it remains in loading state. Maybe I'm missing something, or maybe the implementation isn't quite correct yet.

packages/liveblocks-react/src/__tests__/useInboxNotifications.test.tsx
```diff
@@ -547,12 +552,12 @@ describe("useInboxNotifications: polling", () => {
...
 
     // Wait for the first polling to occur after the initial render
-    jest.advanceTimersByTime(POLLING_INTERVAL);
+    jest.advanceTimersByTime(60_000);
     expect(initialCount).toBe(1);
     await waitFor(() => expect(pollerCount).toBe(1));
 
     // Advance time to simulate the polling interval
-    jest.advanceTimersByTime(POLLING_INTERVAL);
+    jest.advanceTimersByTime(60_000);
```
nvie, line 560: Let's avoid relying on runtime constants in unit tests. If we change the constant, I would want these tests to fail.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1326,19 +1319,9 @@ export class UmbrellaStore<M extends BaseMetadata> {
     query: ThreadsQuery<M> | undefined
   ) {
     const threadsFetcher = async (cursor?: string) => {
-      if (this._client === undefined) {
-        // TODO: Think about other ways to structure this. Throwing a StopRetrying only
-        // makes sense only if we can easily know if the fetcher is going to be wrapped inside the autoRetry function
-        throw new StopRetrying(
-          "Client is required in order to load threads for the room"
-        );
-      }
```
nvie, line 1335: I made `this._client` mandatory, so we don't have to worry about these edge cases anymore. This ugliness is now shifted to the unit tests. Turns out there were only two direct creations of `new UmbrellaStore()` left in the unit tests, so it really wasn't a big deal anymore.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -243,9 +243,17 @@ function makeDeltaPoller_Notifications(store: UmbrellaStore<BaseMetadata>) {
...
+      .waitUntilNotificationsLoaded()
+      .then(() => {
+        poller.enable(pollerSubscribers > 0);
+      })
+      .catch((err) => {
+        // XXX - Think about a better message and if we need to handle this differently
+        console.warn(
+          `Unable to start poller due to failure loading first page of inbox notifications: ${String(err)}`
+        );
+      });
```
nvie, line 257: Should this be structured like this instead?

```ts
void store
      .waitUntilNotificationsLoaded()
      .catch((err) => { … })
      .finally(() => {
        poller.enable(pollerSubscribers > 0);
      });
```

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -243,9 +243,17 @@ function makeDeltaPoller_Notifications(store: UmbrellaStore<BaseMetadata>) {
     // in the `then` body, check again if the number of subscribers if more than 0, and only then
     // if those conditions hold, start the poller
     poller.enable(pollerSubscribers > 0);
```
nvie, line 246: And I assume we'll want to remove this line here, right?

packages/liveblocks-react/src/liveblocks.tsx
nimeshnayaju: You're right! We should run the poller enabling logic on either success and failure. Would also like to spend some time thinking about the overall logic more holistically in our call.

packages/liveblocks-react/src/liveblocks.tsx
nimeshnayaju: Yes! Thanks for catching this!

packages/liveblocks-core/src/lib/Poller.ts
```diff
@@ -1,85 +1,217 @@
...
+   *        |                             |      ^
+   *        |                     on STOP |      | on START
+   *        |                             v      |
+   *   +--------+   if mayPoll()        +----------+      on POLL             +----------+
+   *   | decide |---------------------> | @enabled | -----------------------> | @polling |
+   *   +--------+                       +----------+   after POLL_INTERVAL    +----------+
+   *        ^                                                                      |
+   *        |                                                                      |
+   *        +----------------------------------------------------------------------+
+   */
```
marcbouchenoire, line 114: 😍

packages/liveblocks-core/src/lib/autoRetry.ts
```diff
@@ -1,6 +1,20 @@
...
 
+export class HttpError extends Error {
+  constructor(
+    public message: string,
+    public status: number,
+    public details?: JsonObject
+  ) {
+    super(message);
+  }
+}
```
marcbouchenoire, line 13: Should this live elsewhere maybe? Since `autoRetry` isn't `fetch`-specific. Same for the default `shouldStopRetrying` value, in addition to not retrying 4xx maybe it could also not retry `AbortError` and `TimeoutError` errors to support `AbortSignal`-powered APIs out-of-the-box.

PR #2003, opened 2024-10-22
Detect invalid chars in secret key

### LINKED ISSUE #1909: Cannot convert argument to a ByteString because the character at index 19 has a value of 8230 which is greater than 255. This happens when creating a new room
Error happened while creating a room: TypeError: Cannot convert argument to a ByteString because the character at index 19 has a value of 8230 which is greater than 255

My code is this and I facing an error in nextjs using clerk, while creating room with liveblocks.

export const createDocument = async ({
	userId,
	email,
}: CreateDocumentParams) => {
	const roomId = nanoid();

	try {
		const metadata = {
			creatorId: userId,
			email,
			title: "Untitled",
		};

		const usersAccesses: RoomAccesses = {
			[email]: ["room:write"],
		};
	
		const room = await liveblocks.createRoom(roomId, {
			metadata,
			usersAccesses,
			defaultAccesses: [],
		});

		console.log("after api endpoint", room);
		revalidatePath("/");

		return parseStringify(room);
	} catch (error) {
		console.log(`Error happened while creating a room: ${error}`);
	}
};

nvie:
This PR will update the validation of the provided secret key value in the Node client, by providing a more helpful error message if invalid chars (like `…`) are copy/pasted in there accidentally.

Fixes #1909.

### DIFF
--- a/packages/liveblocks-node/src/__tests__/Session.test.ts
+++ b/packages/liveblocks-node/src/__tests__/Session.test.ts
@@ -20,17 +20,17 @@ describe("authorization (new API)", () => {
         secret: 123,
       })
     ).toThrow(
-      'Invalid value for field "secret". Secret keys must start with "sk_". Please provide the secret key from your Liveblocks dashboard at https://liveblocks.io/dashboard/apikeys.'
+      "Invalid value for field 'secret'. Secret keys must start with 'sk_'. Please provide the secret key from your Liveblocks dashboard at https://liveblocks.io/dashboard/apikeys."
     );
   });
 
-  test("throws when an invalid secret key is provided", () => {
+  test("throws when an public key is provided instead of a secret key", () => {
     expect(() =>
       makeSession({
         secret: "pk_thisisntsecret",
       })
     ).toThrow(
-      'Invalid value for field "secret". Secret keys must start with "sk_". Please provide the secret key from your Liveblocks dashboard at https://liveblocks.io/dashboard/apikeys.'
+      "Invalid value for field 'secret'. Secret keys must start with 'sk_'. Please provide the secret key from your Liveblocks dashboard at https://liveblocks.io/dashboard/apikeys."
     );
   });
 
--- a/packages/liveblocks-node/src/__tests__/Session.test.ts
+++ b/packages/liveblocks-node/src/__tests__/Session.test.ts
@@ -34,6 +34,16 @@ describe("authorization (new API)", () => {
     );
   });
 
+  test("throws when an invalid secret key is provided", () => {
+    expect(() =>
+      makeSession({
+        secret: "sk_this…is…truncated",
+      })
+    ).toThrow(
+      "Invalid chars found in field 'secret'. Please check that you correctly copied the secret key from your Liveblocks dashboard at https://liveblocks.io/dashboard/apikeys."
+    );
+  });
+
   test("default set has no permissions", () => {
     expect(makeSession().hasPermissions()).toEqual(false);
   });
--- a/packages/liveblocks-node/src/utils.ts
+++ b/packages/liveblocks-node/src/utils.ts
@@ -1,5 +1,8 @@
 const DEFAULT_BASE_URL = "https://api.liveblocks.io";
 
+// Valid alphabet for secret/public keys
+const VALID_KEY_CHARS_REGEX = /^[\w-]+$/;
+
 export function getBaseUrl(baseUrl?: string | undefined): string {
   if (
     typeof baseUrl === "string" &&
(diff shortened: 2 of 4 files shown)


packages/liveblocks-node/src/__tests__/Session.test.ts
```diff
@@ -20,17 +20,27 @@ describe("authorization (new API)", () => {
         secret: 123,
       })
     ).toThrow(
-      'Invalid value for field "secret". Secret keys must start with "sk_". Please provide the secret key from your Liveblocks dashboard at https://liveblocks.io/dashboard/apikeys.'
+      "Invalid value for field 'secret'. Secret keys must start with 'sk_'. Please provide the secret key from your Liveblocks dashboard at https://liveblocks.io/dashboard/apikeys."
```
nvie, line 23: (Switching to single quotes in error messages while at it, as those are typically more readable when in a JSON message. We do this everywhere else too.)

packages/liveblocks-node/src/utils.ts
```diff
@@ -36,9 +47,15 @@ export function assertSecretKey(
...
   field: string
 ): asserts value is string {
-  if (!isNonEmpty(value) || !value.startsWith("sk_")) {
+  if (!startsWith(value, "sk_")) {
+    throw new Error(
+      `Invalid value for field '${field}'. Secret keys must start with 'sk_'. Please provide the secret key from your Liveblocks dashboard at https://liveblocks.io/dashboard/apikeys.`
+    );
+  }
+
+  if (!/^[\w-]+$/.test(value)) {
```
SugarDarius, line 56: NB: I will tend to put this RegExp in a const roughly named `SECRET_KEY_RG` and to add a bit of jsdoc to help developers understand what this RegExp represents and for what it's useful such as `Regular Expression to check if the secret key contains one or more word characters or hyphens`. 

It could be helpful for legacy and code intent knowledge sharing.

packages/liveblocks-node/src/utils.ts
```diff
@@ -36,9 +47,15 @@ export function assertSecretKey(
   value: unknown,
   field: string
 ): asserts value is string {
-  if (!isNonEmpty(value) || !value.startsWith("sk_")) {
+  if (!startsWith(value, "sk_")) {
```
SugarDarius, line 50: NB: Shouldn't we strengthen the condition with:
```ts
if (!startsWith(value, "sk_dev_") && !startsWith(value, "sk_prod_")) {
  // ....
}
```

Because if I'm not wrong as it's written it will pass with things like `sk_something_.....` no? Or do am I overthinking here?

SugarDarius (approved):
As far as I'm understanding the context looking good to me ✅ 
Just added two NB (Not Blocking) comments.

Thanks!

packages/liveblocks-node/src/utils.ts
nvie: The `sk_{dev,prod}` pattern is relatively new. I believe there may still exist older `sk_` values that won't have that format unless those keys would be rolled.

> Because if I'm not wrong as it's written it will pass with things like `sk_something_.....` no? Or do am I overthinking here?

In those cases, the key would be syntactically correct, but just not work because it's not a valid value. I think we're good in that case. The point of these checks is to avoid misconfigurations and provide guidance, not to catch all edge cases that aren't valid _today_. Adding a check like `sk_{dev,prod}` would prevent us from changing the key format in the future, so the check is deliberately somewhat loose.

SugarDarius: Oh okay! I wasn't aware of the previous key pattern. Thanks for sharing 🙏🏻 
Yes I agree then in those cases the check is good enough as it is 👍🏻

packages/liveblocks-node/src/utils.ts
nvie: Agreed, I'll put it in a constant. I normally do this as well. Not sure why I left it like this. Thanks for catching 👍  Fixed in e33f0d6ee2b171e1a4b1ad86c64f7dcf54295756.

packages/liveblocks-react/src/__tests__/useInboxNotifications.test.tsx
```diff
@@ -678,43 +678,37 @@ describe("useInboxNotifications: pagination", () => {
...
+    const thread2 = dummyThreadData({ roomId, createdAt: new Date("2021-01-02") }); // prettier-ignore
+    const thread3 = dummyThreadData({ roomId, createdAt: new Date("2021-01-03") }); // prettier-ignore
 
-    const inboxNotificationsPageOne = [
+    const inboxNotificationsPage1 = [
       dummyThreadInboxNotificationData({
         roomId,
-        threadId: threadOne.id,
+        threadId: thread1.id,
+        notifiedAt: new Date("2021-01-03"),
```
nvie, line 689: This unit test could be somewhat unstable, occasionally not passing. I already fixed this on another branch. It's unrelated to this PR, but I'm cherry-picking this fix here now, to ensure the unit tests are fixed on `main` ASAP.

PR #2005, opened 2024-10-22
Tweak package publish script

nvie:
Only push newly created Git commit with the version bump when on `main` branch. On branches/PRs this just causes needless noise. I found myself always reverting this commit immediately after publishing a new version, to avoid merge conflicts and PR noise.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -111,6 +111,16 @@ jobs:
           GIT_TAG: ${{ inputs.tag }}
           GITHUB_URL: "https://github.com/liveblocks/liveblocks"
         run: |
+          echo "==> Pushing changes to GitHub"
+          if [ "$(git current-branch)" = "main" ]; then
+            if !git push origin main; then
+                err "WARNING: Could not push this branch to GitHub!"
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -111,6 +111,16 @@ jobs:
           GIT_TAG: ${{ inputs.tag }}
           GITHUB_URL: "https://github.com/liveblocks/liveblocks"
         run: |
+          echo "==> Pushing changes to GitHub"
+          if [ "$(git current-branch)" = "main" ]; then
+            if !git push origin main; then
+                echo "WARNING: Could not push this branch to GitHub!" >&2
```
SugarDarius, line 117: NB: I would tend to use the word `ERROR` here instead of `WARNING`. To me from my understanding if we can't push then it means something is not right or somewhat right. So it shouldn't just be marked as warning as we're doing `echo "..." >&2` and we're exiting on `2` after.

I mean from my point of view it would be more meaningful.

SugarDarius (approved):

marcbouchenoire (approved):

PR #2006, opened 2024-10-22
Tiptap

CTNicholas:


packages/liveblocks-react-tiptap/src/mentions/MentionsList.tsx
```diff
@@ -0,0 +1,166 @@
...
+      style={{
+        position: strategy,
+        top: 0,
+        left: 0,
+        transform: `translate3d(${Math.round(x)}px, ${Math.round(y)}px, 0)`,
+        minWidth: "max-content",
+        display: props.hide ? "none" : "block"
+      }}>
+      <div className="lb-tiptap-suggestions-list lb-tiptap-mention-suggestions-list">
+        {suggestions === undefined ? <div className="item">Loading...</div> :
```
CTNicholas, line 144: We need to add the real loading placeholder here.

marcbouchenoire: In the Comments version, we don't display the dropdown at all while it loads so this check could be added to line 129 to render `null`

packages/liveblocks-react-tiptap/src/LiveblocksExtension.ts
```diff
@@ -0,0 +1,259 @@
...
+});
+
+export const useLiveblocksExtension = (): Extension => {
+  const room = useRoom();
+
+  // TODO: we don't need these things if comments isn't turned on...
+  // TODO: we don't have a reference to the editor here, need to figure this out
+  useCommentsErrorListener((error) => {
+    // If thread creation fails, we remove the thread id from the associated nodes and unwrap the nodes if they are no longer associated with any threads
+    if (error instanceof CreateThreadError) {
```
jrowny, line 81: NOTE: we have a ticket for this here: https://linear.app/liveblocks/issue/LB-1356/listen-for-comment-creation-fail-events

but didn't want to block the beta release

packages/liveblocks-react-tiptap/src/types.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { DecorationSet } from "@tiptap/pm/view";
+
+export const LIVEBLOCKS_MENTION_KEY = new PluginKey("lb-plugin-mention");
+export const LIVEBLOCKS_MENTION_PASTE_KEY = new PluginKey(
+  "lb-plugin-mention-paste"
+);
+export const LIVEBLOCKS_MENTION_NOTIFIER_KEY = new PluginKey(
+  "lb-plugin-mention-notify"
+);
+export const LIVEBLOCKS_MENTION_TYPE = "liveblocksMention";
```
nimeshnayaju, line 12: The node type used in Lexical is `lb-mention` and `lb-thread-mark`. Do you see any potential issues with the inconsistency? How do you feel about keeping the names the same?

jrowny: I don't see any issue as the two products won't ever mingle, their data is exclusive of each other.

marcbouchenoire (approved):
For the UI side, it's not a blocker so I'll create a ticket about it later for myself, but it would be nice to start re-using `<User />`, `<Avatar />`, and other duplicated components now that we can export things from `@liveblocks/react-ui/_private`.

CTNicholas (commented):
✅ Just to be clear, the examples and starter kit are finished. Probably best to leave all the _after merge_ tasks to me.

PR #2007, opened 2024-10-22
Stabilize failing e2e test

nvie:
This is done by adding some more synchronisations before reading the DOM. It's important, after pressing a button, to first observe the "isSynced" value to flip to `false`, before flipping it to `true`. I believe sometimes it was possible for the `waitUntil()` to observe a `true` value before it got flipped to `false`, effectively not making the test wait long enough for the synchronisation to complete.

Hopefully this makes the E2E test a bit more stable! 🤞


nimeshnayaju (approved):

marcbouchenoire:
Hopefully we don't need to allow even longer timeouts 🤞
Thanks for tackling this, for everyone's sanity!

PR #2016, opened 2024-10-25
Clean up HTTP client internals further

nvie:
This further cleans up HTTP client details in the client, a continuation of the work we already started here.

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -1624,21 +1624,6 @@ export function createRoom<
     delegates.authenticate()
   );
 
-  async function httpPostToRoom(
-    endpoint: "/send-message" | "/text-metadata",
-    body: JsonObject
-  ) {
-    return httpClient1.fetch(
-      endpoint === "/send-message"
-        ? url`/v2/c/rooms/${config.roomId}/send-message`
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -1624,47 +1626,28 @@ export function createRoom<
...
   async function createTextMention(userId: string, mentionId: string) {
-    return httpClient1.fetch(url`/v2/c/rooms/${config.roomId}/text-mentions`, {
-      method: "POST",
-      body: JSON.stringify({
-        userId,
-        mentionId,
-      }),
+    await httpClient1.rawPost(url`/v2/c/rooms/${config.roomId}/text-mentions`, {
+      userId,
+      mentionId,
```
nvie, line 1632: I intentionally did not introduce a behavioral change here. For each of these `.rawGet()`, `.rawPost()`, etc calls we currently do, we should ideally convert those to `.get()`, `.post()` etc on a case-by-case basis, and be sure that it's okay that those calls would throw the response in case of an error. If the call site is not expecting a throw, it may be breaking change, that's why I kept this around for now.

Ideally we convert them all, and then remove the `.rawXxx()` methods from the client.

PR #2023, opened 2024-10-31
yjs package, offline support

jrowny:
Adds offline support to the YJS package, allowing documents to load even faster.


nvie (approved):
Nice and simple implementation by deferring the details to an external package! We've been thinking about offering something similar for threads/comments loading as well.

I would be slightly more comfortable if we added some tests around this to cover a few important edge cases, to prevent future regressions in this area.

In particular, I'd love to see a test that will check that making changes to the document locally before the initial server data has loaded will just work, as well as when indexedDB isn't supported/enabled/allowed by the browser host.

Thanks for adding this speed-up!

PR #2028, opened 2024-11-02
Refactor `Thread` component to be used outside of `RoomProvider`

nimeshnayaju:
This PR refactors the `Thread` component to be usable outside of `RoomProvider`. The `Thread` component internally uses `useRoom()` hook to refer to the room id wherever necessary, which makes it currently unusable outside of `RoomProvider`.


packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
```diff
@@ -1016,11 +1037,12 @@ const ComposerForm = forwardRef<HTMLFormElement, ComposerFormProps>(
...
-      const canComment = self?.canComment ?? true;
-
-      return isSubmitting || disabled || !canComment;
-    }, [isSubmitting, disabled, room]);
+      // XXX - Discussion: Should we check for scopes here? Scopes are currently part of the room state message
+      // XXX - received after room websocket connection, but maybe we enable the composer submission for
+      // XXX - all users. The request will be rejected by the server, so any optimistic update created
+      // XXX - locally will be reverted!
+      return isSubmitting || disabled === true;
+    }, [isSubmitting, disabled]);
```
nimeshnayaju, line 1045: @marcbouchenoire Because we are no longer under `RoomContext` and this information about whether a user can comment or not is extracted from the room state machine received after WebSocket connection is established, my understanding is that we are not able to disable the composer form if the user in fact does not have scopes to write a comment.

How do you feel about loosening the requirements for the composer form to be disabled? The backend endpoint will reject the update, so comment will be added/removed optimistically but will be immediately reverted.

marcbouchenoire: It's not ideal because the delayed removal won't give any feedback about why so it will feel like a bug, but yeah if we can't have this info I don't see another choice 🤔

packages/liveblocks-core/src/http-client.ts
```diff
@@ -1,11 +1,1387 @@
...
+      options.settings
+    );
+  }
+
+  // XXX - Batch for a room will remain in memory even when a room is removed from client.
+  // XXX - Ensure this does not cause issues when the room is remounted, etc.
+  const markInboxNotificationsAsReadBatchByRoom: Map<
+    string,
+    Batch<string, string>
+  > = new Map();
```
nimeshnayaju, line 957: @nvie @marcbouchenoire This batch was previously created as part of the `createRoom` function and was therefore removed from memory when the `Room` object was removed. We had to move this batch from room to the http client for this refactor. We now have a map of room id to the inbox notifications batch but this means, when a room is unmounted, the batch isn't removed automatically. I am not very familiar with the inner workings of a batch myself, so I was curious if you can think of any consequences of doing this.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1475,56 +1505,82 @@ function useEditThreadMetadata<M extends BaseMetadata>() {
...
+  // XXX - with `useEditComment` hook. Note: This would be a breaking change!!!
   return React.useCallback(
-    ({ threadId, body, attachments }: CreateCommentOptions): CommentData => {
+    ({ threadId, body, attachments }: CreateCommentOptions): void => {
       const commentId = createCommentId();
       const createdAt = new Date();
 
+      const userId = client[kInternal].currentUserIdStore.get();
+      // XXX - Should we display a console warning and return early if the user id could not be found?
+      if (userId === undefined) return;
```
nimeshnayaju, line 1522: @marcbouchenoire @nvie What are your thoughts on how we should handle the lack of user id or thread in cache? 

This method currently (in `main`) returns the optimistically created `CommentData` object synchronously. But, if we do decide to return early (due to absence of user id or thread in cache), we'll be forced to update the signature of this method to return void. I think it is okay for this method to return a void object (the `useEditComment` hook returns a method that returns void too), but it would be a breaking change. Curious to hear your thoughts on this!

packages/liveblocks-core/src/http-client.ts
marcbouchenoire: The batching window for these is fairly small (50ms) so I don't think there are big consequences. Even in the scenario where somehow a batching window is opened then a Room is killed before the window is closed, I think it's still fine the make the batched call to the endpoint?

packages/liveblocks-react/src/room.tsx
marcbouchenoire: Could we mimic what happens today with a public key scenario? What does the optimistic `CommentData` look like? What about the real server one? I don't think it's a warning/error case since we support it already via the public key today, even though it's not really useful.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1299,17 +1306,24 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
         result.inboxNotifications
       );
 
+      this._store.set((state) => ({
+        ...state,
+        roomAccesses: {
+          ...state.roomAccesses,
+          [roomId]: result.roomAccess,
+        },
+      }));
```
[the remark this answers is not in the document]
nvie: > only replacing the scopes if the new scope provides higher (write) permissions to the user?

I don't think we'll quickly run into issues with the current implementation, but I think it will be a bit better (more defensive) to implement this in a way that will only ever grow permissions. As if you're throwing all permissions you're ever seeing for a room into a set but never take anything away from there.

The reason is that in theory users can have multiple auth tokens in their auth manager that offer some access to the same room. For example, suppose users have Access Tokens A and B in their local cache:

1. Access token A offers `"engineering:room-a": ["room:write"]`
1. Access token B offers `"engineering:*": ["room:read"]`

Now, if you read threads for room `engineering:room-a`, then the permission hint for that room will be different depending on which access token you use to authorize that request.

This is why I think only ever adding permission hints to the existing set is best: if you have ever seen proof that you have write permission to a room: let's assume (and keep assuming) they have write access. If you see `["room:read"]` at a later time, we should definitely not overwrite the hint. We may still have write access to that room!

It's the job of the auth manager to give you the most permissive token in the cache for a given outgoing request anyway, so we should be able to assume that if the user is creating a new thread in room `engineering:room-a`, they will do so with token A (as provided by the auth manager), whereas if creating a new thread for room `engineering:room-bar` they will be handed token B to do so.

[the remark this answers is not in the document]
nimeshnayaju: Thank you for the thorough explanation 🙏🏼 Definitely a tricky one to get absolutely right. I have updated the logic to only ever add permission to the permission set. So, if a `room:write` permission has been recorded for a room and a new `room:read` permission comes along, we do not replace the permission set with the single `room:read` permission but rather add the `room:read` permission too, so that the new permission set for the room becomes `["room:write", "room:read"]`. It felt simpler to do this. 

```ts
private updateRoomPermissions(permissions: Record<string, Permission[]>) {
  const permissionsByRoom = this._store.get().permissionsByRoom;

  Object.entries(permissions).forEach(([roomId, permissions]) => {
    // Get the existing set of permissions for the room and only ever add permission to this set
    const existingPermissions = permissionsByRoom[roomId];

    if (existingPermissions === undefined) {
      permissionsByRoom[roomId] = new Set(permissions);
    } else {
      permissions.forEach(existingPermissions.add);
    }
  });

  this._store.set((state) => ({
    ...state,
    permissionsByRoom,
  }));
}
```

packages/liveblocks-react-ui/src/components/Composer.tsx
```diff
@@ -706,3 +727,100 @@ export const Composer = forwardRef(
...
+    if (commentId === undefined) {
+      throw new Error(
+        `Cannot create a comment on thread ${threadId} because the thread does not exist in cache.`
+      );
+    } else {
+      throw new Error(
+        `Cannot edit comment ${commentId} on thread ${threadId} because the thread does not exist in cache.`
+      );
+    }
+  }
```
nimeshnayaju, line 783: @marcbouchenoire We throw an error if `Composer` is passed a thread id (and/or comment id) that doesn't exist in the cache. So,

```tsx
return <Composer threadId={"th_1"} commentId={"cm_1"} />
```

will throw an error if the thread id or comment id do not exist in cache.

A side-effect of this change is that a `Thread` component is also passed a thread data and if the associated thread doesn't exist in cache. So,

```tsx
return <Thread thread={thread} />
```

How do you feel about this change?

packages/liveblocks-react-ui/src/components/Composer.tsx
```diff
@@ -555,15 +594,24 @@ export const Composer = forwardRef(
...
     const { preventUnsavedComposerChanges } = useLiveblocksUIConfig();
     const hasResolveMentionSuggestions =
       client[kInternal].resolveMentionSuggestions !== undefined;
     const isEmptyRef = useRef(true);
     const isEmojiPickerOpenRef = useRef(false);
     const $ = useOverrides(overrides);
+    const permissions = useRoomPermissions(useComposerRoomId());
+    const canComment =
+      permissions.has(Permission.CommentsWrite) ||
+      permissions.has(Permission.Write);
```
nimeshnayaju, line 606: I introduced a new internal/private hook called `useRoomPermissions` that returns the set of permissions for the provided room. We use this set of permissions to check the composer should be disabled or not.

packages/liveblocks-core/src/auth-manager.ts
```diff
@@ -35,7 +35,8 @@ export type AuthenticationOptions = {
 );
 
 export function createAuthManager(
-  authOptions: AuthenticationOptions
+  authOptions: AuthenticationOptions,
+  onAuthenticate?: (token: AuthToken) => void
```
nimeshnayaju, line 39: I added a simple `onAuthenticate` callback to the `createAuthManager` function. We update the `currentUserIdStore` on each call to this callback.

```ts
const currentUserIdStore = createStore<string | undefined>(undefined);

const authManager = createAuthManager(options, (token) => {
  const userId = token.k === TokenKind.SECRET_LEGACY ? token.id : token.uid;
  currentUserIdStore.set(() => userId);
});
```

packages/liveblocks-core/src/client.ts
```diff
@@ -158,43 +154,19 @@ export type InternalSyncStatus = SyncStatus | "has-local-changes";
...
-    inboxNotifications: {
-      updated: InboxNotificationData[];
-      deleted: InboxNotificationDeleteInfo[];
-    };
-    threads: {
-      updated: ThreadData<M>[];
-      deleted: ThreadDeleteInfo[];
-    };
-    requestedAt: Date;
-  }>;
```
nimeshnayaju, line 188: All of these http methods are also part of return type of `createHttpClient` method, so these are removed in favour of the `httpClient` internal property, which also holds a bunch of other http methods like `getThreads({ roomId, ... }`, `getInboxNotifications({ roomId, ... })`

packages/liveblocks-react/src/room.tsx
```diff
@@ -1477,54 +1498,74 @@ function useEditThreadMetadata<M extends BaseMetadata>() {
...
+      const { store, onMutationFailure } = getRoomExtrasForClient(client);
+
+      const existing = store
+        .getFullState()
+        .threadsDB.getEvenIfDeleted(threadId);
+      if (existing === undefined) {
+        throw new Error(
+          `Cannot create comment in thread "${threadId}" because the thread does not exist in cache.`
+        );
+      }
```
nimeshnayaju, line 1518: @marcbouchenoire @nvie 

We now throw an error if the thread id passed to the function returned by `useCreateComment` (or `useEditComment`, `useEditThreadMetadata`, `useAddReaction`, etc.) doesn't map to a thread in cache. 

Also, to be consistent with the old behavior, if user id is not defined, we set the user id to `anonymous`

packages/liveblocks-react/src/room.tsx
```diff
@@ -1311,8 +1309,16 @@ function useCommentsErrorListener<M extends BaseMetadata>(
...
 ) => ThreadData<M> {
+  return useCreateRoomThread(useRoom().id);
+}
+
+/**
+ * @private
+ */
+function useCreateRoomThread<M extends BaseMetadata>(
+  roomId: string
+): (options: CreateThreadOptions<M>) => ThreadData<M> {
```
nimeshnayaju, line 1320: You'll notice that these room hooks haven't changed much. I only refactored them into a simpler hook that accepts a room id and does all the same things that it used to do previously. These simpler hooks are also used by our Comment and Composer component so that they can be used outside of RoomProvider.

packages/liveblocks-react-ui/src/components/Composer.tsx
```diff
@@ -555,9 +570,53 @@ export const Composer = forwardRef(
...
+      // Default to room context (creating a new thread)
+      else if (room) {
+        return room.id;
+      }
+
+      // No valid context found
+      throw new Error(
+        "Composer must be a descendant of RoomProvider component"
+      );
+    }
```
nimeshnayaju, line 613: This `getRoomId()` method (in `Composer` component) does a few things to figure out which room id to use. 

In summary, if `commentId` and `threadId` are both passed to the component, it tries to obtain the room id from the comment context. If comment data is not found, then it tries to obtain the room id from the thread context. If neither are found, it tries to look at the room context.

1. In the following example, `Composer` is a descendant of `Comment`, so the composer will use the room id obtained from the comment context.

    ```tsx
    <Comment comment={comment}>
      <Composer commentId={comment.id} threadId={thread.id} /> 
    </Comment />
    ```

2. In the following example, `Composer` is a descendant of `Thread` but not of `Comment`, so the composer will use the room id obtained from the thread context.

    ```tsx
    <Thread thread={thread}>
      <Composer threadId={thread.id} /> 
    </Comment />
    ```

3. In the following example, `Composer` neither a descendant of `Thread` nor of `Comment`, so the composer will use the room id obtained from the RoomProvider context.

    ```tsx
    <RoomProvider roomId={roomId}>
      <Composer /> 
    </Comment />
    ```

If none of `RoomProvider`, `Thread` or `Comment` context were found, we throw an error.

We also do some extra check to ensure that even if a `Composer` is a descendant of a `Thread` component, the thread id of the thread context must match the `threadId` passed to the `Composer`, otherwise, we fall back to the room context.

packages/liveblocks-react-ui/src/components/Composer.tsx
```diff
@@ -555,9 +570,53 @@ export const Composer = forwardRef(
...
+
+    const roomId = getRoomId();
+
+    const createThread = useCreateRoomThread(roomId);
+    const createComment = useCreateRoomComment(roomId);
+    const editComment = useEditRoomComment(roomId);
```
nimeshnayaju, line 619: Notice these private hooks that don't internally rely on `useRoom()` hook. This made the refactoring easier to reason about. The only thing we had to worry about was figuring out where the `roomId` should come from, and that is described here: https://github.com/liveblocks/liveblocks/pull/2028/files#r1843300206

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
```diff
@@ -1001,13 +1028,45 @@ const ComposerForm = forwardRef<HTMLFormElement, ComposerFormProps>(
...
+      // If Composer.Form is within a Room context (creating a new thread)
+      else if (room) {
+        return room.id;
+      }
+
+      // If none of the contexts are available, throw an error
+      throw new Error(
+        "Composer.Form must be a descendant of RoomProvider, Thread, or Comment component."
+      );
+    }
```
nimeshnayaju, line 1062: The process of obtaining room id for a `Composer.Form` is simpler than obtaining one for `Composer` because unlike `Composer`, `Composer.Form` doesn't accept props like `threadId` and `commentId`. So, we'd simply check if the `Composer.Form` is a descendant of `Composer`, in that case, we use the room id passed from `Composer`, otherwise, we check for the thread, comment and room context in order. If no context is available, we throw an error.

packages/liveblocks-core/src/client.ts
```diff
@@ -546,7 +521,22 @@ export function createClient<U extends BaseUserMeta = DU>(
   );
   const baseUrl = getBaseUrl(clientOptions.baseUrl);
 
-  const authManager = createAuthManager(options);
+  const currentUserIdStore = createStore<string | undefined>(undefined);
```
nvie, line 524: Minor nit, but while we're touching this, I think it might be nice not to return the entire store instance in `[kInternal]`, but only expose the "getter":

```tsx
{
  [kInternal]: {
    // currentUserIdStore,                     // ❌
    getCurrentUserId: currentUserIdStore.get,  // ✅
}
```

packages/liveblocks-core/src/client.ts
```diff
@@ -695,6 +686,9 @@ export function createClient<U extends BaseUserMeta = DU>(
   function logout() {
     authManager.reset();
 
+    // Reset the current user id store when the client is logged out
+    currentUserIdStore.set(() => undefined);
```
nvie, line 690: 👍

packages/liveblocks-core/src/http-client.ts
```diff
@@ -1,11 +1,1400 @@
...
+  InboxNotificationDeleteInfoPlain,
+} from "./protocol/InboxNotifications";
+import type { IdTuple, SerializedCrdt } from "./protocol/SerializedCrdt";
+import type { HistoryVersion } from "./protocol/VersionHistory";
+import type { TextEditorType } from "./types/Others";
+import type { Patchable } from "./types/Patchable";
+import type { RoomNotificationSettings } from "./types/RoomNotificationSettings";
 import { PKG_VERSION } from "./version";
 
+export interface RoomHttpApi {
```
nvie, line 51: Same as comment above. Ideally we'd bind `<M>` here at the group level, as there will be less room for errors that way, and we can benefit from the inference.

packages/liveblocks-core/src/http-client.ts
```diff
@@ -179,12 +1575,14 @@ export class HttpClient {
    */
   public async post<T extends JsonObject>(
     endpoint: URLSafeString,
+    authCallback: () => Promise<AuthValue>,
```
nvie, line 1578: As I shared before, I don't really like this approach of passing an auth callback to _every_ method here. To me, it defeats the purpose of having a (single) auth callback (and in a way a good part of the purpose of this HttpClient abstraction).

At this point, if you need to pass a callback into every call, you might as well just pass the `AuthValue` as a param directly and let the call sites figure it out. The reason I added this HttpClient is to avoid call sites from having to think/know about auth.

So I still think there is some value in trying to assign the auth callback in the constructor. If that's not possible for whatever reason, I'd love to understand that reason and I'd be happy to help brainstorm a potential other solution for this.

packages/liveblocks-core/src/http-client.ts
```diff
@@ -1,11 +1,1400 @@
...
+    threads: {
+      updated: ThreadData<M>[];
+      deleted: ThreadDeleteInfo[];
+    };
+    requestedAt: Date;
+    permissionHints: Record<string, Permission[]>;
+  }>;
+}
+
+export function createHttpClient({
```
nvie, line 401: We need to think about naming here a bit. `new HttpClient` and `createHttpClient` sound similar, but one creates an abstract thing that can be used to query (any) backend endpoint, and one uses that thing to expose type-safe helpers.

Maybe rename `createHttpClient` to `createRoomAPI` or `createRoomBackendAPI` or something?

packages/liveblocks-core/src/http-client.ts
```diff
@@ -1,11 +1,1400 @@
...
+  baseUrl,
+  authManager,
+  fetchPolyfill,
+}: {
+  baseUrl: string;
+  authManager: AuthManager;
+  fetchPolyfill: typeof fetch;
+}): ClientHttpApi {
+  async function getAuthValueForRoom(roomId: string) {
+    // TODO: Use the right scope
```
nvie, line 411: I've seen this comment before, so I know you copied it. Do you know what was meant by this?

packages/liveblocks-react-ui/src/components/Composer.tsx
nvie: Wow, a pretty original solution that I did not expect!

I can see how this solution works indeed for the problem at hand, and how it helped you land on the implementation. But I do worry a bit about relying on React contexts for too many things. My areas of worry are: authority, consistency, performance, and (future) use outside of React.

**Authority**
Currently, we rely on a parent component to set/provide the right value. If that component somehow sets an invalid value, or ever fails to update or reset the value correctly in certain edge cases.

**Consistency**
It's technically possible to create a state like this (bare with me if I didn't get it exactly right here):

```tsx
<ComposerRoomIdContext value="room-a">
  <ThreadDataContext value="unrelated-thread-id">
  </ThreadDataContext>
</ComposerRoomIdContext>
```

If the `room-a` and `unrelated-thread-id` form (or can form) an invalid "pair", then the category of bugs that would end up as a result would be extremely hard to reason about.

**Performance**
Not the biggest worry, but stacking so many extra components in the component hierarchy is probably not going to help with performance (and ease of understanding).

### How else?

You're way deeper into the weeds here, but from my (maybe naive) outside understanding of this problem, I'd be inclined to let each component/hook simply ask the central authority which `roomId` a certain `threadId` or `commentId` belongs to. This data is readily available from the pool maintained by the Umbrella store already. Would this be an option here? I feel like it could make this implementation much simpler.

I'm also on this path of thought because of this somewhat loosely related issue LB-1379 we also need to solve for.

So what about we would allow these hooks to do something like this instead?

```tsx
const client = useClient();
client[kInternal].getRoomIdForResource("th_123");  // or
client[kInternal].getRoomIdForResource("cm_123");
```

(Or we make those two lookups, I don't care.) But the point is to ask the authority basically. These lookups could be indexed, and would be super fast O(1), so without any performance overhead or all that component tree stacking. And it would always be consistent/correct, no matter how your state changes.

Happy to talk more about this!

packages/liveblocks-react/src/umbrella-store.ts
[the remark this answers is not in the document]
nvie: Yes, perfect! 👍

packages/liveblocks-core/src/client.ts
marcbouchenoire: @nvie Usually we also need the "susbcriber" in React land to build a reactive hook out of it

nvie: Oh yeah. I think in this case we're not using that though?

marcbouchenoire: We are 👀 https://github.com/liveblocks/liveblocks/pull/2028/files#diff-bc0bfa8da3e750647522fb8b3960dffbe94655e1679cb8978f894bb2c2696254R5

nvie: Oh, nevermind then! I said nothing. This morning, I only checked `react` I think, not `react-ui`. Carry on please! 🙈

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
marcbouchenoire: This does make the primitives somewhat linked to the default components and less independent, could something like an `@internal` `roomId` prop on `Composer.Form` prop also work? Which would be populated by `Composer`.

packages/liveblocks-core/src/http-client.ts
marcbouchenoire: +1, I also find the name "HTTP client" too confusing for what it does, in my ears it's something super low-level like a `fetch` alternative API.

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
nvie: +1, I like this approach. Having the props passed in makes this implementation simpler and more direct. If they're not passed in, we can look it up from the `RoomProvider`. If not, it's an error.

packages/liveblocks-core/src/http-client.ts
nimeshnayaju: Unfortunately, I do not. Guillaume might be the best person to ask about this. I asked Marc too and we both are unsure what the correct scope would be here, if it is not already.

packages/liveblocks-core/src/http-client.ts
nimeshnayaju: We previously used to have a `HttpClient` objects per room (actually two, each with a different auth callback on the constructor). I wanted to keep a single http client for the whole client and use that client to make API calls but that might have been a premature thought.

```ts
const httpClient2 = new HttpClient(config.baseUrl, fetchPolyfill, () =>
  // TODO: Use the right scope
  await authManager.getAuthValue({
    requestedScope: "comments:read",
    roomId,
  });
);
```

If we still want to keep the auth callback on the constructor, we could have a map of `HttpClient` by room but not sure if that would make understanding things difficult. I'd love to discuss more on how we can get the best of all worlds.

nvie: Would it work if we had the following setup?

1. A generic `HttpClient` with an auth callback in the constructor.
2. Have a way of creating a new `HttpClient` from an existing one, using the builder pattern, e.g. `httpClient.withAuth(newCallback)`, returning a new HttpClient instance with a different auth callback.
3. Assign those to each room when it's created.

This way, you end up with an HttpClient instance per room (or one for all rooms), and you still reap the benefits: call sites using the `httpClient` never have to think/know/worry about auth. And which auth method is used by which httpClient instance is also still easy to follow: just look at where the value is created. That's where the auth callback lives.

Let's discuss this idea more in-depth on our call later today 👍

PR #2031, opened 2024-11-04
Return empty threads list if `getThreads` response was a 404

nimeshnayaju:
We introduced a regression in version `2.10.1` by not correctly handling 404 responses when calling `get(Room)Threads` endpoint. This PR reverts that change so that a `404` response is interpreted as an empty list of threads. One caveat (that we also had prior to 2.10.1) is that we use a fake `requestedAt` date created on the client to use for the poller. This is not ideal and would require some deeper thoughts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,13 @@
-## 2.11.1
-
-### `@liveblocks/react`
-
-- Fix regression with `useThreads` that caused the hook to return an error if
-  its associated room did not exist.
-
 ## 2.11.0
 
 ### `@liveblocks/react-ui`
 
 - Upgrade dependencies.
+
+### `@liveblocks/react`
+
+- Fix regression with `useThreads` that caused the hook to return an error if
+  its associated room did not exist.
 
 ## 2.10.2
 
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -2870,29 +2870,42 @@ export function createRoom<
...
-    });
+    } catch (err) {
+      if (err instanceof HttpError && err.status === 404) {
+        return {
+          threads: [],
+          inboxNotifications: [],
+          nextCursor: null,
+          requestedAt: new Date(),
+        };
+      }
```
nimeshnayaju, line 2905: We catch a `404 HttpError` and return an empty threads list and return a fake `requestedAt` date.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 2.11.1
```
marcbouchenoire, line 1: It possibly could go out with `2.11.0` depending on the release schedule (it's not out yet)

nimeshnayaju: Good point and done! Thank you @marcbouchenoire

marcbouchenoire (approved):
Thanks for the quick fix @nimeshnayaju 🙇

PR #2032, opened 2024-11-05
Improve pasting into composers again

marcbouchenoire:
[The last round of pasting improvements](https://github.com/liveblocks/liveblocks/pull/2009) introduced a few silly mistakes (oh how I wish we could create pasting E2E tests), this PR fixes them.

In commit order:
1. When copy-pasting Slate text, we were not leveraging the Slate data and instead were using its serialized HTML form instead, which results in information loss for custom nodes like mentions but also would never match what was copied 1 to 1 (unlike the Slate data)
2. On error we now fallback to the default behavior but we were invoking it twice
3. When deserializing HTML, we now look at spans to guess non-semantic rich text (bold, italic, etc), but we were doing it in a silly way which didn't support multiple attributes on the same span and most importantly we were not doing anything in case of no attribute so we were losing all plain text nodes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Upgrade dependencies.
 - Fix minor appearance issues related to attachments.
+- Fix pasting issues introduced in 2.10.0.
 
 ## 2.10.2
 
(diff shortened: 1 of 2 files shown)


CTNicholas (approved):
I don't know Slate too well, but this appears to make sense 👍

nimeshnayaju (approved):

PR #2035, opened 2024-11-06
Work around for lockfile issue

nvie:


marcbouchenoire (approved):
Seems sufficiently safe and specific to me, thanks! Hopefully we'll find the actual reason at some point.

PR #2036, opened 2024-11-06
Document a small hack

nvie:
…and allow for a possible 6 hour clock time difference buffer.

Follow-up on #2031.


nimeshnayaju (approved):
Thank you 🙏

Issue #2010, opened 2024-10-23
Error while resolving a comment on Lexical editor


barbaldo:
**Describe the bug**

I implemented the solution outlined in the following docs: https://liveblocks.io/dashboard/LIhWQjZ5InMJbK7118hHZ/projects/66a8c337839b83af7783f27a/quickstart and 
https://liveblocks.io/docs/api-reference/liveblocks-react-lexical#AnchoredThreads it all works pretty well however whenever I solve a thread from inside the editor I get this: Unexpected Application Error!
Failed to execute 'observe' on 'ResizeObserver': parameter 1 is not of type 'Element'.

**To Reproduce**

1. Follow the aforementioned docs 
2. Comment anything
3. try to solve said comment

**Expected behavior**

The thread should just disappear

**Additional context**

As with my previous issues, I may be just using Liveblocks wrong or making a mistake, if that's the case I apologize in advance.

jrowny:
At first glance, I don't think you're using it wrong, I think this may be a bug. We will look into it!

barbaldo:
@jrowny fantastic! If you need any more info hit me up!

barbaldo:
Also I just discovered that the same exact behaviour happens in the following situation:

- I have an editor with text and comments inside
- Ctrl + A to select everyhting
- Delete to delete everything
- Same error outlined in the original thread

This brings me to a question I have, is there a way in Lexical to detect if you're deleting commented text? That would be super useful

nimeshnayaju:
Hey @barbaldo, thank you for the extra reproduction steps. We'll look into this shortly and will reach out if there is more information we may need from you. Thank you for reporting the issue 🙏🏼

barbaldo:
Hey @nimeshnayaju , thanks for looking into it! Is is there any news?

barbaldo:
I'd like to add another situation with the same error, basically if you add text and comments, if you try to resolve a comment from the <AnchoredThreads> (Resolve meaning you click the little check icon) it generates the same error

nimeshnayaju:
Hey @barbaldo, I tried reproducing the steps you outlined but wasn't able to. Do you think you could share a screencast and the error log you see? 


https://github.com/user-attachments/assets/d8ed6a64-a5fd-40b8-801b-2964ade2ef68

barbaldo:
@nimeshnayaju trying to upload a screen recording but github is not collaborating....

barbaldo:
[...]

PR #2041, opened 2024-11-06
Move anchored thread lifecycle inside `useLayouEffect` instead of `ref`

nimeshnayaju:
We currently add/remove each anchored thread in a `ref` callback but ref callback, which doesn't seem to work correctly on all React versions (https://github.com/liveblocks/liveblocks/issues/2010). We now add/remove each anchored thread in a `useEffect` hook. The behavior remains the same but is more consistent overall.

```
Unexpected return value from a callback ref in div. A callback ref should not return a function.
```

### DIFF
--- a/packages/liveblocks-react-lexical/src/comments/anchored-threads.tsx
+++ b/packages/liveblocks-react-lexical/src/comments/anchored-threads.tsx
@@ -314,7 +314,7 @@ function ThreadWrapper({
     });
   }
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     const el = divRef.current;
     if (el === null) return;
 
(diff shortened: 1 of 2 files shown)


nvie (approved):
Nice fix! Should we also mention this fix in the changelog?

packages/liveblocks-react-lexical/src/comments/anchored-threads.tsx
```diff
@@ -313,17 +314,19 @@ function ThreadWrapper({
...
   }
 
-  const handleRef = useCallback(
-    (el: HTMLDivElement) => {
-      onItemAdd(thread.id, el);
-      return () => onItemRemove(thread.id);
-    },
-    [thread.id, onItemAdd, onItemRemove]
-  );
+  useEffect(() => {
```
marcbouchenoire, line 317: We could use `useLayoutEffect` here instead since that would be closer to a ref callback in the lifecycle

nimeshnayaju: Good idea! Also added a CHANGELOG like @nvie suggested.

PR #2042, opened 2024-11-07
Add support of Text Mention notification event in @liveblocks/emails

SugarDarius:
Add a set of functions and utilities to help developers build email templates for the Text Mention notification event 👇🏻

- Exposes `prepareTextMentionNotificationEmailAsReact` from `@liveblocks/emails`: it takes a webhook `TextMentionNotificationEvent` and prepare data to be used in an email and converts extracted document content fragment surrounding a mention as React.
- Exposes `prepareTextMentionNotificationEmailAsHtml` from `@liveblocks/emails`: it takes a webhook `TextMentionNotificationEvent` and prepare data to be used in an email and converts extracted document content fragment surrounding a mention as html.

ℹ️ Naming is not set in stone, if you have better understandable suggestions I'd love to have your feedback and insights 🙏🏻
ℹ️ Two examples have added for now:
- `examples/nextjs-lexical-emails-resend` → DX example when using `prepareTextMentionNotificationEmailAsReact` with `Resend` for Lexical text editor
- - `examples/nextjs-tiptap-emails-resend` → DX example when using `prepareTextMentionNotificationEmailAsReact` with `Resend` for Tiptap text editor

Screenshot 2024-11-07 at 09 41 56

<hr />
### Important informations
This branch has been created from `email-templates-functions-and-utilities` before it was merged into `main`. So the list of commit is quit very long. However the files changed are correct and show only the files changed corresponding to add the support of `TextMentionNotificationEvent`.

Thanks 🙏🏻


packages/liveblocks-node/src/client.ts
```diff
@@ -125,6 +137,7 @@ export type RoomData = {
   usersAccesses: RoomAccesses;
   groupsAccesses: RoomAccesses;
   metadata: RoomMetadata;
+  experimental_textEditor?: RoomTextEditor;
```
FlowFlorent, line 140: Just curious what this is, and the reason it's at the room level. 

Does our system allow for a room to have both lexical / tiptap?

In which case, should this be something that is stored somewhere else than the room (but with a relation to the room)

Really coming without knowledge so feel free to point me to documentation or anything 😄

SugarDarius: Hey @FlowFlorent 👋🏻 

The property [`textEditor`](https://github.com/liveblocks/liveblocks-backend/blob/074d088db168ee44832a7237d04316241c10267e/apps/cloudflare/src/mongo/types.ts) is on the `MongoRoom` DB object since 5 month now from what I see. 
As you can see it's made to tell if a room as an associated text editor (lexical or tiptap but never both) or not. I think you should ask @jrowny if you want more info about why it's at the root level 🙂 
You can see the text editor is reported in [LiveblocksPlugin](https://github.com/liveblocks/liveblocks/blob/b1bc5342fafebcc5da7a225786818614681be09c/packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx#L180) for Lexical and in [useLiveblocksExtension](https://github.com/liveblocks/liveblocks/blob/b1bc5342fafebcc5da7a225786818614681be09c/packages/liveblocks-react-tiptap/src/LiveblocksExtension.ts#L89) for Tiptap

In my case I have added on the node client the property `experimental_textEditor` because a few days ago I found that I needed this property from the mongo db object which wasn't formatted in [formatRoom](https://github.com/liveblocks/liveblocks-backend/blob/fe4e9c92c8784cb9886bc7090280567c8db4b79d/apps/cloudflare/src/core/formatters.ts). 
But as it's for more internal cooking (e.g only the dashboard and the package `@liveblocks/emails`) I preferred to mark this property `experimental_` to be free to do whatever w/ it and break it if needed.

packages/liveblocks-node/src/client.ts
```diff
@@ -116,6 +116,18 @@ export type RoomAccesses = Record<
...
 
+const RoomTextEditorType = {
+  Lexical: "lexical",
+  TipTap: "tiptap",
+} as const;
+type RoomTextEditorType =
+  (typeof RoomTextEditorType)[keyof typeof RoomTextEditorType];
+
+export type RoomTextEditor = {
+  type: RoomTextEditorType;
```
nimeshnayaju, line 127: Why not type this directly as `"lexical" | "tiptap"`?

```ts
export type RoomTextEditor = {
  type: "lexical" | "react";
  rootKey: string[];
};
```

SugarDarius: Really good question here!
TBH I do not know at all, I just followed the same typing than we have in `liveblocks-backend`.  But for sire it can be simplified ✅

jrowny: I think this is copied from mongotypes, the reason it's an enum is in case we support many more editors it will be easier to use.

SugarDarius: Done in this [commit](https://github.com/liveblocks/liveblocks/pull/2042/commits/bbb3bd8caff0d1fc2a72e86dea4578c2db1a9c8b)

packages/liveblocks-node/src/client.ts
jrowny: This is not meant to be seen by end users. Instead of experimental, can you just use the @internal annotation like we do for `baseUrl`?

examples/nextjs-lexical/package.json
```diff
@@ -5,7 +5,9 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
+    "dev:email": "email dev",
```
jrowny, line 8: did you mean to add this to the main nextjs-lexical package?

packages/liveblocks-emails/src/index.ts
```diff
@@ -13,12 +13,30 @@ export type {
   ConvertCommentBodyAsHtmlStyles,
   ConvertCommentBodyAsReactComponents,
 } from "./comment-body";
+export type { ResolveRoomInfoArgs } from "./lib/types";
+export type {
+  ConvertLiveblocksTextEditorNodesAsReactComponents,
```
jrowny, line 18: When you have the word "convert", I think it should be "To" instead of "As." ConvertLiveblocksTextEditorNodesToReactComponents.

Also, these names are very long, could we maybe remove Liveblocks from them so they are a bit shorter?

examples/nextjs-lexical/package.json
SugarDarius: Ah thanks I need to remove it 👍🏻

packages/liveblocks-emails/src/index.ts
SugarDarius: For `As` → `To` I will not rename it yet, I already deliver in prod for `ThreadNotificationEvent` with `As` in `ConvertCommentBodyAsReactComponents` or `ConvertCommentBodyAsHtmlStyles`.

But I will think about renaming to remove the `Liveblocks` part 👍🏻

packages/liveblocks-node/src/client.ts
SugarDarius: Let me check

SugarDarius: Thanks for pointing this out @jrowny 🙏🏻 
I made the property more "hidden" in this [commit](https://github.com/liveblocks/liveblocks/pull/2042/commits/586ccbe2026a668fac00de1984bbbcf38978dcb9). 

For now we will keep the `experimental_` part as it's coming from the back-end. I will later to rename it if needed.

CTNicholas (approved):
Excellent work @SugarDarius! Everything's working well, plus docs and examples are ready to go. Let's get this merged!

CTNicholas:
✅

PR #2043, opened 2024-11-07
Implement new hook `useSyncStatus`

nvie:
This implements a new hook, `useSyncStatus()` which is a lot like `useStorageStatus()`, except it will work on the global Liveblocks sync status as a whole (instead of just for the current room's storage). There also is a new `client.getSyncStatus()` method on the Client.

What this means is that if you have (any combination of):
- (Multiple) rooms with Storage
- Threads
- Inbox Notifications
- (Multiple) Text Editors

And _any_ of those have pending changes, then this hook will indicate that sync status globally.

The following APIs are introduced:

```tsx
useSyncStatus();  // "synchronizing" | "synchronized"
useSyncStatus({ smooth: true });
```

Note that unlike `useStorageStatus()`, the returned value will only be `"synchronizing"` or `"synchronized"` (never `"not-loaded"` or `"loading"`). There is also no difference between Suspense and non-Suspense here.

This new hook can be obtained from any place:
- Import directly from `"@liveblocks/react"`
- Import directly from `"@liveblocks/react/suspense"`
- From `createLiveblocksContext()` factory
- From `createRoomContext()` factory

### 🏁 Checklist
- [x] Add unit tests
- [x] Add this hook to E2E test for Storage
- [x] Add this hook to E2E test for Yjs
- [x] Add this hook to E2E test for Comments
- [x] Add this hook to E2E test for Notifications
- [x] Add type-level tests
- [x] Let's make these public APIs for use outside of React too, i.e. `client.getSyncStatus()`
- [x] Make this work for (all) text editors!
- [x] Implement `useSyncStatusListener` hook as well, to more cleanly support the "beforeunload" use case (see ticket)
- [x] Add `usePreventUnsavedChanges` (convenience) hook
- [x] Deprecate `useStorageStatus()` in favor of this new hook?
- [x] Deprecate `useEditorStatus()`?
- [x] Add `useIsEditorReady` hook (to replace our current usage of `useEditorStatus` in examples soon)
- [x] Implement this for attachment uploads and/or draft text in composers
- [x] Update documentation
  - [x] For client
  - [x] For react
  - [x] For react-ui
  - [x] For react-lexical
- [x] Update CHANGELOG
  - [x] For client
  - [x] For react
  - [x] For react-ui
  - [x] For react-lexical

Fixes LB-961.


nvie:
I've implemented this hook to work with all of the following parts of Liveblocks now, see this recording in our E2E test app:

https://github.com/user-attachments/assets/189cfa8f-4e65-43dd-aa41-8c17ac39b1e0

nvie:
Allright, I think I consider this PR mostly done now!

@marcbouchenoire @CTNicholas @nimeshnayaju:
I have a couple of discussion points that I would like your input on to make a joint decision here:

As of this moment, I have implemented both `useSyncStatusListener` and `usePreventUnsavedChanges` hooks. I used both of them in our various E2E test app, and honestly the DX of `usePreventUnsavedChanges` is so much better that I didn't even want to use `useSyncStatusListener` for anything anymore. What do you think?

1. Shall we ship `usePreventUnsavedChanges` out of the box? There is no way to customize its behavior, but I think its default behaviour is pretty nice to have!
1. Do we still want to ship `useSyncStatusListener`? I no longer see a good use case for it, if we have `useSyncStatus({ smooth: true })` and `usePreventUnsavedChanges` already.
1. Speaking of `usePreventUnsavedChanges`… should we expose this as a React hook? It will only prevent changes if the component you use it in is mounted. Therefore, in a real world app, you'll want to do it as high in the tree as possible, probably directly under the `LiveblocksProvider` level in your React tree. This makes me think it should really be a config option on the Client, instead of a hook. How do you feel about making this a config option instead, i.e. `<LiveblocksProvider preventUnsavedChanges={true}>` and `createClient({ preventUnsavedChanges: true })`? This way, we won't need the hook at all and it will even work outside of React, and independent of which components are mounted.

nvie:
Screencast of the "prevent close" in action, here demonstrated on the threads page, but of course this also works for storage, text editors, etc etc:

https://github.com/user-attachments/assets/b8e3f2ca-223d-4ce9-878c-9612e6c7305d

CTNicholas:
Great work so far, it looks nice! My thoughts:

1. Agreed, I think we should add `preventUnsavedChanges` as a config option on `LiveblocksProvider`, and skip the hook for it. 
2. Also happy to skip the hook for `useSyncStatusListener`.
3. I wish there was a way to tell people that "things are okay now" after preventing unsaved changes, for example by sending a toast notification. This might be too much, but we could have an event on `LiveblocksProvider` that triggers _on the first sync after closing the tab was prevented_, e.g. `onPreventUnsavedChangesComplete` (such a mouthful).

```tsx
<LiveblocksProvider 
  preventUnsavedChanges={true} 
  onPreventUnsavedChangesComplete={() => {
    toast("You can close the tab now");
  }
/>
```

marcbouchenoire:
I'm on the same page as you both, the client/provider option seems like the most convenient so far 👍

@CTNicholas One tricky thing with something like `onPreventUnsavedChangesComplete` would be to maybe differentiate the type of update that happened, for example if we count attachment uploads as pending updates, showing a "You can close the tab now" toast could be confusing

CTNicholas:
> @CTNicholas One tricky thing with something like `onPreventUnsavedChangesComplete` would be to maybe differentiate the type of update that happened, for example if we count attachment uploads as pending updates, showing a "You can close the tab now" toast could be confusing

Okay, good point. I've just thought—should `preventUnsavedChanges` also stop you closing the tab if there's text in the composer, and you haven't submitted?

marcbouchenoire:
Ah yeah with this direction it would be obvious to do that 🤔
@nvie It's derailing a bit from the initial plan but do you agree? I haven't looked too deep into the current implementation but would your `setPending` model work for something like calling it from the `Composer` component when it's empty, when it's not, and when it's submitted?

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1179,6 +1199,127 @@ const _useUserThreads_experimental: TypedBundle["useUserThreads_experimental"] =
...
+function useSyncStatus_withClient(
+  client: OpaqueClient,
+  options?: UseSyncStatusOptions
+): SyncStatus {
+  // Normally the Rules of Hooks™ dictate that you should not call hooks
+  // conditionally. In this case, we're good here, because the same code path
+  // will always be taken on every subsequent render here, because we've frozen
+  // the value.
+  /* eslint-disable react-hooks/rules-of-hooks */
+  const smooth = useInitial(options?.smooth ?? false);
```
nimeshnayaju, line 1211: Is there a particular reason why we freeze the value to always be the one set the first?

packages/liveblocks-core/src/room.ts
```diff
@@ -1230,6 +1256,11 @@ export type RoomConfig = {
 
   baseUrl: string;
   enableDebugLogging?: boolean;
+
+  // We would not have to pass this complicated factory/callback functions to
+  // the createRoom() function if we would simply pass the Client instance to
+  // the Room instance, so it can directly call this back on the Client.
+  createSyncSource: () => SyncSource;
```
nimeshnayaju, line 1263: I see that we still pass this factory to the createRoom function. I remember us discussing potentially doing this directly within the client. Do you still plan to do that or was it not easily possible to do it from client after further look?

packages/liveblocks-react/src/liveblocks.tsx
nvie: Yes, if we did not do this, then calling `useSyncStatus({ smooth: true })` on the first render, and `useSyncStatus({ smooth: false })` on the second render would call a different hook, which would be problematic in React due to the fact that hooks rely on call order. (This is exactly the problem that the Rules of Hooks lint rule tries to prevent.) However, if we let the first render decide which branch is going to be taken for all subsequent renders, then this is not an issue.

packages/liveblocks-core/src/room.ts
nvie: I tried it, but it turned out to be harder to do. I needed to add quite a bit of extra complexity around tearing down the room and Storage/Yjs listeners in the leave room exit function, so I decided to keep this implementation instead. I still think that conceptually it should be possible to do this, but the current implementation made it harder than I wanted it to be.

nvie:
OK, thanks for the feedback and great ideas, @CTNicholas and @marcbouchenoire!

> Happy to skip the hook for `useSyncStatusListener`.

OK, I'll remove it.

> Agreed, I think we should add `preventUnsavedChanges` as a config option on `LiveblocksProvider`, and skip the hook for it.

OK, I'll remove the hook and make it a config option then! 👍

> I wish there was a way to tell people that "things are okay now" after preventing unsaved changes, for example by sending a toast notification.

Yeah, I see where you're coming from. The suggested `onPreventUnsavedChangesComplete` option feels a little bit too specific to my liking, but maybe we can find a better API for this and add support for that later? This way, we can keep it out of scope for this PR, because I think we will need to spend a bit more time to land on an API that would feel a bit more stable/final.

> One tricky thing with something like `onPreventUnsavedChangesComplete` would be to maybe differentiate the type of update that happened […]

One implication of the current implementation is that we cannot differentiate the source of the update that happened. You can only tell there are pending updates _somewhere_ in Liveblocks, but not specifically where (i.e. if for Storage, Yjs, or Threads, which Room instance, or which attachment/composer, whatever).

That said, we could easily add support for preventing a tab close if there is text in a composer somewhere that hasn't been submitted yet. The composer would be able to register itself as a sync source when mounted, and indeed mark itself as "pending" if there is text, and "clear" when the text has been submitted. This would be easy. We should probably make that configurable on the Composer I think?

nvie:
> I haven't looked too deep into the current implementation but would your `setPending` model work for something like calling it from the `Composer` component when it's empty, when it's not, and when it's submitted?

@marcbouchenoire I tried to also tie this in with pending attachment uploads and pending "draft" texts in Composer. Could you give me a quick walkthrough of where to make that change exactly? I tried to find where inside the Composer there was an `onChange`-like event to use to detect text/length changes in the text input control, but couldn't immediately find it.

I've exposed a private API like this that we should be able to use roughly like this, in pseudo-code:

```tsx
import { useSyncSource } from "@liveblocks/react/_private";

function Composer() {
  const syncSource = useSyncSource();
  return <input value={text} onChange={() => syncSource.setPending(text.length > 0)} />
}
```

That should be enough to tap into this new client's behavior. Could you help me implement this part inside `@liveblocks/react-ui`?

For pending attachment uploads, we can do a similar thing.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1179,6 +1190,77 @@ const _useUserThreads_experimental: TypedBundle["useUserThreads_experimental"] =
...
+      const newStatus = getter();
+      if (
+        oldStatus.current === "synchronizing" &&
+        newStatus === "synchronized"
+      ) {
+        // Delay delivery of the "synchronized" event
+        timeoutId = setTimeout(() => setStatus(newStatus), config.SMOOTH_DELAY);
+      } else {
+        clearTimeout(timeoutId);
+        setStatus(newStatus);
```
nimeshnayaju, line 1236: Do you know if it is possible to also use `useSyncExternalStore` for this hook too? I am not a 100% certain if it is applicable in this particular case I think it could be possible for "[tearing](https://github.com/reactwg/react-18/discussions/69)" to happen if one uses `useEffect` and `useState` to synchronize with an external store.

packages/liveblocks-react/src/liveblocks.tsx
nimeshnayaju: I was inquiring more from a design perspective. I understand that the way we have currently set up the hook, we want to absolutely freeze the `smooth` option but is it possible to set up the hook in a way that doesn't force `smooth` to be stabilized - maybe [if the `useSyncStatusSmooth_withClient` hook also uses `useSyncExternalStore`](https://github.com/liveblocks/liveblocks/pull/2043/files#r1836749015)?

packages/liveblocks-react/src/liveblocks.tsx
nvie: That would be nice indeed, but I don't know what would be the best approach here. The state being observed here is deliberately decoupled from the external store (via the timeout). We could surely read the immediate status value from the external store and then use it to compute a delayed state transition, i.e. in pseudo-code:

```
const immediateStatus = useSyncStatusImmediately();  // will already be uSES-based
const status = useDelayedStatus(immediateStatus);  // will use timer to smoothen
```

…but in this case, we're already too late because using the `useSyncExternalStore`-based approach will mean that this component will re-render a lot (up to 60 times per second).

Another approach could be to somehow try to shoehorn this into `useSyncExternalStore`, by being smart about the subscription event, and trying to hide/defer certain status change events, but I think fundamentally this is impossible because by the time those events fire, reading the status synchronously will no longer match the status we want to display.

The only solution I see that would allow us to use `useSyncExternalStore` here is to make the external store _be_ the deferred/smooth status, which would require moving this logic into the client instead.

Or can you think of another approach here? Open to any ideas!

packages/liveblocks-react/src/liveblocks.tsx
nvie: Did you mean to change the `getter`/`subscriber` as a pair, i.e. observe a _different_ external store based on the conditional?

```tsx
const subscribe = smooth ? smoothSubscribe : immediateSubscribe;
const getter = smooth ? smoothGetter : immediateGetter;
const status = useSyncExternalStore(subscribe, getter, getter);
```

I think that could work indeed! Would actually be quite nice 🤔 Maybe this is a good argument for creating that "deferred" external store anyway, like I [suggested here](https://github.com/liveblocks/liveblocks/pull/2043#discussion_r1837743310).

Anyway, I'll definitely keep this out of scope for this PR, but I agree it may be nice to set it up like this and be a better React citizen and avoid possible tearing use cases we're not aware of.

CTNicholas:
It's looking really nice—I've given it a good try just now, it works well! I have one piece of feedback:

If you write text in a composer, `useSyncStatus` currently shows `unsynchronized`. While it makes sense to prevent a user closing the page because of this with `preventUnsavedChanges`, I'm in two minds about showing a status badge when a user types text into the composer, as it might look like your app fails to sync. But then also, we don't want to tell people that their composer text is saved, because it isn't. Thoughts @marcbouchenoire @nvie?

https://github.com/user-attachments/assets/89e099f5-cd0b-4c85-a80c-d4f4be757db1

nvie:
Thanks @CTNicholas — and nice you're already using it in a real example! I wanted to chat with you about this thing exactly, but you beat me to it. Maybe we should change the internals to allow a third state on a per-sync-source basis, and separate "I have some local unsaved state" from "I am actually synchronizing with the server".

So that:
- If there are any local changes, it will prevent tabs from closing.
- If there are any actual synchs happening, it will also prevent tabs from closing, but also update the sync status.

Which of these approaches would you prefer?
- We keep the same states, i.e. the `useSyncStore` hook keeps working the same two states as before, but it will no longer show "synchronizing" even if there is draft text in a composer (it will remain "synchronized" in those cases)
- We allow observing the new third state: `"synchronizing" | "synchronized" | "has-local-changes"`. Would you use it for any UI, or is it a needless complication?

CTNicholas:
Hah! I prefer option 1 here, I think 2 is unnecessary complexity for little purpose

CTNicholas:
> > Hah! I prefer option 1 here, I think 2 is unnecessary complexity for little purpose
> 
> A'ight, I just pushed the change that makes it like so. @CTNicholas Can you give your example one more test with this? Does it now behave as-expected in all edge cases? 🙏

Works great, love it! I haven't checked it on Tiptap, as I have a lot do, but I'm assuming it works there too. 

Tiny feedback is that it is can be a tiny bit confusing to enable `smooth` because the option is frozen (when you change it in dev mode you need to refresh the page). This is probably fine to leave I think, as I know the implementation is built around freezing it, and it's a very small thing

nvie:
> Tiny feedback is that it is can be a tiny bit confusing to enable `smooth` because the option is frozen (when you change it in dev mode you need to refresh the page). This is probably fine to leave I think, as I know the implementation is built around freezing it, and it's a very small thing

You mean when updating it HMR won't pick it up, right? That's indeed expected and a result of the current implementation. Nimesh [raised a related issue](https://github.com/liveblocks/liveblocks/pull/2043#discussion_r1836749015) here. Solving that would also solve that smooth HMR issue.

> Works great, love it! I haven't checked it on Tiptap, as I have a lot do, but I'm assuming it works there too.

I also haven't checked explicitly, but given that the behavior is implemented at the Yjs level, it's not editor-specific so indeed it should work.

nvie:
@CTNicholas @marcbouchenoire @nimeshnayaju Last question… should we make `preventUnsavedChanges` default to `true` in the 2.12.0 release, or would that be too aggressive? I think it would be a good UX default, but not sure if we should do it immediately in 2.12.0 or first gather feedback and maybe consider doing it later. What do you all think?

nimeshnayaju:
> @CTNicholas @marcbouchenoire @nimeshnayaju Last question… should we make `preventUnsavedChanges` (at the client level) default to `true` in the 2.12.0 release, or would that be too aggressive? I think it would be a good UX default, but not sure if we should do it immediately in 2.12.0 or first gather feedback and maybe consider doing it later. What do you all think?

My opinion is to default it to `false` for now. I agree it is a good end user experience but it can seem too imposing/opinionated by default for developers.

packages/liveblocks-react/src/liveblocks.tsx
nimeshnayaju: I can see how an implementation might be challenging 🤔 . Let's invest more time on this if there are bug reports from user concerning tearing!

docs/pages/api-reference/liveblocks-react-lexical.mdx
```diff
@@ -830,6 +830,14 @@ the displayed version.
 
 ### useEditorStatus
 
+<Banner title={<>Deprecated</>}>
```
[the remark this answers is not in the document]
marcbouchenoire: They're the same so there's no reason, the banner was probably copied from one that used inline nodes and it was kept as is.

marcbouchenoire:
I like the current state where the client-level option is opt-in but the `Composer` is opt-out.

CTNicholas (approved):
Everything looking good! I'll add it to some examples after we've published the package.

PR #2049, opened 2024-11-08
Fix anchored threads lifecycle for TipTap

jrowny:
This is the exact same fix for TipTap as we had for Lexical since most of the component is similar.


nvie (approved):

nimeshnayaju (approved):
🙏

PR #2053, opened 2024-11-12
prosemirror node api

jrowny:
Adds a prosemirror node API similar to the lexical api, also works with TipTap! Includes some niceities from TipTap like `getText` as well as some extras tiptap doesn't have like`getMarkdown`

new wrapper for getting and modifying doc:
```
withProsemirrorDocument({
   client: new Liveblocks();, roomId: "some-room"
}, (api) => {
     // document API
});
```
and the api can be used like this:

```
(api) => { 
     await api.setcontent( /* Some content */ );
     const markDown = api.toMarkdown();
})
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,6 +22,7 @@ jobs:
             "@liveblocks/react-lexical",
             "@liveblocks/react-tiptap",
             "@liveblocks/node-lexical",
+            "@liveblocks/node-prosemirror",
             "@liveblocks/codemod",
             "@liveblocks/emails",
           ]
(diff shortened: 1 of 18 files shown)


.github/workflows/tests.yml
```diff
@@ -22,6 +22,7 @@ jobs:
             "@liveblocks/react-lexical",
             "@liveblocks/react-tiptap",
             "@liveblocks/node-lexical",
+            "@liveblocks/node-prosemirror",
```
nvie, line 25: I took the liberty of publishing an empty 0.0.1 for this package here: https://www.npmjs.com/package/@liveblocks/node-prosemirror. Now the CI publish step should Just Work™.

nvie (approved):

.github/workflows/tests.yml
jrowny: and that's why you're my favorite person 🙇

PR #2054, opened 2024-11-12
Mark composers as pending updates

marcbouchenoire:
For attachments I'm wondering if the source should be lower since `uploadAttachment` [is technically a room-level API](https://liveblocks.io/docs/api-reference/liveblocks-client#Room.uploadAttachment)?

Fwiw I can't get this to work, even though the `setPending` condition is correct.

### DIFF
--- a/packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
+++ b/packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
@@ -1035,11 +1035,13 @@ const ComposerForm = forwardRef<HTMLFormElement, ComposerFormProps>(
 
     // Mark the composer as a pending update when it isn't empty
     // or when attachments are being uploaded
+    const isPending = !preventUnsavedChanges
+      ? false
+      : !isEmpty || isUploadingAttachments;
+
     useEffect(() => {
-      if (preventUnsavedChanges) {
-        syncSource.setPending(!isEmpty || isUploadingAttachments);
-      }
-    }, [syncSource, isEmpty, preventUnsavedChanges, isUploadingAttachments]);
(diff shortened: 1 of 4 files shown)


nvie (approved):
Thanks for helping me implement this part, @marcbouchenoire! 🙏

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
```diff
@@ -1027,6 +1031,17 @@ const ComposerForm = forwardRef<HTMLFormElement, ComposerFormProps>(
     const ref = useRef<HTMLFormElement>(null);
     const mergedRefs = useRefs(forwardedRef, ref);
     const fileInputRef = useRef<HTMLInputElement>(null);
+    const syncSource = useSyncSource();
+
+    // Mark the composer as a pending update when it isn't empty
+    // or when attachments are being uploaded
+    const isPending = !preventUnsavedChanges
+      ? false
+      : !isEmpty || isUploadingAttachments || attachments.length > 0;
```
nvie, line 1040: @marcbouchenoire One extra change I made here is that I also added the `attachments.length > 0` check here, meaning even if the upload has finished successfully, but the comment has not been submitted yet, it will still be considered "pending".

nvie: Please let me know if that is wrong in any way!

marcbouchenoire: Makes sense! Maybe we could update the various comments that mention attachments uploading explicitly since that's no longer the condition.

PR #2056, opened 2024-11-12
Better load behaviour in Lexical examples

CTNicholas:
Pushing Suspense further down the tree.


examples/nextjs-linear-like-issue-tracker/src/liveblocks.config.ts
```diff
@@ -1,4 +1,4 @@
-import { LiveList, LiveObject } from "@liveblocks/core";
+import { LiveList, LiveObject, ToImmutable } from "@liveblocks/core";
```
nvie, line 1: Yeah, I think this is a mistake. `ToImmutable` is part of our public API surface, and as such should definitely be a public type, exposed through `@liveblocks/client` indeed.

In general, no user should ever have to import/rely on `@liveblocks/core`. If you feel the need to import from `@liveblocks/core` in any of our examples, it's a bug and we should most likely get that fixed. I don't want to set a bad example ourselves here.

nvie (approved):

PR #2057, opened 2024-11-12
Novel example

CTNicholas:
Example using [Novel](https://novel.sh). The `src/components/editor` folder is copied from the Novel repo, with minor changes.


examples/nextjs-tiptap-novel/src/components/editor/ui/menu.tsx
```diff
@@ -0,0 +1,89 @@
...
+//   },
+//   {
+//     font: "Serif",
+//     icon: <FontSerif className="h-4 w-4" />,
+//   },
+//   {
+//     font: "Mono",
+//     icon: <FontMono className="h-4 w-4" />,
+//   },
+// ];
```
marcbouchenoire, line 22: Should we remove this?

examples/nextjs-tiptap-novel/src/app/api/upload/_route.ts
```diff
@@ -0,0 +1,26 @@
+// import { put } from "@vercel/blob";
+// import { NextResponse } from "next/server";
```
marcbouchenoire, line 2: Should the upload feature also be disabled since there's no route to handle it?

examples/nextjs-tiptap-novel/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@liveblocks/client": "^2.12.0",
+    "@liveblocks/node": "^2.12.0",
+    "@liveblocks/react": "^2.12.0",
+    "@liveblocks/react-tiptap": "^2.12.0",
+    "@liveblocks/react-ui": "^2.12.0",
+    "@nanostores/react": "github:ai/react",
```
marcbouchenoire, line 18: 🤔

examples/nextjs-tiptap-novel/src/lib/content.ts
```diff
@@ -0,0 +1,356 @@
+export const defaultEditorContent = {
```
marcbouchenoire, line 1: Could we tweak this? Especially the tweet and the Vercel Blob broken asset, but also maybe removing things like uploads if we disable them etc

marcbouchenoire (approved):

nilsbunger:
Excited to see this coming together. It's a slick example. 

I'm using this demo as inspiration for an editor experience I'm building. I noticed a few small issues:

1. when starting the editor (`npm run dev`), there are a bunch of warnings like this:
```
There are multiple modules with names that only differ in casing.
This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.
```
2. in the editor, if you start doing an @mention, then hit escape so you don't select a user, the cursor becomes stuck on the line - you can no longer arrow-up or down from that line.

Happy to file issues for these but since it's still a PR I thought I could mention them here.

examples/nextjs-tiptap-novel/src/components/editor/ui/menu.tsx
CTNicholas: I've left a bunch of TODOs in because they're from the official Novel editor. I've done this to make it easier for people to diff and see what's changed. Let me know if you think this is a bad idea!

examples/nextjs-tiptap-novel/src/app/api/upload/_route.ts
CTNicholas: Yes!

examples/nextjs-tiptap-novel/src/lib/content.ts
CTNicholas: Yep! Meant to remove this altogether, as Tiptap doesn't ave an initial content option yet (WIP).

examples/nextjs-tiptap-novel/package.json
CTNicholas: Changed it to the normal install, seems to be fine!

CTNicholas:
> Excited to see this coming together. It's a slick example. I'm using this demo as inspiration for an editor experience I'm building. I noticed a few small issues:

Thank you! I've fixed the things you highlighted, except 2. which would require changes to our package.

Issue #2055, opened 2024-11-12
@liveblocks/react-tiptap: components not displaying when using React compiler and/or shouldRenderOnTransaction: false


Ecaratti:
**Describe the bug**

When using either or both React compiler or shouldRenderonTransaction: true on Tiptap's useEditor, the FloatingComposer fails to appear.
 
When re-running my dev server without reactCompiler and with shouldRerenderOnTransaction: true, the component works as expected.

**To Reproduce**

Just set  shouldRerenderOnTransaction: false when using Tiptap's useEditor,
```typescript
  const editor = useEditor({
    extensions: [liveblocks, StarterKit],
    immediatelyRender: false,
    shouldRerenderOnTransaction: false,
  });
```

Issue is similar when setting 

```typescript
  experimental: {
    reactCompiler: true,
  },
``` 

in next.config.ts
/

**Expected behavior**

Floating composer should be displayed when triggering editor.chain().focus().addPendingComment().run()

**Environment (please complete the following information):**

- @tiptap 2.9.1.
- next 15.0.2
- @liveblocks 2.11.0

**Additional context**

The issue spawns from using:
```typescript
  const storage = editor?.storage.liveblocksComments as CommentsExtensionStorage | undefined;
  const showComposer = !!storage?.pendingCommentSelection;
``` 

in the [FloatingComposer](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react-tiptap/src/comments/FloatingComposer.tsx) (lines 32-34).

Tiptap 2.5 introduced performance optimisations which requires the use of useEditorState() to get the latest editor state when shouldRenderTransaction is set to false (cf [announcement ](https://tiptap.dev/blog/release-notes/say-hello-to-tiptap-2-5-our-most-performant-editor-yet) and [doc](https://tiptap.dev/docs/examples/advanced/react-performance#page-title)))

Suggested fix:  in FloatingComposer replace showComposer by:

```typescript
const currentEditorState = useEditorState({
    editor,
    selector: (ctx) => ({
      showComposer: !!ctx.editor?.storage.liveblocksComments?.pendingCommentSelection,
    }),
    equalityFn: (prev, next) => {
      if (!next) return false;
      return prev.showComposer === next.showComposer;
    },
  });
const showComposer = currentEditorState?.showComposer;
``` 

This fixed the issue for me and seems to work either with or without the react compiler and with shouldRerenderOnTransaction either true or false.

[...]

PR #2060, opened 2024-11-13
useEditorState to allow shouldRerenderOnTransaction:false to work

jrowny:
Fix for #2055 

Uses Tiptap's useEditorState with a selector to properly select the editor state when updates happen.


jrowny:
@CTNicholas i also updated the guide here

CTNicholas:
> @CTNicholas i also updated the guide here

Remember we have a React guide too!
- https://github.com/liveblocks/liveblocks/pull/2063

PR #2061, opened 2024-11-13
Tiptap experimental offline support

jrowny:
Adds experimental offline support to Tiptap

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
 
 - Fix an issue with `AnchoredThreads` component not working correctly on certain
   React versions.
-- Fixes an issue where react comonents don't update when
-  shouldRerenderOnTransaction: false is set
+- Fixes an issue where React components don’t update when
+  `shouldRerenderOnTransaction`: false is set
 
 ### `@liveblocks/yjs`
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,12 @@
 
 - Fix an issue with `AnchoredThreads` component not working correctly on certain
   React versions.
+- Fixes an issue where React components don’t update when
+  `shouldRerenderOnTransaction`: false is set
+
+### `@liveblocks/yjs`
+
+- Adds experimental offline support for `LiveblocksYjsProvider`.
```
nvie, line 17: The diff of this PR seems to suggest that experimental offline support already exists in `LiveblocksYjsProvider`, but that we're just exposing it to Tiptap here now. Should that be in the changelog instead?

nvie (approved):

CHANGELOG.md
jrowny: This is because I screwed this branch up.

I'm going to re-make it because this change actually didn't even work, the tiptap option isn't always the right value during startup and the option must be set during the constructor. I will make a new PR with a new approach.

PR #2074, opened 2024-11-18
Add method to get unread inbox notifications count for rooms

FlowFlorent:
#### Descriptions

Adds `room.getUnreadInboxNotificationsCount`

Usage:
```ts
  const room = useRoom();
  const count = await room.getUnreadInboxNotificationsCount();
  
  // OR
  
  const room = client.getRoom("my-room");
  const count = await room.getUnreadInboxNotificationsCount();
```

### DIFF
--- a/packages/liveblocks-core/src/api-client.ts
+++ b/packages/liveblocks-core/src/api-client.ts
@@ -207,6 +207,12 @@ export interface RoomHttpApi<M extends BaseMetadata> {
     inboxNotificationId: string;
   }): Promise<string>;
 
+  getRoomUnreadInboxNotificationsCount({
+    roomId,
+  }: {
+    roomId: string;
+  }): Promise<number>;
+
   getNotificationSettings({
     roomId,
     signal,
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -953,6 +953,14 @@ export type Room<
    */
   getAttachmentUrl(attachmentId: string): Promise<string>;
 
+  /**
+   * Gets the unread count of inbox notifications for the current user and room.
+   *
+   * @example
+   * const count = await room.getUnreadInboxNotificationsCount();
+   */
+  getUnreadInboxNotificationsCount(): Promise<number>;
```
nimeshnayaju, line 962: I suppose this is a method we wish to support, so it is fine to mark it as experimental.

nimeshnayaju (approved):

packages/liveblocks-core/src/api-client.ts
```diff
@@ -982,6 +988,19 @@ export function createApiClient<M extends BaseMetadata>({
...
    * -----------------------------------------------------------------------------------------------*/
+  async function getRoomUnreadInboxNotificationsCount(options: {
+    roomId: string;
+  }) {
+    const { count } = await httpClient.get<{ count: number }>(
+      url`/v2/c/rooms/${options.roomId}/inbox-notifications/count`,
+      await authManager.getAuthValue({
+        requestedScope: "comments:read",
+        roomId: options.roomId,
+      })
```
nimeshnayaju, line 999: @ofoucherot @marcbouchenoire I am not very well versed in the current auth manager but we seem to request cached token for this very scope `comments:read` for a lot of other methods (`createThread`, `deleteThread`) currently. Should we be requesting `room:read` scope in this particular method?

ofoucherot: Yes, for inbox-notifications it should be `room:read`.

FlowFlorent: I agree that `room:read` is the right one.

That said, if you use comments without storage, you'd still want to have the notifications. In which case you need the less powerful permission (`comments:read`) to still provide notifications.

Wondering if this is a missing piece in our system. And we should add a `notifications:write/read` permission.

FlowFlorent:
Closing, more thought needs to be put into this, with an API that likely reuses the project level one using queries

PR #2075, opened 2024-11-19
Tiptap initial content

jrowny:
**TODO after package release**
- [ ] Merge docs https://github.com/liveblocks/liveblocks/pull/2078

---

1. Allows setting the initial content through the hook `const liveblocks = useLiveblocksExtension({ initialContent: "<p> this is initial content! </p>"});` 
2. Allows setting experimental offline support: `const liveblocks = useLiveblocksExtension({ offlineSupport_experimental});`
3. Adds `useIsEditorReady` hook to tiptap package 
4. A warning when user sets the content option on editor

Before this, it was impossible to set initial content.

### DIFF
--- a/examples/nextjs-tiptap/app/tiptap/editor.tsx
+++ b/examples/nextjs-tiptap/app/tiptap/editor.tsx
@@ -19,6 +19,7 @@ export default function TiptapEditor() {
         class: "outline-none flex-1 transition-all",
       },
     },
+    content:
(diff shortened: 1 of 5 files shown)


nvie (approved):
I don't see anything out of the ordinary in this PR, so it looks good to me! 👍

Should we also document these new config options and maybe provide a migration guide for users already using Tiptap?

PR #2076, opened 2024-11-19
Comments formatting UI

marcbouchenoire:
🚧🚧🚧


docs/pages/api-reference/liveblocks-react-ui.mdx
```diff
@@ -351,6 +358,9 @@ Learn more about mutation hooks under
   <PropertiesListItem name="defaultValue" type="CommentBody">
     The composer’s initial value.
   </PropertiesListItem>
+  <PropertiesListItem name="defaultAttachments" type="CommentAttachment[]">
+    The composer's initial attachments.
+  </PropertiesListItem>
```
marcbouchenoire, line 363: There were a few unlisted props that should be documented so I added them while I was there.

packages/liveblocks-react-ui/src/components/Comment.tsx
```diff
@@ -110,6 +111,11 @@ export interface CommentProps extends ComponentPropsWithoutRef<"div"> {
    */
   showAttachments?: boolean;
 
+  /**
+   * Whether to show the composer's formattng controls when editing the comment.
+   */
+  showComposerFormattingControls?: ComposerProps["showFormattingControls"];
```
marcbouchenoire, line 117: I'm naming this "formatting controls" instead of being specific to the toolbar because we'll re-use it for other floating controls like the link edition UI, etc.

packages/liveblocks-react-ui/src/slate/utils/is-selection-collapsed.ts
```diff
@@ -1,8 +0,0 @@
-import type { Selection as SlateSelection } from "slate";
-import { Range as SlateRange } from "slate";
-
-export function isSelectionCollapsed(
```
marcbouchenoire, line 4: This type guard was causing more harm than good by combining the null check and the collapsed one.

packages/liveblocks-react-ui/src/styles/index.css
```diff
@@ -597,16 +607,23 @@
 :is(.lb-comment-body, .lb-composer-editor) {
   color: var(--lb-foreground-secondary);
 
+  &,
+  * {
+    line-height: 1.5;
+  }
```
marcbouchenoire, line 613: This is a possible CSS breaking change.

docs/pages/api-reference/liveblocks-react-ui.mdx
```diff
@@ -351,6 +358,9 @@ Learn more about mutation hooks under
   <PropertiesListItem name="defaultValue" type="CommentBody">
     The composer’s initial value.
   </PropertiesListItem>
+  <PropertiesListItem name="defaultAttachments" type="CommentAttachment[]">
+    The composer's initial attachments.
```
CTNicholas, line 362:
```suggestion
    The composer’s initial attachments.
```

docs/pages/api-reference/liveblocks-react-ui.mdx
```diff
@@ -471,6 +492,16 @@ function Component() {
...
   </PropertiesListItem>
+  <PropertiesListItem name="showAttachments" type="boolean" defaultValue="true">
+    Whether to show attachments.
+  </PropertiesListItem>
+  <PropertiesListItem
+    name="showComposerFormattingControls"
+    type="boolean"
+    defaultValue="true"
+  >
+    Whether to show the composer's formatting controls when editing the comment.
```
CTNicholas, line 503:
```suggestion
    Whether to show the composer‘s formatting controls when editing the comment.
```

docs/pages/api-reference/liveblocks-react-ui.mdx
```diff
@@ -1431,6 +1526,16 @@ function MyComposerAttachments() {
   <PropertiesListItem name="blur" type="function" detailedType="() => void">
     Blur the editor programmatically.
   </PropertiesListItem>
+  <PropertiesListItem name="marks" type="ComposerBodyMarks">
+    Which text marks are currently active and which aren't.
```
CTNicholas, line 1530:
```suggestion
    Which text marks are currently active and which aren‘t.
```

CTNicholas (approved):
✅ Docs

Cool feature! I like the naming you've already chosen.

CTNicholas:
Actually, I have thoughts on the styling. The icons are quite faded out by default, and almost look `disabled`. I'd personally prefer to have them not be faded out, and put a faint background around them when e.g. selecting text that's already bold.

packages/liveblocks-react-ui/src/components/Composer.tsx
```diff
@@ -349,6 +374,97 @@ function ComposerMentionSuggestions({
   ) : null;
 }
 
+function MarkToggle({ mark, shortcut, children, ...props }: MarkToggleProps) {
+  const $ = useOverrides();
+  const label = useMemo(() => {
+    return $.COMPOSER_TOGGLE_MARK(mark);
+  }, [$, mark]);
```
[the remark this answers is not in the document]
marcbouchenoire: I do have a tendency to overuse `useMemo`/`useCallback` but here the `COMPOSER_TOGGLE_MARK` overrides can be set by users so it might be unexpected to run it on every render. I don't think anyone would put super expensive logic in a localization function but still I feel safer controlling its number of invokations

packages/liveblocks-react-ui/src/primitives/Composer/utils.ts
```diff
@@ -175,19 +200,125 @@ export function commentBodyToComposerBody(body: CommentBody): ComposerBody {
...
+export function getSideAndAlignFromFloatingPlacement(placement: Placement) {
   const [side, align = "center"] = placement.split("-");
 
   return [side, align] as const;
 }
 
+export function useOnComposerEditorChangeWithEventSource(
+  editorChangeEventSource: EventSource<void>,
+  callback: () => void
+) {
```
nimeshnayaju, line 225: Is this hook exported purely because we want to also listen to editor `onchange` events inside `Composer.Form`?

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
```diff
@@ -1287,12 +1479,24 @@ const ComposerForm = forwardRef<HTMLFormElement, ComposerFormProps>(
...
+    const toggleMark = useCallback(
+      (mark: ComposerBodyMark) => {
+        toggleEditorMark(editor, mark);
+      },
+      [editor]
+    );
+
+    useOnComposerEditorChangeWithEventSource(editorChangeEventSource, () => {
+      setMarks(getMarks(editor));
+    });
```
nimeshnayaju, line 1491: Related to [this comment](https://github.com/liveblocks/liveblocks/pull/2076#discussion_r1859758326), if we do not want to export a hook just for this use case, we could maybe make this a `useEffect` and mimic the implementation of `useOnComposerEditorChangeWithEventSource` directly here.

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
```diff
@@ -1452,26 +1658,95 @@ const ComposerAttachmentsDropArea = forwardRef<
...
+      },
+      [mark, onClick, onValueChange, toggleMark]
+    );
+
+    const handlePointerDown = useCallback(
+      (event: PointerEvent<HTMLButtonElement>) => {
+        onPointerDown?.(event);
+
+        event.preventDefault();
+        event.stopPropagation();
```
nimeshnayaju, line 1705: Is there a reason why we prevent default behavior and stop propagation on pointer down?

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
```diff
@@ -427,11 +569,12 @@ function ComposerEditorPlaceholder({
 const ComposerMention = forwardRef<HTMLSpanElement, ComposerMentionProps>(
   ({ children, asChild, ...props }, forwardedRef) => {
     const Component = asChild ? Slot : "span";
+    const { isFocused } = useComposer();
     const isSelected = useSelected();
 
     return (
       <Component
-        data-selected={isSelected || undefined}
+        data-selected={(isFocused && isSelected) || undefined}
```
nimeshnayaju, line 577: Is this something we missed previously? Does this mean mentions inside a composer are focusable?

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
```diff
@@ -347,6 +320,175 @@ function ComposerEditorMentionSuggestionsWrapper({
...
+      ) {
+        setHasFloatingToolbarRange(false);
+        setReference(null);
+      } else {
+        setHasFloatingToolbarRange(true);
+
+        const domRange = domSelection.getRangeAt(0);
+        setReference(domRange);
+      }
+    });
```
nimeshnayaju, line 392: Just to make sure I understand this correctly, we listen to changes to the editor (selection, state update, etc) and set the reference element of the floating toolbar to null and immediately in the next animation frame, based on whether a selection range is collapsed or not, we say if the floating toolbar should appear or not.  
And, the reason why we wait for the next animation frame is because the selection change is known only a frame after editor change occurs?

nimeshnayaju (approved):
Amazing job @marcbouchenoire 🚀 . This was quite a big undertaking and you did a great job - the final product looks beautiful! I've tried to review this based on my limited Slate.js knowledge and I only have some general questions.

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
marcbouchenoire: This change is a remnant of when I tried to fix LB-1374 as part of this PR, since I'll fix the problem in another PR I removed this from this one for now.

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
marcbouchenoire: The floating toolbar is only opened if the editor is focused, so we prevent focus from being applied to the toolbar and its elements (for the link edition modal we won't be able to solve this in the same way because it has an input so I'm probably going to revisit this when working on links). I initially only applied this on `MarkToggle` before also adding it to `FloatingToolbar` but it makes more sense to only have it on the toolbar (since `MarkToggle` can be used elsewhere), done in https://github.com/liveblocks/liveblocks/pull/2076/commits/153b32be647f36c32dc81f61c3cf7808191e71ef.

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
marcbouchenoire: We use the [`inline`](https://floating-ui.com/docs/inline) Floating UI middleware for positioning which uses [`getClientRects`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects) to check individual lines instead of the bounds but if we don't wait a frame before updating the reference, somehow the DOM selection returns an empty list from `getClientRects` when you toggle marks (and change the editor's markup). Because of that empty list, `inline` falls back to using `getBoundingClientRect` which makes the toolbar jump around on toggle.

https://github.com/user-attachments/assets/214ee522-58f1-4d8c-8596-0f4e156188ce

packages/liveblocks-react-ui/src/primitives/Composer/utils.ts
marcbouchenoire: In the `Composer.Form` parent but also in children components (e.g. `ComposerEditorMentionSuggestionsWrapper`, `ComposerEditorFloatingToolbarWrapper`, etc).

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
marcbouchenoire: The Composer's `index.tsx` file is already pretty long and complex so I wanted to avoid adding noise just for subscribing, I simplified this by making [a single and generic `useEventSource` hook](https://github.com/liveblocks/liveblocks/pull/2076/commits/2ade5d5747a9fd3409a9e6051e3b004940104d7f).

packages/liveblocks-react-ui/src/utils/use-event-source.ts
```diff
@@ -0,0 +1,17 @@
...
+  callback: () => void
+) {
+  const latestCallback = useLatest(callback);
+
+  useEffect(() => {
+    const unsubscribe = eventSource.subscribe(() => latestCallback.current());
+
+    return unsubscribe;
+  }, [eventSource, latestCallback]);
+}
```
nvie, line 17: Nice! I like this generic hook!

One suggestion here is to only have it rely on the `Observable` API, not the full `EventSource` API, which should ideally be an implementation detail used by the source of the events — i.e. where those events get emitted. Any party interested in listening to those events, are better off just relying on the `Observable` API.

```suggestion
export function useObservable<T>(
  observable: Observable<T>,
  callback: () => void
) {
  const latestCallback = useLatest(callback);

  useEffect(() => {
    const unsubscribe = observable.subscribe(() => latestCallback.current());

    return unsubscribe;
  }, [observable, latestCallback]);
}
```

marcbouchenoire: Makes sense! Done [here](https://github.com/liveblocks/liveblocks/pull/2076/commits/72ff05001f3c767a2b373430f0770df0475e44a1).

packages/liveblocks-react-ui/src/primitives/Composer/contexts.ts
```diff
@@ -82,6 +94,7 @@ export type ComposerEditorContext = {
   validate: (value: SlateElement[]) => void;
   editor: SlateEditor;
   setFocused: Dispatch<SetStateAction<boolean>>;
+  editorChangeEventSource: EventSource<void>;
```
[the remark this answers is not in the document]
nvie: In similar vein to my previous comment, should this become something like this instead?

```suggestion
  onEditorChange: Observable<void>;
```

PR #2079, opened 2024-11-19
v1 liveblocks-react-blocknote

YousefED:
This adds `@liveblocks/react-blocknote` for BlockNote compatibility.

The BlockNote support mostly builds on `@liveblocks/react-tiptap`, but with a simpler API and support for the BlockNote schema (BlockNote comes it's own API for editor contents that's entirely strongly-typed).

The code mostly consists of some helper functions to simplify the BlockNote initialization. The helpers bring down the setup code from this:

```typescript
const liveblocks = useLiveblocksExtension();

  const schema = withLiveblocks(BlockNoteSchema.create());

  const editor = useCreateBlockNote({
    schema,
    _extensions: { liveblocksExtension: liveblocks },
    disableExtensions: ["history"],
  });
```

to this one-liner:

```typescript
const editor = useCreateBlockNoteWithLiveblocks({});
```

## Changes to @liveblocks/react-tiptap
The main change to `@liveblocks/react-tiptap` is to split the Mention node from the Mention extension. This is done because for BlockNote, the Nodes / Marks are configured as part of the BlockNote schema, whereas the other Extensions (i.e.: `notifier` prosemirror plugin etc.) should be passed to the `_extensions` option. The API for tiptap users remains unchanged

## TODO
After a review, these items are open:

- [ ] Add example including Version History. I haven't tested this yet
- [ ] Extensive testing. I focused on getting this to work but the Liveblocks probably knows which details need to be thoroughly tested / where possible caveats are
- [ ] Add documentation
- [ ] Add examples
- [ ] Promote to users of both BlockNote and Liveblocks 🥳


packages/liveblocks-react-blocknote/src/BlockNoteLiveblocksExtension.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const useLiveblocksExtension = (options: LiveblocksExtensionOptions = {}) => {
+  const extension = useTipTapLiveblocksExtension();
+
+  return extension.extend({
+    addExtensions() {
+      const ret = this.parent?.() ?? [];
+
+      // filter out the liveblocks mention extension, because this will be added via the schema instead of as an extension (see schema.ts)
+      return ret.filter((ext) => ext.name !== "liveblocksMention");
```
jrowny, line 14: youn should be able to change line 7 to {mentions: false} to avoid having Liveblocks add its mentions.

packages/liveblocks-react-blocknote/src/initialization/schema.ts
```diff
@@ -0,0 +1,52 @@
...
+  },
+});
+
+
+/**
+ * Adds the Liveblocks Mention Node as inline content to the BlockNote schema
+ * 
+ * This makes sure BlockNote knows about Liveblocks mentions and that you can read/write mentions via the BlockNote API
+ */
+export const withLiveblocksSchema = <
```
jrowny, line 25: ah, this is the part I was missing when trying to get BlockNote working, nice!

packages/liveblocks-react-blocknote/src/version-history/HistoryVersionPreview.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+import type { BlockNoteEditor } from "@blocknote/core";
+import { HistoryVersionPreview as TipTapHistoryVersionPreview } from "@liveblocks/react-tiptap";
+
+type HistoryVersionPreviewProps = Omit<
+  Parameters<typeof TipTapHistoryVersionPreview>[0],
+  "editor"
+> & {
+  editor: BlockNoteEditor<any, any, any>;
```
jrowny, line 10: These components only exist for fixing this type right? I'm wondering if there's a way we can fix the types without having these wrapper components. Maybe adding a declare in index.ts or something? Maybe this is less risky than that idea.

packages/liveblocks-react-tiptap/src/LiveblocksExtension.ts
```diff
@@ -71,7 +72,7 @@ const LiveblocksCollab = Collaboration.extend({
   },
 });
 
-export const useLiveblocksExtension = (): Extension => {
+export const useLiveblocksExtension = (): Extension<LiveblocksExtensionOptions, any> => {
```
jrowny, line 75: we removed the options from the extension and moved them to the hook.

packages/liveblocks-react-tiptap/src/LiveblocksExtension.ts
```diff
@@ -245,6 +246,7 @@ export const useLiveblocksExtension = (): Extension => {
         extensions.push(CommentsExtension);
       }
       if (this.options.mentions) {
+        extensions.push(MentionNode);
```
jrowny, line 249: would it make more sense for MentionExtension to add this node?

packages/liveblocks-react-tiptap/src/mentions/MentionExtension.ts
```diff
@@ -110,86 +108,9 @@ const notifier = ({
...
 };
 
-export const MentionExtension = Node.create<MentionExtensionOptions>({
-  name: LIVEBLOCKS_MENTION_TYPE,
-  group: "inline",
-  inline: true,
-  selectable: true,
-  atom: true,
-
+export const MentionExtension = Extension.create<MentionExtensionOptions>({
```
jrowny, line 111: Just so I understand, this was turned into an extension so you could use the node elsewhere?

jrowny (changes requested):
overall this is super great! Thanks so much for doing this. For compliance reasons, we may need to create the package in our own PR and then bring in the changes once we have it building.

packages/liveblocks-react-tiptap/src/mentions/MentionExtension.ts
YousefED: Yes! In our case, when we create the blocknote schema

packages/liveblocks-react-blocknote/src/version-history/HistoryVersionPreview.tsx
YousefED: It's not just the type, because the tiptap component requires `editor. _tiptapEditor`. 

Because this is an undocumented BlockNote API (to access the underlying tiptap editor), I don't want to ask consumers to access / pass this around.

By building our own "BlockNote" components (even if they just wrap the tiptap ones), we're also more future proof in case we want to make changes to the BlockNote, but not the TipTap implementation

YousefED (commented):
Thanks for the review! Replied to some of the comments. The other ones I'll take into account on the next iteration

jrowny:
Moving to https://github.com/liveblocks/liveblocks/pull/2239 to finish this work

PR #2080, opened 2024-11-19
Add missing functionality to FloatingComposer in Tiptap

### LINKED ISSUE #2065: react-tiptap: FloatingComposer onComposerSubmit prop not working
**Describe the bug**

Passing onComposerSubmit as prop to the FloatingComposer element of the @lexical/react-tiptap package does nothing.

**Proposed fix**

Looks like the prop isn't used at all in the [source file](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react-tiptap/src/comments/FloatingComposer.tsx) of the FloatingComposer.

Looking at the implementation in @lexical/react-lexical, I suggest adding ` props.onComposerSubmit?.(comment, event);` to the handleComposerSubmit function: 

```typescript
  const handleComposerSubmit = useCallback(
    (comment: ComposerSubmitComment, event: FormEvent<HTMLFormElement>) => {
      const { body } = comment;
      if (!editor) {
        return;
      }
      props.onComposerSubmit?.(comment, event);
      if (event.defaultPrevented) return //straight from @lexical/react-lexical
      event.preventDefault();

      const thread = createThread({
        body,
      });
      editor.commands.addComment(thread.id);
    },
    [editor, createThread]
  );
```

jrowny:
- Fixes onComposerSubmit
- adds "escape" to close composer
- adds attachements/metadata support

Fixes #2065

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## 2.12.2 (Not published yet)
 
+### `@liveblocks/react-tiptap`
+
+- Add new options for `useLiveblocksExtension()` to allow setting
+  initialContent, experimental offline support, and the field name
+- Update floating composer to support onComposerSubmit handler and closing the
+  composer with the escape key
+
 ### `@liveblocks/zustand`
 
 - Add support for Zustand v5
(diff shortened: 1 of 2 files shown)


nimeshnayaju (approved):
🚀

PR #2082, opened 2024-11-21
Update Zustand examples

SugarDarius:
Following the add in `v2.12.2` to support Zustand v5 (cf #2077) examples need to be updated as well.
It updates the dependencies and some imports.

LMK I missed something 🙏🏻 

Thanks

### DIFF
--- a/examples/zustand-flowchart/package-lock.json
+++ b/examples/zustand-flowchart/package-lock.json
@@ -8,15 +8,16 @@
       "name": "@liveblocks-examples/zustand-flowchart",
       "license": "Apache-2.0",
       "dependencies": {
-        "@liveblocks/client": "^2.11.0",
-        "@liveblocks/zustand": "^2.11.0",
+        "@liveblocks/client": "^2.12.2",
+        "@liveblocks/zustand": "^2.12.2",
         "@types/node": "18.13.0",
         "@types/react": "18.3.3",
         "@types/react-dom": "18.0.11",
         "next": "13.1.6",
         "react": "18.2.0",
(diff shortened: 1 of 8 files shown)


examples/zustand-flowchart/package-lock.json
```diff
@@ -265,6 +266,34 @@
         "react-dom": ">=17"
       }
     },
+    "node_modules/@reactflow/background/node_modules/zustand": {
+      "version": "4.5.5",
```
nvie, line 270: Seems like reactflow uses Zustand v4, so now both versions will get bundled. Is there a way to either upgrade reactflow to a later version that also uses Zustand, or to keep using Zustand v4 in this example? (Hopefully the upgrade to V5 is backward compatible?)

nvie (approved):
LGTM, but it looks like we're now bundling Zustand v4 _and_ v5 in at least one example here now. It'd be ideal to only bundle one.

examples/zustand-flowchart/package-lock.json
SugarDarius: Yes it's backward compatible. I tested locally and everything was looking fine. I will check if I can updated `reactflow` just in case 👍🏻

SugarDarius: So I checked and `reactflow` do not supports yet `zustand` v5.  But as stated it's fully backward compliant. 
I just updated `reactflow` to its latest version to remove a deprecation warning.

So it's fine from my side 🙂

CTNicholas (approved):

PR #2084, opened 2024-11-21
Simplify implementation of `EventSource.subscribeOnce()`

nvie:
I saw an opportunity to simplify this datastructure so we wouldn't need to maintain two sets here internally. Once is enough. Should make it easier to read. There were extensive tests for the subscribeOnce behavior, and they are all still passing.


jrowny (approved):
<3 net negative PRs. While it may be leaving our code base, I'm adding One Time Observers to my list of potential band names.

marcbouchenoire (approved):
🧹

PR #2091, opened 2024-11-27
Clarify `resolveUsers` doc

FlowFlorent:
#### Description

Specifies that ordering and count are necessary for the return of user objects in `resolveUsers`

```ts
  /**
   * A function that returns user info from user IDs.
   * You should return a list of user objects of the same size, in the same order.
   */
  resolveUsers?: (
    args: ResolveUsersArgs
  ) => OptionalPromise<(U["info"] | undefined)[] | undefined>;
};
```

### DIFF
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -424,6 +424,7 @@ export type ClientOptions<U extends BaseUserMeta = DU> = {
 
   /**
    * A function that returns user info from user IDs.
+   * You should return a list of user objects of the same size, in the same order.
    */
   resolveUsers?: (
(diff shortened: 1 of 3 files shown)


nvie (approved):

marcbouchenoire (approved):
❤️ Could you apply the same change to `resolveRoomsInfo` while we're there? The docs already use the same sentence for it: "You should return a list of room info objects of the same size and in the same order"

marcbouchenoire:
@FlowFlorent I'm merging this now so that it can be part of `2.13.0` which I'm about to release.

FlowFlorent:
Thank you @marcbouchenoire !

PR #2092, opened 2024-11-28
Remove wrong `Timestamp` export

marcbouchenoire:
I just noticed that we currently export the `Timestamp` primitive from `@liveblocks/react-ui`, but it should only be exported from `@liveblocks/react-ui/primitives`. We didn't use it ourselves in other packages so no other changes than the removal.


SugarDarius (approved):
✅

PR #2093, opened 2024-11-28
User notification settings

SugarDarius:
This PR implements a new APIs to get and edit user notification settings → receiving a notification on a specific channel (email, slack, ...). 
```ts
// User notification settings
{
   email: {
      thread: true,
      textMention: true,
      $whatever: true, // coming from the augmentation - custom kinds defined in `ActivitiesData`
   },
   // ... 
}
```
_**⚠️  User Notification Settings are different than Inbox Notification ⚠️**_

The following APIs are introduced:

```tsx
// @liveblocks/client
const settings = await client.getNotificationSettings();
const updatedSettings = await client.updateNotificationSettings({ ... });

// @liveblocks/react - top level import and from `createLiveblocksContext()` factory
const [{ isLoading, error, settings }, update] = useNotificationSettings(); // Suspense available
const onSave = () => {
  update({ ... }); // optimistic update
}

// @liveblocks/node
const settings = await liveblocks.getNotificationSettings({ userId });
const updatedSettings = await liveblocks.updateNotificationSettings({ userId, data: { ... } });
await liveblocks.deleteNotificationSettings({ userId });
```

### 🏁 Checklist

- [x] Expose hooks suspense version
- [x] Expose hooks from `createLiveblocksContext()` factory
- [x] Expose developers utils (e.g check if every kinds are enabled and `DeepPartial`)
- [x] Add unit tests
- [x] Add type-levels tests
- [x] Add hooks e2e test in sandbox
- [x] Check e2e test after merge of backend on `main` (no timeout nor flaky)
- [x] Clean example - Remove `XXX` comments
- [x] Starter kit
- [x] Update documentation
  - [x] client
  - [x] React
  - [x] node client
  - [x] REST API
- [x] Add guide
- [x] Update CHANGELOG.md
  - [x] client
  - [x] React
  - [x] node client


packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -299,10 +301,26 @@ function makeLiveblocksExtrasForClient(client: OpaqueClient) {
...
+        console.warn(
+          `Polling new channel notification settings failed: ${String(err)}`
+        );
+        throw err;
+      }
+    },
+    config.CHANNEL_NOTIFICATION_SETTINGS_INTERVAL,
+    { maxStaleTimeMs: config.CHANNEL_NOTIFICATION_SETTINGS_MAX_TIME }
+  );
+
```
nvie, line 318: I don't think this poller is really needed for your use case? We could remove it if you feel like it's relatively a lot of added complexity for little value, because I don't think notification settings are that likely to change (unlike threads and notifications, etc). When they do, is it really important that connected clients see it reflect immediately?

If you think the polling is nice to have, feel free to keep it in.
But if you feel it adds too much complexity for this use case, feel free to remove it.
At the very least, maybe we could lower the poll interval a bit for this resource, as I don't think it will be necessary to poll every minute. (The way the poller works anyway is that any time the window gets re-focused, it will trigger a poll anyway if the data is stale (as defined by the `MAX_STALE_TIME` value.)

packages/liveblocks-react/src/config.ts
```diff
@@ -19,4 +19,7 @@ export const config = {
 
   NOTIFICATION_SETTINGS_POLL_INTERVAL: 1 * MINUTES,
   NOTIFICATION_SETTINGS_MAX_STALE_TIME: 5 * SECONDS,
+
+  CHANNEL_NOTIFICATION_SETTINGS_INTERVAL: 1 * MINUTES,
+  CHANNEL_NOTIFICATION_SETTINGS_MAX_TIME: 5 * SECONDS,
```
nvie, line 24: See my previous comment. If we keep the poller, I'd suggest to lower the polling interval for this resource to maybe every 10 minutes + 1 minutes respectively or so. The stale time can also be lowered a bit for this one I think.

For full context:

- Interval: 10 minutes
- Max stale time: 1 minute

Means:
- While the window remains focused and active, poll every 10 minutes
- If the window gets re-focused (user went to another app or browser tab, then comes back to this window) then: if the last time the poll happened was <= 1 minute ago, it won't do anything (next "normal" poll will automatically happen in < 10 mins)
- If the last time a poll happened was > minute ago, it will trigger a poll now. After that poll succeeds, the next "normal" poll will happen 10 mins later again.

Also, let's put "stale" in the config variable name at least:

```suggestion
  CHANNEL_NOTIFICATION_SETTINGS_INTERVAL: 10 * MINUTES,
  CHANNEL_NOTIFICATION_SETTINGS_MAX_STALE_TIME: 1 * MINUTES,
```

nvie (approved):
This PR is looking great! Sure wasn't easy to navigate the umbrella store internals, but you did a great job, implementing it pretty similar to all the other managed resources! 💯

I left a few comments, but they're pretty minor suggestions.

packages/liveblocks-react/src/liveblocks.tsx
SugarDarius: Yes and no 🙂 
I'm glad you're pointing this. To me we have to have it because we will offer methods in the `@liveblocks/node` client to update the channel notifications. 

So I think at some use cases where you could update the channel notification settings from the node client such as for example support request or when a user in an app enables an integration like Slack and the app needs to update the channel notification settings by itself and do not expect the user to refresh the app to see the new channel notification settings.

So I'd keep the poller for now because I think it's important to have but based on your suggestion below I will lower the interval value. 
However 10 mins since a bit too much from my perspective. I will tend to see something more in between like 6 mins instead. 

Thanks 🙏🏻

packages/liveblocks-react/src/config.ts
SugarDarius: See my [comment](https://github.com/liveblocks/liveblocks/pull/2093#discussion_r1865900369) above ☝🏻

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -736,12 +794,23 @@ export class UmbrellaStore<M extends BaseMetadata> {
       (hv) => hv
     );
 
+    const channelNotificationSettings = DerivedSignal.from(
+      this.baseChannelNotificationSettings,
+      this.optimisticUpdates,
+      (baseChannelsNOtificationSettings, updates) =>
```
nvie, line 800: NOticed this tiny typo

```suggestion
      (baseChannelsNotificationSettings, updates) =>
```

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -799,6 +868,14 @@ export class UmbrellaStore<M extends BaseMetadata> {
     return this.outputs.versionsByRoomId.subscribe(callback);
   }
 
+  public get4(): ChannelsNotificationSettings {
+    return this.outputs.channelNotificationSettings.get();
+  }
+
+  public subscribe4(callback: () => void): () => void {
+    return this.outputs.channelNotificationSettings.subscribe(callback);
+  }
```
nvie, line 877: Thanks for keeping this consistent with my temporary naming scheme 🙏

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1928,6 +2065,54 @@ function applyOptimisticUpdates_forSettings(
...
+  for (const optimisticUpdate of optimisticUpdates) {
+    switch (optimisticUpdate.type) {
+      case "update-channels-notification-settings": {
+        const incomingSettings = optimisticUpdate.settings;
+
+        for (const channelKey of keys(incomingSettings)) {
+          const key = channelKey;
+          const channelUpdates = incomingSettings[key];
+
+          if (channelUpdates && typeof channelUpdates === "object") {
```
nvie, line 2091: Out of curiosity, what does the `typeof` check here add? If it's non-nullish, would it ever not be an object?

nvie (commented):
Looked at the Umbrella Store updates, and they're perfectly in line with my ongoing refactorings! 👍  Good stuff!

packages/liveblocks-react/src/umbrella-store.ts
SugarDarius: You're right. Thanks for catching this. I think my brain was going to strong in type check on this one

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1209,6 +1208,24 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
     );
 
+    const channelNotificationSettings: LoadableResource<ChannelsNotificationSettingsAsyncResult> =
+      {
+        signal: DerivedSignal.from(
+          (): ChannelsNotificationSettingsAsyncResult => {
+            const asyncResult = this.#channelsNotificationSettings.get();
+            if (asyncResult.isLoading || asyncResult.error) {
+              return asyncResult;
+            }
```
nvie, line 1218: Tiny nit. I renamed `asyncResult` → `result` elsewhere, you may want to do the same rename here for consistency.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1129,6 +1208,24 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
+
+            return {
+              isLoading: false,
+              settings: nn(this.channelNotificationSettings.signal.get()),
+            };
+          }
+        ),
```
nvie, line 1225: Subtle, but this probably needs a `shallow` equality function here:

```suggestion
          },
          shallow,
        ),
```

docs/pages/api-reference/liveblocks-node.mdx
```diff
@@ -1815,6 +1815,59 @@ try {
 }
 ```
 
+### User Notification Settings [#user-notification-settings]
+
+<Banner title="Private beta">
+
+User notification settings is currently in private beta. If you would like
```
FlowFlorent, line 1822: It will be open beta, anyone can download the package and it will be activated for all

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -792,9 +849,11 @@ export class UmbrellaStore<M extends BaseMetadata> {
   //          \                          | |                 | |    & notifications by ID
   //         | \                         | |      Apply      | |
   //         |   `-> OptimisticUpdates --+--+--> Optimistic --+-+--> Notification Settings    (Part 2)
-  //          \                          |        Updates       |
-  //           `------- etc etc ---------+                      +--> History Versions         (Part 3)
-  //                       ^
+  //          \                          |        Updates    |  |
+  //           `------- etc etc ---------+                   |  +--> History Versions         (Part 3)
+  //                       ^                                 |
+  //                       |                                 +-----> Channels Notification Settings (Part 4)
```
FlowFlorent, line 855:
```suggestion
  //                       |                                 +-----> Notification Settings (Part 4)
```

examples/zeni-comments
marcbouchenoire, line 1: 👀

examples/nextjs-notification-settings/app/version-history-dialog.tsx
```diff
@@ -0,0 +1,108 @@
+import * as Dialog from "@radix-ui/react-dialog";
```
marcbouchenoire, line 1: Should we remove things like that from the example? Everything not related or needed for notification settings

examples/nextjs-notification-settings/app/page.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Providers } from "./providers";
+import { useExampleUserId } from "./use-example-user-id";
+
+// Learn how to structure your collaborative Next.js app
+// https://liveblocks.io/docs/guides/how-to-use-liveblocks-with-nextjs-app-directory
+
+export default function Page() {
+  const userId = useExampleUserId();
+  const roomId = useExampleRoomId(
+    "liveblocks:channel-notifications-settings-examples:nextjs"
```
marcbouchenoire, line 16: `channel-notifications-settings-examples` needs to be updated here and I also saw it in other places so a quick find and replace should cover everything

examples/nextjs-notification-settings/README.md
marcbouchenoire, line 1: This file should get a pass: title, badges, description, image, URLs pointing to the repo or external deployments, steps like needing `LIVEBLOCKS_WEBHOOK_SECRET_KEY` or Resend, etc

packages/liveblocks-node/src/client.ts
```diff
@@ -1641,6 +1643,65 @@ export class Liveblocks {
       throw new LiveblocksError(res.status, text);
     }
   }
+
+  /**
+   * Get notification settings for a user for a project.
+   * @param params.userId The user ID to get the notifications settings for.
+   */
+  public async getsNotificationSettings(params: {
```
marcbouchenoire, line 1651:
```suggestion
  public async getNotificationSettings(params: {
```

marcbouchenoire (commented):
The internals look good to me!

I added some comments on the example but I'm not sure how "done" you consider it to be so feel free to ignore them if it's too early.

Repeating something I said in Slack regarding naming, I wonder if us using "user notification settings" publicly is only due to the renaming transition with "room notification settings" or if we would have chosen to add "user" anyway.

PR #2102, opened 2024-12-06
report text editor once connected

jrowny:
Wait for connection before reporting text editor information

### DIFF
--- a/packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
+++ b/packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
@@ -3,7 +3,7 @@ import { CollaborationPlugin } from "@lexical/react/LexicalCollaborationPlugin";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import type { Provider } from "@lexical/yjs";
 import { kInternal, nn, TextEditorType } from "@liveblocks/core";
-import { useClient, useRoom, useSelf } from "@liveblocks/react";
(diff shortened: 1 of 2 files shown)


SugarDarius (approved):
The fix looks good to me ✅ ✅ 

If you want to avoid to report on every lifecycle updates when `room` reference change then you can use a reference boolean. Such as:
```tsx
const isReported = useRef<boolean>(false);
useEffect(() => {
   if (!isReported.current && status === "connected") {
      room[kInternal].reportTextEditor(TextEditorType.Lexical, "root");
      isReported.current = true;
   }
}, [status]);
```

However I'm bothered a bit by the fact we call directly in the plugin provider / extension `kInternal` which represents the `PrivateRoomApi`. I may be going to far but I think it would be good to "protect" these private api by creating another abstraction like 
```tsx
import { useReportTextEditor } from '@liveblocks/react/_private'
const reportTextEditor = useReportTextEditor() // stable identity reference
useEffect(() => {
   reportTextEditor(TextEditorType.Lexical, "root")
}, [status])
```
And more this hook could be directly retain in memory if the report has already been made 🙂 and avoid to call too much the `text-metadata` endpoint.

Thanks 🙏🏻

SugarDarius:
Following my comment above ☝🏻 I created a [branch](https://github.com/liveblocks/liveblocks/compare/private-use-report-text-editor?expand=1) if you're interested 🙂

jrowny:
@SugarDarius your branch looks good, can you make it a PR and we'll merge that one instead?

SugarDarius:
@jrowny done in #2104 🙂

PR #2107, opened 2024-12-10
Protect private room api in text editors

SugarDarius:
This PR implements a set of new private hooks to achieve two goals:
- "Protect" a bit more our `PrivateRoomApi` by create a new layer of abstraction around direct calls to `room[kInternal]` or `client[kInternal`. 
- Use one single source of truth between `@liveblocks/react-tiptap` and `@liveblocks/react-lexical` to avoid any differences of behaviors, avoid code duplication and increase maintainability.

The following private hooks are introduced:
```tsx
useYjsProvider() // Get Yjs provider
useCreateTextMention() // Create a text mention
useDeleteTextMention() // Delete a text mention
useResolveMentionSuggestions() // Get `resolveMentionSuggestions`
useMentionSuggestionsCache() // Get `mentionSuggestionsCache`
```

### 🏁 Checklist

- [x] Check and verify in examples
- [x] Update CHANGELOG.md


packages/liveblocks-react/src/room.tsx
```diff
@@ -797,6 +799,63 @@ function useReportTextEditor(editor: TextEditorType, rootKey: string): void {
...
+    [room]
+  );
+}
+
+/** @private - Internal API, do not rely on it. */
+function useIsResolveMentionSuggestionsDefined(): boolean {
+  const client = useClient();
+  return React.useMemo((): boolean => {
+    return client[kInternal].resolveMentionSuggestions !== undefined;
+  }, [client]);
```
nvie, line 856: We should only use `useMemo` for computationally-expensive expressions. In this case, you're just doing simple member access, so just doing this should be enough:

```suggestion
  return client[kInternal].resolveMentionSuggestions !== undefined;
```

nvie (approved):

packages/liveblocks-react/src/room.tsx
nvie: Additionally, this hook seems a bit over-specific. Should we expose this as `useResolveMentionSuggestions` instead, and doing the `!== undefined` part in the call site? That way the hook seems a bit more composable/flexible to me, and encourages DRYness.

There are other parts of the code base that already access the `resolveMentionSuggestions` prop ([for example](https://github.com/liveblocks/liveblocks/blob/8ecfd777f7995592187f484bc7b4f7ffb5c0a5a5/packages/liveblocks-react/src/use-mention-suggestions.ts#L24)), maybe they can benefit from the same, more generic, hook?

SugarDarius: @nvie can you PTAL? Thanks 🙏🏻

SugarDarius: Ping @nvie when you get a chance 🙂 ☝🏻

CHANGELOG.md
```diff
@@ -1,4 +1,22 @@
...
+
+- Use new private hooks in plugins.
+
+### `@liveblocks/react-tiptap`
+
+- Use new private hooks in extensions.
+
+### `@liveblocks/react-ui`
+
+- Use new private hooks in components.
```
nvie, line 19: For the changelog entry, we may keep it generic and just briefly mention "internal refactorings". I don't think we have to go into more detail here, and would like to reserve the changelog for public user-facing API changes mostly.

nvie (approved):
Looking good, nicely DRY'ed up and reused between different text editor packages now 👍

packages/liveblocks-react/src/room.tsx
nvie: (Sorry I missed these notifications @SugarDarius! Was just going through all my unread mail today and saw I missed these pings — so sorry about that! GitHub is sending me too many emails, I need to improve my notification settings.)

SugarDarius: @nvie no worries 🙂

PR #2115, opened 2024-12-16
Refactor "refs" abstraction to "signals"

nvie:
When we launched Liveblocks [0.18](https://liveblocks.io/blog/whats-new-in-v0-18) (this [PR](https://github.com/liveblocks/liveblocks/pull/384)), internally I introduced an abstraction I called ImmutableRef. That name has been pretty obtuse, although the pattern was solid.

As of today (~2 years later), we've implemented the same patterns in a couple of different places and in a few different ways. Also, to the world this pattern has become known as a "Signal". There is currently a [TC39 proposal](https://github.com/tc39/proposal-signals) (stage 1).

So this PR effectively renames the "Refs" to "Signals", for familiarity. Also, I did a few passes to make them more closely resemble the TC39 proposal (it's not 100% the same yet, but close enough for our use cases), and added some more tests. I've removed all "Refs".

The following Signal classes/helpers are introduced:
- `ISignal` – the minimal _interface_ one needs to implement to have a signal
- The concrete `Signal` class – much like `Signal.State` from the TC39 proposal
- The concrete `DerivedSignal` class – much like `Signal.Computed` from the TC39 proposal
- The concrete `MutableSignal` class – this is much like how we mutate our umbrella store. There is no equivalent of such signal in the TC39 proposal.

@nimeshnayaju Note how conceptually the `ManagedOthers` class (previously named `OthersRef`) now resembles the update pattern we use inside `UmbrellaStore`: you mutate some internal state, and you observe some clean external state. I haven't touched the Umbrella Store in this PR yet, but we could totally DRY this implementation up now to benefit.

This work benefits us in a few ways:

- It should make the existing internals a lot easier to follow if you understand Signals already
- The proposal for Signals is currently in stage 1 (= being considered). This implementation is reasonably close to that implementation and it should be easy to swap our abstraction out for the real thing once it has landed in browsers in the future.
- We can DRY up the same pattern used elsewhere (eg our various "Stores" and the Umbrella Store would also be special cases of this pattern). I will do this in a later PR.
- It'd be a requirement for Storage V2.

### DIFF
--- a/packages/liveblocks-core/src/lib/Signal.ts
+++ b/packages/liveblocks-core/src/lib/Signal.ts
@@ -10,7 +10,7 @@ import type { JsonObject } from "../lib/Json";
 import { compactObject, raise } from "../lib/utils";
 
 const kSinks = Symbol("kSinks");
-const kNotify = Symbol("kNotify");
+const kTrigger = Symbol("kTrigger");
 
 //
 // Before the batch is run, all sinks (recursively all the way down) are marked
@@ -40,20 +40,18 @@ const kNotify = Symbol("kNotify");
 // - If A changes, then all sinks (B, C, and D) will be marked dirty.
 //
 // - Because some of A's sinks are being watched (in this case, D has at least
-//   one subscriber), A will notify B and C that its value has changed.
+//   one subscriber), A will trigger B and C that its value has changed.
 //
 // - Both B and C re-evaluate and may or may not have changed. Three
 //   possibilities:
-//   1. Neither B and C have changed → D will *NOT* be notified
-//   2. Either B or C has changed    → D *will* be notified
-//   3. Both B and C have changed    → D *will* be notified (but only once!)
+//   1. Neither B and C have changed → D will *NOT* be triggered
+//   2. Either B or C has changed    → D *will* be triggered
+//   3. Both B and C have changed    → D *will* be triggered (but only once!)
 //
-// - If in the previous step D has been notified, it will re-evaluate. If it
+// - If in the previous step D has been triggered, it will re-evaluate. If it
 //   has changed itself, it will notify its normal subscriber.
 //
-let activeBatch: {
-  toNotify: Set<AbstractSignal<any>>;
-} | null = null;
+let signalsToTrigger: Set<AbstractSignal<any>> | null = null;
 
 /**
  * Runs a callback function that is allowed to change multiple signals. At the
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/lib/Signal.ts
```diff
@@ -0,0 +1,416 @@
...
+const kNotify = Symbol("kNotify");
+
+//
+// Before the batch is run, all sinks (recursively all the way down) are marked
+// dirty. This already is enough if all we ever used were .get() calls.
+//
+// However, to ensure active subscription notifications also work, we need to
+// keep track of which Signals to notify. Any time the value of a Signal
+// changes, it will trigger an event to all subscribers, but in addition, it
+// will also
```
marcbouchenoire, line 22: "it will also _" 👀

marcbouchenoire (approved):
I haven't found anything in particular to point out except for an incomplete comment 👍

nimeshnayaju:
😍 Great work @nvie. l  I can see this replacing the event source logic used in the umbrella store too. Do you think it could be a fit there?

packages/liveblocks-core/src/lib/Signal.ts
```diff
@@ -0,0 +1,435 @@
...
+// - If A changes, then all sinks (B, C, and D) will be marked dirty.
+//
+// - Because some of A's sinks are being watched (in this case, D has at least
+//   one subscriber), A will trigger B and C that its value has changed.
+//
+// - Both B and C re-evaluate and may or may not have changed. Three
+//   possibilities:
+//   1. Neither B and C have changed → D will *NOT* be triggered
+//   2. Either B or C has changed    → D *will* be triggered
+//   3. Both B and C have changed    → D *will* be triggered (but only once!)
```
nimeshnayaju, line 49: 🤌

nimeshnayaju (approved):
This is such a clean refactor. Really like the new abstraction. Good riddance Refs. 

Also, on a side note, this could be a library in itself - totally see it being useful in many other projects outside Liveblocks.

packages/liveblocks-core/src/lib/EventSource.ts
```diff
@@ -125,37 +111,26 @@ export function makeEventSource<T>(): EventSource<T> {
...
-    _forceClear,
     count,
 
     waitUntil,
-    pause,
-    unpause,
+
+    [Symbol.dispose]: (): void => {
+      _observers.clear();
+    },
```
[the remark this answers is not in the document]
nvie: Exactly! Not sure if in this particular case, using `using` will make much sense — when would you ever create an event source that you dispose of when the variable goes out of scope, but this is great for long-lived event sources that eventually will be cleaned up. Think a room provider unmounting. When explicitly disposed, it will now clear all subscribers, if any. This is great for garbage collection.

nvie:
> Also, on a side note, this could be a library in itself - totally see it being useful in many other projects outside Liveblocks.

I didn't want to introduce a dependency here for now, but it totally could for sure. It's super useful indeed, and there are lots of places in our current code base that are actually Signal or MutableSignal patterns. I'll refactor those next – first wanted to get your approval on this bit.

nvie:
> 😍 Great work @nvie. l I can see this replacing the event source logic used in the umbrella store too. Do you think it could be a fit there?

Absolutely, working on DRYing all that up right now. We have so many places in our packages that benefit from this abstraction. My goal is to rewrite them all so they're all much easier to work with and follow.

PR #2117, opened 2024-12-17
Use #private fields

nvie:
Since #2116 is now merged, we can safely use private fields everywhere (truly private at runtime, and minifiable in builds), instead of using TypeScript-level private field identifiers. This should further reduce the bundle size, and has the added benefit of better runtime isolation.


marcbouchenoire (approved):
#️⃣

PR #2120, opened 2024-12-18
add v2 update options to client

jrowny:
Adds ` {useV2Encoding_experimental:true};` option to LiveblocksYjsProvider allow sending/receiving encoded updates. The backend does not work yet.

Steps before we can merge this PR:
- [x] add v2 options to client
- [x] publish temp package to update core package on backend to grab new types
- [x] update backend to accept v2 instructions
- [x] add v1/v2 tests to this PR once backend is working (there are updates on backend, once that goes out, I will add tests to FE)


packages/liveblocks-yjs/src/doc.ts
```diff
@@ -40,19 +44,26 @@ export default class yDocHandler extends Observable<unknown> {
...
     if (stateVector) {
       if (!readOnly) {
         // Use server state to calculate a diff and send it
         try {
-          const localUpdate = Y.encodeStateAsUpdate(
+          // send v1 or v2update according to client option
+          const encodeUpdate = this.useV2Encoding
+            ? Y.encodeStateAsUpdateV2
+            : Y.encodeStateAsUpdate;
+          const localUpdate = encodeUpdate(
```
SugarDarius, line 66: Mostly for my own knowledge but wasn't it more safe to handle the server update on the same encoding option for every operation? I mean are we sure it's safe to receive a v2 update and potentially make an encode operation on v1 as `useV2Encoding` can be `false`? 🤔 
I read in your back-end PR that messages can be v1 or v2 and won't affect any functionality but I'm wondering if wouldn't be more robust and safe to strictly make the encode operation if `v2 === true && this.useV2Encoding === true` to avoid any dissonances no? 🤔

packages/liveblocks-yjs/src/index.ts
```diff
@@ -236,6 +245,10 @@ export class LiveblocksYjsProvider<
     }
   };
 
+  get useV2Encoding(): boolean {
+    return this.options.useV2Encoding_experimental ?? false;
+  }
```
SugarDarius, line 250: I suppose this getter is required for reason no? Will it be reuse by some other objects?

SugarDarius (approved):

packages/liveblocks-yjs/src/doc.ts
jrowny: Sending updates is completely independent of receiving them. This specific line of code happens to be in a function where we perform a "sync" if we receive state vector which tells our client to respond with an update containing anything since that state vector. State vectors are unaffected by v1 or v2 encoding.

packages/liveblocks-yjs/src/index.ts
jrowny: Not really, it originally had more logic earlier in my development cycle, now it's mostly vestigial/convenience.

nvie (approved):
Seems a bit weird to me that Yjs cannot detect from the payload itself whether it's a v1 or v2 update, but given that fact, this PR looks good to me! 👍

jrowny:
> Seems a bit weird to me that Yjs cannot detect from the payload itself whether it's a v1 or v2 update, but given that fact, this PR looks good to me! 👍

yea, because v1 is just binary and the first bit contains the number of clients who have edited the document. We could guess by setting a unique value in the first bit of a v2 update, which would be unlikely to occur in the first bits of a v2 update. But I think it's just a bit more confusing and prefer to have it be more explicit.

PR #2123, opened 2024-12-18
Refactor more of our state management to signals

nvie:
This PR refactors more of our state management towards Signals.

It's a big change as always with the umbrella store, and I certainly do not consider this done yet, but all the remaining todos in the comments I left I consider optional and could also be tackled in a later PR. I think this is already a good step forward, and all tests are (still) passing.

@nimeshnayaju I'd love to have a chat about next steps here, as I'm figuring out how to best organize the code in a readable way, in particular with the way we do the reads of the "AsyncResult" versions of Threads, Notifications, Settings, Versions, etc. And how to make our Paginated resources real signals too (they currently aren't convertible 1:1).


packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -532,22 +533,35 @@ export class SinglePageResource {
...
+ *        'room-def': { threads: "replies_and_mentions" },
+ *        'room-xyz': { threads: "none" },
+ *      }
  */
-export type UmbrellaStoreState<M extends BaseMetadata> = {
+type SettingsByRoomId = Record<RoomId, RoomNotificationSettings>;
+
+type PermissionHintsByRoomId = Record<RoomId, Set<Permission>>;
+
+export type CleanThredifications<M extends BaseMetadata> =
```
SugarDarius, line 558:
```suggestion
export type CleanThreadifications<M extends BaseMetadata> =
```
no?

SugarDarius (approved):
From what I see and understand the changes are looking okay to me 🙂 
I will surely need your help to apply them correctly in #2093 though ^^

packages/liveblocks-react/src/umbrella-store.ts
nvie: Sure! Will update the name of this contraction!

PR #2125, opened 2024-12-19
Drop support for React 17

nvie:
Drops support for React 17 in the next version.

There was a lot of internal complexity necessary to support batch updates and avoiding the dreaded [zombie child problem](https://liveblocks.io/docs/platform/troubleshooting#stale-props-zombie-child)). The issue is that these bugs are subtle, and it's hard to garantee we correctly wrap the provided batch wrapper around every possible state update we do. I don't even think we have been doing this correctly with all new APIs we have introduced since 0.18.

Rather than trying to meticulously track and wrap all state updates, dropping support for React 17 is much easier path. If users are stuck on React 17 or lower, they can still decide not to upgrade Liveblocks and keep using 2.14.0, which is the last version that supports React 17.

After upgrading to React 18, users should not experience subtle state update bugs, and should also notice a performance improvement.

It also allows us to drop the shims for `useSyncExternalStore`, `useId`, `useLayoutEffect`, etc, shrinking our bundle size and removing a dependency.


packages/liveblocks-react/src/use-sync-external-store-with-selector.ts
nvie, line 1: It looks like React only now natively ships the `useSyncExternalStore` hook, and not our beloved `useSyncExternalStoreWithSelector`, so I decided to vendor-in this implementation, straight from the shim. (This code is not touched, only adjusted from Flow to TypeScript here.)

SugarDarius (approved):

CTNicholas (approved):

marcbouchenoire (approved):
You got all the ones I had in mind for `react-ui` 👌

nvie:
Thanks for polishing the upgrade guide, @CTNicholas!

PR #2128, opened 2024-12-20
Centralize shared Rollup configs

marcbouchenoire:
This PR started in https://github.com/liveblocks/liveblocks/tree/tiptap-floating-toolbar, where `@rollup/plugin-typescript` was throwing warnings about our usage of ambient types with Tiptap. We only use `@rollup/plugin-typescript` as a `configPlugin` to author our Rollup configs directly in TypeScript, so I started this PR by moving our configs to vanilla JS (`.mjs` since Rollup expects a module) so we could get rid of it. Then I thought it would be a good time to centralize our 3 copy-pasted configs across `react-ui`, `react-tiptap`, and `react-lexical`, so it's now living in `shared`.

I tried pinning all dependencies to avoid differences between then and now and somehow one issue remained that didn't exist before: we preserve `"use client"` directives but somehow they were being duplicated, so I added a small custom plugin that specifically removes duplicated `"use client"` directives.

use-client


nvie (approved):
I don't know enough about the Rollup details specifically, but I love the consolidation into a single shared dev dependency — nice work! 👍

SugarDarius (commented):
I do not know that much about rollup but from what I see and read it's looking  good ✅

PR #2130, opened 2024-12-20
Refactor more store internals to signals

nvie:
Next steps of refactoring of the umbrella store internals. Like last time, there is more refactoring to do here and I'm not done yet, but this PR is a good midpoint that should not change any existing behavior and should be good to merge as-is.

Some of the ideas we had in the past are starting to take shape now slowly. The most important one here being that I've started to isolate each "chunk" of the umbrella store into its own ministore.

Each ministore has:

- A single signal for the data it's managing
- A bunch of mutations that make sense for that particular data structure

The umbrella store just holds together these ministores mostly.

I want to eventually split these up into separate modules for even clearer separation of concerns, but I decided not to yet, to avoid unnecessarily complex merge conflicts with Aurel's current WIP. We can break them out later.

### DIFF
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -413,7 +413,7 @@ function useInboxNotifications_withClient<T>(
     };
   }, [poller]);
 
-  // XXX There is a disconnect between this getter and subscriber! It's unclear
+  // XXX_vincent There is a disconnect between this getter and subscriber! It's unclear
   // why the getInboxNotificationsLoadingState getter should be paired with
   // subscribe1 and not subscribe2 from the outside! (The reason is that
   // getInboxNotificationsLoadingState internally uses `get1` not `get2`.) This
@@ -946,7 +946,7 @@ function useUserThreads_experimental<M extends BaseMetadata>(
     };
   }, [poller]);
 
-  // XXX There is a disconnect between this getter and subscriber! It's unclear
+  // XXX_vincent There is a disconnect between this getter and subscriber! It's unclear
   // why the getUserThreadsLoadingState getter should be paired with subscribe1
   // and not subscribe2 from the outside! (The reason is that
   // getUserThreadsLoadingState  internally uses `get1` not `get2`.) This is
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -479,10 +479,10 @@ function useMarkInboxNotificationAsRead_withClient(client: OpaqueClient) {
       client.markInboxNotificationAsRead(inboxNotificationId).then(
         () => {
           // Replace the optimistic update by the real thing
-          store.updateInboxNotification(
+          store.markInboxNotificationRead(
             inboxNotificationId,
-            optimisticUpdateId,
-            (inboxNotification) => ({ ...inboxNotification, readAt })
+            readAt,
+            optimisticUpdateId
           );
         },
         () => {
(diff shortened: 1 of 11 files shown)


packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -588,9 +604,225 @@ export type CleanNotifications = {
   notificationsById: Record<string, InboxNotificationData>;
 };
 
+function createStore_forNotifications() {
```
SugarDarius, line 607: I really like this new approach a lot 👍🏻

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1063,44 +1210,36 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
   public createThread(
-    optimisticUpdateId: string,
+    optimisticId: string,
     thread: Readonly<ThreadDataWithDeleteInfo<M>>
   ): void {
     // Batch 1️⃣ + 2️⃣
     batch(() => {
-      this.removeOptimisticUpdate(optimisticUpdateId); // 1️⃣j
-      this.#mutateThreadsDB((db) => db.upsert(thread)); // 2️⃣
+      this.optimisticUpdates.remove(optimisticId); // 1️⃣j
```
marcbouchenoire, line 1241: It was here before but:
```suggestion
      this.optimisticUpdates.remove(optimisticId); // 1️⃣
```

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -588,9 +604,225 @@ export type CleanNotifications = {
...
+
+    // XXX_vincent Remove this eventually
+    force_set: (
+      mutationCallback: (lut: NotificationsLUT) => void | undefined | boolean
+    ) => signal.mutate(mutationCallback),
+    invalidate: () => signal.mutate(),
+  };
+}
+
+function createStore_forRoomNotificationSettings() {
```
SugarDarius, line 706: Can't we type the return type of `createStore_*` functions?
I'd like to ensure we always set `signal` and `invalidate` and `force_set` as optional such as:
```ts
type Store<T> = {
    signal: SignalAsReadonly,
    invalidate: () => void,
    force_set?: () => void
} & T;
function createStore_whatever(): Store<{...}> {}
```

marcbouchenoire (approved):
My understanding of the umbrella store is not at 100% after the multiple refactorings passes but what I'm reading here makes sense! (and is easier to compare between different types of stores given that there are less one-off methods now)

SugarDarius (approved):
It's going to the right direction! I really like it 🚀

packages/liveblocks-react/src/umbrella-store.ts
nvie: Thanks, I'll just remove those 1️⃣ and 2️⃣ labels now, since the code has become so small it's easy to see at a glance.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -413,13 +413,12 @@ function useInboxNotifications_withClient<T>(
...
   }, [poller]);
 
-  // TODO(vincent+nimesh) There is a disconnect between this getter and
-  // subscriber! It's unclear why the getInboxNotificationsLoadingState getter
-  // should be paired with subscribe1 and not subscribe2 from the outside! (The
-  // reason is that getInboxNotificationsLoadingState internally uses `get1`
-  // not `get2`.) This is strong evidence that
-  // getInboxNotificationsLoadingState itself wants to be a Signal! Once we
-  // make it a Signal, we can simply use `useSignal()` here! ❤️
+  // XXX_vincent There is a disconnect between this getter and subscriber! It's unclear
```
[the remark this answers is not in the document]
nvie: Yep. They only break CI if it's `XXX`, that's why I added `XXX_vincent` here. So CI wouldn't break right now for this PR, but I will tackle these next.

packages/liveblocks-react/src/umbrella-store.ts
nvie: I will do it once the actual types have been settled. The `force_set` and `invalidate` methods should both go away soon (they're anti-patterns), but some code currently still relies on it.

🔜

packages/liveblocks-react/src/umbrella-store.ts
nvie: Yeah, should make the updates much more "local" and thus easier to reason about.

PR #2139, opened 2024-12-23
Fix two re-rendering regressions

nvie:
This PR fixes two re-rendering regressions I introduced with the 2.15.0 refactorings:

- A forgotten `shallow` caused unnecessary re-renders in `useThreads` and `useInboxNotifications`
- The `useInboxNotifications` only got re-rendered when threads changed (not when only notifications changed)

I found these after doing more cleanup (removing the noise made these issues stand out). I'm going to open a separate PR for that cleanup work, but wanted these fixes to stand out from that noise.

### DIFF
--- a/packages/liveblocks-react/src/umbrella-store.ts
+++ b/packages/liveblocks-react/src/umbrella-store.ts
@@ -33,6 +33,7 @@ import {
   MutableSignal,
   nanoid,
   nn,
+  shallow,
   Signal,
   stringify,
 } from "@liveblocks/core";
@@ -617,7 +618,7 @@ function createStore_forNotifications() {
   }
 
   function markAllRead(readAt: Date) {
-    return signal.mutate((lut) => {
+    signal.mutate((lut) => {
       for (const n of lut.values()) {
         n.readAt = readAt;
       }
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -617,7 +618,7 @@ function createStore_forNotifications() {
   }
 
   function markAllRead(readAt: Date) {
-    return signal.mutate((lut) => {
```
marcbouchenoire, line 620: Was this `return` unnecessary?

marcbouchenoire (approved):
Did a test catch the regression?

packages/liveblocks-react/src/umbrella-store.ts
nvie: None of the other similar methods return it. It's a `void` anyway, so it doesn't matter.

SugarDarius (approved):
Looking good ✅

nvie:
> Did a test catch the regression?

Nope! These were found after I cleanup things and by removing the noise the bug became apparent.

Thinking about it more, actually I don't think technically it's a bug in 2.15.0 after all, as in practice, we're still invalidating the entire umbrella store whenever _any_ resource updates, so technically this isn't a bug yet. But one of the refactoring goals still is to only update rerender a resource when that resource changes, so it would show up as a bug once we make that change.

PR #2141, opened 2024-12-23
Remove `flushSync` fallback

marcbouchenoire:
I found a leftover from the React <18 clean up. For additional context, we only use [`flushSync`](https://react.dev/reference/react-dom/flushSync) in [a utility to support exit animations](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react-ui/src/utils/Persist.tsx).


nvie (approved):
🧹

SugarDarius (approved):

PR #2146, opened 2025-01-02
Implement conditional signal subscriptions

nvie:
This PR implements dynamic tracking of signals for `DerivedSignal`.

Previously:

```ts
const counter = new Signal(0);
const isEven = DerivedSignal.from(counter, n => n % 2 === 0);
const parity = DerivedSignal.from(isEven, even => even ? "even" : "odd");
```

This syntax is still convenient and succinct for simple signals, but because the dependent signals needed to be passed in at declaration time, they are static relationships.

This means that conditional or nesting of signal dependencies was not possible before. This PR makes signals automatically dependent on any signal consumed in the evaluation function.

For example, the above code can now also be written as:

```ts
const counter = new Signal(0);
const isEven = DerivedSignal.from(() => counter.get() % 2 === 0);
const parity = DerivedSignal.from(() => isEven.get() ? "even" : "odd");
```

The previous syntax where you can pass in signals statically and their values get passed into the evaluation function is still supported for convenience and backward-compatibility. You can use either way, or even combine the two approaches.

The syntax

```ts
DerivedSignal.from(signalX, signalY, signalZ, (x, y, z) => { ... });
```

is now just sugar for:

```ts
DerivedSignal.from(() => {
  const x = signalX.get();
  const y = signalY.get();
  const z = signalZ.get();
  ...
});
```

This syntax is both more powerful and more closely aligned with the [TC39 signals proposal (stage 1)](https://github.com/tc39/proposal-signals) as well.

It works through signal read tracking. Any time a derived signal's evaluation function is executed, all signals that are read from (= called `.get()` on) during its execution are being tracked and the derived signal is automatically registered as sinks of those dependent signals.

This makes it possible to keep track of a list of signals, and dynamically listen to a subset of them. Or to make a signal track a list of signals itself as values. We need this functionality to track user/room threads response statuses by querykeys, for example, which we'll implement in a future PR.

### DIFF
--- a/packages/liveblocks-core/src/lib/DefaultMap.ts
+++ b/packages/liveblocks-core/src/lib/DefaultMap.ts
@@ -1,0 +1,71 @@
+/**
+ * Like ES6 map, but takes a default factory function which will be used
+ * to create entries for missing keys on the fly.
+ *
+ * Useful for code like:
+ *
+ *   const map = new DefaultMap(() => []);
+ *   map.getOrCreate('foo').push('hello');
+ *   map.getOrCreate('foo').push('world');
+ *   map.getOrCreate('foo')
+ *   // ['hello', 'world']
+ *
+ */
+export class DefaultMap<K, V> extends Map<K, V> {
+  #_factoryFn?: (key: K) => V;
+
+  constructor(
+    factoryFn?: (key: K) => V,
+    entries?: readonly (readonly [K, V])[] | null
+  ) {
+    super(entries);
+    this.#_factoryFn = factoryFn;
+  }
+
+  get #factoryFn() {
+    if (this.#_factoryFn === undefined) {
+      throw new Error("DefaultMap used without a factory function");
+    }
+    return this.#_factoryFn;
+  }
+
+  /**
+   * Gets the value at the given key, or creates it.
+   *
+   * Difference from normal Map: if the key does not exist, it will be created
+   * on the fly using the factory function, and that value will get returned
+   * instead of `undefined`.
+   */
+  getOrCreate(key: K, factoryFn?: (key: K) => V): V {
+    let value = super.get(key);
+    if (value === undefined) {
+      value = factoryFn !== undefined ? factoryFn(key) : this.#factoryFn(key);
+      this.set(key, value);
+    }
+    return value;
+  }
+
+  /**
+   * Sets the value at the given key.
+   *
+   * Difference from normal Map: if the second arg is a function, it will be
+   * called with the current value as its argument and the return value will be
+   * set instead. If the returned value is `undefined`, it will remove the key,
+   * meaning the next time the key is accessed it will get re-created by the
+   * factory function.
+   */
+  set(key: K, value: V): this;
+  set(key: K, valueFn: (prev: V) => V | undefined): this;
+  set(key: K, value: V | ((prev: V) => V | undefined)): this {
+    if (typeof value === "function") {
+      value = (value as (prev: V) => V)(this.get(key) ?? this.#factoryFn(key));
+    }
+
+    if (value === undefined) {
+      super.delete(key);
+      return this;
+    } else {
+      return super.set(key, value);
+    }
+  }
+}
--- a/packages/liveblocks-core/src/lib/DefaultMap.ts
+++ b/packages/liveblocks-core/src/lib/DefaultMap.ts
@@ -1,5 +1,7 @@
+import { raise } from "./utils";
+
 /**
- * Like ES6 map, but takes a default factory function which will be used
+ * Like ES6 map, but takes a default (factory) function which will be used
  * to create entries for missing keys on the fly.
  *
  * Useful for code like:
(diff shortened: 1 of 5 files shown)


packages/liveblocks-core/src/lib/DefaultMap.ts
```diff
@@ -0,0 +1,71 @@
...
+ *   map.getOrCreate('foo').push('world');
+ *   map.getOrCreate('foo')
+ *   // ['hello', 'world']
+ *
+ */
+export class DefaultMap<K, V> extends Map<K, V> {
+  #_factoryFn?: (key: K) => V;
+
+  constructor(
+    factoryFn?: (key: K) => V,
```
SugarDarius, line 18: I would tend to put a comment here to explain that `factoryFn` is optional in the constructor because in the `getOrCreate` method we can use an override.

packages/liveblocks-core/src/lib/DefaultMap.ts
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * Difference from normal Map: if the second arg is a function, it will be
+   * called with the current value as its argument and the return value will be
+   * set instead. If the returned value is `undefined`, it will remove the key,
+   * meaning the next time the key is accessed it will get re-created by the
+   * factory function.
+   */
+  set(key: K, value: V): this;
+  set(key: K, valueFn: (prev: V) => V | undefined): this;
+  set(key: K, value: V | ((prev: V) => V | undefined)): this {
```
SugarDarius, line 59: Shouldn't we also let developers use a `factoryFn` of their own to be aligned with the behaviors of `getOrCreate`? 🤔 
I surely misunderstood something here. But from my point of view during the review I will to align behaviors between `set` and `getOrCreate`.

packages/liveblocks-core/src/lib/__tests__/DefaultMap.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    const map2 = new DefaultMap<string, string[]>((key) => [key + "!"]);
+    map2.getOrCreate("foo").push("hello");
+    map2.getOrCreate("foo").push("world");
+    map2.getOrCreate("bar").push("bye");
+    expect(new Map(map2)).toEqual(
+      new Map([
+        ["foo", ["foo!", "hello", "world"]],
+        ["bar", ["bar!", "bye"]],
+      ])
+    );
```
SugarDarius, line 40: This one gave me headaches to fully understand. But it helped me to understand that `factoryFn` is only run once if get returns `undefined` at first 😅

packages/liveblocks-core/src/lib/DefaultMap.ts
nvie: Good point, can add!

It should be noted that I took this DefaultMap implementation and its unit tests straight from our backend, where we already have it defined here: https://github.com/liveblocks/liveblocks-backend/blob/main/shared/liveblocks-server/src/lib/DefaultMap.ts

Once this is published, I'll move the backend over to import it from `@liveblocks/core` as well.

packages/liveblocks-core/src/lib/__tests__/DefaultMap.test.ts
nvie: Yep, that's exactly the point of `getOrCreate`!

packages/liveblocks-core/src/lib/DefaultMap.ts
nvie: I wrote this code long ago, and upon re-reading it, I think I may have made this a bit _too_ fancy here. I'm going to remove the callback arg to `.set()`. The purpose of the `DefaultMap` is to add `.getOrCreate()` to maps, next to `.get()` and `.set()`. We should not have to make `.set()` do anything more fancy like this!

I'll also make a similar adjustment in the backend. Thanks for the extra eyes on this!

packages/liveblocks-core/src/lib/DefaultMap.ts
SugarDarius: Good to know thanks 🙏🏻

packages/liveblocks-core/src/lib/DefaultMap.ts
SugarDarius: Ah okay! Yeah in this case I agree! Thanks 🙏🏻

nvie: In other words, instead of `map.set('foo', inc)`, one could always write `map.set('foo', map.getOrCreate('foo') + 1)` instead, which is less magic and more clear.

nvie: Done in 45542725a6ce7613fdf1010e4e39712c25c391fc.

nvie:
@SugarDarius Back-port of these changes to the backend: https://github.com/liveblocks/liveblocks-backend/pull/691/files

marcbouchenoire (approved):
:chef-kiss:

PR #2152, opened 2025-01-06
Another pass at refactoring the umbrella store

nvie:
In this PR, I added some cleanups of the umbrella store internals. I've exposed the AsyncResult getters as their own derived signal in the case of inbox notifications. Next up, I described [this refactoring idea](https://github.com/liveblocks/liveblocks/commit/0ba92e608249a8a00adc04182fdcce77dbccd1db#diff-23360c9df12391b58a2cd9fb409aed3b610672d07c35b4c09f9e421143d9de02), and then implemented it in 59ab0b44c0391ffb8069c95c3ed0cccbf83209c6.

In the same way, we can structure access of room version history, and user / room threads as well. I will add that in a later stage, but first wanted to get feedback/approval on this change.


packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1,3 +1,55 @@
...
+// - `waitUntilRoomNotificationSettingsLoaded()`
+//
+// ------------------------------------------------------------------------
+//
+// These signals also whenever one of their sources change:
+//
+// 1. The loading state changes (from a fetch).
+// 2. The base settings themselves are updated (side-effect of a fetch).
+// 3. Any optimistic updates are applied.
+//
```
nvie, line 52: This idea is now implemented (in 59ab0b44c0391ffb8069c95c3ed0cccbf83209c6). I've left this comment in, because I think room/user threads can follow suit in similar fashion.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1506,9 +1527,7 @@ export class UmbrellaStore<M extends BaseMetadata> {
     batch(() => {
       this.historyVersions.invalidate();
       this.notifications.invalidate();
-      this.optimisticUpdates.invalidate();
       this.permissionHints.invalidate();
-      this.roomNotificationSettings.invalidate();
```
nvie, line 1511: We're getting closer to no longer always invalidating everything. These two could now be removed. To get rid of the rest, more (similar) refactoring needs to happen first.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -961,10 +975,63 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
+      }, shallow);
+
+      return { signal, waitUntilLoaded: resource.waitUntilLoaded };
+    });
+
     this.outputs = {
       threadifications,
       threads,
       notifications,
+      loadingNotifications,
```
nvie, line 1034: Note that we now have `outputs.notifications` and `outputs.loadingNotifications`. These are almost the same. The main difference is:

- `outputs.notifications` signal provides access to the (clean) notifications directly (you can use `sortedNotifications` or `notificationsById` directly)
- `outputs.loadingNotifications` signal provides access to the "async result" version of this (which is how they are exposed through our hooks).

Ideally, we would be able to get rid of `outputs.notifications` here, but it's used to power the `useInboxNotificationsThread` hook still. We'll have to take a deeper look to see if/how we could get rid of this.

I still think having all of these is still potentially confusing, in part because the current names aren't clear enough:
- Input signals:
  - `notifications` (the base "registry", populated by fetches or incoming webhook updates)
  - `#notifications` (tracking pagination state for inbox notifications)
- Output signals:
  - `threadifications` (derived from base `notifications` + optimistic updates)
  - `notifications` (derives from `outputs.threadifications`)
  - `loadingNotifications` (derives from `outputs.notifications`)

Cleaning that structure up is still TBD.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -433,67 +452,69 @@ export class PaginatedResource {
...
+    return this.#cachedPromise;
   }
 }
 
+// TODO Find better name?
+type ManagedResource<T> = {
+  signal: ISignal<T>;
+  waitUntilLoaded: () => UsablePromise<void>;
+};
+
```
SugarDarius, line 501: `LoadableSignal` maybe ? 🤔

SugarDarius (approved):
It's looking good from my side ✅ 
The umbrella store tends to become more stronger and robust, I like what I'm seeing here 👍🏻  It's a great work 🙏🏻 

I just one question for timing purpose. When do you think you'll end the refactoring?
I'm asking because we would need to go further on #2093 soon which works on the umbrella store if you remember 🙂

packages/liveblocks-react/src/umbrella-store.ts
nvie: I think eventually we can combine this abstraction with the "resource" abstraction we already have, so they aren't two separate abstractions that each need to have a name. First I need to rewrite the threads parts similarly.

nvie:
> I think however we might need to start to split the umbrella store file, it's very big to digest 😅

Agreed, but I would like to do it only after I finish the ongoing refactorings / when I feel things are in a steady place, ready to be split up into logical and well-named chunks, and the final pattern has been established.

> I just one question for timing purpose. When do you think you'll end the refactoring?

Since it's a really complex change where the end goal isn't completely clear to myself either, I'm approaching each individual PR as just one step in the right direction, and good-to-be-merged-as-is without breaking changes. I could stop here right now, or take 10 more such steps. Sometimes during the refactoring another, better, pattern emerges, but then I have to apply it consistently everywhere, so unfortunately the work is a bit noisy.

> I'm asking because we would need to go further on #2093 soon which works on the umbrella store if you remember 🙂

Since you're working on a feature, that's always more important. My onging refactorings should not be a blocker. If you cannot (or don't want to) back merge my refactorings into your branch, I'm happy to do so, as I offered earlier. Your work is leading here.

As soon as you work lands on `main`, I'll include your newly added "resources" part of the refactorings automatically. This is the unfortunate downside of maintaining two long-lived branches.

SugarDarius:
> Since you're working on a feature, that's always more important. My ongoing refactorings should not be a blocker. If you cannot (or don't want to) back merge my refactorings into your branch, I'm happy to do so, as I offered earlier. Your work is leading here.

It was just a question here to have a picture of the timing 🙂 For sure I will back merge this branch if you're ready to merge it 🙂 
To me it's important to have them on my PR 🙂

PR #2153, opened 2025-01-07
Fix `useLayoutEffect` warnings

marcbouchenoire:
We removed the isomorphic `useLayoutEffect` workaround in https://github.com/liveblocks/liveblocks/pull/2125 because I thought `useLayoutEffect` stopped giving SSR warnings from React 18.0.0 (March 2022), but it actually stopped from 18.3.0 (April 2024) after [this PR](https://github.com/facebook/react/pull/26395) landed. Since we support all 18 versions, I'm bringing the workaround back in this PR.


packages/liveblocks-react-tiptap/.eslintrc.js
```diff
@@ -26,6 +26,12 @@ module.exports = {
           "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='use']",
         message: "use is only available on React >=19.",
       },
+      {
+        selector:
+          "ImportDeclaration[source.value='react'] ImportSpecifier[imported.name='useLayoutEffect']",
+        message:
+          "useLayoutEffect triggers a warning when executed on the server on React <=18.2.0. Import it from '@liveblocks/react/_private' instead.",
```
nvie, line 33: ❤️

nvie: If you find it convenient, maybe you can put this definition in https://github.com/liveblocks/liveblocks/blob/main/shared/eslint-config/restricted-syntax.js, so it will use the same lint check in all projects? (Only if convenient, not a must.)

nvie (approved):
Nice, thanks for fixing this!

packages/liveblocks-react-tiptap/.eslintrc.js
marcbouchenoire: The only reason I didn't do that was to tweak the message for `@liveblocks/react` to not reference itself (suggesting `./lib/use-layout-effect` instead of `@liveblocks/react/_private`)

marcbouchenoire: Actually, [both are possible](https://github.com/liveblocks/liveblocks/pull/2153/commits/075636ca05d0fea6a022c141f09ad5bd466cb0b5) and they don't conflict each other. 🥳

PR #2161, opened 2025-01-09
DRY up type guards

nvie:
Just a bit of DRYing up of logic used in multiple projects.

### DIFF
--- a/packages/liveblocks-core/src/lib/guards.ts
+++ b/packages/liveblocks-core/src/lib/guards.ts
@@ -1,14 +1,18 @@
-import { isPlainObject } from "./utils";
+export function isPlainObject(
+  blob: unknown
+): blob is { [key: string]: unknown } {
+  // Implementation borrowed from pojo decoder, see
+  // https://github.com/nvie/decoders/blob/78849f843193647eb6b5307240387bdcff7161fb/src/lib/objects.js#L10-L41
+  return (
+    blob !== null &&
+    typeof blob === "object" &&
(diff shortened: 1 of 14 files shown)


marcbouchenoire (commented):
Hopefully none of these changes leak out in public space and infer differently on TS 4.* 🙏

nvie:
@marcbouchenoire These should all be private APIs (only exposed from `core`). In `client` we should indeed not do this.

packages/liveblocks-core/src/lib/guards.ts
```diff
@@ -0,0 +1,18 @@
...
+    blob !== null &&
+    typeof blob === "object" &&
+    Object.prototype.toString.call(blob) === "[object Object]"
+  );
+}
+
+/**
+ * Check if value is of shape { startsWith: string }
+ */
+export function isStartsWith(blob: unknown): blob is { startsWith: string } {
```
marcbouchenoire, line 16: Tiny nitpick but maybe we could name it `isStartsWithOperator` or similarly? Or do you want to keep it not specific to operators? At a glance it's pretty similar to `startsWith()`.

marcbouchenoire (approved):
✨

packages/liveblocks-core/src/lib/guards.ts
nvie: Agreed, that's actually a better name. Will rename before merging! 👍

PR #2163, opened 2025-01-09
Support filtering thread metadata by absence

nvie:
Filtering threads by metadata using `useThreads({ query })` (or `useUserThreads_experimental({ query })`) now supports explicit filtering for metadata absence using `null`.

For example, you can now use `{ query: { metadata: { color: null } } }` to filter threads that do not have a `color` attribute in their metadata.

> [!NOTE]  
> Although officially never supported before, due to a bug in previous implementations, filtering for metadata absence was sort of possible by using explicit-`undefined`. This is no longer supported. If you relied on this behavior, please ensure you change `undefined` to `null` (which will actually also filter the threads in the backend properly).

```tsx
// Filter pinned threads that don't have a `color` set
useThreads({ query: { metadata: { pinned: true, color: null } } });

// These two queries are now equivalent
useThreads({ query: { metadata: { pinned: true } } });
useThreads({ query: { metadata: { pinned: true, color: undefined } } });
```

### Checklist

- [x] Deploy [this backend change](https://github.com/liveblocks/liveblocks-backend/pull/693) first
- [x] Write upgrade guide (included in #2167)

Fixes LB-1333

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,30 @@
 ## vNEXT (Not yet published)
 
+### `@liveblocks/react`
+
+Filtering threads by metadata using `useThreads({ query })` (or
+`useUserThreads_experimental({ query })`) now supports explicit filtering for
+metadata absence using `null`.
+
+For example, you can now use `{ query: { metadata: { color: null } } }` to
+filter threads that do not have a `color` attribute in their metadata.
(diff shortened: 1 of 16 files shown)


SugarDarius (approved):
Looking good from what I understand ✅

PR #2165, opened 2025-01-10
Fix room threads query before room creation

SugarDarius:
This PR is fixing the fetch to room threads when a new room is created.

Screenshot 2025-01-10 at 11 31 20

The fix is made to first get the room and then wait for the status to equals to `connected` before making the fetch to the room's threads when the room is freshly created.

Before 👇🏻 

https://github.com/user-attachments/assets/c61af510-8173-4eb7-8ae4-0ab64884d366

After 👇🏻 

https://github.com/user-attachments/assets/1f6bc80f-78ac-4ea2-bf3d-e5d63de91f22

It may not be the best way as I may have missed something. However I'm pushing this PR to get your thoughts on this. 

Also it may seem overkill at first read because the delta query works right after but on slow networks it may not.

Thanks 🙏🏻


nvie:
This is a problem we talked about before. Relying on the `connected` status is a bit brittle unfortunately. There really is no good event here that would be super robust. What matters most to me is that if a room does not exist, that should not be an error state: simply returning an empty list from the backend would be fine. We talked about this [here on Slack](https://liveblocks.slack.com/archives/C038S22AG3S/p1730749979038379?thread_ts=1730722293.839109&cid=C038S22AG3S) in the past. I don't think we actually created the fix, but I would be a proponent of simply changing the backend's response from a 404 to a 200 with an empty thread list, which would make this problem go away in the client (including all existing clients that haven't updated yet).

SugarDarius:
Noted 🙂 
Seems fair to return an empty list of threads with a 200 http status 👌🏻 

Thanks 🙏🏻 
Closing this PR then 🙂

PR #2167, opened 2025-01-10
Add generic error listener for client to respond to unhandled API errors

nvie:
This PR now makes errors that happen as a result of using Comments & Notifications show up in `useErrorListener`.

In addition, this hook previously required a RoomProvider, but no longer requires this room-specific context: it will now see errors from all rooms.

```ts
// ❌ Before: this would require a RoomProvider and would notify about errors for that room only
// ✅ Now: this requires a LiveblocksProvider and will notify about errors for any room
useErrorListener((error: LiveblocksError) => {
  /* show toast, or notify Sentry, Datadog, etc */
  const { type, roomId } = error.context;
});
```

The following API remains backward-compatible (it only receives errors for that room), but may now also receive new error types for Comments & Notifications as well:

```ts
// 👌 Same as before, but might now receive more errors
const unsubscribe = room.subscribe("error", (err: LiveblocksError) => {
  const { type, roomId } = error.context;
});
```

Also, this new lower-level API now exists on the client, for projects not using React.

```ts
// ✨ This is a new API
const unsubscribe = client.events.error.subscribe((err: LiveblocksError) => {
  const { type, roomId } = error.context;
});
```

### Checklist

- [ ] Get customer feedback
- [x] Write migration guide
- [x] ~Add back `useRoomErrorListener` as a backward-compatible hook (will help the migration guide)~
- [x] Undo addition of `useRoomErrorListener` after all and recommend code snippet in migration guide
- [x] Make accessing all errors type-safe using error codes
- [x] Update example in docs with full list of error cases
- [ ] Add more unit test coverage for these errors
- [x] Should we console log any errors if there aren't any subscribers to `client.events.error`? Seems a nice idea.

Fixes LB-1550


docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -1332,9 +1332,20 @@ contexts.
 <Header />
 ```
 
-### useErrorListener [@badge=RoomProvider]
+### useErrorListener [@badge=Both]
 
-Listen to potential room connection errors.
+<Banner>
+
+This behavior of this hook changed in 2.16. Before, this hook needed to be
```
marcbouchenoire, line 1339:
```suggestion
The behavior of this hook changed in 2.16. Before, this hook needed to be
```

packages/liveblocks-core/src/lib/EventSource.ts
```diff
@@ -23,9 +23,11 @@ export type Observable<T> = {
 
 export type EventSource<T> = Observable<T> & {
   /**
-   * Notify all subscribers about the event.
+   * Notify all subscribers about the event. Will return `false` if there
+   * weren't any subscribers at the time the .notify() was called, or `true` if
+   * there was at least one subscriber.
```
marcbouchenoire, line 28: 👌

packages/liveblocks-react-lexical/src/comments/comment-plugin-provider.tsx
```diff
@@ -55,6 +52,15 @@ export const ThreadToNodesContext = createContext<ThreadToNodesMap | null>(
...
 );
 
+function useCommentsErrorListener(callback: (error: LiveblocksError) => void) {
+  const client = useClient();
+  const savedCallback = useLatest(callback);
+
+  useEffect(() => {
+    return client.events.error.subscribe(savedCallback.current);
+  }, [client, savedCallback]);
+}
```
marcbouchenoire, line 62: Isn't this the same as the new `useErrorListener`?

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -1332,9 +1332,20 @@ contexts.
...
+nested under a RoomProvider, and would only notify you about connection errors
+for that room. Since 2.16, this hook will notify you about any Liveblocks error
+(not just connection errors), and for any room.
+
+</Banner>
+
+Listen to potential Liveblocks errors. Examples of errors include room
+connection errors, or errors creating or deleting threads, editing thread
```
marcbouchenoire, line 1347:
```suggestion
connection errors, errors creating or deleting threads, editing thread
```

packages/liveblocks-react/src/types/index.ts
```diff
@@ -288,6 +286,17 @@ export type SharedContextBundle<U extends BaseUserMeta> = {
      */
     useIsInsideRoom(): boolean;
 
+    /**
+     * useErrorListener is a React hook that allows you to respond to potential room
+     * connection errors.
```
marcbouchenoire, line 291: These should be updated like in `liveblocks.tsx`:
```
 * useErrorListener is a React hook that allows you to respond to any
 * Liveblocks error, for example room connection errors, errors
 * creating/editing/deleting threads, etc.
```

marcbouchenoire (changes requested):
I added some comments mostly about the docs/comments, the changes themselves look good to me 👍

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -1332,9 +1332,20 @@ contexts.
...
+metadata, etc.
 
 ```ts
 import { useErrorListener } from "@liveblocks/react/suspense";
```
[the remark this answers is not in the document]
nvie: @CTNicholas I've updated the example here: https://github.com/liveblocks/liveblocks/blob/cf3c3b15bc887a4ab658dc713ea4ab72206d6c8d/docs/pages/api-reference/liveblocks-react.mdx#useerrorlistener-badgeboth. Still not super happy with this. Would you mind helping me iterate on some tweaks to make this a bit more friendly?

packages/liveblocks-react-lexical/src/comments/comment-plugin-provider.tsx
nvie: Thanks for catching this. I DRYed this up with the normal listener in 5ede3f82761433277ef7124b10abf428949b8818 👍

nvie: Actually… I think with https://github.com/liveblocks/liveblocks/pull/2167#issuecomment-2589615909 now in place, this will actually trigger this DX warning when "just" using our Lexical integration now 🤔 Also it will silence [these dev warnings](https://github.com/liveblocks/liveblocks/pull/2167/commits/e0a0a9396fd6396ee46bfce94a134ad54140e83f) by default now when you use our Lexical package 🤔

nvie: Going to revert [this change](https://github.com/liveblocks/liveblocks/pull/2167#issuecomment-2589615909) to avoid this issue.

nvie: Done in ef3f1e7c89f56fda06e8974dc0aaa0f846cab547.

CTNicholas:
Good changes! Do you think you we need `useRoomErrorListener`? I feel like we're only adding it but backwards compatibility. We could just recommend this in the upgrade guide:

```tsx
function App() {
  const room = useRoom();
  
  useErrorListener((error) => {
    if (error.roomId !== room.id) {
      return false;
    }
    
    // ...
  });
}
```

nvie:
> Good changes! Do you think you we need `useRoomErrorListener`? I feel like we're only adding it for backwards compatibility. We could just recommend this in the upgrade guide:

I think you're right, this is a great call. Will remove it and recommend this instead 👍

PR #2171, opened 2025-01-13
Add new Relax<T> to help define discriminated union types

nvie:
Relaxes a discriminated union type definition, by explicitly adding properties defined in any other member as `never`. We used to do this manually, but with this helper, TypeScript will automatically do it.

This makes accessing the members much more relaxed in TypeScript.

For example:

```ts
type MyUnion = Relax<
  | { foo: string }
  | { foo: number; bar: string }
  | { qux: boolean }
>;
```

Normally, these will all be type errors: 🤔

```ts
const union: MyUnion = ...;
union.foo; // Error: Property 'foo' does not exist on type 'MyUnion'
union.bar; // Error: Property 'bar' does not exist on type 'MyUnion'
union.qux; // Error: Property 'qux' does not exist on type 'MyUnion'
```

With `Relax<T>`, accessing fields is much easier: 😇

```ts
const union: MyUnion = ...;
union.foo; // string | number | undefined
union.bar; // string | undefined
union.qux; // boolean
union.whatever; // Error: Property 'whatever' does not exist on type 'MyUnion'
```

### DIFF
--- a/packages/liveblocks-core/src/auth-manager.ts
+++ b/packages/liveblocks-core/src/auth-manager.ts
@@ -1,6 +1,7 @@
 import { StopRetrying } from "./connection";
 import { isPlainObject } from "./lib/guards";
 import type { Json } from "./lib/Json";
+import type { Relax } from "./lib/Relax";
 import type {
   Authentication,
   CustomAuthenticationResult,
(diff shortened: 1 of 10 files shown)


marcbouchenoire (approved):
Lovely ❤️ That was probably my biggest (only?) gripe with working with discriminated unions 🙂

PR #2173, opened 2025-01-13
Examples: Text editor toolbars

CTNicholas:
This updates most examples to use the floating toolbar. I'll upgrade the Notion AI example after we have stage 1 of AI Copilots. The Novel editor has its own toolbar, so I'm leaving that.


examples/nextjs-linear-like-issue-tracker/src/components/Editor.tsx
```diff
@@ -76,17 +76,20 @@ export function Editor({
...
+                    <ContentEditable className="outline-none editor-styles" />
+                  }
+                  placeholder={
+                    <div className="absolute top-0 left-0 pointer-events-none text-neutral-500 whitespace-nowrap">
+                      Start typing here…
+                    </div>
+                  }
+                  ErrorBoundary={LexicalErrorBoundary}
+                />
+                <EditorFloatingToolbar />
```
marcbouchenoire, line 91: Since the issue is fixed in the latest version, we can move this back outside of the conditional.

marcbouchenoire (approved):
Should we keep both toolbars in `nextjs-lexical` and `nextjs-tiptap`? I guess we could since they're more of a base/kitchen sink than realistic examples.

CTNicholas:
> Should we keep both toolbars in `nextjs-lexical` and `nextjs-tiptap`? I guess we could since they're more of a base/kitchen sink than realistic examples.

In Liveblocks examples, ignoring Novel, I think we should only show our toolbar from now, and we should try to make it work for every use case. In Yjs examples, happy to use custom toolbars.

marcbouchenoire:
> In Liveblocks examples, ignoring Novel, I think we should only show our toolbar from now, and we should try to make it work for every use case. In Yjs examples, happy to use custom toolbars.

Oh I meant that those two examples have both a **static** _and_ a **floating** toolbar.

PR #2174, opened 2025-01-14
remove pending selection, as its not needed

jrowny:
I incorrectly assumed that bluring (unfocusing) the editor removed its selection state and stored that in pendingCommentSelection. That is unnecessary complication because the selection state is properly stored.

this PR simplifies the approach by only storing a boolean, rather than a selection. editor.state.selection is used where pendingCommentSelection was used previously.

this also fixes (most) of the issues with commenting while a collaborator is editing. Thanks @yousefED for the tip!


nvie (approved):
Approving now, but I'd defer to Nimesh if you want a proper review of this, as I'm unfamiliar with Tiptap internals.

nimeshnayaju (approved):

PR #2176, opened 2025-01-17
Poll for Comments & Notifications when window regains focus

nvie:
This PR makes the poller trigger a poll whenever a `focus` event happens on the browser window (and the data is stale). We already triggered a poll when the window became _visible_, but this is slightly different. It also refreshes data if, for example, two windows are open side by side, and you change focus to one of those windows.


CTNicholas (approved):
Thank you!

marcbouchenoire (approved):

PR #2186, opened 2025-01-20
Display toolbar buttons names by default

marcbouchenoire:
Tiny improvement to make the case of `<Toolbar.Button name="foo" />` behave more like expectations:

```
// 2.16.0 
<Toolbar.Button name="Bold" /> // An empty (broken) button

// With this PR
<Toolbar.Button name="Bold" /> // A text-only button with a "Bold" label
```

For additional context, `name` is required (used in the button’s mandatory tooltip and its `aria-label` value), `icon` and `children` aren’t.

Other cases (with/without an `icon`, with custom `children`, etc) aren't affected by this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## vNEXT (not yet published)
 
+## v2.16.1
+
+### `@liveblocks/react-lexical` and `@liveblocks/react-tiptap`
+
+- `<Toolbar.Button />` and `<Toolbar.Toggle />` now display their `name`
+  visually if `children` and `icon` aren’t set.
+
 ## v2.16.0
 
 Our error listener APIs will now receive more errors in general, including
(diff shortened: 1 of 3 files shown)


SugarDarius (approved):
✅ 
Just for my own personal knowledge is it expected that we have the same component not mutualized between `@liveblocks/react-lexical` and `@liveblocks/react-tiptap` and get a single source of truth?

CTNicholas (approved):
Thanks for making the change!

marcbouchenoire:
> Just for my own personal knowledge is it expected that we have the same component not mutualized between @liveblocks/react-lexical and @liveblocks/react-tiptap and get a single source of truth?

@SugarDarius A lot is shared but not everything, some parts are editor-specific, and because I know that we have this duplication problem in other pieces of those 2 packages I thought it would be easier to not do anything for now until a global "share everything" PR to fix it in one go. (soon™)

PR #2191, opened 2025-01-22
Remove unused example dependencies

marcbouchenoire:
This PR removes unused dependencies in text editor examples and updates the Resend examples to match recent changes.


CTNicholas (approved):
Ahhh, thanks for doing this!

PR #2193, opened 2025-01-23
Use base url from env vars

SugarDarius:
# 🚧🚧🚧 Draft 🚧🚧🚧

A simple util function to read env vars to set base url in clients (browser & node).
This util will do direct reads (to keep static replacement) for env vars from `process.env` and `import.meta.env` and will use what it finds to set the base url. 

If a base url is set in the code (hard coded) then it always this value used. Otherwise we try to read the different env vars we seek (explicitly named to keep control).

Please note it's a first draft maybe we want something else more generic idk. 
Tested with:
- nextjs examples 
- vite examples

Thanks 🙏🏻


packages/liveblocks-core/src/lib/environment.ts
```diff
@@ -0,0 +1,33 @@
...
+        ? process.env.LIVEBLOCKS_BASE_URL
+        : undefined;
+    case "NEXT_PUBLIC_LIVEBLOCKS_BASE_URL":
+      return typeof process !== "undefined"
+        ? process.env.NEXT_PUBLIC_LIVEBLOCKS_BASE_URL
+        : undefined;
+    case "VITE_LIVEBLOCKS_BASE_URL":
+      return typeof import.meta.env !== "undefined"
+        ? import.meta.env.VITE_LIVEBLOCKS_BASE_URL
+        : undefined;
```
nvie, line 29: Unfortunately, although it looks like `import.meta` is accessing the property `meta` on a global "object" `import`, this is not the case :(

It's a special syntactical construct that has to be understood by the JavaScript runtime/parser. This construct is only defined if your project uses `type: "module"`, which I reckon most Vite projects will, but we cannot assume this is the case for _all_ of our customers projects using Liveblocks.

What this means is that our file `dist/index.js` cannot be parsed/understood by Node projects that are not using `type: "module"` yet.

You can see this happen for yourself if you:

- Check out this branch
- Go into the `liveblocks-core` directory
- Run `turbo run build`
- Then try to "run" the output: `node dist/index.js` (it will break 😞)  
  (but note that running `node dist/index.mjs` won't break!)

Also the `process.env` runtime variable isn't a normal object.

We can talk more in-depth on a call if you'd like.

SugarDarius: I just did a test by using a stringified evalued function (aka `new Function("return import.meta.env") however now even it not breaks anymore the parsing / interpretation we're going back to the fact they are not anymore statically replaced :/

It looks like a dead end besides creating instead of our plugin to be used only in dev to a direct injection :/

SugarDarius: Following your suggestions @nvie it looks like we have something on the right direction 🙂 
I add to make some tweaks for testing and I centralized the source of truth for `getBaseUrl`. 

Thank you so much 🙏🏻 
Feel free to make edits if you need to. I gave enough credits to this PR for now 😅 

The description of PR is updated. Tested with local examples ✅

nvie: @SugarDarius I did another pass at this. I found another approach that would avoid the need for module mocking. (I personally dislike mocking modules in Jest a lot, as this has cost me days of my professional life debugging issues when I wasn't aware stuff was being mocked.)

I think BASE_URL wise, this PR is good to go now! However… before we merge this, I would like to nail a couple more things DX wise. Reading only the `BASE_URL`, but _not_ supporting other env vars feels broken/confusing to me.

I tried implementing this more broadly, but started to realize I was going into a rabbit hole that was a bit deeper than I initially expected. So I need to finish this another time.

But ideally, on the server:

```ts
// Should automatically use BASE_URL and SECRET_KEY from .env file
// Should be fine if BASE_URL is missing, but should error if SECRET_KEY is not present
const liveblocks = new Liveblocks();

// Should still work, and take precedence over env vars
const liveblocks = new Liveblocks({ baseUrl, secret });
```

Ditto for the client:

```ts
// Should be fine if BASE_URL is missing, but should only read PUBLIC_KEY if `authEndpoint` isn't set
// If `authEndpoint` isn't set _and_ PUBLIC_KEY is missing, it should error
createClient({ });

// Fine if BASE_URL is missing, but should ignore NEXT_LIVEBLOCKS_PUBLIC_KEY if present
createClient({ authEndpoint });
createClient({ baseUrl, publicApiKey });  // Should still work and take precedence over envs
createClient({ baseUrl, authEndpoint });  // Should still work and take precedence over envs
```

(Same exact behavior for `<LiveblocksProvider>`.)

Differences are subtle, but if anything fails, we should fail with excellent error messages that make sense in the context the user is in. Just to say I'd like to do a holistic pass over this when we decide to add this!

packages/liveblocks-core/src/env.ts
```diff
@@ -0,0 +1,60 @@
...
+ * we avoid using `import.meta` in any source code and instead use
+ * a placeholder (`__IMPORT_META__`), which we replace as follows:
+ * - ESM output: `import.meta`
+ * - CJS output: `null` (causing runtime error instead of parse error)
+ */
+
+// Just so TypeScript doesn't trip up on `__IMPORT_META__`
+declare let __IMPORT_META__: { env: Record<string, string | undefined> };
+
+const _ = <T>(fn: () => T): T | undefined => {
```
SugarDarius, line 31: Would you be kindly all right to keep a real name for the try-catch wrapper @nvie ? Usually we use the character `_` for unused variables in code bases. So I'd prefer to have a name like it was before to better understand the code intent and what it means 😅  I agree `safeGet` wasn't maybe the best but at leasts it's meaningful when reading it event if it's only used in this file.

packages/liveblocks-core/src/lib/environment.ts
SugarDarius: Nicely done using the `__IMPORT_META__` global constant 👨🏻‍🍳 👍🏻  🙂

packages/liveblocks-core/src/env.ts
SugarDarius: Ping @nvie ^^

marcbouchenoire (approved):
I haven't followed too closely the various experimentations but the result looks great!

packages/liveblocks-core/src/env.ts
nvie: Done in 2e0a59a452a60c0cb03836f22a4c04424199c614.

nvie (changes requested):
I think `BASE_URL`-wise, this PR is looking good! However… before we merge this, I would like to nail a couple more things DX wise. Reading only the `BASE_URL`, but not supporting other env vars feels broken/confusing to me.

I tried implementing this more broadly last week, but started to realize I was going into a rabbit hole that was a bit deeper than I initially expected.

But ideally, on the server:

```ts
// Should automatically use BASE_URL and SECRET_KEY from .env file
// Should be fine if BASE_URL is missing, but should error if SECRET_KEY is not present
const liveblocks = new Liveblocks();
```
```ts
// Should still work, and take precedence over env vars
const liveblocks = new Liveblocks({ baseUrl, secret });
```

Ditto for the client:

```ts
// Should be fine if BASE_URL is missing, but should only read PUBLIC_KEY if `authEndpoint` isn't set
// If `authEndpoint` isn't set _and_ PUBLIC_KEY is missing, it should error
createClient({ });
```
```ts
// Fine if BASE_URL is missing, but should ignore NEXT_LIVEBLOCKS_PUBLIC_KEY if present
createClient({ authEndpoint });
createClient({ baseUrl, publicApiKey });  // Should still work and take precedence over envs
createClient({ baseUrl, authEndpoint });  // Should still work and take precedence over envs
```

(And same exact behavior for `<LiveblocksProvider>`.)

Differences are subtle, but if anything fails, we should fail with excellent error messages that make sense in the context the user is in.

Just to say I'd like to do a holistic pass over this before we merge this PR.

PR #2197, opened 2025-01-24
Update schema info pages

CTNicholas:


nvie (approved):
Thanks for updating! This was indeed very outdated info.

PR #2199, opened 2025-01-24
spread set to support devtools

hans-lizihan:
### Fixing a bug

- Liveblock zustand assumes `set` in zustand to only have 1 argument. but actually it takes 3. This is breaking integration with other middlewares like devtools.

#### How to test it
may need to update example with liveblock zustand to include devtools


nvie:
Thanks for the contribution, @hans-lizihan! Closing this PR in favor of #2199, so we can add some tiny additions before merging.

PR #2203, opened 2025-01-27
Ensure not server side for every suspense hooks

SugarDarius:
~~This PR fixes an issue in the umbrella store to avoid to fetch on the server on first render.~~

~~This bug has been discovered during an example implementation when using suspense hooks where the error `To use Liveblocks client in a non-DOM environment with a url as auth endpoint, you need to provide a fetch polyfill.` was hit.~~

~~The fix is to return a resolved promise in `waitUntilLoaded` method of `PaginatedResource` and `SinglePageResource` where the `window` object is undefined. It will avoid to make any fetch on the server on first render when use with suspense hooks.~~

After re-thinking and finding a deep more the root cause. It was appearing it's in fact a matter of DX where some suspense hooks wasn't clearly explaining that they cannot be used without being wrapped in a `<ClientSideSuspense />`. 

The fix now ensures that every suspense hook aren't run on the server and update the error message.

Thanks 🙏🏻

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### `@liveblocks/react`
 
-- Fix bug in internal store to avoid to fetch on the server on first render.
+- Improve error message if hooks are accidentally called server side
 
 ### `@liveblocks/zustand`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## vNEXT (not yet published)
 
+### `@liveblocks/react`
+
+- Fix bug in internal store to avoid to fetch on the server on first render.
+
+### `@liveblocks/zustand`
+
 - Fix bug in Zustand typing in case the multi-argument form of `set()` is used
   (thanks [@hans-lizihan](https://github.com/hans-lizihan))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## vNEXT (not yet published)
 
+## v2.16.2
+
 ### `@liveblocks/react`
 
 - Improve error message if hooks are accidentally called server side
--- a/packages/liveblocks-react/src/lib/ssr.ts
+++ b/packages/liveblocks-react/src/lib/ssr.ts
@@ -1,0 +1,8 @@
+export function ensureNotServerSide(): void {
+  // Error early if suspense is used in a server-side context
+  if (typeof window === "undefined") {
+    throw new Error(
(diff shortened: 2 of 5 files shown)


nvie (approved):
@nimeshnayaju for context, @SugarDarius and I spotted this one after debugging an unexpected server-side initiated auth attempt that should not be happening, triggered by the use of `useInboxNotifications()`, ultimately caused by [this line](https://github.com/liveblocks/liveblocks/blob/7df5914e8e93c3ff038562edbd6b370c6bf481fd/packages/liveblocks-react/src/liveblocks.tsx#L415) that executes during a render — in this particular case triggered during SSR.

Of course, this line is only intended to ever be executed on the client, so I suggested to simply define the `.waitUntilLoaded()` promises as no-ops on the server.

@nimeshnayaju Do you agree with this approach, or do you maybe have another idea?

nimeshnayaju:
> @nimeshnayaju for context, @SugarDarius and I spotted this one after debugging an unexpected server-side initiated auth attempt that should not be happening, triggered by the use of `useInboxNotifications()`, ultimately caused by [this line](https://github.com/liveblocks/liveblocks/blob/7df5914e8e93c3ff038562edbd6b370c6bf481fd/packages/liveblocks-react/src/liveblocks.tsx#L415) that executes during a render — in this particular case triggered during SSR.
> 
> Of course, this line is only intended to ever be executed on the client, so I suggested to simply define the `.waitUntilLoaded()` promises as no-ops on the server.
> 
> @nimeshnayaju Do you agree with this approach, or do you maybe have another idea?

Thank you for explaining the issue. I can see how our current approach can cause this issue. I don't recall completely but I believe this used to not be an issue - we'd only fetch inside `useEffect` and not on render. The promise status would be updated based on whether the promise has resolved or rejected, etc, so, we didn't need to perform the `window === undefined` check, but I am also okay with the solution implemented. Did you happen to brainstorm other solutions too? Would love to know how you decided on this particular solution.

nimeshnayaju (approved):

nvie (changes requested):
Actually maybe there is a slightly better way to address this. I'll take a quick stab at it this morning when I'm behind a keyboard.

nvie:
@nimeshnayaju After having another look at this problem this morning, I realized that these hooks should really not be run on the server at all in the first place. Rather than working around it, we instead now throw an error message (like we do in all Storage/Presence hooks too) when these Suspense hooks are called server-side, and point the user to the relevant documentation, so they can wrap their components in a ClientSideSuspense wrapper to avoid this issue.

That was the real problem, but it was really easy to make this mistake. So the fix here was to just improve the DX.

PR #2212, opened 2025-02-04
Handle unmounted composers during submit

marcbouchenoire:
If a composer is unmounted during its `onComposerSubmit` callback, our own `onSubmitEnd` callback will still be triggered afterwards but it will crash because the composer's editor no longer exists and can't be cleared/blurred.


gmichopoulos (commented):
Confirmed that this solves the issue for me!

SugarDarius (commented):
It works as a fast and quick solution however I will tend to not use a `try/catch` block for doing the job. 
If we have underlaying errors coming from either `clear()`, `blur` or `clearAttachments` but the component isn't unmounted we won't be able to track them in the `try/catch` block.

I think it would be better to use an explicit state management with a reference such as:
```tsx
import { useCallback, useEffect, useRef } from 'react'

export function useIsMounted(): () => boolean {
  const isMounted = useRef(false)

  useEffect(() => {
    isMounted.current = true

    return () => {
      isMounted.current = false
    }
  }, [])

  return useCallback(() => isMounted.current, [])
}
```
to keep track of the mounted state of component. 
Then you could call `onSubmitEnd`'s intructions based on this state: 
```tsx
const onSubmitEnd = () => {
  if (!isMounted) return;
    try {
      clear();
      blur();
      clearAttachments();
    } catch (error) {
      // Log error if needed.
      console.error("Error during cleanup:", error);
    }
    setSubmitting(false);
}
```

It makes the code also more predictable as it will avoid to execute functions on unmounted components.

WDYT?

nvie (approved):
LGTM!

marcbouchenoire:
> It makes the code also more predictable as it will avoid to execute functions on unmounted components.

@SugarDarius I should clarify that the error isn't about React state and happens across multiple components so doing this might not work for all cases. The main problem is that `blur`/`clear` live in `<Composer.Form />` where we have the Slate instance and `<Composer.Editor />` renders the DOM node for that Slate instance, if `<Composer.Editor />` is unmounted, we could still have access to the Slate instance in `<Composer.Form />` where the callbacks are triggered but Slate calls like `blur` will crash because the editor's DOM node no longer exists.

One other fix that could be more explicit than a `try...catch` would be to check if the editor's DOM node exists before calling Slate methods, I'll try that.

marcbouchenoire:
@SugarDarius Checking for the DOM node synchronously or using refs wasn't reliable, so [I just moved the `try...catch` to be more explicit](https://github.com/liveblocks/liveblocks/pull/2212/commits/0d1624b77fb07a23ec46945a8d130ce864b590a5).

SugarDarius:
> One other fix that could be more explicit than a try...catch would be to check if the editor's DOM node exists before calling Slate methods, I'll try that.

Yes I agree 👍🏻

> or using refs wasn't reliable
Oh 😮 that's sad 🙁  so I suppose that even doing roughly this:
```tsx
// Composer Editor
const editableRef = useRef<HTMLDivElement>(null);
const { setEditableNode } = useComposerEditorContext()

useEffect(() => {
  return () => { setEditableNode(null) }
}, [])

return (
<Slate>
   <Editable ref={(node) => { setEditableNode(node) }} />
</Slate>
)

// Composer Form
const { editor, editableNode } = useComposerEditorContext()
const blur = () => {
  if (editableNode && ReactEditor.hasDOMNode(editor, editableNode) {
    // ...
  }
}
```
It won't work neither?

Thanks for making the the `try/catch` blocks more explicit 🙏🏻

SugarDarius (approved):

marcbouchenoire:
@SugarDarius I tried something similar yesterday (manually checking for the existence of `ref.current.querySelector("[data-slate-editor]")` from `Composer.Form`) and it had false positives.

PR #2221, opened 2025-02-07
Rename `OptionalPromise<T>` → `Awaitable<T>`

nvie:
Renames our internal helper type `OptionalPromise<T>` to `Awaitable<T>`. I wanted to add this type only to realize we already had this. I've seen the name `Awaitable` used in many other libraries too, it seems to be the more common name in the community for this type — also, see [this SO answer](https://stackoverflow.com/a/56129545/148872).

Also avoids reading `OptionalPromise<T>` as meaning `Promise<T> | undefined`.

### DIFF
--- a/packages/liveblocks-client/src/index.ts
+++ b/packages/liveblocks-client/src/index.ts
@@ -5,6 +5,7 @@ import { PKG_FORMAT, PKG_NAME, PKG_VERSION } from "./version";
 detectDupes(PKG_NAME, PKG_VERSION, PKG_FORMAT);
 
 export type {
+  Awaitable,
   BaseActivitiesData,
   BaseMetadata,
   BaseUserMeta,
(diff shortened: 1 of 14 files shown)


packages/liveblocks-client/src/index.ts
```diff
@@ -5,6 +5,7 @@ import { PKG_FORMAT, PKG_NAME, PKG_VERSION } from "./version";
 detectDupes(PKG_NAME, PKG_VERSION, PKG_FORMAT);
 
 export type {
+  Awaitable,
```
nvie, line 8: I've also decided to re-export this in `@liveblocks/client`, because this type is part of our public API already (in some return types).

marcbouchenoire (approved):
I was never a fan of this naming (I'm only blaming myself for that) 👍

SugarDarius (commented):
✅

PR #2224, opened 2025-02-07
Fix a small typo in the readme

andreaselia:
I updated the link within the repository as that's what it currently is, but maybe it would be better linking to the website?

https://liveblocks.io/docs/platform/sync-datastore


CTNicholas:
Ahhh, thanks for highlighting this @andreaselia! I forgot to add `https://liveblocks.io` to a bunch of links, fixing it over here: 
- https://github.com/liveblocks/liveblocks/pull/2225

PR #2228, opened 2025-02-11
Fix tiptap<->room sync issue

jrowny:
We use a singleton map to store yjs providers across re-renders because we should only run a yjs sync when the room changes, but we don't have yjs providers at the `<RoomProvider>` level. 

The problem in this case is that yjsProvider map was referencing a room instance that no longer exists. The Lexical extension tries to solve this using a ref (but I'm not sure if that works, I should investigate)

This PR takes adds a new factory method, willDestroy notification on room, and fixes a few types.

- getYjsProviderForRoom() returns a LiveblocksYjsProvider that will be destroyed when the room is destroyed. Providers are stored in a weakMap until the room is destroyed.
- Tiptap now uses that function (lexical should too)
- Changes one example (the rest need to be verified/changed in the future)
- Updates documentation for the new factory function


packages/liveblocks-react-tiptap/src/LiveblocksExtension.ts
```diff
@@ -332,20 +332,30 @@ export const useLiveblocksExtension = (
       ];
     },
     addStorage() {
-      if (!providersMap.has(room.id)) {
+      /* NOTE:
+      the purpose of the providersMap is to keep the same provider alive across renders
+      re-instantiating the provider when the room hasn't changed can cause the yjs doc state to be out of sync
+      with liveblocks yjs state. In this instance, we can just check if the room has changed, because the instance
+      from useRoom will be shallowly equal to the previous instance.
```
nvie, line 339:
```suggestion
      from useRoom will be referentially equal to the previous instance.
```

packages/liveblocks-yjs/src/index.ts
```diff
@@ -43,7 +43,7 @@ export class LiveblocksYjsProvider<
   extends Observable<unknown>
   implements IYjsProvider
 {
-  private room: Room<P, S, U, E, M>;
+  public room: Room<P, S, U, E, M>;
```
nvie, line 46: Not sure if it's really needed to make this a public field. My understanding is that `room` is only ever assigned in the constructor, so effectively this could remain a private property and can even be marked as a `private readonly` property to make that more obvious that its value won't change over the lifetime of the YjsProvider (and we can do the same for the other readonly properties on this class, like `rootDoc` etc).

This matters, because… 1/2

packages/liveblocks-react-tiptap/src/LiveblocksExtension.ts
```diff
@@ -332,20 +332,30 @@ export const useLiveblocksExtension = (
...
+      */
+      let provider = providersMap.get(room.id);
+      if (provider && provider.room !== room) {
+        // The room has changed, destroy the old provider, so a new one can be made
+        provider.destroy();
+        providersMap.delete(room.id);
+      }
```
nvie, line 346: 2/2

…it means this solution can likely be implemented in a simpler way, by not keying these in a Map from room _ID_ to providers, but instead by using a WeakMap of Room _instances_ to providers.

This has been our go-to trick to extend stuff on the Room in other packages, without the Room having to know about it. And if the Room instance goes out of existance, the weakmap will also garbage-collect these provider instances.

nvie (commented):
Here is an example of my suggested approach: 310bfc6256d4a64c00bea49cdfae056ff7d952b9.

I did not push it directly to your branch. Could you test if this also solves the issue, because this approach has worked like a charm for us everywhere else.

jrowny:
I think that without the destroy getting called, the old rooms will always
have a reference from the listeners in the provider and therefore will
never be removed from the weakmap. I can try that approach tomorrow and see
if I can make it work.

nvie:
> I think that without the destroy getting called, the old rooms will always have a reference from the listeners in the provider and therefore will never be removed from the weakmap. I can try that approach tomorrow and see if I can make it work.

Ah yes, correct. The Room however is explicitly destroyed by the Client, so we could emit an event that you can subscribe to to properly destroy the provider as well when this happens. That way, it should work perfectly.

nvie:
I've given this a shot in the last two commits I added.

1. Uses a WeakMap, associating the YjsProvider instance to the Room instance, not the room ID
2. Destroys the YjsProvider instance when the Room instance gets destroyed by the client explicitly (this happens when you unmount the last instance of the Room), unregistering any remaining subscribers
3. The WeakMap will take care of the garbage collection

packages/liveblocks-core/src/room.ts
```diff
@@ -3088,11 +3097,16 @@ export function createRoom<
       reconnect: () => managedSocket.reconnect(),
       disconnect: () => managedSocket.disconnect(),
       destroy: () => {
-        syncSourceForStorage.destroy();
```
jrowny, line 3091: is syncSourceForStorage no longer needed here?

nvie: Huh? Oops! Thanks for catching that, definitely unintended!

CTNicholas (commented):
Nice! Should we add a `useYjsProvider` hook to `@liveblocks/react`? Then `useRoom` and the argument won't be necessary.

jrowny:
> Nice! Should we add a `useYjsProvider` hook to `@liveblocks/react`? Then `useRoom` and the argument won't be necessary.

I think we want to keep the yjs dependency separate, which is why we don't include it there.

docs/pages/api-reference/liveblocks-yjs.mdx
```diff
@@ -408,3 +405,18 @@ awareness.emit("destroy", true);
 [`Awareness.off("destroyed")`]: #Awareness.off.destroyed
 [`Awareness.once("destroyed")`]: #Awareness.once.destroyed
 [`Awareness.emit("destroyed")`]: #Awareness.emit.destroyed
+
+## getYjsProviderForRoom [#getYjsProviderForRoom]
+
+Factory function to get a LiveblocksYjsProvider for a room. This provider will
```
nvie, line 411:
```suggestion
Gets the current or creates a new LiveblocksYjsProvider for a room. This provider will
```

nvie (approved):
Nice work, I like this final solution a lot, it's a lot simpler to understand and reason about now 👍

packages/liveblocks-core/src/room.ts
```diff
@@ -3088,11 +3097,17 @@ export function createRoom<
...
         context.yjsProvider?.off("status", yjsStatusDidChange);
+        syncSourceForStorage.destroy();
         syncSourceForYjs.destroy();
         uninstallBgTabSpy();
         managedSocket.destroy();
+
+        // Unregister all registered callbacks
+        for (const source of Object.values(eventHub)) {
+          source[Symbol.dispose]();
+        }
```
nvie, line 3110: I added this because I realized there was a cleanup bug here, but feel free to remove it from this PR, if you think this is causing the awareness error. I can try to tackle that in a separate PR instead of blocking this PR with it. Whatever you prefer!

nvie (approved):
Nice, I like this fix better! 👌

PR #2231, opened 2025-02-12
Fix `Thread` overrides

marcbouchenoire:
**Context:** Overrides can be set globally via [`LiveblocksUIConfig`](https://liveblocks.io/docs/api-reference/liveblocks-react-ui#LiveblocksUIConfig) or via `overrides` props on components directly.

This PR fixes how the `Thread` component (which is made of `Comment` and `Composer` components) wasn't correctly passing down its overrides internally.


SugarDarius (approved):

CTNicholas (approved):
Nice!

PR #2232, opened 2025-02-12
Fix quadratic buildup of operations

rudi-c:
The way operations are accumulated into the `activeBatch.reverseOps` unfortunately has quadratic behavior, as can be seen in the profile below (somewhat of an extreme test, where we load 4 square miles of Mapbox building data).

Screenshot 2025-02-11 at 12 56 53 PM

This is because calling unshift repeatedly is O(n^2).

Testing:

I ran the unit tests. I'm not I fully trust them though. For example, a PR where the only diff is to replace `activeBatch.reverseOps.unshift(...reverse);` with `activeBatch.reverseOps.unshift(...Array.from(reverse));` passes the unit tests also. That being said, more obvious mistakes like failing to reverse the input to `context.undoStack.push` do fail unit tests.


jrowny:
thanks for the PR, we'll take a look shortly!

nvie:
Hi @rudi-c, this change makes a lot of sense, and I can see how the current implementation can be unperformant in your use case. This reminded me of [some work I started last year](https://github.com/liveblocks/liveblocks/commit/ae9df12bc46c556a02feb2bf201eb1fdcea92094) but never finished, where I wanted to replace these arrays with a `Deque` datastructure to make the `.unshift()` efficient. (There are a couple more places in the code base that would benefit from this datastructure.)

We definitely want to merge this. But before we do, I would like to make absolutely sure we're not introducing a regression, so I'll have to verify that there is indeed adequate test coverage. I'll try to get to that next week.

rudi-c:
Yeah, either a reverse array or a deque data structure solves the problem -- this feels like a question of which one is clearer/more maintainable, and I'll let you make the call on that one :)

nvie (approved):
Hi @rudi-c, thanks again for pointing this out, and for your contribution. I pushed a couple of commits, where I cherry-picked my Deque implementation from last year, and used it instead of the arrays internally. I did run a few benchmarks, and turns out `deque.pushLeft(many)` is even faster than a `array.push(...many)`. Apparently the spread operator adds a ton of overhead on large arrays, because all those items need to be copied to the stack first before the function can be called.

Before I merge this, do you want to give this version a try and do some benchmarking yourself? I just want to make sure this alternative fix also solves your problem as expected.

To try it out, you can install this version:

```tsc
npx create-liveblocks-app@latest --upgrade v2.18.2-test1
```

Thanks! 🙏

rudi-c:
@nvie I upgraded our packages to try it in app, and the time to import data from my benchmark into LiveBlocks went down from >5s to <2s. Thanks a lot!

nvie:
Amazing — thanks for helping us improve this part!

PR #2233, opened 2025-02-13
Mention node

jrowny:
In preparation for blocknote support, we're separating the mention node so it can be used outside of the tiptap extension.


jrowny:
I'll double check tomorrow, but the schema should be the same as the attributes and type name are the same

nvie (approved):

PR #2234, opened 2025-02-13
Update Codeowners

SugarDarius:
I'm updating the codeowners file to better redirect in case of internal or external PRs. 
Please feel free to update this file as I'm made the edits based on the git blame. So some lines may be off.

Thanks!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,5 @@
 # Public packages
+/packages/	@nvie
 /packages/liveblocks-core/	@nvie @sugardarius @marcbouchenoire
 /packages/liveblocks-client/	@nvie @sugardarius @marcbouchenoire
 /packages/liveblocks-react/	@nvie @sugardarius @marcbouchenoire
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,7 +1,17 @@
 # Public packages
-/packages/   @nvie
-/packages/liveblocks-node/   @nvie @ofoucherot @FlowFlorent
-/packages/liveblocks-react-ui/   @marcbouchenoire
+/packages/liveblocks-core/	@nvie @sugardarius @marcbouchenoire
```
nvie, line 2: I'd prefer to keep the fallback in, but fine with the more specific overloads (later rules will override this one for specific packages):

```suggestion
/packages/   @nvie
/packages/liveblocks-core/	@nvie @sugardarius @marcbouchenoire
```

.github/CODEOWNERS
```diff
@@ -1,7 +1,17 @@
...
+/packages/liveblocks-client/	@nvie @sugardarius @marcbouchenoire
+/packages/liveblocks-react/	@nvie @sugardarius @marcbouchenoire
+/packages/liveblocks-react-ui/	@marcbouchenoire @jrowny
```
marcbouchenoire, line 5: I'm not sure @jrowny wants to be there, but maybe I'm wrong.

SugarDarius: I put @jrowny because of the git blame telling me he worked on the version history components 😅

PR #2252, opened 2025-02-20
Revert fix avoid duplicated email data

SugarDarius:
Hey 👋🏻 

After syncing with Florent the previous fix I made in #2218 is not valid and irrelevant. 
Basically it changes nothing at all for developers and end users and do not provides a sustainable way to avoid duplication of `emailData` between two webhook events because it's a 1:1 relation between a thread and an inbox notification. 

That's why it requires a new tech design thinking where would potentially introduce in the inbox notification a new key where we could store activities related to the `thread` notification kind. I will write a tech design about it.

Thanks 🙏🏻


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## vNEXT (not yet published)
 
+### `@liveblocks/emails`
+
+- Revert the deduplication logic in email notification preparation functions
```
marcbouchenoire, line 5: I don't think we need to repeat the exact names since this package doesn't contain other things, but we could mention the version of the last change 

```md
- Revert deduplication logic introduced in `v2.18.0` as it provided no measurable benefits while increasing complexity.
```

marcbouchenoire (approved):
Too bad we can't have this behavior yet, next time 🤞

nvie (approved):

PR #2253, opened 2025-02-20
Log input data when `JSON.stringify()` calls fail

nvie:
This PR:

- Renames the existing `stringify` → `stableStringify`
- Adds a new `stringifyOrLog` which is a drop-in replacement for `JSON.stringify()`, but which will report the input data to the console in case that stringify call fails. This should help us debug certain complex use cases.


marcbouchenoire (approved):
🕵️

PR #2261, opened 2025-02-20
Modernize TS config for all of our packages

nvie:
This is a long overdue PR:

- Switches all packages from `moduleResolution: "node"` (which is _ancient_!) to `moduleResolution: "node16"` (which is the recommended setting for all libraries)
  - All imports are rewritten accordingly, requiring a file extension now. For example: `import {} from "./foo"` → `import {} from "./foo.js"`
- Makes all packages `type: "module"` by default. This changes the default from CJS to ESM.
  - All files are renamed accordingly `*.js` → `*.cjs`
  - All files are renamed accordingly `*.mjs` → `*.js`
- Makes all React component builds use `_jsx()` in their output bundles, rather than `React.createElement()`
- Consolidate all common TS configs (DRYs them all up)


SugarDarius (commented):
It would be nice to have such modernize TS config, however based on the error you get from the `e2e/next-sandbox` tests I wonder what would be the impact on end developers and their projects? 

Does it means they will be forced to either rename `liveblocks.config.ts` → `liveblocks.config.mts` or to defined in their `package.json` the type `module`. 
So it became a kind of breaking change when developers are import types (e.g `import type { LiveList } from "@liveblocks/core";`) right?

nvie:
Hmm. I'll investigate. This wasn't supposed to be a breaking change. This should not have an impact on end users. Not sure what's up here, I'll dig in a bit.

nvie:
Found the issue: the e2e app's `tsconfig.json` was extending our base TS config template (which I changed). Because of it, it changed the moduleResolution for that project to `node16` instead of `node10`. I put this back explicitly in 320e3a524a4351d4181d44ce58d0814ad7a3f9de, and now it's working again. So indeed there is no breaking change with this.

That said, we should really upgrade to node16 for the e2e test app as well, which I can take care of (but that's just to make it more modern too, not because it's required now).

schema-lang/codemirror-language/src/syntax.d.grammar.ts
nvie, line 1: @marcbouchenoire I had to make a couple of tweaks to this project to make it work with the Lezer `syntax.grammar` import, but it seems to work now. Can you verify I didn't break anything I'm not aware of?

marcbouchenoire: It did break the tests, how did you make Jest happy with `import` in other repos?

marcbouchenoire (approved):
I'm happy about all changes except having to use `.js` extensions but if that's recommended then so be it, I thought this was recommended when TS was used directly as a bundler, and not something like `tsup` like in our case.

schema-lang/codemirror-language/src/syntax.d.grammar.ts
nvie: You mean locally? On CI the tests are passing, and they are [for me](https://gist.github.com/nvie/34a595fd1eeed1b526c4f3b777daea29) 😬

nvie: Turns out these tests were also not passing on `main`, and CI isn't running these.

nvie: So the import issue was already there on `main`, but this PR should fix them. Jest annoyingly assumes all files are CJS even still today, and even when `type: "module"` is set in the project. So the way to make Jest work with these ESM files is to put this hack in the global/shared Jest config: https://github.com/liveblocks/liveblocks/pull/2261/files#diff-771df2d411e1733071196271fe6618c6cca911be447e57d8c4f897153ffd35c0

nvie: We should really just leave Jest behind and switch to Vitest, which has much saner defaults and works with ESM out of the box.

SugarDarius:
I always thought using `node` or `node16` as `moduleResolution` or `module` was for back-end oriented and node targeted code libraries.

Using `bundler` here seems more legit because of:
- We're building packages meant to be used in bundled apps mostly (except the node packages).
- We're using `tsup` and `rollup`
- Final import resolution is bundlers' job when using our packages in bundled apps.

So it seems fine to switch to `esnext` / `bundler` 👍🏻

nvie:
I'll give it a try and let you know if there are any issues with it, but I do indeed think it's fine. In the recommended guidelines I think I may have gotten too focused on the "library" vs "application" part.

nvie:
It worked! 🎉 The net diff is much smaller now. Thanks for the feedback here @marcbouchenoire + @SugarDarius! 🙌

SugarDarius (approved):

PR #2265, opened 2025-02-21
Better DX for user notification settings

SugarDarius:
This PR updates the DX of user notification settings by using a creator factory to represents a similar of a `Proxy` like based on a private symbol `kPrivate`. 

### Context:
We're getting back from the backend an object for user notification settings where potentially some channels aren't existing because there are no notification kinds enabled on them at the project level. 
But we can't generate a specific type strictly equals to the truth of the project notification settings. So we can't guess which channels are enabled or not in advance. That's why we decided first to define the `UserNotificationSettings` as it is today where all channels are defined on it.

### Problematic
When developers use for example the util `isNotificationChannelEnabled(settings.slack)` it can throw an ugly error because the channel `webPush` may not be enabled so the key doesn't exists on the project but developers are trying to access

### Solution
To mitigate this problematic and offer a better DX I'm introducing a creator proxy factory around user notification settings 👉🏻 `createUserNotificationSettings()`. It's not a real `Proxy` built-in object but more a similar where it store a partial of a plain user notification settings (remote fetched from the back-end) under `[kPrivate].__plain__` and will define for each channels a getter. Each getter will check the plain value and returns `null` with an error log if a channel do not exists in the object get back from the back-end.

```ts
// Project Notification Settings
// email channel enabled ✅
// webPush channel disabled ❌

// Proxy creation
const settings = createUserNotificationSettings({
  email: {
    thread: true,
    textMention: true,
    $fileUploaded: true,
  },
})

console.log(settings.emails) // ✅ 👉🏻 { thread: true, textMention: true, $fileUploaded: true } 
console.log(settings.webPush) // ✅  👉🏻 `null` with an error log "In order to use the 'webPush' channel, please set up your project first. For more information https://liveblocks.io/docs/errors/enable-a-notification-channel"
```
👉🏻 It provides a better DX because developers are aware of they can't access the slack channel because they need to enable it on their project.

CleanShot 2025-02-25 at 15 23 22@2x

#### Pros:
- Better DX more explicit error and invites developers to enable the channel in their project
- Better typing to express data remotely fetched from the back-end
- Makes our code more explicit
- The plain value isn't enumerable
- Our notification settings are readable only
#### Cons:
- Makes the code of our customers a bit more verbose (see the example)
- When updating a user notification settings object internally it needs to be patched -- meaning we need to update the plain value hidden under `[kPrivate].__plain__`. I created an helper for it when updating from an optimistic update: `patchUserNotificationSettings()` -- it could be completed with other helpers if needed.

### 🏁 Checklist

- [x] Validate usage
- [x] Add doc url
- [x] Update documentation
- [x] Update changelog

Fixes: [LB-1757](https://linear.app/liveblocks/issue/LB-1757/improve-dx-when-using-public-packages)


CTNicholas (approved):
Great work! I'm not too familiar with our internals, but ✅ for behaviour, docs, examples.

packages/liveblocks-core/src/api-client.ts
```diff
@@ -1411,8 +1416,8 @@ export function createApiClient<M extends BaseMetadata>({
 
   async function updateUserNotificationSettings(
     settings: PartialUserNotificationSettings
-  ): Promise<UserNotificationSettings> {
-    return httpClient.post<UserNotificationSettings>(
+  ): Promise<UserNotificationSettingsPlain> {
```
nvie, line 1419: I noticed this public API is using the type `UserNotificationSettingsPlain` which is marked as `@internal`. Every type that is part of a public API (arguments and return type) should also be a public type ideally. Any reason for this switch?

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
```diff
@@ -47,19 +89,120 @@ type DeepPartialWithAugmentation<T> = T extends object
...
+       * In the TypeScript standard library definitions, the built-in interface for a property descriptor
+       * does not include a specialized type for the “this” context in the getter or setter functions.
+       * As a result, both the ⁠get and ⁠set methods implicitly have ⁠this: any.
+       * The reason is that property descriptors in JavaScript are used across various objects with
+       * no enforced shape for ⁠this. And so the standard library definitions have to remain as broad as possible
+       * to support any valid JavaScript usage (e.g `Object.defineProperty`).
+       *
+       * So we can safely tells that this getter is typed as `this: UserNotificationSettings` because we're
+       * creating a well known shaped object → `UserNotificationSettings`.
+       */
```
nvie, line 143: This comment wouldn't be needed IMO. Using the `this:` argument type here makes sense and is fine without further explanation.

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
```diff
@@ -26,11 +27,52 @@ export type NotificationChannelSettings = {
...
-  [C in NotificationChannel]: NotificationChannelSettings;
+  [C in NotificationChannel]: NotificationChannelSettings | null;
+} & {
+  /**
+   * @private
+   *
+   * `UserNotificationSettings` with private internal properties
+   * to store the plain settings and methods.
+   */
+  [kPrivate]: PrivateUserNotificationSettingsApi;
```
nvie, line 75: I'd prefer to keep this outside of the public `UserNotificationSettings` type. TypeScript-wise, these should be implementation details invisible to the user. We can, however, define a new internal type for this, eg `ProxiedUserNotificationSettings` which is truly private (not exported nor used in any public APIs), and then use that to `as`-cast the places where TypeScript will fail if you remove it here.

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
```diff
@@ -26,11 +27,52 @@ export type NotificationChannelSettings = {
...
+const kPrivate = Symbol("user-notification-settings-private");
+
+/**
+ * @private
+ *
+ * Private properties and methods internal to `UserNotificationSettings`.
+ * As a user of Liveblocks, you should NEVER USE ANY OF THESE DIRECTLY,
+ * because bad things will happen.
+ */
+export type PrivateUserNotificationSettingsApi = {
```
nvie, line 58: This type doesn't seem to be used. Can you do a pass over all exported types that aren't used outside this module? Makes it much easier to verify we don't accidentally export private impl details.

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
```diff
@@ -26,11 +27,52 @@ export type NotificationChannelSettings = {
...
+  __plain__: UserNotificationSettingsPlain;
```
nvie, line 59: Also was a bit puzzled by the use of the nested value `__plain__`. Is this really needed, or could you basically just remove this intermediate nesting and use `type PrivateUserNotificationSettingsApi = UserNotificationSettingsPlain` here directly? Maybe a refactoring artefact, I wasn't sure, but I didn't see exactly the value the nesting provided immediately.

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
```diff
@@ -26,11 +27,52 @@ export type NotificationChannelSettings = {
...
+  [C in NotificationChannel]?: NotificationChannelSettings;
+};
+
+/**
+ * @private
+ *
+ * Symbol to brand some properties and methods
+ * as internal and private in `UserNotificationSettings`
+ */
+const kPrivate = Symbol("user-notification-settings-private");
```
nvie, line 49: Maybe rename this to `kPlain` in the end?

nvie (commented):
Thanks for adding the E2E test, and for the good documentation here. This definitely provides a much better DX, so I'm happy with this. Also the error message looks great with the link to our error page to provide more details 👍 

I did leave a couple of suggestions regarding the implementation of the types and exports and what APIs are public/private, but nothing too substantial—these should be simple to address I think.

Overall, nice work, @SugarDarius! 👌

e2e/next-sandbox/pages/user-notification-settings/index.tsx
```diff
@@ -112,58 +112,74 @@ function Settings() {
...
-        onUpdate={() =>
-          updateSettings({
-            webPush: {
-              thread: settings ? !settings.webPush.thread : false,
-              textMention: settings ? !settings.webPush.textMention : false,
-            },
-          })
-        }
-      />
+      {settings !== undefined && settings.email !== null ? (
```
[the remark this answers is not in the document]
SugarDarius: Done in [4fb40bd](https://github.com/liveblocks/liveblocks/pull/2265/commits/4fb40bd3e9e1ae061622f1ba03036994a7379ae9)

examples/nextjs-notification-settings/app/settings/_components/user-notifications-settings.tsx
```diff
@@ -9,205 +9,202 @@ import {
...
-  const handleChangeEmailChannel$fileUploadedKind = (
-    checked: boolean
-  ): void => {
-    updateNotificationSettings({
-      email: {
-        $fileUploaded: checked,
-      },
-    });
-  };
+  // Each returns `true` if the current user has at least one notification kind enabled on that channel
```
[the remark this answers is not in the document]
SugarDarius: No it's the comment which isn't good. The role the utility `isNotificationChannelEnabled` is to check for a channel that every kind is enabled. It allows developers to create one check box for every kinds and define UIs where users can use this check to enable or disable every kinds at once.

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
```diff
@@ -47,19 +89,120 @@ type DeepPartialWithAugmentation<T> = T extends object
...
+ * const isEmailChannelEnabled = isNotificationChannelEnabled(settings.email);
+ * ```
  */
 export function isNotificationChannelEnabled(
-  settings: NotificationChannelSettings
+  settings: NotificationChannelSettings | null
 ): boolean {
-  return values(settings).every((enabled) => enabled === true);
+  return settings !== null
+    ? values(settings).every((enabled) => enabled === true)
```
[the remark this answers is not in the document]
SugarDarius: See my [comment](https://github.com/liveblocks/liveblocks/pull/2265#discussion_r1977186930) above ☝🏻

examples/nextjs-notification-settings/package-lock.json
```diff
@@ -8,14 +8,14 @@
...
-        "@liveblocks/react-ui": "^2.18.3",
-        "@liveblocks/yjs": "^2.18.3",
+        "@liveblocks/client": "^2.18.4-uns2",
+        "@liveblocks/core": "^2.18.4-uns2",
+        "@liveblocks/emails": "^2.18.4-uns2",
+        "@liveblocks/node": "^2.18.4-uns2",
+        "@liveblocks/react": "^2.18.4-uns2",
+        "@liveblocks/react-tiptap": "^2.18.4-uns2",
+        "@liveblocks/react-ui": "^2.18.4-uns2",
+        "@liveblocks/yjs": "^2.18.4-uns2",
```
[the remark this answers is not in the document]
SugarDarius: I'd like to run the script to update the examples right away after publishing a new version. I do not think it has a huge impact to keep it like it is and I'd like the example to be merge with a workable version. It will be a matter of minutes before updating the examples after the publish

packages/liveblocks-core/src/api-client.ts
SugarDarius: `UserNotificationSettingsPlain` is marked as `@private` not `@internal`. Also `createApiClient` isn't exported from `@liveblocks/core` index file. So I'm not sure sure to well understand you here 🤔 Can you elaborate?

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
SugarDarius: I'd like to keep it if it's okay for you. Calling `this:` in a getter isn't a "natural" behavior even if it makes sense. And for legacy and knowledge sharing I think it's good to keep it to avoid someone removes it from the list of params and thinks it's safe and good to just do `this[kPrivate]` without declaring from the list of params.

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
SugarDarius: I think I'm seeing what you mean. I just followed our current conventions we had on `kInternal` properties. I can remove the nested value indeed 👍🏻

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
SugarDarius: Let think about it 👍🏻

packages/liveblocks-core/src/protocol/UserNotificationSettings.ts
SugarDarius: Done in [1df1f19](https://github.com/liveblocks/liveblocks/pull/2265/commits/1df1f19662f1312520c66a13ab40426a5cfd8c23)

LMK if it was what you were thinking about.

SugarDarius:
@nvie I think I addressed your feedback like you were expressing it. PTAL thanks 🙏🏻

CHANGELOG.md
```diff
@@ -1,5 +1,42 @@
...
+// webPush channel disabled ❌
+
+const [{ settings }] = useNotificationSettings();
+
+console.log(settings.emails); // ✅ 👉🏻 { thread: true, textMention: true, $fileUploaded: true }
+console.log(settings.webPush); // ✅  👉🏻 `null` with an error log "In order to use the 'webPush' channel, please set up your project first. For more information: https://liveblocks.io/docs/errors/enable-a-notification-channel"
+```
+
+The main benefits are your code will become more predictable and will avoid to
+create a breakage into your app if someone is disabling channels by mistake.
```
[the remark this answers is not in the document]
CTNicholas: My preference for the changelogs is to go with whatever is easiest for everyone! This is a low-effort place to briefly document every change. Just one or two lines is fine.

Also we don't need to highlight this difference between the two versions in the docs, as we haven't even officially released the first version.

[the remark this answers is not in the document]
SugarDarius: Noted ✅

packages/liveblocks-core/src/api-client.ts
nvie: In that case it's fine, I think I misread this as if it was an exported type, but if not it's fine. Btw, `@private` has no significant meaning I think (mostly used for IDEs and human consumption), but `@internal` has a technical meaning: it will ensure that the type will never be published in any of the `*.d.ts` files. This can be extremely useful to not accidentally publish types, but it can also break your types if you're not careful.

For example, if you have:

```ts
export type A = { foo: Foo };

/** @internal */
export type Foo = Whatever;
```

Then TypeScript will simply put

```
declare type A = { foo: Foo };
```

In the `*.d.ts` output, which will lead a TS error "Unknown type 'Foo'".

nvie: In this case, because we're not exporting those types that are now marked `@internal`, you may be able to remove those annotations altogether, since non-exported types are internal automatically.

nvie (approved):
LGTM!

PR #2268, opened 2025-02-28
Add better error handling to `@liveblocks/node` methods

### LINKED ISSUE #1978: The error thrown by getRoom() in @liveblocks/node encodes an error object in the message key
**Describe the bug**

Initialising a liveblocks/node instance and calling getRoom on a non-existent roomId throws an error. I expected this to be a LiveblocksError class, but it seems to be something else with the following object encoded as a string within `error.message`:

```ts
{
  error: 'ROOM_NOT_FOUND',
  message: 'Room not found',
  suggestion: 'Please use a valid room ID, room IDs are available in the dashboard: https://liveblocks.io/dashboard/rooms',
  docs: 'https://liveblocks.io/docs/api-reference/rest-api-endpoints'
}
```

This means to check this error, I need to run 

```ts
JSON.parse(err.message).error === 'ROOM_NOT_FOUND')
```

which is obviously very fragile

**Expected behavior**

I'd expect this to return a LiveblocksError, so I can rely on the types. The encoded inner error is very strange, and means I need to write a bunch of defensive / slow code to see if the string can be parsed as a json object.

**Environment (please complete the following information):**

```
"@liveblocks/node": "^2.8.1"
```

nvie:
This PR adds two quality-of-life improvements to `@liveblocks/node`:

1. Fixes weirdly-encoded error messages in `LiveblocksError` where the message string seems to be the whole response's JSON body. This was making it harder than necessary to programmatically handle specific errors.
2. Adds support for passing an `AbortSignal` to every method (optional).

Fixes #1978

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -160,6 +160,10 @@ type M = DM;
 type S = DS;
 type U = DU;
 
+export type RequestOptions = {
+  signal?: AbortSignal;
+};
+
 /**
  * Interact with the Liveblocks API from your Node.js backend.
  */
@@ -178,7 +182,11 @@ export class Liveblocks {
     this.#baseUrl = new URL(getBaseUrl(options.baseUrl));
   }
 
-  async #post(path: URLSafeString, json: Json): Promise<Response> {
+  async #post(
+    path: URLSafeString,
+    json: Json,
+    options?: RequestOptions
+  ): Promise<Response> {
     const url = urljoin(this.#baseUrl, path);
     const headers = {
       Authorization: `Bearer ${this.#secret}`,
@@ -189,11 +197,16 @@ export class Liveblocks {
       method: "POST",
       headers,
       body: JSON.stringify(json),
+      signal: options?.signal,
     });
     return res;
   }
 
-  async #put(path: URLSafeString, json: Json): Promise<Response> {
+  async #put(
+    path: URLSafeString,
+    json: Json,
+    options?: RequestOptions
+  ): Promise<Response> {
     const url = urljoin(this.#baseUrl, path);
     const headers = {
       Authorization: `Bearer ${this.#secret}`,
@@ -204,13 +217,15 @@ export class Liveblocks {
       method: "PUT",
       headers,
       body: JSON.stringify(json),
+      signal: options?.signal,
     });
   }
 
   async #putBinary(
     path: URLSafeString,
     body: Uint8Array,
-    params?: QueryParams
+    params?: QueryParams,
+    options?: RequestOptions
   ): Promise<Response> {
     const url = urljoin(this.#baseUrl, path, params);
     const headers = {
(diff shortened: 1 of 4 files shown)


packages/liveblocks-node/src/client.ts
```diff
@@ -1689,27 +1846,64 @@ export class Liveblocks {
...
+      text = await res.text();
+    } catch {
+      text = FALLBACK;
+    }
+    let obj: JsonObject;
+    try {
+      obj = JSON.parse(text) as JsonObject;
+    } catch {
+      obj = { message: text };
+    }
```
SugarDarius, line 1900: What's the diff between doing this ☝🏻 and
```ts
import { tryParseJson } from "@liveblocks/core"
const obj = tryParseJson(text) ?? { message: text }
```
Just because of the type casting in `tryParseJson` returning `Json` and not a `JsonObject`?
Asking for my own knowledge mostly 😅

marcbouchenoire (approved):
Looks good!

~~Would it be weird to include `signal` in the existing `params` since those don't necessarily mean URL params?~~

packages/liveblocks-node/src/client.ts
```diff
@@ -520,13 +561,16 @@ export class Liveblocks {
...
    * @param roomId The id of the room to return.
    * @returns The room with the given id.
+   * @param options.signal (optional) An abort signal to cancel the request.
    */
-  public async getRoom(roomId: string): Promise<RoomData> {
-    const res = await this.#get(url`/v2/rooms/${roomId}`);
+  public async getRoom(
+    roomId: string,
+    options?: RequestOptions
+  ): Promise<RoomData> {
```
nimeshnayaju, line 569: Do you think `getRoom()` (and perhaps, `getThread`, `getComment`) should return a `null` object if the room was not found? Not sure how big of a breaking change this would be though.

packages/liveblocks-node/src/client.ts
nimeshnayaju: I think it is because we almost always return a `JsonObject` from the server, but in the case we don't, `reason`, `suggestion`, `error` are set to `undefined`.


```ts
return new LiveblocksError(
 (obj.message || FALLBACK) as string,
 res.status,
 obj.reason as string | undefined,
 obj.suggestion as string | undefined,
 obj.error as string | undefined
);
```

packages/liveblocks-node/src/client.ts
nvie: I think that would be nice honestly. Dealing with custom errors from libraries is never a great DX, especially not when they may be "expected" errors like 404s. I'd prefer `null`, but ideally we'd make that a breaking change and write an upgrade guide in another release. I don't want to introduce a breaking change now.

packages/liveblocks-node/src/client.ts
```diff
@@ -204,40 +217,62 @@ export class Liveblocks {
...
+      headers,
+      signal: options?.signal,
+    });
     return res;
   }
 
-  async #get(path: URLSafeString, params?: QueryParams): Promise<Response> {
+  async #get(
+    path: URLSafeString,
+    params: QueryParams,
```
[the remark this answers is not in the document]
nvie: Can do 👍  I don't have a strong opinion on this one.

packages/liveblocks-node/src/client.ts
nvie: I don't think there is a difference, might actually change this – thanks for the suggestion 👍

nvie:
> Would it be weird to include `signal` in the existing `params` since those don't necessarily mean URL params?

`params` are `QueryParams`, so I would expect them to show up in the URLs. It'd be strange to me to put the `signal` in there.

PR #2270, opened 2025-03-03
Composer improvements related to mentions

marcbouchenoire:
### Fix mention suggestions not opening around whitespace

Fixes https://liveblocks.slack.com/archives/C04TFCME5MF/p1740731569791229

We allowed whitespace in mentions in `2.18.3` but prevented it after `@`, and it led to cases like `hello @| world` not working. This PR fixes that, `hello @| world` works, `hello @ |world` doesn't.

### Improve mention suggestions floating placement

Fixes https://discord.com/channels/913109211746009108/1341426138425327626/1341427760425472162

Scrolling now only happens as a last resort, when both sides don't have enough space.

#### Before

https://github.com/user-attachments/assets/863d35f5-9251-4373-990e-a46964444399

#### After

https://github.com/user-attachments/assets/6b9105f6-abf9-4d99-b177-a6a413070e2a

### DIFF
--- a/packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
+++ b/packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
@@ -302,11 +302,15 @@ function ComposerEditorMentionSuggestionsWrapper({
     update();
 
     // Reset the mention suggestions after the placement update
-    requestAnimationFrame(() => {
+    const animationFrame = requestAnimationFrame(() => {
       mentionSuggestions.style.overflowY = "auto";
       mentionSuggestions.style.maxHeight =
         "var(--lb-composer-floating-available-height)";
     });
+
+    return () => {
+      cancelAnimationFrame(animationFrame);
+    };
   }, [userIds?.length, isOpen, elements.floating, update]);
 
   return (
--- a/packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
+++ b/packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
@@ -257,6 +257,8 @@ function ComposerEditorMentionSuggestionsWrapper({
     placement,
     x,
     y,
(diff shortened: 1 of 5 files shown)


packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
```diff
@@ -279,6 +281,34 @@ function ComposerEditorMentionSuggestionsWrapper({
...
+    // Trigger a placement update
+    update();
+
+    // Reset the mention suggestions after the placement update
+    requestAnimationFrame(() => {
+      mentionSuggestions.style.overflowY = "auto";
+      mentionSuggestions.style.maxHeight =
+        "var(--lb-composer-floating-available-height)";
+    });
+  }, [userIds?.length, isOpen, elements.floating, update]);
```
SugarDarius, line 310: What happens in case the `userIds` list quickly change?
I would tend to naturally cancel the animation frame in the cleanup. 
```tsx
useLayoutEffect(() => {
  // ...
  let id: number | undefined = undefined
  id  = requestAnimationFrame(() => {
    // ...
  })
  
  return () => {
    if (id) {
      cancelAnimationFrame(id)
    }
  }
}, [userIds.length])
```

My point is to prevent any visual glitch if  `userIds.length` is updated but we already scheduled an animation frame. I'd like to prevent quick jumps on the available height. 

Feel free to ignore this comment if you think there is no chance of visual glitches though 🙂

SugarDarius (approved):

packages/liveblocks-react-ui/src/primitives/Composer/index.tsx
marcbouchenoire: The list itself re-renders when the users change so the default styles always get re-applied but it doesn't hurt to be explicit 👍

PR #2271, opened 2025-03-03
Add optional `useRoom({ allowOutsideRoom: true })` API

nvie:
This adds an optional `useRoom({ allowOutsideRoom: true })` API to replace the `useRoomOrNull` private/internal hook that some users have started relying on. This new API should be more easily discovered, reduce the number of public APIs we expose at the top-level, yet still provide correct typing.

```ts
// Inferred TypeScript types
useRoom()                            // returns Room
useRoom({ allowOutsideRoom: true })  // returns Room | null
```

When used outside of a RoomProvider context:

```ts
// Runtime behavior
useRoom()                            // throws
useRoom({ allowOutsideRoom: true })  // returns null
```

See also, [this Slack message](https://liveblocks.slack.com/archives/C083U8FAUF8/p1740491333528449?thread_ts=1740396195.278319&cid=C083U8FAUF8) for context.

### DIFF
--- a/packages/liveblocks-react/test-d/augmentation.test-d.tsx
+++ b/packages/liveblocks-react/test-d/augmentation.test-d.tsx
@@ -259,6 +259,14 @@ declare global {
   expectType<number>(room.getPresence().cursor.x);
   expectType<number>(room.getPresence().cursor.y);
   expectError(room.getPresence().nonexisting);
+
+  expectType<string>(classic.useRoom({ allowOutsideRoom: false }).id);
+  expectType<string | undefined>(
+    classic.useRoom({ allowOutsideRoom: true })?.id
+  );
+  expectType<string | undefined>(
+    classic.useRoom({ allowOutsideRoom: !Math.random() })?.id
+  );
 }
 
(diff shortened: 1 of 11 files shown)


packages/liveblocks-react/test-d/augmentation.test-d.tsx
```diff
@@ -259,6 +259,14 @@ declare global {
   expectType<number>(room.getPresence().cursor.x);
   expectType<number>(room.getPresence().cursor.y);
   expectError(room.getPresence().nonexisting);
+
+  expectType<string>(classic.useRoom({ allowOutsideRoom: false }).id);
+  expectType<string | undefined>(
+    classic.useRoom({ allowOutsideRoom: true })?.id
+  );
+  expectType<string | undefined>(
+    classic.useRoom({ allowOutsideRoom: !Math.random() })?.id
```
CTNicholas, line 268: This will always be `false`, is it meant to be this? `Math.random() < 0.5`. Same in other places.

CTNicholas (approved):
Nice! Just one comment

PR #2275, opened 2025-03-04
Pass `userId` to the request body when calling `markThreadAsResolved` and `markThreadAsUnresolved`

nimeshnayaju:
I noticed that our `markThreadAsResolved` and `markThreadAsUnresolved` methods were not passing `userId` field to the request body, so the endpoint always failed with an error.

### DIFF
--- a/packages/liveblocks-node/src/__tests__/client.test.ts
+++ b/packages/liveblocks-node/src/__tests__/client.test.ts
@@ -221,1555 +221,1730 @@ describe("client", () => {
       }
     });
   });
-
-  test("should return a list of room when getRooms receives a successful response", async () => {
-    const client = new Liveblocks({ secret: "sk_xxx" });
-    await expect(client.getRooms()).resolves.toEqual({
-      nextPage: "/v2/rooms?startingAfter=1",
-      data: [room],
-    });
-  });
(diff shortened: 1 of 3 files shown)


packages/liveblocks-node/src/__tests__/client.test.ts
```diff
@@ -221,1555 +221,1730 @@ describe("client", () => {
       }
     });
   });
-
-  test("should return a list of room when getRooms receives a successful response", async () => {
-    const client = new Liveblocks({ secret: "sk_xxx" });
-    await expect(client.getRooms()).resolves.toEqual({
-      nextPage: "/v2/rooms?startingAfter=1",
-      data: [room],
+  describe("get rooms", () => {
```
nimeshnayaju, line 224: Updated the tests to use a `describe` block as well for easier navigation of the test file as it was getting very large.

FlowFlorent (approved):

SugarDarius (approved):
LGTM ✅ 

I've just updated the changelog line as we released yesterday the `v2.20.0`. 
You fix will come in the next one I think 🙂

marcbouchenoire (approved):

PR #2276, opened 2025-03-05
New tab for iframes

CTNicholas:
NextAuth doesn't work in iframes

CleanShot 2025-03-05 at 13 04 33@2x


SugarDarius (approved):
Thank you 🖤

PR #2279, opened 2025-03-07
Explain text editor is permanent

CTNicholas:


nvie (approved):

stevenfabre (approved):
Left a couple of comments. Looks good otherwise! Thanks for doing this so quickly

PR #2281, opened 2025-03-10
Fix tutorial bundling

CTNicholas:
Merge straight after:
- https://github.com/liveblocks/liveblocks.io/pull/2538


tutorial/react/getting-started/nesting-data-types/initial/App.tsx
```diff
@@ -11,7 +11,10 @@ export default function App() {
   const publicApiKey = "{% LIVEBLOCKS_PUBLIC_KEY %}";
 
   return (
-    <LiveblocksProvider publicApiKey={publicApiKey}>
+    <LiveblocksProvider
+      publicApiKey={publicApiKey}
+      key={Math.random() /* Fixes a tutorial bug, don't do this */}
```
marcbouchenoire, line 16: Is there no other way to fix this? Could we trigger a Sandpack refresh/reload when you click on the solution button instead?

CTNicholas: I'm actually doing that too, just forgot to commit it on the other PR. Doesn't work without the `key` still. I'll mess around a bit and see if I can find a better way.

CTNicholas: No, I don't see a way around this. This is the only way that isn't buggy, and I prefer this solution to people writing in every few works saying it doesn't work

marcbouchenoire (approved):

PR #2288, opened 2025-03-12
Refactor core internal logic for thread notification event email

SugarDarius:
## What's going on?
This PR refactors the core internal logic in `@liveblocks/emails` for thread notification event. It makes the code more dry and factorized to be more scalable and ready to handle future work and features. It's a necessary first before scaling our functions.

### Convert comment body
Now the two functions `convertCommentBodyAsReact` and `convertCommentBodyAsHtml` are replaced by `convertCommentBody` which is more generic and use the same single source of truth to convert bodies into a format like html, React or whatever. 
It exposes a new generic type `ConvertCommentBodyElements` where we can define each elements we want to convert the comment body 👇🏻 
```tsx
/**
 * Protocol:
 * Comment body elements to be converted to a custom format `T`
 */
export type ConvertCommentBodyElements<T, U extends BaseUserMeta = DU> = {
  /**
   * The container element used to display comment body blocks.
   */
  container: (args: CommentBodyContainerElementArgs<T>) => T;
  /**
   * The paragraph element used to display paragraphs.
   */
  paragraph: (args: CommentBodyParagraphElementArgs<T>, index: number) => T;
  /**
   * The text element used to display text elements.
   */
  text: (args: CommentBodyTextElementArgs, index: number) => T;
  /**
   * The link element used to display links.
   */
  link: (args: CommentBodyLinkElementArgs, index: number) => T;
  /**
   * The mention element used to display mentions.
   */
  mention: (args: CommentBodyMentionElementArgs<U>, index: number) => T;
};
```

### Prepare Thread Notification Email
The function `prepareThreadNotificationEmailBaseData` has been replaced by `prepareThreadNotificationEmail` which now handles extracting data, resolving room info, resolving users and converting comment bodies all at once. It's a new abstraction layer which makes more sense. 
And now our functions `prepareThreadNotificationEmailAsHtml` and `prepareThreadNotificationEmailAsReact` are just using simply: 
```tsx
export async function prepareThreadNotificationEmailAsReact(client, event, options): Promise<ThreadNotificationEmailDataAsReact | null> {
  // ...
  const data = await prepareThreadNotificationEmail<ReactNode, BaseUserMeta>(
    client,
    event,
    { /* ... options ... */ },
    {
      /* elements to convert comment body */
    }
  )
  
  return data
}
```
It will allow us to expose more functions we'd like such as `prepareThreadNotificationEmailAsMarkdown` or `prepareThreadNotificationEmailAsCustom`.

> [!IMPORTANT]
> Please note that with this new implementation we're not forced anymore to expose `htmlBody` and `reactBody` for the returned comments. We can now simply return a single `body` property no matter what format we're using to convert the comment body. So `htmlBody` and `reactBody` are now marked as deprecated to avoid any breaking changes in our customers code. 

Thanks 🙏🏻 

> [!TIP]
> Our examples using our prepare functions for thread notification event are updated to the latest test version `v2.21.0-emails3` FYI. It will be better for you if you want to do some testing.

Fixes [LB-1806](https://linear.app/liveblocks/issue/LB-1806/refactor-core-internal-of-thread-notification-functions)


CTNicholas (commented):
Hey! I’m not too familiar with the internals here, so I haven’t reviewed it.

Quick question—why are we renaming `reactBody/htmlBody` to `body`? If we do this, we need an upgrade guide, as it's a breaking change 👍

CTNicholas:
> I explained it the description of this PR 😅 under the `Important` section. Also it's not a breaking change I don't want to introduce it right now. So I'm keeping the backward compatibility with `reactBody` and `htmlBody` and I marked them as deprecated.

Right, I saw that! My question is—why is the customer's API changing? Is this better for them?

We have [this function](https://liveblocks.io/docs/api-reference/liveblocks-client#stringify-comment-body) already where users can insert `comment.body`, which is the `CommentBody` type. Would adding new `.body` types increase confusion for users? 

Before this PR, you have:
- `comment.body` → always CommentBody
- `comment.htmlBody` → always HTML
- `comment.reactBody` → always JSX

After this PR, you have:
- `comment.body` → ???

I believe in being really careful with API naming so that things are as clear as possible. Though maybe I'm wrong here, what do you think?

SugarDarius:
> Right, I saw that! My question is—why is the customer's API changing? Is this better for them?

Before this refactor the logic was too opinionated and too restrictive IMO -- so yes it's better I think.
So now t's changing because of the refactoring we will allow in the future to let customers convert the comment body format in a format of their own -- whatever they want (you can read more on this [slack thread](https://liveblocks.slack.com/archives/C07JBLYB9GW/p1741597103430419)). 
And so now our `as` functions converting in React or html are becoming on-top abstractions. See my code example in the PR's description.

After time and some reflexion I'm not really happy with having `htmlBody` and `reactBody`. From a DX standpoint when you're converting the body into another format no matter what it is you don't really care about the naming, you care most about the type you're getting from the result of the function. For example if I call `prepareThreadNotificationEmailAsReact` then the body of the comments should be of type `React.ReactNode` this is what's the most important.
Moreover the function aren't returning a `CommentData` type they are returning a `CommentEmailData` type which has not the same shape. 
I want to expose a new function named `prepareThreadNotificationEmailAsCustom` to let customers use their own format as an on-top abstraction as stated and we will not be able to know what type it will be. For example if it's a custom object on their own called `MyCompanyFormat` we won't be able to produce a property key called `commentEmailData.myCompanyFormatBody: MyCompanyFormat`.

I just want to be coherent and clear across the functions.

I hope I'm clear 😅 LMK if not 🙂

CTNicholas:
> I hope I'm clear 😅 LMK if not 🙂

That makes sense, thank you for explaining 🙂

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## vNEXT (not yet published)
 
+## `@liveblocks/emails`
+
+- Implement new abstraction protocol to convert comment bodies.
+- Implement a new core logic for thread notification event.
```
marcbouchenoire, line 6: I don't think we need to mention these two here since they have no public effect.

docs/pages/api-reference/liveblocks-emails.mdx
```diff
@@ -348,6 +348,9 @@ export async function POST(request: Request) {
         createdAt: Date <Fri Dec 15 2023 14:15:22 GMT+0000 (Greenwich Mean Time)>,
 
         // The formatted comment, pass it to React `children`
+        body: { /* ... */},
+
+        // Deprecated: the formatted comment, pass it to React `children` (use `body` instead)
         reactBody: { /* ... */ },
```
marcbouchenoire, line 354: I don't think we should keep it visible here in the docs, imo we can remove it completely and rely on `@deprecated` tags for previous users.

CHANGELOG.md
SugarDarius: I will tend to keep at least `Implement a new core logic for thread notification event.` to explain more about than just the deprecation. It gives more context on why `reactBody` and `htmlBody` ^^

docs/pages/api-reference/liveblocks-emails.mdx
SugarDarius: @CTNicholas any feedback on this here? From my perspective I always thought it's important to keep visually in the deprecation in the docs. Maybe I'm wrong and as @marcbouchenoire said just rely on the `@deprecated` mark in the code?

CTNicholas: I think it's okay to mention it here, since we are essentially showing a `console.log` of the real object, and users will see this too 👍

PR #2293, opened 2025-03-17
New APIs to mutate Storage from Node

nvie:
This PR adds some new APIs that let you (mass) mutate room's storage contents straight from Node.

The new APIs should feel very familiar if you're already using `useMutation()` in React.

```ts

const liveblocks = new Liveblocks(/* options */);

// Mutate a single room
await liveblocks.mutateStorage(
  "my-room",
  async ({ root }) => {
    // Arbitrarily mutate the room's storage contents using Live structures
    root.get("animals").push("🐟");
  }
);
```

Mass-mutating multiple rooms is also possible. The first argument can be anything you can pass to `.getRooms()` to select/filter the rooms. You can filter by room ID prefix, by metadata, or by ownership:

```ts
// Mass-mutate rooms
await liveblocks.massMutateStorage(
  { query: { roomId: { startsWith: "my-prefix-" } } },
  // Or: { query: { metadata: { color: "red" } } }
  // Or: { groupIds: ["engineering"] }

  async ({ room, root }) => {
    // Arbitrarily mutate the room's storage contents using Live structures
    root.get("animals").push("🐟");
  }
);
```

When running mutations from the backend, make sure to not change the storage in a way that connected clients don't expect. Connected users will not see a presence.

By default, mass mutations will be executed concurrently, mutating 8 rooms in parallel. You can control this behavior with the `concurrency` option. If you which to run mutations serially, set `concurrency` to 1.

If your mutation function is asynchronous, then mutations will get flushed to the server periodically, using the `flushInterval` option (defaults to 1000ms), can be any value, but should be at least 100ms. This is similar to the `throttle` setting in the browser client, but it cannot go as low as 16ms, because this isn't using websockets.


packages/liveblocks-node/src/client.ts
```diff
@@ -520,6 +558,33 @@ export class Liveblocks {
...
+      );
+      for (const room of data) {
+        yield room;
+      }
+      if (!nextCursor) {
+        break;
+      }
+      cursor = nextCursor;
+    }
+  }
```
nvie, line 586: This is an abstraction over calling `.getRooms()` without having to think about pagination. I kept it private for now, but we can consider making it a public/supported API. What do you think @nimeshnayaju @CTNicholas ?

CTNicholas: Cool idea, I can see this being helpful.

CTNicholas:
Looks great! 

So by default, it will flush every `1000ms`? Do you think users will see their Storage update in 2 or 3 different flushes, and think our function is buggy or not working right? I'm wondering if the default should just run at the end of the callback instead.

packages/liveblocks-node/src/client.ts
nimeshnayaju: Love generator functions and is perfect for this use case. 

I also had a thought - if it made sense to use `yield` (generator functions) instead of the `flush` API. I am not suggesting we do that but @nvie have you considered that or thought of such alternative?

packages/liveblocks-node/src/client.ts
```diff
@@ -1891,6 +1971,188 @@ export class Liveblocks {
...
+    // pool.
+    // TODO Implement the handling of fix-ops:
+    // const data = (await res.json()) as {
+    //   messages: readonly (
+    //     | ServerMsg<JsonObject, BaseUserMeta, Json>
+    //     | readonly ServerMsg<JsonObject, BaseUserMeta, Json>[]
+    //   )[];
+    // };
+    // return data;
+  }
```
nimeshnayaju, line 2155: I am curious how this would work. Would we update the `root` as we flush and if there are corrections to make to the storage?

packages/liveblocks-node/src/client.ts
nvie: @nimeshnayaju Not super sure what you mean. Do you mean to accept a coroutine instead of a regular callback function as the 2nd argument, so users could "yield" back to trigger a flush? I think that's not obvious enough to most users. Also, `flush()` no longer exists anymore. Flushing happens automatically now, you don't need to control it explicitly anymore.

nvie:
> So by default, it will flush every `1000ms`? Do you think users will see their Storage update in 2 or 3 different flushes, and think our function is buggy or not working right? I'm wondering if the default should just run at the end of the callback instead.

Ah yes, that may be better actually. That's equivalent to letting `flushInterval` default to positive infinity.

packages/liveblocks-node/src/client.ts
nvie: @CTNicholas OK, then I'll think I'll make this `.iterRooms()` helper public too as part of this change 👍

packages/liveblocks-node/src/client.ts
nvie: Yep, it would be the same as how the browser client works. The edge case this fixes for is small. I think 99% of users won't ever notice anything for this use case, but it's something that in theory (and thus eventually for some users) might be causing ordering issues when updating LiveLists from a mutation that are also simultaneously being edited by users.

packages/liveblocks-node/src/client.ts
nimeshnayaju: @nvie I meant an alternative API where you could `yield` to `flush` changes, but I don't think it would have the same ergonomics that we now have with the automatic flush.

nimeshnayaju (approved):

SugarDarius (approved):

packages/liveblocks-node/src/client.ts
nvie: Made the `.iterRooms()` API public/supported in b7ad1958f4a3dedba12486fe96f8a34d9d88a8e2.

CTNicholas:
> Ah yes, that may be better actually. That's equivalent to letting `flushInterval` default to positive infinity.

Sounds good to me!

CTNicholas:
> @CTNicholas I just updated this in [c0af0fc](https://github.com/liveblocks/liveblocks/commit/c0af0fcfe30ca7120f63a85514543a8be606ba7f). Do you think we should still keep the `flushInterval` control option at all, or maybe not have this option to start with?

Great, thanks! It will be helpful with AI streaming, we have a couple of customers that do this with Storage. 

However, I do think a `flush` function would be easier to understand than a background interval. We could have a simple snippet in our docs, they could run it as each chunk is received from an AI stream. This is Vercel AI SDK:

```tsx
import { streamText } from "ai";

await liveblocks.mutateStorage(
  "my-room",
  async ({ root, flush }) => {
    const result = streamText({
      model: "gpt-4.5-preview",
      prompt: 'Invent a new holiday and describe its traditions.',
    });
    
    for await (const textPart of result.textStream) {
      root.get("ai-value").set(textPart);
      flush();
    }    
  }
);

```

nvie:
> If I were choosing, I would personally swap the interval for this.

Hmm. It's surely possible, but I just swapped out the explicit `flush` for the auto-flushing because I thought the manual flushing was deemed a bit too complex to deal with? I'd love to figure this API out before we publish this.

One thing to consider is that even if we end up adding a `flush()` API, it's not going to guarantee an HTTP request is made immediately. There still will be _some_ throttling going on. For example:

```ts
for (let i = 0; i < 10; i++) {
  root.set("a", i);
  flush();
}
```

This will definitely not trigger 10 HTTP requests. Calling `flush()` at best just means "it's okay to flush all changes up to this point".

Options we have:

1. No intermediate flushes. Just one HTTP request at the end of the mutation.
1. Explicit `flush()` (but as explained above).
1. Automatic flushing, based on a configurable interval. Ideally with a good opinionated default.
1. A mix of explicit and automatic flushing? Allowing both?

Adding explicit `flush()` also is a one-way door: we have to keep the API. Starting without an explicit `flush()` means we could always add the API later. For the same reason, maybe we should also start without a config option for it?

I think going with option (3) and picking a great default has my preference so far after considering most use cases. Maybe 1000ms was too much as a default, and we should lower it a bit?

Curious to hear your thoughts and preferences, @CTNicholas, @nimeshnayaju, @SugarDarius, @jrowny.

SugarDarius:
> Curious to hear your thoughts and preferences 

After some reflexions I'd prefer option 3. It feels more manageable and configurable to have both automatic and explicit flushing. We will cover a lot of use cases and let developers create graceful recovery behaviors such as force flushing in case of failure:
```tsx
await liveblocks.mutateStorage(
  "my-room",
  async ({ root, flush }) => {
    try {
      const replicates = await db.tables('replicates').select('*')
      for (const replicate of replicates) {
        const data = await formatReplicate(replicate)
        // Let the automatic flush do its job preiodically
        root.get('my-replicate').set(data)
      }
    }
    catch (err) {
      // Explicit (force) flushing in case of error
      flush()
    }   
  }
);
```

What I tend to see if:
- A default flush interval option set to `~100ms/500ms`. 
- An explicit `flush` option to call which will bypass the current `debouncer` and will send the message immediately.

nvie:
> After some reflexions I'd prefer option 3.

Did you mean option 4?

SugarDarius:
> Did you mean option 4?
Sorry, yes option `4` 👍🏻

nvie:
This form:

```ts
try {
  …
} finally {
  // Explicit (force) flushing in case of error
  flush()
}
```

Would behave identically to:

```ts
try {
  …
} finally {
  // Explicit (force) flushing in case of error
  flush()
}
```

So an explicit `flush()` would not be needed to do this.

> _"An explicit `flush` option to call which will bypass the current `debouncer` and will send the message immediately."_

As I described above, I think allowing to bypass a debouncer is not a good idea in general and I cannot think of a use case where this amount of timing control would be required. We don't even offer this level of flush control with our WebSockets from the browser (our most real time environment).

nimeshnayaju (approved):

nimeshnayaju:
@nvie Have you considered as an alternative to the `flush` API, that we flush immediately after each update (might be too expensive to make an HTTP call so often - maybe we look into streaming the updates? - I know we discussed this briefly privately on Slack) and a`batch` function that accepts a callback so that it's easier to visually track what the upcoming `flush` entails. I guess, this suggestion is closer to option 4.

> Options we have:
> 
> No intermediate flushes. Just one HTTP request at the end of the mutation.
> Explicit flush() (but as explained above).
> Automatic flushing, based on a configurable interval. Ideally with a good opinionated default.
> A mix of explicit and automatic flushing? Allowing both?

nvie:
> _"Have you considered as an alternative to the `flush` API, that we flush immediately after each update (might be too expensive to make an HTTP call so often - maybe we look into streaming the updates? - I know we discussed this briefly privately on Slack) and a`batch` function that accepts a callback so that it's easier to visually track what the upcoming `flush` entails. I guess, this suggestion is closer to option 4."_

So in the `for (let i=0; i<10; i++) { root.set("a", i) }` example, it would trigger 10 HTTP requests? I think it's a bad idea. Also opening a streaming connection to the server without a clear end bound seems like asking for trouble, as I don't think it will allow the DO to hibernate or limit the ability to accept other HTTP requests in the mean time.

CTNicholas:
We've had a chat about this on Slack, and we think it's best to:
- Automatically flush `sync` changes then switch to an interval for `async` changes
- An interval of `200ms`
- No configuration

Reasoning behind this:
- We want it to feel real-time. Flushing `sync` changes immediately helps with this.
- If you await something, it's generally going to be a fetch, or similar, which takes a long time anyway. Adding a delay of just `200ms` is not going to feel noticeable.
- Because we have the immediate flush, and the short delay on `async`, it will always feel pretty much real-time, so we don't need to offer a configuration option.

PR #2296, opened 2025-03-19
Only loosen dupe detection of exactly-same versions when inside a testing context

nvie:
In prod/dev contexts, we definitely still want to error when this situation occurs.

### DIFF
--- a/packages/liveblocks-core/src/dupe-detection.ts
+++ b/packages/liveblocks-core/src/dupe-detection.ts
@@ -39,7 +39,7 @@ export function detectDupes(
 
   if (!g[pkgId]) {
     g[pkgId] = pkgBuildInfo;
-  } else if (g[pkgId] === pkgBuildInfo) {
+  } else if (process.env.NODE_ENV === "test" && g[pkgId] === pkgBuildInfo) {
     // Allow it, see https://github.com/liveblocks/liveblocks/pull/1004
   } else {
     const msg = [


packages/liveblocks-core/src/dupe-detection.ts
```diff
@@ -39,7 +39,7 @@ export function detectDupes(
 
   if (!g[pkgId]) {
     g[pkgId] = pkgBuildInfo;
-  } else if (g[pkgId] === pkgBuildInfo) {
+  } else if (process.env.NODE_ENV === "test" && g[pkgId] === pkgBuildInfo) {
```
SugarDarius, line 42: Can we strengthen the check a bit more?
I may be wrong but the term `test` is very generic, we can think about customers having test environments containing extended values like `test-preprod`, `test_staging`, `test*`...

So I would make the check such as
```ts
else if (process.env.NODE_ENV.includes("test") && g[pkgId] === pkgBuildInfo) {
  // ...
}
```
Like this we're not forcing customers to set the `NODE_ENV` to be exactly `test` even if it will be the major case.
WDYT?

SugarDarius (approved):

packages/liveblocks-core/src/dupe-detection.ts
nvie: I think that's a bit brittle and arbitrary. Really under all normal conditions, we want to error in this situation. The only exception where this is too aggressive is in a testing situation. Definitely in a Jest test setup, but maybe also in Vitest, I don't know for sure.

An alternative approach could be to simply remove this condition and always throw, _except_ when a well-known env var/flag is set, i.e. `NO_LIVEBLOCKS_DUPE_WARNING`. But I think that's just overthinking it too much and I want to stay pragmatic here.

nvie:
Not pursuing this.

PR #2303, opened 2025-03-21
Export more icons in `@liveblocks/react-ui`

marcbouchenoire:
Since we're doing another release, we might as well do this now.


CTNicholas (approved):
Nice! Thanks

PR #2311, opened 2025-03-24
Subscribe/unsubcribe

marcbouchenoire:
🚧🚧🚧

### DIFF
--- a/packages/liveblocks-react/src/room.tsx
+++ b/packages/liveblocks-react/src/room.tsx
@@ -388,6 +388,8 @@ function makeRoomContextBundle<
     useEditThreadMetadata,
     useMarkThreadAsResolved,
     useMarkThreadAsUnresolved,
+    useSubscribeToThread,
+    useUnsubscribeFromThread,
     useCreateComment,
     useEditComment,
     useDeleteComment,
@@ -451,6 +453,8 @@ function makeRoomContextBundle<
       useEditThreadMetadata,
       useMarkThreadAsResolved,
       useMarkThreadAsUnresolved,
+      useSubscribeToThread,
+      useUnsubscribeFromThread,
       useCreateComment,
       useEditComment,
       useDeleteComment,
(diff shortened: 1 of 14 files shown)


packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1278,6 +1406,34 @@ export class UmbrellaStore<M extends BaseMetadata> {
     });
   }
 
+  /**
+   * Creates a existing subscription, replacing the corresponding
+   * optimistic update.
+   */
```
SugarDarius, line 1412:
```suggestion
  /**
   * Creates an existing subscription, replacing the corresponding
   * optimistic update.
   */
```

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1813,52 +1978,128 @@ function applyOptimisticUpdates_forThreadifications<M extends BaseMetadata>(
...
+            // Create subscriptions for all existing threads in the room
+            subscriptions[subscriptionKey] = {
+              kind: "thread",
+              subjectId: thread.id,
+              createdAt: new Date(),
+            };
+          } else if (update.settings.threads === "none") {
+            // Delete subscriptions for all existing threads in the room
+            delete subscriptions[subscriptionKey];
+          } else if (update.settings.threads === "replies_and_mentions") {
```
SugarDarius, line 2042: Maybe you could dry up all this logic part to make it more readable and understandable by using inner closures such as `applyThreadSettings()`, `shouldSubscribeToThread()`, `subscribe()` and `unsubscribe()`? 🤔

SugarDarius (approved):
What a job 🚀 👏🏻 
Very small comments on my side, LGTM ✅

packages/liveblocks-react/src/room.tsx
```diff
@@ -277,33 +278,42 @@ function makeRoomExtrasForClient(client: OpaqueClient) {
...
     store,
     onMutationFailure,
+    pollThreadsForRoomId: (roomId: string) => {
+      const threadsPoller = threadsPollersByRoomId.getOrCreate(roomId);
+
+      // If there's a threads poller for this room, immediately trigger it
+      if (threadsPoller) {
+        threadsPoller.markAsStale();
+        threadsPoller.pollNowIfStale();
+      }
```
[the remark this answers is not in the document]
marcbouchenoire: We had [a discussion](https://liveblocks.slack.com/archives/C08DGF1N7EG/p1744889343110589?thread_ts=1744884626.140219&channel=C08DGF1N7EG&message_ts=1744889343.110589) about it with @nvie, I was initially thinking of `forcePoll` too but there are other factors like deduping that could still make a forced poll skipped so I moved to the more explicit option, especially since there are very few use-cases for it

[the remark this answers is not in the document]
SugarDarius: I see ✅ , thanks for sharing this discussion 🙏🏻

packages/liveblocks-react/src/umbrella-store.ts
marcbouchenoire: We don't do this for other `applyOptimisticUpdates` methods so imo it would create more noise to introduce different patterns in the same file (I see these large files, like `room.ts(x)`, as good candidates to _not_ abstract too much in them) rather than follow the existing

marcbouchenoire: I should have said "misdirection" rather than "noise", I meant the fact that you can't read from top to bottom like the other `applyOptimisticUpdates` anymore which I personally like, that said I agree with the `shouldSubscribeToThread` one: I called it `isThreadParticipant` and kept it external since it's not specific to subscribing, and moved some things to make them slightly more readable: https://github.com/liveblocks/liveblocks/commit/e11bc474b73e84b0a2b33b1a45b7394f870a4f99

PR #2318, opened 2025-03-25
Rename “room notification settings” to “room subscription settings”

marcbouchenoire:
We have settled on the name “room **subscription** settings” to replace “room **notification** settings”, avoiding confusion with "**user** notification settings". 😵‍💫

This PR updates every instance of the old naming (hopefully, it's a lot 🥲) in the Liveblocks packages, in a non-breaking way (we'll remove the deprecated methods/hooks/types in 3.0). I also added codemods to do the rename automatically already.

~~I haven't made the change yet but I think we should rename public types like `UserNotificationSettings` to `NotificationSettings` at the same time (also in a non-breaking way of course), they were only named like that because of the confusion with "room notification settings", anyone opposed to that?~~ Done in https://github.com/liveblocks/liveblocks/pull/2318/commits/88ef29ce0d79e5900fc387c202faa36b24eb8490.

Documentation will be done in a separate PR.

Closes partly LB-1845 (the backend/endpoints still need to be handled)
Closes LB-1849


packages/liveblocks-core/src/types/RoomSubscriptionSettings.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export type RoomSubscriptionSettings = {
+  threads: RoomThreadsSubscriptionSettings;
+  textMentions: RoomTextMentionsSubscriptionSettings;
+};
+
+/**
+ * @deprecated Renamed to `RoomSubscriptionSettings`
+ */
+export type RoomNotificationSettings = RoomSubscriptionSettings;
```
SugarDarius, line 13: I'm wondering if it shouldn't be considered as a protocol like `InboxNotification`, `VersionHistory`, `Comments` or `UserNotificationSettings` no? 
In fact I'm struggling a bit to understand what are the things to put in `types` and the things to be put in `protocols`.

packages/liveblocks-react/src/room.tsx
```diff
@@ -2291,29 +2395,29 @@ function useHistoryVersionsSuspense(): HistoryVersionsAsyncSuccess {
...
+  // because changing the error type would be a breaking change
   const client = useClient();
   const room = useRoom();
   return useCallback(
-    (settings: Partial<RoomNotificationSettings>) => {
+    (settings: Partial<RoomSubscriptionSettings>) => {
       const { store, onMutationFailure } = getRoomExtrasForClient(client);
       const optimisticId = store.optimisticUpdates.add({
-        type: "update-notification-settings",
+        type: "update-subscription-settings",
```
SugarDarius, line 2412: I will to add the prefix `room` such as `update-room-subscription-settings` to clearly communicate the intent here. Just having `update-subscription-settings` is a bit too generic to me IMO and we could potentially think we're updating the settings for the user at project level. 

Fell to ignore my comment though, it's just me and my head who likes clean separation of concerns 😅

packages/liveblocks-core/src/types/LiveblocksError.ts
```diff
@@ -74,10 +74,15 @@ type CommentsOrNotificationsErrorContext =
         | "MARK_ALL_INBOX_NOTIFICATIONS_AS_READ_ERROR"
         | "DELETE_ALL_INBOX_NOTIFICATIONS_ERROR";
     }
+  // TODO: Deprecated, remove this once "room notification settings" hooks are removed
   | {
       type: "UPDATE_NOTIFICATION_SETTINGS_ERROR";
       roomId: string;
     }
+  | {
+      type: "UPDATE_SUBSCRIPTION_SETTINGS_ERROR";
```
SugarDarius, line 83: Same here I would tend to use the term `room` such as `UPDATE_ROOM_SUBSCRIPTION_SETTINGS_ERROR` to well separate between what's about a user (project level) and what's about a room (room level).

SugarDarius (approved):
Looking good ✅ 

> I haven't made the change yet but I think we should rename public types like UserNotificationSettings to NotificationSettings at the same time (also in a non-breaking way of course), they were only named like that because of the confusion with "room notification settings", anyone opposed to that?

In the end I will still keep the term `user` to keep a clean separation of concerns. It helps a lot to quickly identify the subject of the settings. Also in the end after some I think the hooks aren't using the good naming. We should use `useUserNotificationSettings` instead of `useNotificationSettings` because otherwise it's too wide and not explicit enough from a developer standpoint even if `NotificationSettings` is shorter. 
But I can overthink here 🤔 Happy to have any other feedback about it.

packages/liveblocks-core/src/types/RoomSubscriptionSettings.ts
marcbouchenoire: I'm not entirely sure either, but it comes from the backend so I'm moving it to `protocol`.

marcbouchenoire:
> In the end I will still keep the term user to keep a clean separation of concerns. It helps a lot to quickly identify the subject of the settings. Also in the end after some I think the hooks aren't using the good naming. We should use useUserNotificationSettings instead of useNotificationSettings because otherwise it's too wide and not explicit enough from a developer standpoint even if NotificationSettings is shorter.
But I can overthink here 🤔 Happy to have any other feedback about it.

@SugarDarius I thought that was one of the goal of the rename and imo it's no longer a problem after it. I'm biased because I don't like the `User*` naming convention client-side (both "room subscription settings" and "user notification settings" are for a specific user, I personally don't see how one needs it but not the other (but it does make sense with endpoints and in the backend)), the only other place where we have it is `useUserThreads` and I think we should try to find a clearer API for it when we make it public, without "user" in the name.

packages/liveblocks-core/src/types/LiveblocksError.ts
marcbouchenoire: I agree that the `roomId` key wasn't enough to make the distinction when the name was `UPDATE_NOTIFICATION_SETTINGS_ERROR` but imo after the rename it's no longer a problem.

SugarDarius:
> I don't like the User* naming convention client-side (both "room subscription settings" and "user notification settings" are for a specific user, I personally don't see how one needs it but not the other (but it does make sense with endpoints and in the backend)), the only other place where we have it is useUserThreads and I think we should try to find a clearer API for it when we make it public, without "user" in the name.

@marcbouchenoire agreed 👍🏻 
I just thought about it again and in the end even if `useUserNotificationSettings` is the most explicit, `useNotificationSettings` stays clear, concise and directly states what it manages as it implies a user context (notification settings are typically per user). So let's keep it that way and rename the public types in the protocol. 

LMK if you need insight or help from me for this.`UserNotificationSettings` is used in a lot of places so do not hesitate 🙂

ofoucherot (approved):
🚀

PR #2336, opened 2025-04-01
Add subscriptions to store

marcbouchenoire:
This PR adds (thread) subscriptions to the umbrella store. Non-store changes are/will be done in https://github.com/liveblocks/liveblocks/pull/2311.

Thread subscriptions are returned from the existing threads/inbox notifications `c` endpoints (single, paginated, and deltas), and they can be updated by subscribe/unsubscribing from individual threads, or by changing your room subscription settings.

[Backend counterpart PR](https://github.com/liveblocks/liveblocks-backend/pull/937)


packages/liveblocks-react/src/ThreadDB.ts
```diff
@@ -182,15 +182,17 @@ export class ThreadDB<M extends BaseMetadata> {
...
   ): ThreadData<M>[] {
     const index = direction === "desc" ? this.#desc : this.#asc;
     const crit: ((thread: ThreadData<M>) => boolean)[] = [];
     if (roomId !== undefined) {
       crit.push((t) => t.roomId === roomId);
     }
-    crit.push(makeThreadsFilter(query));
+    if (query !== undefined) {
+      crit.push(makeThreadsFilter(query));
+    }
```
marcbouchenoire, line 195: Some `findMany` use-cases don't need advanced filtering so it's now optional.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1847,6 +1937,55 @@ function applyOptimisticUpdates_forRoomSubscriptionSettings(
   return roomSubscriptionSettingsByRoomId;
 }
 
+/**
+ * Applies optimistic updates to subscriptions in a stable way.
+ */
+function applyOptimisticUpdates_forSubscriptions(
```
marcbouchenoire, line 1943: I left some `TODO`s for non-implemented optimistic updates that I think we could stay without given how non-critical they are.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -985,6 +1054,15 @@ export class UmbrellaStore<M extends BaseMetadata> {
...
     );
 
+    const threadSubscriptions = DerivedSignal.from(
+      threadifications,
+      this.subscriptions.signal,
+      (t, s) => ({
+        subscriptions: s,
+        ...t,
+      })
+    );
```
marcbouchenoire, line 1064: We don't have generic subscriptions yet, only `thread`s, and the only entry points for them are threads/inbox notifications endpoints so we're piggy-backing on `threadifications` changes.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1992,17 +1999,33 @@ function useSubscribeToRoomThread(roomId: string) {
...
+      const optimisticId = store.optimisticUpdates.add({
+        type: "subscribe-to-thread",
+        threadId,
+        subscribedAt,
+      });
 
       client[kInternal].httpClient.subscribeToThread({ roomId, threadId }).then(
         () => {
-          // TODO: Apply result of subscribeToThread
+          // TODO: subscribeToThread returns the thread, should it return the subscription (we could use it here instead of using subscribedAt)?
```
marcbouchenoire, line 2013: The `c` endpoints for (un)subscribing currently return the thread (like the non-`c` ones), but the returning thread doesn't contain any info about the newly-created subscription so it can't be applied as truth. Currently I just use the optimistic date as truth if it's a success.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -579,6 +606,14 @@ export type CleanNotifications = {
   notificationsById: Record<string, InboxNotificationData>;
 };
 
+export type CleanThreadSubscriptions = {
+  /**
+   * Thread subscriptions by key (kind + subject ID).
+   * e.g. `thread:${string}`, `$custom:${string}`, etc
+   */
+  subscriptions: SubscriptionsByKey;
+} & CleanNotifications;
```
nvie, line 615: I really dislike the `&` operator in TypeScript for how opaque it makes type definitions. It barely stands out from looking at this type definition here that apparently on CleanThreadSubscriptions there also exist `sortedNotifications` and `notificationsById` fields. Would you consider duplicating these field definitions here, for readability/transparency?

nvie (approved):
Most excellent work, @marcbouchenoire! 😍

packages/liveblocks-react/src/room.tsx
```diff
@@ -1992,17 +1999,33 @@ function useSubscribeToRoomThread(roomId: string) {
...
+          store.createSubscription(
+            {
+              kind: "thread",
+              subjectId: threadId,
+              createdAt: subscribedAt,
```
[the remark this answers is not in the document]
marcbouchenoire: At the next thread/inbox notification fetch (any of them: delta, paginated, etc) yeah

packages/liveblocks-react/src/umbrella-store.ts
marcbouchenoire: Good point, I can also use this as an opportunity to comment on why do we even have notifications here

SugarDarius (approved):
🚀 ✅

PR #2356, opened 2025-04-09
Check if body has element before removing child

nperez0111:
When unmounting in BlockNote, we remove the elements from React in a different order than Tiptap.
This caused an error to be thrown when trying to remove the file from the document's body (the element is already unmounted at that point).
Which was throwing this error: `Failed to execute 'removeChild' on node`

The resolution is pretty straight forward though, just check whether the element exists in the body, before trying to remove it from the DOM.

This issue was found in our discord: https://discord.com/channels/928190961455087667/1359077522632999065/1359077522632999065

### DIFF
--- a/packages/liveblocks-react-tiptap/src/mentions/MentionExtension.ts
+++ b/packages/liveblocks-react-tiptap/src/mentions/MentionExtension.ts
@@ -205,7 +205,9 @@ export const MentionExtension = Extension.create<MentionExtensionOptions>({
             },
 
             onExit() {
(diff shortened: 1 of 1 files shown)


raghav-thirdaiautomation:
hi @nperez0111, waiting for PR merge...
thanks

raghav-thirdaiautomation:
hi @jrowny @nimeshnayaju,
we are using live block pro and waiting for this fixes, can you please check this fixes?? 
Thank you

jrowny (approved):

PR #2361, opened 2025-04-11
Use declaration maps for improved local development

marcbouchenoire:
This PR aims to allow navigating between the various `packages` monorepo projects by using TS' [declaration maps](https://www.typescriptlang.org/tsconfig/#declarationMap).

1. [tsup doesn't support them](https://tsup.egoist.dev/#generate-typescript-declaration-maps--d-ts-map) 😢
2. The declaration maps shouldn't be shipped in the production builds since we don't ship the `src` folders anyway

### tsup

When generating **production** builds: no changes, the built-in `dts` option is used.

When generating **local** builds: the `dts` option and code-splitting are disabled, and we manually run `tsc` to output declarations and their declaration maps. The `dts` option outputs **bundled** `.d.ts` **and** `.d.cts` files, while `tsc` keeps the same file structure and only outputs `.d.ts` → it shouldn't change anything in local development except for running ["are the types wrong"](https://arethetypeswrong.github.io/) in the `lint:package` scripts locally.

### Rollup

Not done yet.

### Questions

- Is piggy-backing on `NODE_ENV` really a good idea?

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -89,7 +89,7 @@ jobs:
 
       - name: Build all packages
         run: |
-          npm run build -- --filter='./packages/*'
+          NODE_ENV="production" npm run build -- --filter='./packages/*'
 
       - name: Publish packages
         env:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,7 +55,7 @@ jobs:
         uses: oven-sh/setup-bun@v1
 
       - name: Build
-        run: npm run build -- --filter ${{ matrix.pkg }}
+        run: NODE_ENV="production" npm run build -- --filter ${{ matrix.pkg }}
 
       - name: Ensure no files are changed
         run: |
--- a/packages/liveblocks-client/settings.json
+++ b/packages/liveblocks-client/settings.json
@@ -1,6 +1,0 @@
-{
-    "editor.defaultFormatter": "esbenp.prettier-vscode",
-    "editor.formatOnSave": true,
-    "editor.tabSize": 2,
-    "editor.detectIndentation": false
-}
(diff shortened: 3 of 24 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -89,7 +89,7 @@ jobs:
 
       - name: Build all packages
         run: |
-          npm run build -- --filter='./packages/*'
+          NODE_ENV="production" npm run build -- --filter='./packages/*'
```
nvie, line 92: Not really sure how I feel about relying on NODE_ENV for this.

Would it be an option to simply always let the declaration maps be generated (also in production), but simply not include them in the NPM bundle by excluding their file extension?

```
  "files": [
    "dist/**",
    "!**.d.ts.map",  👈
    "README.md"
  ],
```

nvie (approved):
Great idea, thanks for working on this! 😍

.github/workflows/publish-packages.yml
nvie: This exclude pattern works, I just tested it locally 👍

marcbouchenoire: The biggest reason is that I'm not 100% confident that the different `.d.ts` outputs don't lead to differences, but if they don't, I'll find a way to duplicate them as `.d.cts` and I agree having a single path would be easier and more predictable

marcbouchenoire: I tried to simplify things with the "single output" option but I think I've reached a dead-end: generating the `.d.ts` files with `tsc` separately from the code leads to different file structures and publint links that to an "Internal resolution error" with `node16` resolution (see image in PR description) 😢 Fixing this seems like a rabbit hole I don't want to go into atm but there's still the option of two outputs:
- local: break `node16` resolution but gain declaration maps
- production: same output as before this PR
And we could use something else than `process.env.NODE_ENV` to switch between local and production

PR #2363, opened 2025-04-14
Subscribe/unsubscribe: `@liveblocks/node` methods

marcbouchenoire:
This PR adds `@liveblocks/node` methods for the new APIs added with subscribe/unsubscribe:

- [x] `client.subscribeToThread`
- [x] `client.unsubscribeFromThread`
- [x] `client.getThreadSubscriptions`

I'll do the docs afterwards in https://github.com/liveblocks/liveblocks/pull/2323, and I haven't tackled LB-1857, we can do it in a dedicated PR.

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -44,6 +44,8 @@ import type {
   ThreadDataPlain,
   ToImmutable,
   URLSafeString,
+  UserSubscriptionData,
+  UserSubscriptionDataPlain,
 } from "@liveblocks/core";
 import {
   checkBounds,
@@ -53,6 +55,7 @@ import {
   convertToInboxNotificationData,
   convertToSubscriptionData,
   convertToThreadData,
+  convertToUserSubscriptionData,
   createManagedPool,
   createNotificationSettings,
   LiveObject,
@@ -1385,6 +1388,39 @@ export class Liveblocks {
       throw await LiveblocksError.from(res);
     }
     return (await res.json()) as Promise<ThreadParticipants>;
+  }
+
+  /**
+   * Gets a thread's subscriptions.
+   *
+   * @param params.roomId The room ID to get the thread subscriptions from.
+   * @param params.threadId The thread ID to get the subscriptions from.
+   * @param options.signal (optional) An abort signal to cancel the request.
+   * @returns An array of subscriptions.
+   */
+  public async getThreadSubscriptions(
+    params: { roomId: string; threadId: string },
+    options?: RequestOptions
+  ): Promise<{ data: UserSubscriptionData[] }> {
+    const { roomId, threadId } = params;
+
+    const res = await this.#get(
+      url`/v2/rooms/${roomId}/threads/${threadId}/subscriptions`,
+      undefined,
+      options
+    );
+
+    if (!res.ok) {
+      throw await LiveblocksError.from(res);
+    }
+
+    const { data } = (await res.json()) as {
+      data: UserSubscriptionDataPlain[];
+    };
+
+    return {
+      data: data.map(convertToUserSubscriptionData),
+    };
   }
 
   /**
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -38,6 +38,8 @@ import type {
   RoomSubscriptionSettings,
   SerializedCrdt,
   StorageUpdate,
+  SubscriptionData,
+  SubscriptionDataPlain,
   ThreadData,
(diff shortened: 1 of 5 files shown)


marcbouchenoire:
> Do you think there is a potential for confusion here?

It can be confusing yeah I agree, but on the backend I wasn't a fan of "delete subscription" to mean "unsubscribe" and I think I feel the same for "create subscription", they feel too internal to me. I'm also not a fan of them not matching the client/React APIs. 🤔

nvie:
Makes sense, as long as it had been considered and the new method names were chosen deliberately, I'm on board with this!

packages/liveblocks-node/src/client.ts
```diff
@@ -1384,6 +1390,39 @@ export class Liveblocks {
...
+   *
+   * @param params.roomId The room ID to get the thread subscriptions from.
+   * @param params.threadId The thread ID to get the subscriptions from.
+   * @param options.signal (optional) An abort signal to cancel the request.
+   * @returns An array of subscriptions.
+   */
+  public async getThreadSubscriptions(
+    params: { roomId: string; threadId: string },
+    options?: RequestOptions
+  ): Promise<{ data: UserSubscriptionData[] }> {
```
nvie, line 1404: Will this method (ever) need pagination?

packages/liveblocks-node/src/client.ts
```diff
@@ -1612,6 +1651,59 @@ export class Liveblocks {
...
+    const res = await this.#post(
+      url`/v2/rooms/${roomId}/threads/${threadId}/unsubscribe`,
+      { userId: params.data.userId },
+      options
+    );
+
+    if (!res.ok) {
+      throw await LiveblocksError.from(res);
+    }
+  }
```
nvie, line 1705: It's a good practice to call `res.json()` even if you don't need to return anything in this `void` function.

The reason is that `res.ok` only checks the status header, and it does not start consuming the response body. This might cause the server to hold on to the response body stream longer than needed, because it hasn't been consumed by the client yet.

(I haven't checked the backend implementation, but if you haven't defined a response body in the backend for this endpoint, it might be good to at the very least return a minimal `{"ok":true}` JSON body response there. Ideally, _all_ endpoint responses are JSON.)

nvie (approved):
Left only some minor comments, but this is looking great! 👌

packages/liveblocks-node/src/client.ts
marcbouchenoire: We'll likely add it post-launch, and the `{ data }` shape doesn't require breaking changes

packages/liveblocks-node/src/client.ts
nvie: Yeah, ideally we stop using empty 204 responses entirely in the backend. It makes it a backward incompatible change if we need to change it. I have been through that a couple times and think at this point that every endpoint should return a JSON object in the body response, no exception. It just makes it a lot easier to extend later. cc @FlowFlorent

FlowFlorent: Traditionally our DELETEs do not return the resource so that's why it feels ok to use 204. But honestly returning a 200 so that we can extend it later sounds like a lot easier time, on board with that!

FlowFlorent (approved):

FlowFlorent:
The more I think about it the more I am leaning towards `createThreadSubscription` and `deleteThreadSubscription`

otherwise I wonder if we should update the endpoints to be:

POST ..../subscribe
POST ..../unsubscribe

And thus it'd make sense to have `unsubscribe` and `subscribe` as method names.

@nvie thanks for your suggestion!

marcbouchenoire:
> otherwise I wonder if we should update the endpoints to be:
>
> POST ..../subscribe
> POST ..../unsubscribe

@FlowFlorent Not sure I see what you mean, the endpoints are currently `/subscribe` and `/unsubscribe`

My preference is still on `subscribe`/`unsubscribe`, when we'll have generic subscriptions we'll probably go with `createSubscription` and `deleteSubscription` for lack of a better generic name but that means we would have `createSubscription` and `createThreadSubscription` which would just do `createSubscription({ kind: "thread", ... })`, `subscribeToThread` does the same but imo the naming helps it.

nvie:
@marcbouchenoire This makes a lot of sense, I like the latest naming you went for here 👍

packages/liveblocks-node/src/client.ts
marcbouchenoire: Are we okay with the inconsistency between this and all our other 204s? If so I'll update the endpoint and add `await res.json()` here.

PR #2370, opened 2025-04-16
Avoid using `Symbol.dispose`

nvie:
Fixes an issue where [some browsers have not implemented support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#:~:text=11-,dispose,-Experimental) for this well-known symbol yet.

This could lead to runtime issues like this in some browsers:
image (2)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,15 @@
 
 ## v2.23.1
 
+### `@liveblocks/client`
+
+- Fix potential runtime error in browsers that do not support `Symbol.dispose`
+  yet.
+
 ### `@liveblocks/node`
 
-- Fix a bug in `.mutateStorage()` and `.massMutateStorage()` where mutating storage could potentially corrupt the storage tree.
+- Fix a bug in `.mutateStorage()` and `.massMutateStorage()` where mutating
+  storage could potentially corrupt the storage tree.
 
 ## v2.23.0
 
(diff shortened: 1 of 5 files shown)


marcbouchenoire (approved):

marcbouchenoire:
@nvie Maybe you could update `CHANGELOG.md` in this PR already since we want to release this as `2.23.1` today

SugarDarius (approved):
FYI If you don't want to get completely rid of it and keep using `Symbol.dispose` we could add our own polyfill: 
```tsx
function setSymbolPolyfillWhenRequired(name: string) {
  if (!Object.getOwnPropertyDescriptor(Symbol, name)) {
    Object.defineProperty(Symbol, name, {
      value: Symbol(`liveblocks:symbol:${name}`),
      configurable: false,
      enumerable: false,
      writable: false,
    });
  }
}

setSymbolPolyfillWhenRequired('dispose')
setSymbolPolyfillWhenRequired('asyncDispose')

const foo = { [Symbol.dispose]: () => {console.log("bar") }}
foo[Symbol.dispose]() // 👉🏻 logs `bar`
```

nvie:
> Maybe you could update `CHANGELOG.md` in this PR already since we want to release this as `2.23.1` today

Yes, I will update the CHANGELOG, after #2372 is merged into `main`.

nvie:
> FYI If you don't want to get completely rid of it and keep using `Symbol.dispose` we could add our own polyfill:

We did not really need the actual use of `Symbol.dispose` yet. I used it just so that if you would do `using room = createRoomSomehow()` at one point, it would just work, but it was obviously too soon. We can always change it again later. Until then, YAGNI.

PR #2372, opened 2025-04-17
Fix: ensure unique session is always assigned correctly

nvie:
This fixes an issue with the newly introduced `.mutateStorage()` and `.massMutateStorage()` APIs, where the transient session's unique actor ID wasn't read correctly, and which could lead to storage corruptions.


marcbouchenoire (approved):
🫣

SugarDarius (approved):

PR #2376, opened 2025-04-17
Add `getUserRoomSubscriptionSettings` method

FlowFlorent:
### Description

Introduces `getUserRoomSubscriptionSettings`, a paginated endpoints that lists a user's room subscription settings.

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -1852,7 +1852,7 @@ export class Liveblocks {
   }
 
   /**
-   * Returns the room subscription settings for a user.
+   * Returns all room subscription settings for a user.
    * @param params.userId The user ID to get the room subscription settings from.
    * @param options.signal (optional) An abort signal to cancel the request.
    */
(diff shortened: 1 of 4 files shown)


packages/liveblocks-node/src/client.ts
```diff
@@ -1851,6 +1851,41 @@ export class Liveblocks {
     return this.getRoomSubscriptionSettings(params, options);
   }
 
+  /**
+   * Returns the room subscription settings for a user.
```
marcbouchenoire, line 1855:
```suggestion
   * Returns all room subscription settings for a user.
```

packages/liveblocks-node/src/client.ts
```diff
@@ -1851,6 +1851,41 @@ export class Liveblocks {
...
+    params: {
+      userId: string;
+    },
+    options?: RequestOptions
+  ): Promise<
+    Page<
+      {
+        roomId: string;
+      } & RoomSubscriptionSettings
+    >
```
marcbouchenoire, line 1869: I don't think we need to export it (unless someone asks) but we could make this a named type `UserRoomSubscriptionSettings` like `ThreadParticipants`

PR #2379, opened 2025-04-23
Unintuitive behavior around calling .undo() while history is paused?

nvie:
Added this test case to show the current behavior of `.undo()` while history is paused, which might be unintuitive. The "solution" is to always call `.resume()` right before `.undo()`, but I'm considering if we should just make this the default behavior instead of making users responsible for this.

For example, take this sequence / undo stack:
- Change A
- Change B
- `.pause()`
- Change C1
- Change C2
- `.undo()`

Calling the `.undo()` here will currently first "wipe" the C1 and C2 changes, then undo B, resulting in A. I think this is a little unintuitive at least. Is there an actual use case for this behavior, or was it an oversight?

If we'd change `.undo()` to always call `.resume()` first, the behavior would be that the changes since the last `.pause()` would be undone instead, resulting in B.

Other edge cases would also be fine I think:
- Calling `.resume()` while there is no paused history is a no-op
- Calling `.resume()` while there is a paused history but it does not contain any changes will also be a no-op. In that case, it would also still undo B.

Thoughts?

### DIFF
--- a/packages/liveblocks-core/src/__tests__/room.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.test.ts
@@ -1177,6 +1177,23 @@ describe("room", () => {
     expect(room.getPresence()).toEqual({ x: 0 });
   });
 
+  test("undoing while history is paused", () => {
+    const { room } = createTestableRoom({});
+
+    room.updatePresence({ x: 0 }, { addToHistory: true });
+    room.updatePresence({ x: 1 }, { addToHistory: true });
+    expect(room.getPresence()).toEqual({ x: 1 });
+
+    room.history.pause();
+    room.updatePresence({ x: 2 }, { addToHistory: true });
+    expect(room.getPresence()).toEqual({ x: 2 });
+
+    room.history.undo();
+
+    expect(room.getPresence()).toEqual({ x: 0 });
+    //                                      ^ NOTE: Without the .pause() call, this would be 1
+  });
+
   test("undo redo with presence that do not impact presence", () => {
     const { room } = createTestableRoom({});
     // room.connect();  // Seems not even needed?


SugarDarius:
From my understanding the actual behavior is correct and legit. 
If you pause the history after `Change B` and then call `.undo()` I would not be surprise to get a `Change A` result.

Pausing the history means "after pausing no matter what happens I do not want to store any changes in the history stack". So the undo behavior is doing its job correctly. 

If we're  calling`.resume()` right before calling `.undo()` then IMO I do not see the point of calling `.pause()`  because  the undo action will just wipe out all the changes you made since calling `.undo()` which is not in reality an undo action. It's just a replace/reset action. You're putting back the history right like it was before pausing. 

In fact the case you're describing is a kinda `freeze` action in the history. If I want to get back to `Change B` after `Change C1` and `Change C2` I would rather call a .freeze()` method such as:
```tsx
// → Change A
// → Change B
room.history.freeze()
  // 👉🏻 Like a `.pause()` but when calling `.undo()` after it will not result as `Change A`
// → Chance C
// → Change D
room.history.undo()
  // 👉🏻 When called it will result as `Change B`. 
```
At least is how I see the thing 😅 And also I strongly think it's an explicit developer choice to make, not us to implement it as a default behavior under the hood.

nvie:
> Pausing the history means "after pausing no matter what happens I do not want to store any changes in the history stack". So the undo behavior is doing its job correctly.

FWIW, this is not actually what is happening. It's not _ignoring_ those changes. It's _buffering_ / _grouping_ all changes after `.pause()` — which is why I think `pause()` is a bit of a misnomer. Because when you call `.resume()` all changes will still get added to the undo stack, just as one entry instead of many.

nvie:
I already started thinking holistically about a better undo/redo API, which should tackle all pieces of feedback we have received over the years about the current shortcomings. Ideally we can tackle this in a fully backward-compatible way, but also offering better primitives. I will polish my notes and write this up in a proposal / tech design soon (probs when I'm back from my holiday).

PR #2396, opened 2025-04-29
never skip applying updates

jrowny:
Fix a bug where clients could become desynchronized after an id collision. 

This fix removes code that was an optimization to avoid applying an update that was already applied, but the optimization is moot because yjs applyUpdate already handles this situation just fine.

This does not fix the id collision, but the result is now only used for the status hooks to track pending state. This is a less bad bug than desync, so I'm pushing this fix while we rethink the way pending state works.


nimeshnayaju (approved):
🚀

PR #2408, opened 2025-05-06
Improvements for read-only users in `Thread` and `Comment`

marcbouchenoire:
We weren't reading permissions in some UI parts within the default Comments components that are permission-based, so interacting with them led to an optimistic update immediately reverted (since the endpoints do read permissions):
- `Thread`: Toggling the resolved status of a thread
- `Comment`: Adding/updating reactions

### After

Buttons are removed (adding a reaction), unless they convey something useful for read-only users (seeing the reactions or the resolved status), in which case they're only disabled.

### DIFF
--- a/packages/liveblocks-react-ui/src/components/Thread.tsx
+++ b/packages/liveblocks-react-ui/src/components/Thread.tsx
@@ -1,14 +1,16 @@
 "use client";
 
-import type {
-  BaseMetadata,
-  CommentData,
-  DM,
-  ThreadData,
+import {
+  type BaseMetadata,
+  type CommentData,
+  type DM,
+  Permission,
+  type ThreadData,
 } from "@liveblocks/core";
 import {
   useMarkRoomThreadAsResolved,
   useMarkRoomThreadAsUnresolved,
+  useRoomPermissions,
   useRoomThreadSubscription,
 } from "@liveblocks/react/_private";
 import * as TogglePrimitive from "@radix-ui/react-toggle";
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react-ui/src/components/Thread.tsx
```diff
@@ -241,6 +243,13 @@ export const Thread = forwardRef(
       }
     }, [unreadIndex]);
 
+    const permissions = useRoomPermissions(thread.roomId);
+    const canComment =
+      permissions.size > 0
+        ? permissions.has(Permission.CommentsWrite) ||
+          permissions.has(Permission.Write)
+        : true;
```
SugarDarius, line 251: NB: I would tend to create a kind of helper to mutualize the code even it's a small check. I find it relevant to have the same source of truth. 
It could take the form of a new private hook such as roughly:
```tsx
const { 
  canWrite, 
  canRead, 
  canWriteComment, 
  canReadComment,
  canWritePresence 
} = useCheckRoomPermissions()
```
It could handle at the same time the default and the granular permissions:
```tsx
export function useCheckPermissions() {
  const permissions = useRoomPermissions()
  // Case for no permissions defined
  if (permissions.size <= 0) {
    return {
      /* ... */
    }
  }
  
  // Compute granular permissions
  return {
    /* ... ... */
  }
}
```
WDYT? No blocking to though 🙂

SugarDarius (approved):

nimeshnayaju (approved):

packages/liveblocks-react-ui/src/components/Thread.tsx
marcbouchenoire: I saw different logic in https://github.com/liveblocks/liveblocks/blob/13f3fe51252ad9aeab4748b2ae4e2437ed7d4368/packages/liveblocks-react-ui/src/components/Composer.tsx#L699 and some movement on it in the `copilot-chat` branch so I thought it was too early for a common abstraction. Ideally we can do this later 👍

PR #2416, opened 2025-05-20
Change `LiveblocksUIConfig` casing

marcbouchenoire:
I noticed that we didn't stick to Pascal case for this niche context provider, with 3.0 and all the upcoming `Ai*` things, it feels like a good time to fix that. Or do you think the breaking change is not worth it? If it was a main API I would vote for keeping the inconsistency to lower the upgrade cost, but this one being really niche/advanced makes me think it's fine.

### DIFF
--- a/docs/pages/api-reference/liveblocks-react-ui.mdx
+++ b/docs/pages/api-reference/liveblocks-react-ui.mdx
@@ -2140,16 +2140,16 @@ import { Icon } from "@liveblocks/react-ui";
 Find a full list of available icons in
 [our GitHub repo](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react-ui/src/icon.ts).
 
-#### LiveblocksUIConfig
+#### LiveblocksUiConfig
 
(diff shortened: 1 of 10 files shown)


SugarDarius:
FWIW to me the casing for this case is correct. `UI` represents two distinct words `user+interface` so it's fine for me to keep `LiveblocksUIConfig`.

marcbouchenoire:
> FWIW to me the casing for this case is correct. `UI` represents two distinct words `user+interface` so it's fine for me to keep `LiveblocksUIConfig`.

It's the same with `AiChat` and `prepareThreadNotificationEmailAsHtml` though, they're also acronyms of real words. I think we initially chose this for readability (`prepareThreadNotificationEmailAsHTML` looks nicer because it's at the end but `AiChat` and `LiveblocksUiConfig` are easier to read when it's in the middle), and now imo we should be consistent.

SugarDarius:
Makes sense!

SugarDarius (approved):

CTNicholas (approved):

PR #2418, opened 2025-05-20
Implement tool-calling support on the client side

nvie:
This PR lands the client-side updates to match the [backend tool-calling support](https://github.com/liveblocks/liveblocks-backend/pull/1036). I've also added a light example to the todo app, which we can iterate on to improve the way the `execute` / `render` properties will work exactly.

What's **currently implemented**:
- [x] Data types & protocol is stable now
- [x] Support for tool registration via the `<AiChat tools={...} />` prop
- [x] Each tool definition should have an `execute`, or `render` function, or both
- [x] The abort button will show while the assistant message is in either `generating` or `awaiting-tool` status
- [x] Aborting an assistant message while in generating phase will work like before: failing the message itself
- [x] Aborting an assistant message while in awaiting-tool phase will abort all tool calls, _and_ fail the message itself
- [x] The `respond` function is passed to `render`

What's **not implemented yet** (can all be tackled later, no blockers to merge):
All moved to follow-up PRs:
- [ ] Multi-tab support isn't there yet. Currently not only the designated client will execute: any open tab will. If you have two tabs open, both will execute the side effect. This is TBD.
- [ ] Making these APIs nicer to work with.
- [ ] Better TS support via inferred `args` and `result` types.
- [ ] Any other outstanding XXX's for internal clean ups.

### DIFF
--- a/packages/liveblocks-core/src/ai.ts
+++ b/packages/liveblocks-core/src/ai.ts
@@ -13,6 +13,7 @@ import * as console from "./lib/fancy-console";
 import { isDefined } from "./lib/guards";
 import type { Json, JsonObject } from "./lib/Json";
 import { nanoid } from "./lib/nanoid";
+import type { Resolve } from "./lib/Resolve";
 import { shallow, shallow2 } from "./lib/shallow";
 import { batch, DerivedSignal, MutableSignal, Signal } from "./lib/signals";
 import { SortedList } from "./lib/SortedList";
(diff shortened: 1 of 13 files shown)


packages/liveblocks-core/src/ai.ts
```diff
@@ -678,17 +763,28 @@ export type Ai = {
...
     ): DerivedSignal<UiChatMessage[]>;
     getChatById: (chatId: string) => AiChat | undefined;
+
+    // XXX This "tool definition" registry should not be signal-based I think?
+    // XXX We should make it similar to "global knowledge"
     getToolDefinitionΣ(
       chatId: string,
       toolName: string
-    ): Signal<ClientToolDefinition | undefined>;
+    ): Signal<AiToolDefinition | undefined>;
```
nvie, line 787: @nimeshnayaju What do you think? I don't think this needs to be a signal, does it?

marcbouchenoire (approved):
The changes look right to me from a high-level view (I'll let others chime in for the details) 👍

nimeshnayaju (approved):

packages/liveblocks-core/src/ai.ts
nimeshnayaju: The reason I made it signal based is so that if the render component changed, the tool call ui in the assistant message also changes. If it is not defined as a signal and if during development, you change the tool call's render method, the UI doesn't change.

nvie: Ohh, I see. Hmm, I hadn't even considered that. Once we support APIs for easy tool registration, where you thinking to support something like this:

```tsx
function SomeComponent() {
  return (
    <Stuff>
      {condition
        ? <RegisterAiTool toolName="doSomething" render={Component1} />
        : <RegisterAiTool toolName="doSomething" render={Component2} />}
    </Stuff>
  );
}
```

And that simply by toggling the `condition`, all chats would re-render and would show either Component1 or Component2 where "doSomething" was called?

Maybe I don't see the use case for this yet.

nimeshnayaju: It is not just for this use case (which I agree isn't common) but for faster iteration during development. If you do not make it a signal, and say during development, you update the render component of a tool call, the updated UI won't be displayed until a re-render of `AssistantMessage` takes place. I added this a while ago, so I could have missed something then, so feel free to remove the signal subscription part and validate it if you'd like.

nvie: Hmm, I thought hot module reloading would take care of the re-rendering in those cases, but maybe I'm missing something.

nvie: I just tried changing it, and you're right! It no longer updates dynamically when we do this. I guess I misunderstood how HMR works internally in that case. Either way, it's clear to me now that we indeed need to keep this signal-based — thanks for adding this in, because it does indeed make the DX a lot better 👍

PR #2419, opened 2025-05-22
Implement `AiMessage.Content` primitive

nvie:
This new `<AiMessage.Content />` primitive is introduced as an internal API only for now, but could eventually be made public once we're happy with how it's working. We also reuse it now in the kitchen sink which allowed us to avoid quite a bit of duplication of imporant logic that needs to be correct for setting up the tool invocation context (e.g. how the `respond()` function gets computed and passed into tool invocation instances).

The `AiMessage.Content` component is a primitive, in that it does not interfere with styling. It's mainly used as a helper to loop over the message's `content` (or content-so-far) array, and rendering the provided parts with the provided component slots (e.g. `TextPart`, `UploadedImagePart` (eventually), `ReasoningPart`). The one missing part here that cannot directly displayed is the `ToolInvocationPart`: users will have to use a `render` message on a tool to define those UIs.

@marcbouchenoire Is this looking good to you?
@nimeshnayaju Feel free to take a stab at defining the "default" `TextPart`, etc components (which aren'[t currently implemented yet](https://github.com/liveblocks/liveblocks/blob/bc8d7a28d75f4a791e77180516f99f986d942eff/packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx#L27-L29), but which is fine for now as AiChat will provide explicit versions for now internally).

Thanks for all the help here! 🙏

### DIFF
--- a/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
+++ b/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
@@ -66,7 +66,7 @@ export const AiChatAssistantMessage = memo(
       ) {
         if (message.contentSoFar.length === 0) {
           children = (
-            <div className="lb-ai-chat-message-thinking lb-ai-chat-pending">
+            <div className="lb-ai-chat-message-thinking lb-ai-chat-streaming">
               {$.AI_CHAT_MESSAGE_THINKING}
             </div>
           );
@@ -133,13 +133,6 @@ function AssistantMessageContent({
   message: UiAssistantMessage;
   components: Partial<GlobalComponents> | undefined;
 }) {
-  const content = message.content ?? message.contentSoFar;
-
-  // A message is considered to be in "reasoning" state if it only contains reasoning parts and no other parts.
-  const isReasoning =
-    content.some((part) => part.type === "reasoning") &&
-    content.every((part) => part.type === "reasoning");
-
   return (
     <div className="lb-ai-chat-message-content">
       <AiMessage.Content
@@ -153,10 +146,10 @@ function AssistantMessageContent({
             />
           ),
 
-          ReasoningPart: ({ part }) => (
+          ReasoningPart: ({ part, isStreaming }) => (
             <ReasoningPart
               text={part.text}
-              isPending={isReasoning}
+              isStreaming={isStreaming}
               components={components}
             />
           ),
@@ -242,11 +235,11 @@ function CodeBlock({
  * -----------------------------------------------------------------------------------------------*/
 function ReasoningPart({
   text,
-  isPending,
+  isStreaming,
   components,
 }: {
   text: string;
-  isPending: boolean;
+  isStreaming: boolean;
   components: Partial<GlobalComponents> | undefined;
 }) {
   const [isOpen, setIsOpen] = useState(false);
@@ -259,10 +252,10 @@ function ReasoningPart({
       <CollapsiblePrimitive.Trigger
         className={classNames(
           "lb-collapsible-trigger",
-          isPending && "lb-ai-chat-pending"
+          isStreaming && "lb-ai-chat-streaming"
         )}
       >
-        {/* TODO: If `isPending` is true, show "Reasoning…"/"Thinking…", otherwise show "Reasoned/thought for x seconds"? */}
+        {/* TODO: If `isStreaming` is true, show "Reasoning…"/"Thinking…", otherwise show "Reasoned/thought for x seconds"? */}
         Reasoning
         <span className="lb-collapsible-chevron lb-icon-container">
           <ChevronRightIcon />
--- a/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
+++ b/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
@@ -66,7 +66,7 @@ export const AiChatAssistantMessage = memo(
       ) {
         if (message.contentSoFar.length === 0) {
           children = (
-            <div className="lb-ai-chat-message-thinking lb-ai-chat-streaming">
+            <div className="lb-ai-chat-message-thinking lb-ai-chat-pending">
               {$.AI_CHAT_MESSAGE_THINKING}
             </div>
           );
@@ -252,7 +252,7 @@ function ReasoningPart({
       <CollapsiblePrimitive.Trigger
         className={classNames(
           "lb-collapsible-trigger",
-          isStreaming && "lb-ai-chat-streaming"
+          isStreaming && "lb-ai-chat-pending"
         )}
       >
         {/* TODO: If `isStreaming` is true, show "Reasoning…"/"Thinking…", otherwise show "Reasoned/thought for x seconds"? */}
--- a/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
+++ b/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
@@ -161,6 +161,10 @@ function AssistantMessageContent({
               components={components}
             />
           ),
+
+          ToolInvocationPart: ({ children }) => (
+            <div className="lb-ai-chat-message-tool-invocation">{children}</div>
+          ),
         }}
       />
     </div>
--- a/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
+++ b/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
@@ -1,23 +1,10 @@
-import type {
-  AiAssistantContentPart,
-  AiToolDefinitionRenderProps,
-  AiToolInvocationPart,
-  Json,
-  MessageId,
-  UiAssistantMessage,
-} from "@liveblocks/core";
-import { kInternal } from "@liveblocks/core";
-import { useClient } from "@liveblocks/react";
-import { useSignal } from "@liveblocks/react/_private";
+import type { UiAssistantMessage } from "@liveblocks/core";
 import { Lexer } from "marked";
 import {
   type ComponentProps,
-  createContext,
   forwardRef,
   memo,
   type ReactNode,
-  useCallback,
-  useContext,
   useEffect,
   useMemo,
   useRef,
@@ -35,6 +22,7 @@ import {
   type GlobalOverrides,
   useOverrides,
 } from "../../overrides";
+import { AiMessage } from "../../primitives";
 import * as CollapsiblePrimitive from "../../primitives/internal/Collapsible";
 import {
   type BlockToken,
@@ -89,30 +77,21 @@ export const AiChatAssistantMessage = memo(
         } else {
           children = (
             <AssistantMessageContent
-              content={message.contentSoFar}
-              chatId={message.chatId}
-              messageId={message.id}
+              message={message}
               components={components}
             />
           );
         }
       } else if (message.status === "completed") {
         children = (
-          <AssistantMessageContent
-            content={message.content}
-            chatId={message.chatId}
-            messageId={message.id}
-            components={components}
-          />
+          <AssistantMessageContent message={message} components={components} />
         );
       } else if (message.status === "failed") {
         // Do not include the error message if the user aborted the request.
         if (message.errorReason === "Aborted by user") {
           children = (
             <AssistantMessageContent
-              content={message.contentSoFar}
-              chatId={message.chatId}
-              messageId={message.id}
+              message={message}
               components={components}
             />
           );
@@ -120,9 +99,7 @@ export const AiChatAssistantMessage = memo(
           children = (
             <>
               <AssistantMessageContent
-                content={message.contentSoFar}
-                chatId={message.chatId}
-                messageId={message.id}
+                message={message}
                 components={components}
               />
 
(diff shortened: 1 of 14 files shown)


packages/liveblocks-react-ui/src/primitives/index.ts
```diff
@@ -1,5 +1,7 @@
 export type { ComposerBodyMark, ComposerBodyMarks } from "../types";
 export * as AiChatComposer from "./AiChatComposer";
+// XXX Move into /internal instead of exposing it as a Primitive here right now?
+export * as AiMessage from "./AiMessage";
```
nvie, line 4: Any preferences for where to export this? It's a primitive, but also a private API, so I couldn't decide. cc @marcbouchenoire @nimeshnayaju

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+/* -------------------------------------------------------------------------------------------------
+ * ToolInvocationPart
+ * -----------------------------------------------------------------------------------------------*/
+// XXX Rename to AiToolInvocationContext?
+const AiToolDefinitionRenderContext =
+  createContext<AiToolDefinitionRenderProps | null>(null);
+
+// XXX Rename to useAiToolInvocationContext?
+export function useAiToolDefinitionRenderContext() {
+  const context = useContext(AiToolDefinitionRenderContext);
```
nvie, line 41: I haven't renamed these yet, wanted to get your opinion first, @marcbouchenoire.

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  AiMessageContentComponents,
+  AiMessageContentProps,
+} from "./types";
+
+const AI_MESSAGE_CONTENT_NAME = "AiMessageContent";
+
+const defaultMessageContentComponents: AiMessageContentComponents = {
+  TextPart: () => "Default TextPart is not implemented yet",
+  ReasoningPart: () => "Default ReasoningPart is not implemented yet",
+  UploadedImagePart: () => "Default UploadedImagePart is not implemented yet",
```
nvie, line 29: @nimeshnayaju Feel free to update these however you see fit. You can just push your changes to this PR if you'd like.

packages/liveblocks-react-ui/src/primitives/AiMessage/types.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface AiMessageContentComponents {
+  /**
+   * The component used to display text parts.
+   */
+  TextPart: ComponentType<{ part: AiTextPart }>;
+
+  /**
+   * The component used to display uploaded image parts.
+   */
+  UploadedImagePart: ComponentType<{ part: AiUploadedImagePart }>;
```
nvie, line 20: @nimeshnayaju Should we remove all "UploadedImage" stuff before the launch? I'd prefer to add it back in after the launch once we are ready to support it?

packages/liveblocks-react-ui/src/primitives/index.ts
marcbouchenoire: I personally prefer from `_private` like we do for the `AiChatComposer`, the distinction between `@liveblocks/react-ui` and `@liveblocks/react-ui/primitives` is only to signal to users what kind of components they are importing (especially when we have a `Comment` **primitive** and a `Comment` **default component**), but imo we ourselves don't need it.

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
marcbouchenoire: I think I like `AiToolInvocationContext`/`useAiToolInvocationContext` but maybe `AiToolDefinitionRenderProps` should use the same terms then, `AiToolInvocationRenderProps` (or even `AiToolInvocationProps` since `execute` doesn't get "props" but just "args")?

packages/liveblocks-react-ui/src/components/internal/AiChatUserMessage.tsx
```diff
@@ -19,14 +20,14 @@ export interface AiChatUserMessageProps extends ComponentProps<"div"> {
   overrides?: Partial<GlobalOverrides>;
 }
 
+function PlainTextPart({ part }: { part: AiTextPart }) {
```
marcbouchenoire, line 23: Not now but when it will become public, each part's props type should be exported to avoid `{ part: AiTextPart }` in user code

packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
```diff
@@ -154,60 +131,41 @@ export const AiChatAssistantMessage = memo(
...
+              text={part.text}
+              components={components}
+              className="lb-ai-chat-message-text"
+            />
+          ),
+
+          ReasoningPart: ({ part }) => (
+            <ReasoningPart
+              text={part.text}
+              isPending={isReasoning}
```
marcbouchenoire, line 163: Could we give `isReasoning` (and in the future things like the time it took/is taking) as a prop?

```
ReasoningPart: ({ part, status }) => status === "reasoning" ? "Reasoning…" : "Reasoned"
```

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
nvie: Good point, and I agree 👍

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const props = {
+    ...rest,
+
+    // It only makes sense and is safe to call `respond()` in "executing" state.
+    respond: part.status === "executing" ? respond : noop,
+  };
+  return (
+    // XXX What to do about this style since this is a primitive?
+    // XXX Nimesh thinks to delete it
+    <div className="lb-ai-chat-message-tool">
```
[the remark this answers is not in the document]
nvie: OK, just wanted to confirm that with you. We were thinking the same. I'll remove it 👍

packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
marcbouchenoire: Yeah maybe not `status`, I do think the concept of "pending when it's the only part" could be first-class especially if/when it comes with duration info (`{ part, isPending, duration /* The reasoning total duration in ms, when pending and when done */ }`) but that not a requirement now.

For `isTheOnlyPart`, since this component does the looping, would it make sense to give `{ part, index, content }` similarly to `Array.map` to allow `isTheOnlyPart` and anything else?

nvie: Oh yes, I like passing `index` (and `duration`!) as well. Duration would require a bit more tracking, but it would indeed be pretty nice to have (for all parts, not just reasoning).

What do you think, @nimeshnayaju ?

packages/liveblocks-react-ui/src/primitives/index.ts
nimeshnayaju: Same! It's a private API for now, so I'd keep it in `_private`. The goal is to move it to `primitives` once it's public.

packages/liveblocks-react-ui/src/primitives/AiMessage/types.ts
nimeshnayaju: Yes! I'd like these removed before the launch. We still have few unknowns on attachments on both client and server side.

packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
nimeshnayaju: I can see how `index` makes sense - although if we decide to do grouping, not sure if the idea of index would still make sense. 

`isPending` makes a lot of sense too as it applies to all parts. 

`duration`, on the other hand, is more complex and would most definitely be a value generated on the server. It is also possible for the same assistant message to have interleaving reasoning and text parts (e.g. `thinking`, `text`, `thinking`), so I think this will require more thought.

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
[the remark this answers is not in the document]
nvie: In private DM on Slack, @marcbouchenoire and I talked a bit more about removing this, and it turns out removing the CSS for the margin would still be needed for our AiChat implementation, but not be a problem for end users (as they could always add any spacing/wrapping inside their own `render` implementation.

For now, we decided the following:

```tsx
components={{
  TextPart,
  ReasoningPart,

  /**
   * This part only will be marked @internal.
   */
  ToolInvocationPart: ({ part, children }) => {
    if (!children) {
      return null
    }
  
    // `children` is the result from the tool's `render` (with the context, the execute/respond logic, etc)
    return <div class="lb-...">{children}</div>
  }
}}
```

This will:
- Allow ourselves to still inject the `lb-ai-chat-message-tool` class name and apply the default margin there.
- Not be visible to end users, who can already control wrapping/margins.
- Default to `Fragment`.

cc @nimeshnayaju

packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
nvie: > _"I can see how index makes sense - although if we decide to do grouping, not sure if the idea of index would still make sense."_

Good thought. Maybe we need two indexes? A group index, and an index within the group (starting at 0 in each group).

nvie: > _"duration, on the other hand, is more complex and would most definitely be a value generated on the server. It is also possible for the same assistant message to have interleaving reasoning and text parts (e.g. thinking, text, thinking), so I think this will require more thought."_

Absolutely, we can come back to this idea later.

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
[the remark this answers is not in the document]
nimeshnayaju: I had a similar thought yesterday. I think this is fine for now but for further discussion - I am also wondering, from a customization point of view, if it makes sense to export a `ToolInvocationPart` component (privately, at first) because tool invocations require some internal APIs and are purely logic related code and let users render other message parts the way they wish and group however they want.

Internally we don't do anything fancy inside of `AiMessage.Content` component - we simply map over the parts and render default component - the main reason we have it now is for `ToolInvocationPart`.

cc: @marcbouchenoire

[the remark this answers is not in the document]
marcbouchenoire: With the other primitives' `components` we export what they do by default:
```
<Comment.Body
  components={{
    Mention: (props) => <Comment.Mention {...props} /> // This is changing nothing
  }}
/>
```

I could see the same for the parts:
```
<AiMessage.Content
  components={{
    TextPart: (props) => <AiMessage.TextPart {...props} /> // This is changing nothing,
    ToolInvocationPart: (props) => <div><AiMessage.ToolInvocationPart {...props} /></div> // Just adding a wrapper
  }}
/>
```

Which could allow using just the parts and loop yourself:
```
{m.parts.map(part => {
  // ...

  if (part.type === "tool-invocation") {
    return <div><AiMessage.ToolInvocationPart part={part} /></div>
  }
})}
```

One change would be that the `execute`/`respond`/`render`/etc logic would live in `<AiMessage.ToolInvocationPart />`. This is all hypothetical and for later though.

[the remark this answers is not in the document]
nvie: This idea is now implemented in ec1900d5016a2c5d6c55156ee3b0c9db86dc17b6.

packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
nvie: I've implemented the `isPending` prop, but named it `isStreaming` instead. While at it, I'm also passing down the `index` as a prop, in case the part wants it. See 207402cbd701f0411aa6cd34500f0d13e1fb9e24.

marcbouchenoire: I like it 👍 For `lb-ai-chat-pending`, this is a generic class that applies a shimmer effect so it can be used for anything loading/pending (for example we might use it on the title in `AiTool` if the spinner is too aggressive), I think it still makes sense to call it "pending" and not restricting it to "streaming"

nvie: The boolean should reflect that the part's data can still be expected to change (so indeed useful to implement the shimmer effect). I wouldn't mind going back to pending, but I wanted to kind of avoid it since messages can also be said to be "pending" if they are still in either "generating" or "awaiting-tool" state. And a tool invocation can be "pending" if it's in "receiving" or "executing" status. So my main worry is… is it going to be clear enough what it means for a part to be "pending"? Maybe we should flip the bool and call it `isFinalized`?

marcbouchenoire: Sorry I meant for the CSS class specifically, `isStreaming` makes sense on the JS side but this CSS class is used in other contexts so it's not tied to this streaming concept. Hopefully that makes sense, imagine a class named `.loading` which would make any element it's applied to fade in and out, we would use it in many contexts without naming it differently for each.

nvie: Ohhhh – whoops! My bad for renaming it. I wanted to make it consistent as I thoughtit was only used here. I'll rename the CSS class back then, and keep the rest 👍

PR #2421, opened 2025-05-26
Prepare `Markdown` and `Collapsible` primitives

marcbouchenoire:
Similarly to https://github.com/liveblocks/liveblocks/pull/2419, this PR pushes the `Markdown` and `Collapsible` primitives further to remove their duplicate implementations, but they're still in `_private` for now.

@nimeshnayaju You told me you felt comfortable with the `Markdown` one but imo `Collapsible` is ready too, let me know if not. Especially given that its API matches [Radix](https://www.radix-ui.com/primitives/docs/components/collapsible)/[Base UI](https://base-ui.com/react/components/collapsible), I only added support for uncontrolled/controlled modes.


packages/liveblocks-react-ui/src/primitives/Markdown.tsx
```diff
@@ -1,25 +1,105 @@
...
+import type { ComponentPropsWithSlot } from "../types";
+
 export type MarkdownComponents = {
-  CodeBlock: ComponentType<{ language: string | null; code: string }>;
-  Anchor: ComponentType<ComponentPropsWithoutRef<"a">> | "a";
+  CodeBlock: ComponentType<MarkdownComponentsCodeBlockProps>;
+  Link: ComponentType<MarkdownComponentsLinkProps>;
+  Heading: ComponentType<MarkdownComponentsHeadingProps>;
+  Image: ComponentType<MarkdownComponentsImageProps>;
+  Blockquote: ComponentType<MarkdownComponentsBlockquoteProps>;
```
marcbouchenoire, line 18: I prefer the opinionated/"editorialized" slots like `Heading` with a `level` prop instead of `h1`, `h2`, `h3`, etc, especially with cases like `CodeBlock` which could mix `pre`/`code`, curious to hear other thoughts on this.

packages/liveblocks-react-ui/src/primitives/Markdown.tsx
```diff
@@ -586,23 +627,18 @@ function InlineTokenComp({
         ));
       }
 
-      const Anchor = components?.Anchor ?? defaultComponents.Anchor;
+      const Link = components?.Link ?? defaultComponents.Link;
```
marcbouchenoire, line 630: I forgot the difference between "Anchor" and "Link" when I added this last time: "Anchor" is a way to override _any_/_all_ links in the **default components** (e.g. needing to use `next/link` instead of `a`), "Link" was used in the `Comment.Body` primitive to [allow customizing](https://liveblocks.io/docs/api-reference/liveblocks-react-ui#primitives-Comment.Body-Link) `CommentBody` links (e.g. style them, add tooltips, etc), the same exists for links inside `Composer.Editor`. Since the case of `Markdown` is the same as `Comment.Body`/`Composer.Editor`, I'm changing it to "Link".

packages/liveblocks-react-ui/src/primitives/Collapsible/index.tsx
```diff
@@ -149,3 +170,16 @@ export const Content = forwardRef<HTMLDivElement, ContentProps>(
     );
   }
 );
+
+if (process.env.NODE_ENV !== "production") {
+  CollapsibleContent.displayName = COLLAPSIBLE_CONTENT_NAME;
+  CollapsibleRoot.displayName = COLLAPSIBLE_ROOT_NAME;
+  CollapsibleTrigger.displayName = COLLAPSIBLE_TRIGGER_NAME;
+}
```
nimeshnayaju, line 178: @marcbouchenoire Why do we need these? My understanding was that if `displayName` is not set, React infers it from the name of the function that defines the component. In our case, we're okay with the function name being used for debugging purpose, no?

packages/liveblocks-react-ui/src/primitives/Collapsible/index.tsx
```diff
@@ -9,39 +9,58 @@ import {
...
     forwardedRef
   ) => {
+    const [isOpen, onOpenChange] = useControllableState(
+      // If the collapsible is neither controlled nor uncontrolled, it defaults to controlled as open.
+      controlledOpen === undefined && defaultOpen === undefined
+        ? true
+        : controlledOpen,
+      controlledOnOpenChange,
+      defaultOpen
+    );
```
nimeshnayaju, line 49: I'm okay to keep us but I try to not complicate the component logic with more things than needed. In this case, I'm okay with keeping the component simply uncontrolled because it makes the internal logic much simpler and the primitives as dumb as possible. I know `useControllableState` handles a lot of the complexity but it also adds an extra layer of code to manage and account for.

packages/liveblocks-react-ui/src/primitives/Markdown.tsx
```diff
@@ -1,25 +1,105 @@
...
+  Link: ComponentType<MarkdownComponentsLinkProps>;
+  Heading: ComponentType<MarkdownComponentsHeadingProps>;
+  Image: ComponentType<MarkdownComponentsImageProps>;
+  Blockquote: ComponentType<MarkdownComponentsBlockquoteProps>;
+
+  // Paragraph
+  // Inline (text, strong, em, code, del)
+  // Table
+  // List
+  // Separator (hr)
```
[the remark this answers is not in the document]
nimeshnayaju: Good point - tables and lists are complex. I haven't thought about it in detail but how does the following signature for `Table` look for you @marcbouchenoire ?

```ts
Table: ({ headings, rows }: { headings: ReactNode[]; rows: ReactNode[][] }) => {
  return (
    <table>
      <thead>
        <tr>
          {headings.map((heading, index) => (
            <th key={index}>{heading}</th>
          ))}
        </tr>
      </thead>
      <tbody>
        {rows.map((row, index) => {
          return (
            <tr key={index}>
              {row.map((cell, index) => (
                <td key={index}>{cell}</td>
              ))}
            </tr>
          );
        })}
      </tbody>
    </table>
  );
}
```

packages/liveblocks-react-ui/src/primitives/Markdown.tsx
```diff
@@ -168,47 +248,29 @@ export function BlockTokenComp({
...
-          break;
+      const Heading = components?.Heading ?? defaultComponents.Heading;
+
+      if (token.depth < 1 || token.depth > 6) {
+        return null;
       }
+
       return (
-        <HeadingTag>
+        <Heading level={token.depth as 1 | 2 | 3 | 4 | 5 | 6}>
```
nimeshnayaju, line 265: I didn't want to do a `as`, so I used a `switch-case` statement, but I am also okay with this as it might be more pragmatic.

packages/liveblocks-react-ui/src/primitives/Markdown.tsx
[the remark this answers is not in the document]
marcbouchenoire: Lovely, I think it'd be great!

packages/liveblocks-react-ui/src/primitives/Collapsible/index.tsx
marcbouchenoire: Would you say the same for a public API? It's still private for now but the goal is to have it public at some point, and even now I had in mind to add `defaultOpen`/`open`/`onOpenChange` to `AiTool` which is collapsible and maybe already public. imo requiring your own `useState` for a public API like this would be a bit much, supporting controlled/uncontrolled seems to be an expectation for stateful components (this one but also `Switch`, `Checkbox`, `Input`, etc) in "behavior" libraries like Radix.

packages/liveblocks-react-ui/src/primitives/Collapsible/index.tsx
marcbouchenoire: I thought it was because of `forwardRef` but at this point we're just moving these from primitive to primitive and I don't even know if it's needed anymore, plus we don't do it for the high-level components even though they're also using `forwardRef`. I'll test this today but I'll be happy to remove all this boilerplate from all primitives 👍

packages/liveblocks-react-ui/src/primitives/AiMessage/types.ts
```diff
@@ -25,7 +25,7 @@ type ReasoningPartProps = {
 };
 
 /** @internal */
-type ToolInvocationPartProps = {
+type AiMessageContentToolInvocationPartProps = {
```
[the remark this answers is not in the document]
marcbouchenoire: Yeah in this context it could be confusing 🤔 I've been adding these (and sometimes "Components" in "dot-less" cases like `Markdown`) just as a way to allow us to ship the defaults attached and not separately, because when we do this we need two props types. I don't think that's super clear so here's an example from `Comment.Body`:

```tsx
type CommentBodyMentionProps = {
  userId: string;
};

type CommentMentionProps = ComponentPropsWithSlot<"span">;

<Comment.Body
  components={{
    Mention: (props) => ...
    //        ^^^^^ CommentBodyMentionProps
  }}
/>

<Comment.Body
  components={{
    Mention: (props) => <Comment.Mention>...</Comment.Mention>
    //                   ^^^^^^^^^^^^^^^ This component is what's used by default if `Mention` isn't customized, it's a `span` and can receive `CommentMentionProps` (different from `CommentBodyMentionProps`)
  }}
/>
```

packages/liveblocks-react-ui/src/primitives/Collapsible/index.tsx
nimeshnayaju: For public API, I agree it makes sense to provide both controlled and uncontrolled props. If the goal is to make these public or use it directly on a public API, please feel free to ignore my comments! Thanks for explaining @marcbouchenoire.

packages/liveblocks-react-ui/src/primitives/Collapsible/index.tsx
marcbouchenoire: It's not exactly because of `forwardRef`, it's because of the anonymous `() => {}` function inside, if I remove the manual `displayName` then it turns into "Anonymous".

If I use a named function like `const AiMessageContent = forwardRef(function AiMessageContent() { ... })`, it does show up but its name is messed with by the bundler 😢



Any other ideas?

nimeshnayaju: 😵‍💫 Wow that is not ideal - thanks for looking into this! Thank god React 19 removes the need to use `forwardRef`

Issue #2425, opened 2025-05-26
Entry point for `_private`, `suspense` and `primitives` is esm vs cjs


paolobueno:
**Describe the bug**

The `package.json#main` field for `@liveblocks/react-ui/primitives`, `@liveblocks/react/_private` and `@liveblocks/react/suspense` are configured as `index.js` files and as such in our jest+babel setup we ran into either the `SyntaxError: Cannot use import statement outside a module` when directly importing them, or the [liveblocks module duplication](https://liveblocks.io/docs/errors/dupes) error if trying to transpile on the fly (by whitelisting via jest `transformIgnorePatterns`)

As a workaround setting the module entry points to the `index.cjs` files with the following diffs (via [patch-package](https://www.npmjs.com/package/patch-package))  made them work fine:

`@liveblocks+react+2.24.2.patch`
```patch
diff --git a/node_modules/@liveblocks/react/_private/package.json b/node_modules/@liveblocks/react/_private/package.json
index 2f1164e..ec64e2e 100644
--- a/node_modules/@liveblocks/react/_private/package.json
+++ b/node_modules/@liveblocks/react/_private/package.json
@@ -1,4 +1,4 @@
 {
-  "main": "../dist/_private.js",
+  "main": "../dist/_private.cjs",
   "types": "../dist/_private.d.ts"
 }
diff --git a/node_modules/@liveblocks/react/suspense/package.json b/node_modules/@liveblocks/react/suspense/package.json
index 737bae7..7d6a8d9 100644
--- a/node_modules/@liveblocks/react/suspense/package.json
+++ b/node_modules/@liveblocks/react/suspense/package.json
@@ -1,4 +1,4 @@
 {
-  "main": "../dist/suspense.js",
+  "main": "../dist/suspense.cjs",
   "types": "../dist/suspense.d.ts"
 }
```

`@liveblocks+react-ui+2.24.2.patch`
```patch
diff --git a/node_modules/@liveblocks/react-ui/primitives/package.json b/node_modules/@liveblocks/react-ui/primitives/package.json
index 0f9f66a..4eda174 100644
--- a/node_modules/@liveblocks/react-ui/primitives/package.json
+++ b/node_modules/@liveblocks/react-ui/primitives/package.json
@@ -1,4 +1,4 @@
 {
-  "main": "../dist/primitives/index.js",
+  "main": "../dist/primitives/index.cjs",
   "types": "../dist/primitives/index.d.ts"
 }
```

[...]

PR #2426, opened 2025-05-27
Use CJS outputs in nested package.json for subpath imports

marcbouchenoire:
This PR fixes https://github.com/liveblocks/liveblocks/issues/2425.

### Context

Some of our packages use subpath imports and to support them with TypeScript's `node` module resolution we added [nested `package.json` files](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react/suspense).

In 2.19, we moved our packages to `"type": "module"`, moving our `.js` outputs from CJS to ESM, but we still shipped CJS as `.cjs` correctly including for subpath imports, [via the `"exports"` field in our packages' root `package.json`](https://github.com/liveblocks/liveblocks/blob/1eb7d94e6355e093c39c53f5ffa22be9baa32e96/packages/liveblocks-react/package.json#L9-L43).

But it appears that some tools like Jest incorrectly read from the nested `package.json` files when dealing with subpath imports instead of reading `"exports"` in the root `package.json`. Since 2.19, these `package.json` files are pointing to ESM `.js` files and so they break in CJS environments.

### Fix

We update the nested `package.json` files to point to `.cjs` files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 ## v2.24.3
 
-This release fixes an issue with subpath imports (e.g.
-`@liveblocks/react/suspense`) and CommonJS in our packages which could happen
-with certain bundlers.
+### `@liveblocks/react` and `@liveblocks/react-ui`
+
+- Fix an issue with subpath imports (e.g. `@liveblocks/react/suspense`) and
+  CommonJS in our packages which could happen with certain bundlers.
 
 ## v2.24.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## vNEXT (not yet published)
 
+## v2.24.3
+
+This release fixes an issue with subpath imports (e.g.
+`@liveblocks/react/suspense`) and CommonJS in our packages which could happen
+with certain bundlers.
+
 ## v2.24.2
 
 ### `@liveblocks/react-ui`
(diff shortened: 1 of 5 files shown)


nvie (approved):
Thanks for getting this fixed! 🙏

marcbouchenoire:
@SugarDarius Definitely not expected for `_private/package.json` 😭 Not sure why the root one would change though, I'll look into it.

marcbouchenoire:
@SugarDarius I just tried myself on `main` and I have no changes, what's your `turbo --version`? And does the issue happen every time?

marcbouchenoire:
@SugarDarius I'm on `2.3.3` too and it's [the monorepo's version](https://github.com/liveblocks/liveblocks/blob/1a37a402a13d7153501c77f2a7f2e6f1eaec7db8/package.json#L30) so that should be expected, but I don't know why I don't have the issue too.

SugarDarius:
I was thinking about it too @nvie. I was looking at the configs and everything look okay. Yup I will make the try!

SugarDarius:
I guess we have our answer here 👍🏻 
It's well a bad Turbo cache, it works correctly now after running `rm -rf .turbo/`. TIL regularly clean the Turbo cache. 

Thanks @marcbouchenoire & @nvie 🙏🏻

nvie:
> TIL regularly clean the Turbo cache

Well... the thing is that normally you shouldn't! I think it means we are missing a properly declared dependency here somewhere.

SugarDarius:
FYI @marcbouchenoire @nvie I've opened the PR: https://github.com/liveblocks/liveblocks/pull/2430

PR #2427, opened 2025-05-27
Catch errors caused by outdated tool call renderings

nvie:
This is the current ugly implementation. It's not shippable as-is, but at the very least it avoids crashing the app for now.

Screen Shot 2025-05-27 at 16 36 39@2x


packages/liveblocks-react-ui/src/primitives/ErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
+import type { ComponentType, ErrorInfo, ReactNode } from "react";
```
marcbouchenoire, line 1: Do you think we'll export this publicly? If so, I could see it in `@liveblocks/react` like `ClientSideSuspense`, otherwise maybe we could move it in `/src/utils` here rather than `/src/primitives` like other private utilitarian components, https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react-ui/src/utils/Portal.tsx for example.

marcbouchenoire (approved):
Good idea to defer the rendering question to another PR 👍

packages/liveblocks-react-ui/src/primitives/ErrorBoundary.tsx
nvie: Maybe! But yeah, I didn't know where to best put it. I'll move it to utils for now 👍

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
```diff
@@ -98,9 +99,18 @@ function ToolInvocation({
...
-      <tool.render {...props} />
-    </AiToolInvocationContext.Provider>
+    <ErrorBoundary
+      fallback={
+        <p style={{ color: "red" }}>
+          Failed to render tool call result for '{part.toolName}'. See console
+          for details.
+        </p>
+      }
+    >
```
nimeshnayaju, line 109: Does this mean that if someone uses `ToolInvocation` directly (instead of using `AiChat`), they'll see the error fallback?

nimeshnayaju (approved):

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
nvie: Nobody is supposed to (ever) use `ToolInvocation` directly as it's so finicky to get right. That's why we built `AiMessage.Content`. But yes, if they'd use `AiMessage.Content`, then this is also the UI they'd get. I realize this cannot be the end state. Would love some help on finalizing this in the best way possible. Both the default UI to use in `AiChat`, as well as the best default to use for non-`AiChat` use cases. We can open another PR to change that 👍

PR #2428, opened 2025-05-28
Don't show `AiTool` spinner for tools without `execute` methods

marcbouchenoire:
This PR hides the `AiTool` "executing" spinner for tools which don't have `execute` methods.

We don't want users to have access to the `execute` callback in `render` since that's counter-intuitive, maybe we could pass its existence like `executable: boolean`, but for now I'm using `kInternal` to keep the callback just for us.

### DIFF
--- a/packages/liveblocks-react-ui/src/components/AiTool.tsx
+++ b/packages/liveblocks-react-ui/src/components/AiTool.tsx
@@ -1,8 +1,9 @@
-import type {
-  AiToolExecuteCallback,
-  AiToolTypePack,
-  JsonObject,
-  ToolResultData,
+import {
+  type AiToolExecuteCallback,
+  type AiToolTypePack,
+  type JsonObject,
+  kInternal,
+  type ToolResultData,
 } from "@liveblocks/core";
 import type { ComponentProps, ReactNode } from "react";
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react-ui/src/components/AiTool.tsx
```diff
@@ -174,7 +175,11 @@ const noop = () => {};
 export const AiTool = Object.assign(
   forwardRef<HTMLDivElement, AiToolProps>(
     ({ children, title, icon, className, ...props }, forwardedRef) => {
-      const { status, toolName } = useAiToolInvocationContext();
+      const {
+        status,
+        toolName,
+        [kInternal]: { execute },
```
nvie, line 181: I wouldn't mind exposing this on the context as an `isAutoExecuting: boolean` if you would prefer that.

nvie (approved):
Thank you! 🙏

packages/liveblocks-react-ui/src/components/AiTool.tsx
marcbouchenoire: I think my preference would be to wait before someone asks for it, since in non-generic cases (unlike us with `AiTool`) you would know if a tool has an `execute` method or not, since you'd be the one defining it just above `render`.

PR #2433, opened 2025-05-29
Alternative syntax for passing down inferred types to `AiTool.Confirmation`

nvie:
What do you think? This way of passing down the inferred type information to the `AiTool.Confirmation` feels a bit nicer to me:

```tsx
// ❌ Previously (still supported)
render: ({ $types }) => (
  <AiTool.Confirmation<typeof $types>
    confirm={({ ids }) => ({ ok: true })}
    cancel={() => ({ ok: false })}
    />
)
```

```tsx
// ✅ Alternative, using props, and renamed from $types to types
render: ({ types }) => (
  <AiTool.Confirmation
    types={types}
    confirm={({ ids }) => ({ ok: true })}
    cancel={() => ({ ok: false })}
    />
)
```

I think it would make the example snippets look a ton better.

### DIFF
--- a/e2e/next-ai-kitchen-sink/app/todo/page.tsx
+++ b/e2e/next-ai-kitchen-sink/app/todo/page.tsx
@@ -214,7 +214,8 @@ export default function Page() {
 
                         render: ({ status, result, $types }) => (
                           <AiTool>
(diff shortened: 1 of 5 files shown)


CTNicholas:
`types={types}` definitely looks better!

PR #2436, opened 2025-05-30
Refactor core internal logic for text mention notification event email

SugarDarius:
## What's going on?
This PR refactors the core internal logic in `@liveblocks/emails` for text mention notification event. It makes the code more dry and factorized to be more scalable and ready to handle future work and features. It's a necessary first before scaling our functions.

### Convert mention content
Now the two functions `convertTextEditorNodesAsHtml` and `convertTextEditorNodesAsReact` which was responsible for converting a mention content are replaced by `convertMentionContent` which is more generic and use the same single source of truth to convert mention contents into a format like html, React, markdown or whatever. 
It exposes a new generic type `ConvertMentionContentElements` where we can define each elements we want to convert the mention content 👇🏻 
```tsx
/**
 * Protocol:
 * Mention content elements to be converted to a custom format `T`
 */
export type ConvertMentionContentElements<T, U extends BaseUserMeta = DU> = {
  /**
   * The container element used to display mention content blocks
   */
  container: (args: MentionContentContainerElementArgs<T>) => T;
  /**
   * The mention element used to display the mention itself.
   */
  mention: (args: MentionContentMentionElementArgs<U>, index: number) => T;
  /**
   * The text element used to display the text surrounding the mention.
   */
  text: (args: MentionContentTextElementArgs, index: number) => T;
};
```

### Prepare Text Mention Notification Email
The function `prepareTextMentionEmailBaseData` has been replaced by `prepareTextMentionNotificationEmail` which now handles extracting data, resolving room info, resolving users and converting comment bodies all at once. It's a new abstraction layer which makes more sense. 
And now our functions `prepareTextMentionNotificationEmailAsHtml` and `prepareTextMentionNotificationEmailAsReact` are just using it simply: 
```tsx
export async function prepareTextMentionNotificationEmailAsReact(client, event, options): Promise<TextMentionNotificationEmailDataAsReact | null> {
  // ...
  const data = await prepareTextMentionNotificationEmail<ReactNode, BaseUserMeta>(
    client,
    event,
    { /* ... options ... */ },
    {
      /* elements to convert mention content */
    }
  )
  
  return data
}
```
It will allow us to expose more functions we'd like such as `prepareTextMentionNotificationEmailAsMarkdown`, `prepareTextMentionNotificationEmailAsCustom`, or `prepareTextMentionNotificationEmailAsWhatever`

> [!IMPORTANT]
> Please note that with this new implementation we're not forced anymore to expose `htmlContent` and `reactContent` for the returned mention content. We can now simply return a single `content` property no matter what format we're using to convert the mention content. It offers a better DX and a better type inference.

Thanks 🙏🏻 

Fixes [LB-1810](https://linear.app/liveblocks/issue/LB-1810/refactor-code-internal-of-text-notifications-functions)


marcbouchenoire (approved):
Thanks for working on it!

PR #2440, opened 2025-06-02
3.0: Group mentions groundwork

marcbouchenoire:
This PR applies all breaking changes needed for us to introduce group mentions later. It also applies a few adjacent non-breaking changes to better prepare the arrival of group mentions (e.g. introducing the `{ kind: "user", id: string }` format everywhere even if there's no other `kind` yet, adding support for returning objects instead of just IDs in `resolveMentionSuggestions` even though it's a backward-compatible change, etc.).

More context on the changes themselves: https://www.notion.so/liveblocks/Group-Mentions-1e482084c8128095a7fbec38fa54ca9f?source=copy_link#1f382084c81280b79cd9c2ce5ab8f03b

### Changes

The [updated 3.0 upgrade guide](https://github.com/liveblocks/liveblocks/pull/2440/files#diff-b0e499270e83973add7b37ebb8d6ad9f3aa7c158bb5e294a7670cc6798aab7ef) is a good human-readable outline of all changes in this PR.

For the Comments side of things, there will be a backend PR before the launch to make sure the new mention format with `kind` is accepted and existing mentions are returned with `kind: "user"`.

### To do

- [ ] The E2E tests are failing because creating a comment/thread with the new `{ type: "mention", kind: "user", id: "123" }` format isn't accepted by the backend yet

### DIFF
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -4820,24 +4820,40 @@ client.resolvers.invalidateMentionSuggestions();
 
 ## Utilities
 
-### getMentionedIdsFromCommentBody [#get-mentioned-ids-from-comment-body]
+### getMentionsFromCommentBody [#get-mentions-from-comment-body]
 
-Returns an array of each user’s ID that has been mentioned in a `CommentBody`
-(found under `comment.body`).
+Returns an array of mentions from a `CommentBody` (found under `comment.body`).
 
 ```ts
-import { getMentionedIdsFromCommentBody } from "@liveblocks/client";
+import { getMentionsFromCommentBody } from "@liveblocks/client";
 
-const mentionedIds = getMentionedIdsFromCommentBody(comment.body);
+const mentions = getMentionsFromCommentBody(comment.body);
+```
+
+An optional second argument can be used to filter the returned mentions. By
+default, if it’s not provided, all mentions are returned, including future
+mention kinds (e.g. group mentions in the future).
+
+```tsx
+// All mentions (same as `getMentionsFromCommentBody(commentBody)`)
+getMentionsFromCommentBody(commentBody);
+
+// Only user mentions with an ID of "123"
+getMentionsFromCommentBody(
+  commentBody,
+  (mention) => mention.kind === "user" && mention.id === "123"
+);
+
+// Only mentions with an ID which starts with "prefix:"
+getMentionsFromCommentBody(commentBody, (mention) => (
+  mention.id.startsWith("prefix:")
+);
 ```
 
 Here’s an example with a custom `CommentBody`.
 
 ```ts
-import {
-  CommentBody,
-  getMentionedIdsFromCommentBody,
-} from "@liveblocks/client";
+import { CommentBody, getMentionsFromCommentBody } from "@liveblocks/client";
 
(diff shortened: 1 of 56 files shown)


docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -4820,24 +4820,40 @@ client.resolvers.invalidateMentionSuggestions();
 
 ## Utilities
 
-### getMentionedIdsFromCommentBody [#get-mentioned-ids-from-comment-body]
```
marcbouchenoire, line 4823: I always struggle to decide what should be kept as "deprecated" in the docs for posterity vs what can be removed/replaced, `getMentionedIdsFromCommentBody` is a very niche utility so I'm removing it.

CTNicholas:
Docs look good! 

- Could you check if there any email guides in `/guides` that need updating?
- Could you add a section to the upgrade guide "Does this affect you?" and just highlight who this affects, e.g. "If you're using Comments or ... continue reading".

packages/liveblocks-emails/src/mention-content.tsx
```diff
@@ -86,8 +87,12 @@ export async function convertMentionContent<T, U extends BaseUserMeta = DU>(
   const blocks: T[] = nodes.map((node, index) => {
     switch (node.type) {
       case "mention": {
+        if (node.kind !== "user") {
+          return assertNever(node.kind, "Unknown mention kind");
+        }
+
```
SugarDarius, line 93: It would be nice to add a test in `__tests__/mention-content.test.tsx` to ensure when the mention kind is not okay we do well assert never for this case 🙂

packages/liveblocks-react/src/use-mention-suggestions.ts
```diff
@@ -8,14 +8,38 @@ import {
...
+
+  if (typeof suggestions[0] === "string") {
+    return (suggestions as string[]).map((id) => ({
+      kind: "user" as const,
+      id,
+    }));
+  }
+
+  return suggestions as MentionData[];
+}
```
SugarDarius: Small suggestion if you want to make the code shorter
```tsx
function normalizeMentionSuggestions<T extends string | MentionData>(
  suggestions: T[]
): MentionData[] {
  return suggestions.map((item) =>
    typeof item === "string"
      ? { kind: "user" as const, id: item }
      : (item as MentionData)
  );
}
```

marcbouchenoire: I had to do it slightly differently, done in https://github.com/liveblocks/liveblocks/pull/2440/commits/8369e4717aa6371e4b94117459edbce05fdc2f82.

packages/liveblocks-emails/src/mention-content.tsx
marcbouchenoire: That feels like over-testing to me, since it's a single condition and there's only one path to it 🤔

marcbouchenoire:
> Could you add a section to the upgrade guide "Does this affect you?" and just highlight who this affects, e.g. "If you're using Comments or ... continue reading".

@CTNicholas Done in https://github.com/liveblocks/liveblocks/pull/2440/commits/d69d62e08691a46873d72e1d1c10048edd4daf14, it's a bit long because of the variety of changes in 3.0 🤔

PR #2441, opened 2025-06-02
Avoid illegal dependency on React from core

nvie:
Open to better ideas for this, because now it no longer enforces that a user's render implementation actually returns valid JSX.

### DIFF
--- a/packages/liveblocks-core/src/ai.ts
+++ b/packages/liveblocks-core/src/ai.ts
@@ -1,5 +1,4 @@
 import type { JSONSchema7 } from "json-schema";
-import type { ComponentType } from "react";
 
 import { getBearerTokenFromAuthValue } from "./api-client";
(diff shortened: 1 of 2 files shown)


jrowny (approved):

marcbouchenoire (approved):
I don't have a better idea 😕

SugarDarius:
Dumb thought here: can't we just say it will return a `JSX.Element` as TypeScript supports JSX out of the box?
Of course we would have to update the `tsconfig.json` config but in the end I would see
```
(props: AiToolInvocationProps<A, R>) => JSX.Element | null
```

nvie:
That would be great, but I think it would require JSX to be enabled in _our user's_ tsconfig.json files then, not? Not an issue for those using React, but it would be a parse error if you don't?

PR #2442, opened 2025-06-03
Implement `useSendAiMessage` hook

nimeshnayaju:
This PR implements `useSendAiMessage` hook that sends a message in the provided chat. The hook is of the following signature:

```ts
useSendAiMessage(
  chatId: string,
  options?: { copilotId?: string }
): (message: string) => void;
```

This hook returns a method to send a message. The method internally creates an optimistic user message and assistant message and dispatches the 'ask-in-chat' command to the WebSocket server. 

Additionally, I've updated the `Empty` component in `AiChat` to include the `chatId` and `copilotId` props - these are the same properties that are passed to the `AiChat` component.

### DIFF
--- a/e2e/next-ai-kitchen-sink/app/todo/page.tsx
+++ b/e2e/next-ai-kitchen-sink/app/todo/page.tsx
@@ -216,9 +216,10 @@ export default function Page() {
                           additionalProperties: false,
                         },
 
-                        render: ({ status, result, $types }) => (
+                        render: ({ status, result, types }) => (
                           <AiTool>
-                            <AiTool.Confirmation<typeof $types>
+                            <AiTool.Confirmation
+                              types={types}
                               variant="destructive"
                               confirm={({ ids }) => {
                                 const deletedTitles = todos
--- a/e2e/next-ai-kitchen-sink/app/todo/page.tsx
+++ b/e2e/next-ai-kitchen-sink/app/todo/page.tsx
@@ -4,10 +4,13 @@ import { defineAiTool } from "@liveblocks/core";
 import {
   ClientSideSuspense,
   LiveblocksProvider,
+  useSendAiMessage,
 } from "@liveblocks/react/suspense";
 import { useCallback, useState } from "react";
 import { Popover } from "radix-ui";
 import { AiChat, AiTool } from "@liveblocks/react-ui";
+import { PlusIcon } from "../icons/plus-icon";
+import { ListIcon } from "../icons/list-icon";
 
 export default function Page() {
   const [todos, setTodos] = useState<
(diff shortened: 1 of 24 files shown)


e2e/next-ai-kitchen-sink/app/todo/page.tsx
```diff
@@ -271,3 +277,27 @@ export default function Page() {
...
+      <button
+        onClick={() => sendMessage("Add a todo titled 'Buy milk'")}
+        className="inline-flex items-center justify-center gap-2 border border-[var(--lb-foreground-subtle)] px-2 py-2 rounded-lg cursor-pointer"
+      >
+        <PlusIcon className="size-4" />
+        Add new todo
+      </button>
+    </div>
+  );
+}
```
nimeshnayaju, line 303: Here's how we could implement the `Empty` component in `AiChat`.

packages/liveblocks-react-ui/src/components/AiChat.tsx
```diff
@@ -207,7 +209,7 @@ export const AiChat = forwardRef<HTMLDivElement, AiChatProps>(
               {$.AI_CHAT_MESSAGES_ERROR(error)}
             </div>
           ) : messages.length === 0 ? (
-            <Empty />
+            <Empty chatId={chatId} copilotId={copilotId} />
```
nimeshnayaju, line 223: Passing `chatId` and `copilotId` to the `Empty` component so that users don't have to drill these props themselves.

e2e/next-ai-kitchen-sink/app/todo/page.tsx
```diff
@@ -271,3 +277,27 @@ export default function Page() {
     </main>
   );
 }
+
+function AiChatEmptyComponent({ chatId }: { chatId: string }) {
```
marcbouchenoire, line 281: We should export `AiChatComponentsEmptyProps` (and `AiChatComponentsLoadingProps` too even if it's empty for now I think) publicly from `"@liveblocks/react-ui"` and use it here.

marcbouchenoire (approved):
Love it ❤️

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1114,6 +1117,60 @@ function useDeleteAiChat() {
   );
 }
 
+/**
+ * Returns a function to send a message in an AI chat.
+ *
+ * @example
+ * const sendMessage = useSendAiMessage(chatId);
+ * sendMessage("Hello, Liveblocks AI!");
+ */
```
nvie, line 1126: @nimeshnayaju FYI, I just went over this for the rest of the hooks, and fixed the package linting issues that reported some hooks were missing doc strings, so while I was in that mode, I also added it here for your newly added hook 👍

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1114,6 +1117,60 @@ function useDeleteAiChat() {
...
+
+      void client[kInternal].ai.askUserMessageInChat(
+        chatId,
+        { id: newMessageId, parentMessageId: lastMessageId, content },
+        targetMessageId,
+        {
+          stream: false,
+          copilotId: copilotId as CopilotId | undefined,
+        }
+      );
```
nvie, line 1168: There's quite some duplication here with what we have in `AiChatComposer` in the other package too. I would love to see if we could instead expose this logic as a private method in core, so we can avoid having to expose the `messagesStore` and duplicate the optimistic message creation, and calling `askUserMessageInChat`. Fine for now, but something to clean up soon.

nvie (approved):
Lovely work, @nimeshnayaju! 😍

CTNicholas (approved):

packages/liveblocks-react/src/liveblocks.tsx
nimeshnayaju: Agreed. The duplication is intentional for now and the goal is to eventually use this hook inside the `AiChatComposer` but that is currently not possible due to certain requirements of the `AiChatComposer` component.

e2e/next-ai-kitchen-sink/app/todo/page.tsx
nimeshnayaju: Good point. Also added JSDoc comments to the props.

PR #2447, opened 2025-06-03
Add concept for `RegisterAiTool` component

nvie:
Like we have support for `<RegisterAiKnowledge />`, this PR adds a similar `<RegisterAiTool />` helper.

The current API looks like this:

```tsx
// Globally
<RegisterAiTool name="listTodos" tool={defineAiTool()({ /* ... */ })} />

// Scoped to a single chatId
<RegisterAiTool name="listTodos" tool={defineAiTool()({ /* ... */ })} chatId="my-chat" />
```

### Behavior

The behavior should be pretty intuitive:

1. If you provide a `chatId`, then it will only be made available to that `AiChat` instance. Other AiChat instances for different chat IDs will not have access to these tools.
2. If you do not provide a `chatId`, then it will be made available to all `AiChat` instances.
3. Mounting this component registers the tool.
4. Unmounting it removes the tool.
5. Registering tools via `<AiChat chatId={chatId} tools={{ ... }} />` is also still possible. Internally it just uses this component to register them now (scoped to its `chatId`).

If there are existing tool invocations in a chat but the AiChat instance doesn't have access to the tool because it isn't registered, or it only registered for another chat ID, then the chat will not render anything. (This is the normal behavior for missing/unknown tool invocations.)

### The `tool={}` prop

Notice how I had to use a `tool={defineAiTool()({ ... })}` prop here unfortunately for now 😞

For the same type inference reasons that we need the double-parens approach to `defineAiTool()()`, I could not directly offer this API:

```tsx
// ❌ Not possible unless we would sacrifice type inference support
<RegisterAiTool
  name="listTodos"
  description="List todos"
  execute={() => {}}
  render={() => {}}
  />
```

### DIFF
--- a/packages/liveblocks-core/src/ai.ts
+++ b/packages/liveblocks-core/src/ai.ts
@@ -925,13 +925,13 @@ export type Ai = {
     key?: string
   ) => void;
   /** @private This API will change, and is not considered stable. DO NOT RELY on it. */
-  registerChatTool: (
+  registerTool: (
     chatId: string,
     name: string,
     tool: AiOpaqueToolDefinition
   ) => void;
   /** @private This API will change, and is not considered stable. DO NOT RELY on it. */
-  unregisterChatTool: (chatId: string, tool: string) => void;
+  unregisterTool: (chatId: string, tool: string) => void;
 };
 
 /** @internal */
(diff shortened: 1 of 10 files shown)


marcbouchenoire (approved):
`tool={defineAiTool()({ ... })}` is unfortunate indeed but given that, the changes look good to me 👍

packages/liveblocks-core/src/ai.ts
```diff
@@ -354,64 +370,94 @@ function now(): ISODateString {
   return new Date().toISOString() as ISODateString;
 }
 
+// Symbol used to register tools globally. These tools are not scoped to
+// a particular chatId and made available to any AiChat instance.
+const kWILDCARD = Symbol("*");
```
nimeshnayaju, line 375: Makes sense!

PR #2448, opened 2025-06-04
Fix bug when on-screen chats are getting deleted

nvie:
Fixes LB-2061.
See also, [this Slack thread](https://liveblocks.slack.com/archives/C07VAF08PAN/p1748607630600139).

## Solution we decided on
1. Deleting chats will mark them as deleted in-memory instead of deleting the chat object.
2. Those chats will get a `deletedAt` field.
3. Re-render of component using `useAiChat()` is triggered, allowing UI updates to show chat is deleted.
4. ~An event is fired for the deletion. You can use the new `useAiChatDeletedListener()` hook to subscribe and respond (i.e. to redirect to another page).~ **Edit:** Decided to remove this after all in 03f0fe5f78db29ea52d03d4fa22f0151aef31bd5.

### DIFF
--- a/e2e/next-ai-kitchen-sink/app/chats/[chatId]/page.tsx
+++ b/e2e/next-ai-kitchen-sink/app/chats/[chatId]/page.tsx
@@ -123,10 +123,17 @@ function Chat({ chatId }: { chatId: string }) {
     };
   }, []);
 
+  const title = chat.title || "Untitled";
+  const titleClasses = chat.deletedAt
+    ? "line-through opacity-50"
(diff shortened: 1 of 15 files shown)


nvie:
> Are you not exporting the new hook from `suspense` and the factories as well because of it being experimental?

I actually just thought of that and wanted to come back to it. Fortunately we have [the linters](https://github.com/liveblocks/liveblocks/actions/runs/15441560466/job/43460368863?pr=2448) 😇 
Will fix this before merging.

PR #2449, opened 2025-06-04
Expose way(s) to control when `AiTool` is collapsed/expanded

marcbouchenoire:
I started this PR wanting to expose `open`/`defaultOpen`/`onOpenChange` on `AiTool` to control its internal `Collapsible`, but I went through 2 questions about that:

### ~~Question 1: Naming~~
I remembered that on the `Composer` component, I opted for the opposite of "open" with "collapsed" at the time, the main reason being that "open" is clear when there's the name "collapsible" next to it (that's why [most](https://ui.shadcn.com/docs/components/collapsible) [UI](https://www.radix-ui.com/primitives/docs/components/collapsible) [libraries](https://base-ui.com/react/components/collapsible) use it), but less so when it's next to "AiTool" or "Composer".

Does that make sense?

**Answer:** We're going with "collapsed".

### ~~Question 2: A possible simpler/higher-level API~~

As I started using the props myself, I realized that we already have some state with the `status` value and it could make building a "AiTool starts expanded to show details about what it's doing, then collapses when it's done" behavior really easy.

```
<AiChat
  tools={{
    slowTool: defineAiTool()({
      parameters: { /* ... */ },
      execute: async ({ id }) => {
          await wait(5000)
          return { ok: true };
       },
       render: ({ status }) => (
          <AiTool collapsed={status === "executed"}>
             <AiTool.Inspector />
          </AiTool>
       ),
    }),
  }}
/>
```

The issue with that is that `collapsed` is a controlled value, and because `status` won't change and isn't tied to `onCollapsedChange`, this `AiTool` won't expand on click after it's collapsed. What if we would make a mixed API between controlled and uncontrolled where a prop like `collapsed` would collapse/expand when it changes, **but** we would internally also collapse/expand based on clicks. Is that too weird? Or confusing? 

**Answer:** This PR implements this idea of `collapsed` being "semi-controlled".

### DIFF
--- a/packages/liveblocks-react-ui/src/utils/use-controllable-state.ts
+++ b/packages/liveblocks-react-ui/src/utils/use-controllable-state.ts
@@ -2,11 +2,29 @@ import { console } from "@liveblocks/core";
 import { useCallback, useEffect, useRef, useState } from "react";
 
 export function useControllableState<T>(
+  /**
+   * The value used as an uncontrolled fallback if neither `value` nor `defaultValue` are provided.
+   */
+  fallbackValue: T,
+
+  /**
+   * The controlled value.
+   */
   value?: T,
+
+  /**
+   * The event handler called when the value changes, if the value is controlled.
+   */
   onChange?: (value: T) => void,
+
+  /**
+   * The default uncontrolled value.
+   */
   defaultValue?: T
 ) {
-  const [uncontrolledValue, setUncontrolledValue] = useState(defaultValue);
+  const [uncontrolledValue, setUncontrolledValue] = useState(
+    defaultValue ?? fallbackValue
+  );
   const isControlled = value !== undefined;
   const wasControlled = useRef(isControlled);
 
--- a/packages/liveblocks-react-ui/src/utils/use-controllable-state.ts
+++ b/packages/liveblocks-react-ui/src/utils/use-controllable-state.ts
@@ -139,12 +139,7 @@ export function useSemiControllableState<T>(
   /**
    * The event handler called when the uncontrolled value changes.
    */
-  onChange: ((value: T) => void) | undefined,
-
-  /**
-   * An optional comparison function.
-   */
-  isEqual: (a: T, b: T) => boolean = Object.is
+  onChange: ((value: T) => void) | undefined
 ) {
   const [uncontrolledValue, setUncontrolledValue] = useState(value);
   const lastChange = useRef<"uncontrolled" | "controlled">("controlled");
@@ -153,7 +148,7 @@ export function useSemiControllableState<T>(
 
   // Listen to `value` changes during the render phase to avoid
   // having to always sync `uncontrolledValue` on every change.
-  if (!isEqual(lastValue.current, value)) {
+  if (!Object.is(lastValue.current, value)) {
     lastValue.current = value;
     lastChange.current = "controlled";
   }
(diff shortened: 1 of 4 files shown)


packages/liveblocks-react-ui/src/utils/use-controllable-state.ts
```diff
@@ -2,11 +2,29 @@ import { console } from "@liveblocks/core";
 import { useCallback, useEffect, useRef, useState } from "react";
 
 export function useControllableState<T>(
+  /**
+   * The value used as an uncontrolled fallback if neither `value` nor `defaultValue` are provided.
+   */
+  fallbackValue: T,
+
+  /**
+   * The controlled value.
```
nvie, line 16: Setting this to `undefined` makes it uncontrolled, right? That's the definition of whether something is controlled or uncontrolled? Could that be documented?

packages/liveblocks-react-ui/src/utils/use-controllable-state.ts
```diff
@@ -2,11 +2,29 @@ import { console } from "@liveblocks/core";
...
+
+  /**
+   * The event handler called when the value changes, if the value is controlled.
+   */
   onChange?: (value: T) => void,
+
+  /**
+   * The default uncontrolled value.
+   */
   defaultValue?: T
```
nvie: Could it suffice to make `defaultValue` mandatory and remove `fallbackValue`?

- Uncontrolled: (only) use a `defaultValue`, no `value`
- Controlled: have a `defaultValue`, and a `value` (but `defaultValue` would not get used unless `value` gets set to `undefined` at a later time)

nvie (commented):
> I remembered that on the `Composer` component, I opted for the opposite of "open" with "collapsed" at the time, the main reason being that "open" is clear when there's the name "collapsible" next to it, but less so when it's next to "AiTool" or "Composer".
> Does that make sense?

Yes! Makes a ton of sense to me in this context, good thinking!

> As I started using the props myself, I realized that we already have some state with the `status` value and it could make building a "AiTool starts expanded to show details about what it's doing, then collapses when it's done" behavior really easy.

Yes, makes a lot of sense to, I like it 👍

packages/liveblocks-react-ui/src/utils/use-controllable-state.ts
marcbouchenoire: I thought that it would mean giving up on the console warning because we wouldn't be able to differentiate between a user-set value and a default one, but the warning is based on if `value` changes so yeah, definitely possible. Done in https://github.com/liveblocks/liveblocks/pull/2449/commits/9e4fa022b9a95b69fa062e28c15f0d3e08c77304.

I also made the 2nd and 3rd arguments value optional instead of key optional to make it clear how this hook should be used (receiving the 3 props from a component, not using it manually and partially)

packages/liveblocks-react-ui/src/utils/use-controllable-state.ts
marcbouchenoire: Yes, I improved the comments a bit in https://github.com/liveblocks/liveblocks/pull/2449/commits/9e4fa022b9a95b69fa062e28c15f0d3e08c77304.

packages/liveblocks-react-ui/src/utils/use-controllable-state.ts
nvie: Oh nice, I wanted to also suggest to make them mandatory to specify. Lovely changes! ❤️

PR #2452, opened 2025-06-06
Fix state loss when using an inlined `render` function

nvie:
We had a bug that would cause state loss when using inlined `render: () => { ... }` functions that would use state. This bug would not happen if you moved it into an external component.

For example, if you would write:

```tsx
render: ({ status, toolCallId }) => {
  const [color] = useState(() => nextColor());
  return (
    <div style={{ color }}>
      Hello world from {toolCallId} ({status})!
    </div>
  );
},
```

Any time this tool would get re-rendered, it would actually get unmounted and remounted, causing its local state to be lost and being assigned a new color, which is not the desired behavior of course.

## Before
❌ Before, when render props would update, local state was lost.
Notice in video: color also changes when status changes.

https://github.com/user-attachments/assets/1e545317-f2b2-467d-b67b-5810ddff013c

## After
✅ Now, when render props would update, the component stays mounted now, keeping local state.
Notice in video: color no longer changes when status changes.

https://github.com/user-attachments/assets/3eb4eca2-b797-4e61-8f28-47e8d0c30156

### DIFF
--- a/packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
+++ b/packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
@@ -36,7 +36,7 @@ const defaultMessageContentComponents: AiMessageContentComponents = {
  * ToolInvocationPart
  * -----------------------------------------------------------------------------------------------*/
 
-function CustomTool(props: {
+function StableRenderFn(props: {
   renderFn: FunctionComponent<
     AiToolInvocationProps<JsonObject, ToolResultData>
   >;
@@ -104,7 +104,7 @@ function ToolInvocation({
       }
     >
       <AiToolInvocationContext.Provider value={props}>
-        <CustomTool
+        <StableRenderFn
           renderFn={
             tool.render as FunctionComponent<
               AiToolInvocationProps<JsonObject, ToolResultData>


packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
```diff
@@ -72,20 +81,19 @@ function ToolInvocation({
...
+    const { type: _, ...rest } = part;
+    return {
+      ...rest,
+      respond,
+      types: undefined as never,
+      [kInternal]: {
+        execute: tool?.execute,
+      },
+    };
+  }, [part, respond, tool?.execute]);
```
nvie, line 94: This is not strictly part of the fix, but I noticed that we were passing it in as `value={props}` to the context provider below, which I believe would cause any component subscribing to that context would get re-rendered, even if none of the fields actually changed.

nvie: Happy to revert this or discuss it elsewhere if you think it's controversial or not doing anything, @marcbouchenoire.

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
```diff
@@ -36,6 +36,15 @@ const defaultMessageContentComponents: AiMessageContentComponents = {
...
  * -----------------------------------------------------------------------------------------------*/
 
+function CustomTool(props: {
+  renderFn: FunctionComponent<
+    AiToolInvocationProps<JsonObject, ToolResultData>
+  >;
+  props: AiToolInvocationProps<JsonObject, ToolResultData>;
+}) {
+  return props.renderFn(props.props);
+}
```
nvie: The final trick is this:

- Rather than invoking `<tool.render />`, instead we invoke `<CustomTool renderFn={tool.render} />`.
- This gives React a stable reference (the value of `CustomTool` is constant), and thus a place to associate state in the tree.
- The implementation will still "just" call `tool.render(props)`, returning the user-provided JSX.

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
```diff
@@ -36,6 +36,15 @@ const defaultMessageContentComponents: AiMessageContentComponents = {
  * ToolInvocationPart
  * -----------------------------------------------------------------------------------------------*/
 
+function CustomTool(props: {
```
marcbouchenoire: Any particular reason to choose "custom" if this would be applied to all tools?

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
marcbouchenoire: I have nothing against this 👍

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
```diff
@@ -36,6 +36,15 @@ const defaultMessageContentComponents: AiMessageContentComponents = {
...
+  renderFn: FunctionComponent<
+    AiToolInvocationProps<JsonObject, ToolResultData>
+  >;
+  props: AiToolInvocationProps<JsonObject, ToolResultData>;
+}) {
+  return props.renderFn(props.props);
```
nvie, line 45: Note that to support this pattern, we "call" this render function here inlined, meaning it will only work for functional components, and won't work with class components. That's why I also changed from `ComponentType` to `FunctionComponent` in this PR. I think that's fine — who is still using class components anyway, right?

marcbouchenoire (approved):
Thanks for fixing this 🙏 And I'm happy to see that it doesn't require complexity!

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
nvie: Honestly, it was a working title and I forgot to get back to this and clean this name up. Renamed `CustomTool` → `StableRenderFn` now — thanks for the reminder 👍

nvie:
> Thanks for fixing this 🙏 And I'm happy to see that it doesn't require complexity!

Ha! On this journey, we iterated on 2 earlier solutions that had loads more complexity (check the individual commits to see a bit of that exploration). Until the solution worked, and we started simplifying it until this was all that remained. Mostly @nimeshnayaju to thank for that! ❤️

PR #2459, opened 2025-06-09
Rename `status` -> `stage` on InvocationPart types

nvie:
Fixes LB-2080.

Using `stage` implies there's only a one-way change direction:
- `receiving` → `executing` → `executed`

Unlike a status, which might "change back" to a previous status.

Thanks for the excellent name suggestion, @CTNicholas! 🙏


CTNicholas (approved):
Thanks!

PR #2464, opened 2025-06-12
Finalize tool invocation result type

nvie:
WIP (will update description later).


e2e/next-ai-kitchen-sink/app/todo/page.tsx
```diff
@@ -229,50 +227,68 @@ export default function Page() {
...
+                        render: ({ stage, result, types }) => {
+                          if (result?.type === "success") {
+                            console.log(result.data.deletedTitles);
+                          } else if (result?.type === "error") {
+                            console.error(result.error);
+                          } else if (result?.type === "cancelled") {
+                            // No extra fields available here
+                          } else {
+                            // If we get here, then stage isn't "executed"
+                          }
```
nvie: Note that the type of the result we pass to render is completely superfluous at this point. If you don't like the status field, you no longer _have_ to use it. You can also distinguish these cases by looking at other fields.

This same block could also be expressed as:

```tsx
if (result?.data) {
  console.log(result.data.deletedTitles);
} else if (result?.error) {
  console.error(result.error);
} else if (result?.cancelled) {
  // No extra fields available here
} else {
  // If we get here, then stage isn't "executed"
}
```

I think I actually would prefer that style!

nimeshnayaju (approved):

SugarDarius (approved):
:chef-kiss: ✅

PR #2465, opened 2025-06-12
Fix `<AiChat />` not scrolling all the way to the bottom

nimeshnayaju:
This PR fixes the following two issues:
1. Pressing 'Scroll to bottom' icon button wouldn't scroll the chat container all the way to the bottom (and leave a small space behind)
2. When a tool call collapsible was the last message in a chat container, and when the chat was refreshed, the chat container wouldn't scroll all the way to the bottom.

### DIFF
--- a/packages/liveblocks-react-ui/src/components/AiChat.tsx
+++ b/packages/liveblocks-react-ui/src/components/AiChat.tsx
@@ -258,7 +258,7 @@ export const AiChat = forwardRef<HTMLDivElement, AiChatProps>(
                 className="lb-ai-chat-scroll-indicator-button"
                 tabIndex={isScrollIndicatorVisible ? 0 : -1}
                 aria-hidden={!isScrollIndicatorVisible}
-                disabled={!isScrollIndicatorVisible}
+                aria-disabled={!isScrollIndicatorVisible}
                 onClick={() => scrollToBottom("smooth")}
               >
                 <span className="lb-icon-container">
--- a/packages/liveblocks-react-ui/src/styles/index.css
+++ b/packages/liveblocks-react-ui/src/styles/index.css
@@ -502,7 +502,6 @@
       opacity: 0;
       transition-duration: calc(3 * var(--lb-transition-duration));
       transition-property: block-size, opacity, content-visibility;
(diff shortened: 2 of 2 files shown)


packages/liveblocks-react-ui/src/components/AiChat.tsx
```diff
@@ -258,7 +258,7 @@ export const AiChat = forwardRef<HTMLDivElement, AiChatProps>(
                 className="lb-ai-chat-scroll-indicator-button"
                 tabIndex={isScrollIndicatorVisible ? 0 : -1}
                 aria-hidden={!isScrollIndicatorVisible}
-                disabled={!isScrollIndicatorVisible}
```
nimeshnayaju, line 261: Disabling the button as soon as the scroll visible become invisible was causing the scroll to be interrupted mid scroll when the 'Scroll to bottom' icon button was pressed.

packages/liveblocks-react-ui/src/styles/index.css
```diff
@@ -519,7 +518,6 @@
       &:where([data-state="open"]) {
         block-size: auto;
         opacity: 1;
-        content-visibility: auto;
```
nimeshnayaju, line 522: Including `content-visibility` was causing issue with chat container not scrolling all the way to the bottom and if a tool call collapsible was the last message in the chat. @marcbouchenoire I recall you mentioned a need to include this but removing it did not break any animations and functionality - maybe I'm missing something?

marcbouchenoire: I can't recall exactly what was the issue without it, I tried your branch with/without the Tailwind reset styles, on all browsers, and I also can't see anything breaking 🙌 I did a Safari/Firefox/reset pass a few weeks ago so maybe that "fixed" the need for `content-visibility`, regardless I feel better without it anyway.

packages/liveblocks-react-ui/src/components/AiChat.tsx
marcbouchenoire: Oh that's strange, but since it's `aria-hidden` already we can remove this line entirely.

PR #2475, opened 2025-06-16
Tweaks and moving `RegisterAi*` components to `@liveblocks/react`

marcbouchenoire:
This PR tweaks various things in the docs and does two bigger changes:
- Move `RegisterAiKnowledge` and `RegisterAiTool` to `@liveblocks/react` instead of `@liveblocks/react-ui`
- Reduces the number of times we say "actions" instead of "tools": "tools" is still confusing as a concept but ideally we can use the docs to teach how tools can take many forms and do many things, imo we should keep saying things like "tools can run actions" while avoiding completely replacing "tools" by "actions", if that makes sense

I'm not sure that I'm fully done here but given the size of the changes and that @nvie was planning to work on docs too, I'd prefer to merge this asap instead of keeping branches.


CTNicholas (approved):
Taken a close look through it all, perfect, thank you!

PR #2476, opened 2025-06-16
Fix bug where calling `respond()` inside `render()` would potentially make multiple network calls

nvie:
Fixes LB-2113.

It should only ever make a network call once, and immediately become a no-op for any subsequent calls.

### DIFF
--- a/packages/liveblocks-core/src/ai.ts
+++ b/packages/liveblocks-core/src/ai.ts
@@ -400,7 +400,9 @@ function createStore_forChatMessages(
   // auto-execute the execute() function for.
   const autoExecutableMessages = new Set<MessageId>();
 
-  const seenToolCallIds = new Set<string>();
+  // Used to keep track of any tool invocation that should be kicked off by
+  // this client instance.
+  const seenToolInvocationIds = new Set<string>();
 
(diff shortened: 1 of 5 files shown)


SugarDarius (approved):
LGTM ✅

packages/liveblocks-core/src/ai.ts
```diff
@@ -524,32 +531,42 @@ function createStore_forChatMessages(
...
+            .getToolΣ(toolInvocation.name, message.chatId)
             .get();
 
           const respondSync = <R extends JsonObject>(
             ...args: OptionalTupleUnless<R, [result: ToolResultResponse<R>]>
           ) => {
             const [result] = args;
+            if (executedToolInvocationIds.has(toolInvocation.invocationId)) {
+              console.warn(
+                `Ignoring respond(): tool '${toolInvocation.name}' (${toolInvocation.invocationId}) has already executed`
```
nvie: Actually, this isn't the correct fix. Please ignore, will update the PR with the proper fix.

nvie:
The previous fix wasn't covering all cases. Just updated the PR to fix all cases.

marcbouchenoire (approved):

PR #2478, opened 2025-06-17
Update AI docs pages

CTNicholas:


docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -1340,6 +1371,17 @@ const sendAiMessage = useSendAiMessage("my-chat-id");
 sendAiMessage("Hello!");
 ```
 
+### useAiChatMessages
+
+Returns a list of every message sent to an AI chat, identified by its ID.
```
marcbouchenoire: "in/from an AI chat" maybe? "sent to" sounds like it's only the user ones to me

marcbouchenoire (approved):
🙌

PR #2479, opened 2025-06-17
Only connect to AI socket when using AI features

nvie:
Fixes LB-2070.

### DIFF
--- a/packages/liveblocks-core/src/ai.ts
+++ b/packages/liveblocks-core/src/ai.ts
@@ -830,8 +830,9 @@ export type Ai = {
   [kInternal]: {
     context: AiContext;
   };
-  connect: () => void;
-  reconnect: () => void;
+  connectInitially: () => void;
+  // connect: () => void;
+  // reconnect: () => void;
   disconnect: () => void;
   getStatus: () => Status;
 
@@ -1144,9 +1145,16 @@ export function createAi(config: AiConfig): Ai {
     }
   });
 
+  function connectInitially() {
+    if (managedSocket.getStatus() === "initial") {
+      managedSocket.connect();
+    }
+  }
+
   async function sendClientMsgWithResponse<T extends ServerAiMsg>(
     msg: DistributiveOmit<ClientAiMsg, "cmdId">
   ): Promise<T> {
+    connectInitially();
     if (managedSocket.getStatus() !== "connected") {
       await managedSocket.events.didConnect.waitUntil();
     }
@@ -1262,8 +1270,8 @@ export function createAi(config: AiConfig): Ai {
         context,
       },
 
-      connect: () => managedSocket.connect(),
-      reconnect: () => managedSocket.reconnect(),
+      connectInitially,
+      // reconnect: () => managedSocket.reconnect(),
       disconnect: () => managedSocket.disconnect(),
 
       getChats,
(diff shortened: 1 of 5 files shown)


packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -222,6 +222,25 @@ export function getLiveblocksExtrasForClient<M extends BaseMetadata>(
   };
 }
 
+// Connect to the AI socket whenever this hook is called, to use in all AI-related hooks.
+//
+// The internal `ManagedSocket` no-ops when calling `connect()` if it is already connected,
+// so we don't need any conditional logic here. And we don't call `disconnect()` in cleanup
+// here because we don't want to disconnect whenever a single hook unmounts, instead we
+// disconnect when `LiveblocksProvider` unmounts.
```
marcbouchenoire, line 230: We could update this comment now that `connect()` doesn't exist anymore

SugarDarius (approved):

packages/liveblocks-core/src/ai.ts
```diff
@@ -1144,9 +1145,16 @@ export function createAi(config: AiConfig): Ai {
     }
   });
 
+  function connectInitially() {
+    if (managedSocket.getStatus() === "initial") {
```
marcbouchenoire, line 1149: I initially did that (directly in `useEnsureAiConnection`) but thought the state machine handled not reconnecting automatically, was the issue that it's not the case?

packages/liveblocks-core/src/client.ts
```diff
@@ -583,8 +583,25 @@ export function createClient<U extends BaseUserMeta = DU>(
...
+          throw new StopRetrying("Cannot use AI Copilots with a public token");
+        } else if (resp.token.parsed.k === TokenKind.SECRET_LEGACY) {
+          // TODO Make error message more instructive, include a link to docs
+          throw new StopRetrying("Cannot use AI Copilots with legacy tokens");
+        } else {
+          if (!resp.token.parsed.ai) {
+            // TODO Make error message more instructive, include a link to docs
+            throw new StopRetrying("AI Copilots not enabled for this account");
+          }
+        }
```
marcbouchenoire: 🙏

packages/liveblocks-react/src/liveblocks.tsx
nvie: The `.connectInitially()` is _not_ to avoid the "no-op" ness of connect in either of `connecting`, `connected`, `reconnecting`, or `reconnect` states. Instead, it is mostly to prevent the behavior in case the managed socket is in `disconnected` state. In that state, calling `.connect()` would normally be like `.reconnect()`, but in this case, we still want that to be a no-op.

This is important because if the connection attempt lead to an error like "Cannot use public key for AI Copilots" or "AI Copilots not enabled for your account" and the machine would be in disconnected state, we absolutely don't want to trigger a reconnect.

packages/liveblocks-core/src/ai.ts
nvie: That would have worked, but only when called from non-Suspense hooks.
The issue with the suspense version is this structure:

```tsx
useEffect(() => { /* I kick off a promise */ }, []);
use(waitUntilSomething);
```

Would cause the component to await a promise that will never be started, as the effect would not get a chance to run. The fix is to not just connect here, but more importantly to connect here: https://github.com/liveblocks/liveblocks/blob/18bb0b362a7485e82910927993a807b5d9153822/packages/liveblocks-core/src/ai.ts#L1157

PR #2486, opened 2025-06-18
Fix/update a few examples

marcbouchenoire:
This PR moves the last 2 remaining CRA examples to Vite (because we should and also because of how 3.0 needs TS 5.0) and silences a few type issues with Lexical.

Once it's merged I'll update the liveblocks.io gallery since some files have moved and changed names (and we no longer need to pass a public API key to iframes since all examples now use environment variables)


CTNicholas (approved):
Thanks for the fixes!

PR #2489, opened 2025-06-19
Group mentions

marcbouchenoire:
This PR introduces group mentions and a few other group-related changes. It's a big one so I'll try my best to give context and highlight important changes in this description.

## Context

Group mentions are available everywhere (Comments, Text Editor, and `@liveblocks/emails`) as a new mention kind, similar to user mentions:

```tsx
// Mentioning a user ID
{
  kind: "user",
  id: "user-0"
}

// Mentioning a group ID
{
  kind: "group",
  id: "group-0"
}
```

Groups are managed on Liveblocks (via the REST API or `@liveblocks/node`) as a mapping of group ID to group members (user IDs), they don’t contain any display data like a name, an avatar, etc.

It’s also possible to “hardcode” group members dynamically into mentions, to build dynamic mentions like `@here` for example.

```tsx
// Mentioning dynamic group members
{
  kind: "group",
  id: "here",
  userIds: ["user-0", "user-1", …]
}
```

Like user mentions, group mentions are created by end users by selecting them in mentions suggestions dropdowns after typing `@`. Our users give us these suggestions via [`resolveMentionSuggestions`](https://liveblocks.io/docs/api-reference/liveblocks-client#createClientResolveMentionSuggestions):

```tsx
// Today's API of returning IDs only works for user mentions but remains available
resolveMentionSuggestions: async ({ text }) => {
  // Search users based on `text` and return their IDs
  return ["user-0", "user-1", …]
}

// The new object-based API is the only way to use group mentions 
resolveMentionSuggestions: async ({ text }) => {
  // Search users and groups based on `text` and return them as objects
  return [
    { kind: "group", id: "here", userIds: /* Get the IDs of users currently present */ ["user-0""] },
    { kind: "group", id: "group-0" },
    { kind: "group", id: "group-1" },
    { kind: "user", id: "user-0" },
    { kind: "user", id: "user-1" },
  ]
}
```

### 🆕 Group mentions are handled out-of-the-box in `@liveblocks/react-ui` and the Text Editor packages

| Mentions                                                                                                                              | Suggestions dropdown                                                                                                                    |
|---------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|
| mentions | dropdown |

### 🆕 A new resolver `resolveGroupsInfo`

Even when groups are managed on Liveblocks, they don't contain display information (name, avatar, etc.), just IDs. Instead display information is set client-side via `resolveGroupsInfo`, like how [`resolveUsers`](https://liveblocks.io/docs/api-reference/liveblocks-client#createClientResolveUsers) is used for users' display information, or the same with [`resolveRoomsInfo`](https://liveblocks.io/docs/api-reference/liveblocks-client#createClientResolveRoomsInfo) for rooms.

This resolver brings the usual related things: a globally augmentable `GroupInfo` type and its `BaseGroupInfo` default, `useGroupInfo(groupId)` to consume it in React, `client.resolvers.invalidateGroupsInfo` to clear the cache, etc.

Similar to `resolveUsers`, we optionally handle `name` and `avatar` out-of-the-box in the default components (and `description`, optionally too), but any property can be set and then used with `useGroupInfo`.

### 🆕 Group management methods in `@liveblocks/node`

Methods like `Liveblocks.createGroup` or `Liveblocks.getUserGroups` to work with managed groups.

### 🔨 Support group mentions in thread inbox notifications in `@liveblocks/react-ui` and `@liveblocks/emails`

It's expected to work like this but group mentions in thread inbox notifications work the same way as user mentions: if a user is part of a group that gets group mentioned in a thread, they'll see an inbox notification saying "<author> mentioned you in…".

### 🔨 Support group mentions in `stringifyCommentBody` and `@liveblocks/emails` functions

The [`stringifyCommentBody`](https://liveblocks.io/docs/api-reference/liveblocks-client#stringify-comment-body) utility, [`prepareThreadNotificationEmailAsReact`](https://liveblocks.io/docs/api-reference/liveblocks-emails#prepare-thread-notification-email-as-react), [`prepareThreadNotificationEmailAsHtml`](https://liveblocks.io/docs/api-reference/liveblocks-emails#prepare-thread-notification-email-as-html), [`prepareTextMentionNotificationEmailAsReact`](https://liveblocks.io/docs/api-reference/liveblocks-emails#prepare-text-mention-notification-email-as-react), and [`prepareTextMentionNotificationEmailAsHtml`](https://liveblocks.io/docs/api-reference/liveblocks-emails#prepare-text-mention-notification-email-as-html) now all support group mentions in the same way. They get a new `resolveGroupsInfo` option (the same signature as the client-side one above) and mentions in them now receive a `group` value holding the result of `resolveGroupsInfo`, similar to how they receive a `user` value holding the result of `resolveUsers`.

```tsx
// `@liveblocks/client` and `@liveblocks/node`
const stringComment = await stringifyCommentBody(comment.body, {
  format: "markdown",
  resolveUsers: async ({ userIds }) => {
    const usersData = await __getUsersFromDB__(userIds);

    return usersData.map((userData) => ({
      name: userData.name, // "Nimesh"
      avatar: userData.avatar.src, // "https://..."
    }));
  },
  resolveGroupsInfo: async ({ groupIds }) => {
    // Same as `resolveUsers` but for groups
  }, 
  elements: {
    // Show the user's or group's `name` or fallback to the mention's ID
    mention: ({ element, user, group }) =>
      `@{user?.name ?? group?.name ?? element.id}`,
  },
});

// `@liveblocks/emails`
const emailData = await prepareThreadNotificationEmailAsReact(
  liveblocks,
  webhookEvent,
  {
    resolveUsers: async ({ userIds }) => {
      const usersData = await __getUsersFromDB__(userIds);

      return usersData.map((userData) => ({
        name: userData.name, // "Nimesh"
        avatar: userData.avatar.src, // "https://..."
      }));
    },
    resolveGroupsInfo: async ({ groupIds }) => {
      // Same as `resolveUsers` but for groups
    }, 
    components: {
      // Show the user's or group's `name` or fallback to the mention's ID
      Mention: ({ element, user, group }) => (
        <span>@{user?.name ?? group?.name ?? element.id}</span>
      ),
  }
);
```

### 🔨 Support `tenantId` params in `@liveblocks/node`

New and existing `@liveblocks/node` methods which need to receive tenant IDs (if tenants are used) now accept `tenantId` params.

### 🌅 The `getThreadParticipants` `@liveblocks/node` method and related endpoint are now marked as deprecated

Instead, thread subscriptions or the `getMentionsFromCommentBody` utility can be used.

### Groups info vs groups

We have two group-related stores: the "groups info" store and the "groups" one.
The "groups info" store is user-defined and public (e.g. `useGroupInfo`) but the "groups" one is internal, it contains the mappings of managed groups so that we know which groups the current user is part of based on group IDs. It's populated in two ways:

- On demand as group mentions become visible (in batches)
  - **Example:** Arriving on a screen with a text editor containing a group mention of `group-0` and two threads containing two group mentions of `group-0` and `group-1`, the mappings of `group-0` and `group-1` will be fetched in one call.
- All of them at once when fetching inbox notifications for the first time (as part of the inbox notifications response, not a separate call)
  - We need all mappings to be able to show "You've been mentioned" inbox notifications for group mentions in threads without any flashes or jumps

Both ways use the same store so mappings already fetched with inbox notifications won't be re-fetched on demand afterwards.

## To do

- [x] Update documentation
  - [x] Deprecated `getThreadParticipants` endpoint in REST API and method in `@liveblocks/node`
  - [x] Groups info resolver and related changes in `@liveblocks/client` and `@liveblocks/react`
  - [x] `resolveMentionSuggestions` changes in `@liveblocks/client` and `@liveblocks/react`
  - [x] Primitives changes in `@liveblocks/react-ui`
  - [x] Comment rendering changes in `stringifyCommentBody` and `@liveblocks/emails`
  - [x] Text mention changes in `@liveblocks/emails`
  - [x] Group management endpoints in REST API
  - [x] Group management methods in `@liveblocks/node`
  - [x] `tenantId` param in REST API
  - [x] `tenantId` param in `@liveblocks/node`
  - [x] Tenant ID (done in https://github.com/liveblocks/liveblocks/pull/2578)
- [x] Finalize tests
  - [x] Update existing mention-related tests to account for group mentions
  - [x] Add tests for `tenantId` params in `@liveblocks/node`
  - [x] Add tests for private `useGroup` hook and how `useInboxNotifications` fills the user's groups in the first call
- [x] Update changelog
  - [x] Update the documentation links for group mentions and tenants 
- [x] Create guide (or use the feature/product pages?)
  - [x] What to do to start using group mentions (group management, `resolveMentionSuggestions`, primitives, `@liveblocks/emails`, etc)
- [x] Update starter kit
  - [x] Add server action to sync dummy/local groups to Liveblocks
  - [x] Update `resolveMentionSuggestions` to offer group mentions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,13 @@ and [tenants](http://liveblocks.io/docs/authentication/tenants) in the docs.
 - Support returning group mention suggestions in `resolveMentionSuggestions`.
 - Support group mentions in `stringifyCommentBody`, it now accepts a
   `resolveGroupsInfo` option that passes the results to mentions as `group`.
+- Add query filters `roomId` and `kind` on the `getInboxNotifications` method.
 
 ### `@liveblocks/react`
 
 - Add `useGroupInfo` hook to use `resolveGroupsInfo` in React, same as `useUser`
   for `resolveUsers`.
+- Add query filters `roomId` and `kind` on the `useInboxNotifications` hook.
 
 ### `@liveblocks/react-ui`
 
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -3579,7 +3579,29 @@ console.log(subscriptions);
   </PropertiesListItem>
 </PropertiesList>
 
-<PropertiesListEmpty title="Arguments">_None_</PropertiesListEmpty>
+<PropertiesList title="Options">
+  <PropertiesListItem name="roomId" type="string">
+    Only return inbox notifications for the given room. [Learn
+    more](#filtering-by-room-id).
+  </PropertiesListItem>
+</PropertiesList>
+
+#### Filtering by room ID [#filtering-by-room-id]
+
+You can filter inbox notifications by those that are associated with a specific
+room or kind, by passing a `string` to `query.roomId` or `query.kind`.
+
+```ts
+// Filtering for inbox notifications that are associated with a specific room or kind
+const { inboxNotifications } = await client.getInboxNotifications({
+  query: {
+    // +++
+    roomId: "room1",
+    kind: "thread",
+    // +++
+  },
+});
+```
 
 ### Client.getInboxNotificationsSince
 
--- a/packages/liveblocks-core/src/api-client.ts
+++ b/packages/liveblocks-core/src/api-client.ts
@@ -402,7 +402,10 @@ export interface RoomHttpApi<M extends BaseMetadata> {
 }
 
 export interface NotificationHttpApi<M extends BaseMetadata> {
-  getInboxNotifications(options?: { cursor?: string }): Promise<{
+  getInboxNotifications(options?: {
+    cursor?: string;
+    query?: { roomId?: string; kind?: string };
+  }): Promise<{
     inboxNotifications: InboxNotificationData[];
     threads: ThreadData<M>[];
     subscriptions: SubscriptionData[];
@@ -412,6 +415,7 @@ export interface NotificationHttpApi<M extends BaseMetadata> {
 
   getInboxNotificationsSince(options: {
     since: Date;
+    query?: { roomId?: string; kind?: string };
     signal?: AbortSignal;
   }): Promise<{
     inboxNotifications: {
@@ -1525,8 +1529,17 @@ export function createApiClient<M extends BaseMetadata>({
   /* -------------------------------------------------------------------------------------------------
    * Inbox notifications (User-level)
    * -----------------------------------------------------------------------------------------------*/
-  async function getInboxNotifications(options?: { cursor?: string }) {
+  async function getInboxNotifications(options?: {
+    cursor?: string;
+    query?: { roomId?: string; kind?: string };
+  }) {
     const PAGE_SIZE = 50;
+
+    let query: string | undefined;
+
+    if (options?.query) {
+      query = objectToQuery(options.query);
+    }
 
     const json = await httpClient.get<{
       threads: ThreadDataPlain<M>[];
@@ -1543,6 +1556,7 @@ export function createApiClient<M extends BaseMetadata>({
       {
         cursor: options?.cursor,
         limit: PAGE_SIZE,
+        query,
       }
     );
 
@@ -1565,8 +1579,15 @@ export function createApiClient<M extends BaseMetadata>({
 
   async function getInboxNotificationsSince(options: {
     since: Date;
+    query?: { roomId?: string; kind?: string };
     signal?: AbortSignal;
   }) {
+    let query: string | undefined;
+
+    if (options?.query) {
+      query = objectToQuery(options.query);
+    }
+
     const json = await httpClient.get<{
       threads: ThreadDataPlain<M>[];
       inboxNotifications: InboxNotificationDataPlain[];
@@ -1580,7 +1601,7 @@ export function createApiClient<M extends BaseMetadata>({
     }>(
       url`/v2/c/inbox-notifications/delta`,
       await authManager.getAuthValue({ requestedScope: "comments:read" }),
-      { since: options.since.toISOString() },
+      { since: options.since.toISOString(), query },
       { signal: options.signal }
     );
     return {
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -3830,6 +3830,24 @@ function Inbox() {
 }
 ```
 
+#### Querying inbox notifications [#useInboxNotifications-query]
+
+It’s possible to return inbox notifications that match a certain query with the
+`query` option. You can filter inbox notifications based on their associated
+room ID or kind.
+
+```tsx
+// Returns threads that match the entire `query`, e.g. { roomId: "room1", ... }
+const { inboxNotifications } = useInboxNotifications({
+  query: {
+    // Filter for roomId
+    roomId: "room1",
+    // Filter for kind
+    kind: "thread",
+  },
+});
+```
+
 #### Pagination [#useInboxNotifications-pagination]
 
 By default, the `useInboxNotifications` hook returns up to 50 notifications. To
--- a/packages/liveblocks-core/src/client.ts
+++ b/packages/liveblocks-core/src/client.ts
@@ -203,7 +203,10 @@ export type NotificationsApi<M extends BaseMetadata> = {
    * const data = await client.getInboxNotifications();  // Fetch initial page (of 20 inbox notifications)
    * const data = await client.getInboxNotifications({ cursor: nextCursor });  // Fetch next page (= next 20 inbox notifications)
    */
-  getInboxNotifications(options?: { cursor?: string }): Promise<{
+  getInboxNotifications(options?: {
+    cursor?: string;
+    query?: { roomId?: string; kind?: string };
+  }): Promise<{
     inboxNotifications: InboxNotificationData[];
     threads: ThreadData<M>[];
     subscriptions: SubscriptionData[];
@@ -236,6 +239,7 @@ export type NotificationsApi<M extends BaseMetadata> = {
    */
   getInboxNotificationsSince(options: {
     since: Date;
+    query?: { roomId?: string; kind?: string };
     signal?: AbortSignal;
   }): Promise<{
     inboxNotifications: {
--- a/packages/liveblocks-react/src/__tests__/umbrella-store/index.test.ts
+++ b/packages/liveblocks-react/src/__tests__/umbrella-store/index.test.ts
@@ -43,7 +43,11 @@ describe("Umbrella Store", () => {
     ).toEqual(LOADING); // versions by room ID
 
     // Sync async-results getters
-    expect(store.outputs.loadingNotifications.signal.get()).toEqual(LOADING);
+    expect(
+      store.outputs.loadingNotifications
+        .getOrCreate(JSON.stringify({ roomId: "room-a" }))
+        .signal.get()
+    ).toEqual(LOADING);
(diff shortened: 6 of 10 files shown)


packages/liveblocks-emails/src/lib/batch-resolvers.ts
```diff
@@ -0,0 +1,152 @@
...
+  const index = ids.indexOf(id);
+
+  return results?.[index];
+}
+
+/**
+ * Batch calls to a resolver callback (which expects an array of IDs
+ * and returns an array of results) into a single call.
+ */
+export class BatchResolver<T> {
```
SugarDarius, line 28: I wonder if we shouldn't move this into `@liveblocks/core`? We may need it into some other places no?

packages/liveblocks-emails/src/lexical-editor.ts
```diff
@@ -450,3 +485,22 @@ export function findLexicalMentionNodeWithContext({
...
+    };
+  } else if (isSerializedGroupMentionNode(node)) {
+    return {
+      kind: "group",
+      id: node.attributes.__groupId,
+      userIds: node.attributes.__userIds,
+    };
+  }
+
+  assertNever(node, "Unknown mention kind");
```
SugarDarius, line 505: Do we really need to `assertNever` here? The function `findLexicalMentionNodeWithContext` should always ensure to return either a `SerializedLexicalMentionNode` or `SerializedLexicalGroupMentionNode`.

If not we shouldn't call `getMentionDataFromLexicalNode`

packages/liveblocks-emails/src/thread-notification.tsx
```diff
@@ -76,26 +82,84 @@ export const getUnreadComments = ({
...
+
+    if (!nextCursor) {
+      break;
+    }
+
+    cursor = nextCursor;
+  }
+
+  return groups;
+}
```
SugarDarius, line 111: For later but this DX makes me think it would be nice to have an async iterator for our users who doesn't want to handle the `nextCursor` such as
```tsx
for await (const data of client.getUserGroupsAsync({ userId }) {
  console.log(data)
}
```
where data are yielded each a next cursor is available.

packages/liveblocks-node-lexical/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const MENTION_CHARACTER = "@";
```
SugarDarius: This constant is used in multiple packages. Surely something we should add in a central place such as in `@liveblocks/core`.

SugarDarius (commented):
Looking good so far for the parts I reviewed:
-  `@liveblocks/emails` ✅ 
- `@liveblocks/node` ✅ 
- `@liveblocks/react` ✅ Feel free to open a discussion to create optimistic subscriptions. It seems an interesting point.

Thanks 🙏🏻

packages/liveblocks-emails/src/lib/batch-resolvers.ts
marcbouchenoire: It's backend-specific and in the frontend we already have [`Batch` and `BatchStore`](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-core/src/lib/batch.ts) so I'd rather centralize it if or when we do need it in two places.

packages/liveblocks-emails/src/thread-notification.tsx
marcbouchenoire: Yep we already have [`iterRooms`](https://liveblocks.io/docs/api-reference/liveblocks-node#iter-rooms) and [`iterInboxNotifications`](https://liveblocks.io/docs/api-reference/liveblocks-node#iter-users-userId-inboxNotifications) like this and we could do the same here but I'd rather do it later than do too many things in the initial release.

packages/liveblocks-emails/src/lexical-editor.ts
marcbouchenoire: Maybe not with an explicit union type as the input like here, but I find that it makes future refactorings like adding new mention kinds (e.g. `kind: "copilot"` later) feel a tiny bit "safer" since it ties the input/output types to the implementation.

packages/liveblocks-node-lexical/src/constants.ts
marcbouchenoire: Done in https://github.com/liveblocks/liveblocks/pull/2489/commits/c27f4e28c65b64236963935564e7f9ad61ee303d.

CHANGELOG.md
```diff
@@ -2,6 +2,52 @@
 
 ## v3.7.0
 
+This release introduces group mentions (e.g. `@engineering`) across all packages
+and first-class support for tenants. Learn more about [group mentions](#) and
```
nvie: Reminder to not forget to link these to the actual doc URLs.

packages/liveblocks-core/src/api-client.ts
```diff
@@ -1526,6 +1546,14 @@ export function createApiClient<M extends BaseMetadata>({
       }
     );
 
+    const groups = json.groups.map(convertToGroupData);
+
+    // Instead of being returned publicly, the user's groups are put in
+    // a separate store which is also used for on-demand fetching.
+    for (const group of groups) {
+      groupsStore.fill(group.id, group);
```
nvie: Since `.fill()` calls `update` which in turn set a signal, this loop means the same signal might get updated multiple times (triggering a notification waterfall after every update potentially, depending on configrued listeners). A best practice is to wrap such updates in a `batch()` which will only trigger a single notification waterfall at the end.

You could wrap it right here, or you could make `.fill()` take an array instead and do the batch wrapping there (closer to the actual signal), which I think would be nicest to avoid bleeding that complexity in here.

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -587,6 +654,62 @@ const client = createClient({
...
+}
+
+// A group mention suggestion with fixed group members
+{
+  kind: "group",
+  id: "group-1",
+  members: ["user-1", "user-2"],
+}
+```
+
```
nvie: The way I understand it, the id `"here"` used in other examples is not a special ID. Instead, what makes it special is the presence of a `userIds` key. This makes sense.

But what happens if you do this: `{ kind: "group", id: "actuallyExistingGroupId", userIds: ["a", "b", "c"] }`? I think the local `userIds` take precedence and will cause the group's members not be looked up in that case, right?

If so, that is definitely the behavior I would expect. Maybe good to explicitly document that?

nvie (approved):
Wow—an amazing PR. I read through most of it (☕) and have barely any comments to add here. It's really well executed, super consistent with every API we already have out there, it cleanly extends our existing mention mechanism, and works exactly as I would expect. Outstanding execution, @marcbouchenoire 😍!

CHANGELOG.md
marcbouchenoire: Waiting on https://github.com/liveblocks/liveblocks/pull/2655.

docs/pages/api-reference/liveblocks-client.mdx
marcbouchenoire: Yep as expected `userIds` takes precedence over existing group IDs as well, I'll do a pass to be more explicit about it in the docs 👍

packages/liveblocks-core/src/api-client.ts
marcbouchenoire: Great catch, since this API is internal (and new for this use-case) I'll improve it to support arrays 👍

marcbouchenoire: Done in https://github.com/liveblocks/liveblocks/pull/2489/commits/d86f4b186cf71c27344b589739ea0d365ec43a7e.

PR #2491, opened 2025-06-24
Improve AI docs

nvie:
I found a couple of bugs in the docs, and added some more:

Screen Shot 2025-06-24 at 15 43 47@2x

### DIFF
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -1013,7 +1013,7 @@ In this example, until the Confirm button is clicked, the AI chat will remain in
 
 This example is for illustrative purposes only. In practice, using our
 [`AiTool.Confirmation`](/docs/api-reference/liveblocks-react-ui#AiTool.Confirmation)
-tool is preferred.
+tool is preferred for building confirm/cancel flows.
 
 Like with the `execute` function, the `respond` function should be called with a
 value of this shape:
--- a/docs/pages/api-reference/liveblocks-client.mdx
+++ b/docs/pages/api-reference/liveblocks-client.mdx
@@ -281,7 +281,7 @@ const client = createClient({
         Authentication: "<your own headers here>",
         "Content-Type": "application/json",
       },
-      // Don't forget to pass `room` down. Note that it
+      // Don’t forget to pass `room` down. Note that it
       // can be undefined when using Notifications.
       body: JSON.stringify({ room }),
     });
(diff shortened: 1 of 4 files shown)


marcbouchenoire (approved):
Pre-approving for future changes too ✅

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -891,18 +891,142 @@ Note that the function should be called like this `defineAiTool()({ ... })`
...
+  use, to enable showing a UI while the tool call `args` object is being
+  streamed in, but before the tool call can actually be made.
+- `executing` The tool call is currently executing, or is ready to be. In this
+  stage, the `args` are fully known, but the result of the tool call is not
+  known yet.
+- `executed` The tool call has been executed, and the result is known. This
+  happens after your `execute` function was run, or after you called `respond()`
+  inside `render`. In this stage, the `result` object will be available.
+
+The render component will automatically re-render when its stage changes.
```
marcbouchenoire, line 923: `defineAiTool` isn't React-specific but since this page is for `@liveblocks/client` I think we would could specify that `render` is React-specific.

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -891,18 +891,142 @@ Note that the function should be called like this `defineAiTool()({ ... })`
...
+          }}
+        >
+          Confirm
+        </button>
+        <button
+          onClick={() => {
+            respond({ cancel: true });
+          }}
+        >
+          Confirm
```
marcbouchenoire:
```suggestion
          Cancel
```

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -891,18 +891,142 @@ Note that the function should be called like this `defineAiTool()({ ... })`
...
+  },
+});
+```
+
+In this example, until the Confirm button is clicked, the AI chat will remain in
+“executing” stage, awaiting the result of this tool call.
+
+This example is for illustrative purposes only. In practice, using our
+[`AiTool.Confirmation`](/docs/api-reference/liveblocks-react-ui#AiTool.Confirmation)
+tool is preferred.
```
marcbouchenoire: Only for simple yes/no confirmations, other valid use-cases like questions/forms aren't covered by it.

marcbouchenoire (commented):
Great additions!

docs/pages/api-reference/liveblocks-client.mdx
nvie: Ah, good point. I just wanted to document this, then realized that currently, we only have React APIs that would do anything with a tool definition. Either you pass it as a prop to `AiChat` directly, or to `<RegisterAiTool />`, both Reactisms. So currently, it's not really possible to use it outside of React anyway.

docs/pages/api-reference/liveblocks-client.mdx
nvie: Fixed in eef738722e2298edb5059dea2caa8711255d7c7d 👍

PR #2492, opened 2025-06-24
`AiTool` and reasoning part improvements

marcbouchenoire:
This PR tweaks a few things:
- Reasoning parts now auto-collapse when they're done, and they start collapsed when loading chats with reasoning parts that were already done
- `AiTool` has a new `collapsible` prop which will make `AiTool` no longer collapsible (it stays in its current `collapsed` state
- Improve `AiTool` header spacings

Now that 3.0 is launched, are we going back to regular smaller releases? I added docs changes in this PR but that means they'll be published as soon as this PR lands. I can move them to their own PR but over many small PRs that seems like a recipe to forget about docs changes.

### DIFF
--- a/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
+++ b/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
@@ -4,6 +4,7 @@ import {
   forwardRef,
   memo,
   type ReactNode,
+  useEffect,
   useState,
 } from "react";
 
@@ -26,6 +27,7 @@ import type {
 import * as Collapsible from "../../primitives/Collapsible";
 import { classNames } from "../../utils/class-names";
 import { ErrorBoundary } from "../../utils/ErrorBoundary";
+import { useSemiControllableState } from "../../utils/use-controllable-state";
 import { Prose } from "./Prose";
 
 type UiAssistantMessage = WithNavigation<AiAssistantMessage>;
@@ -146,8 +148,17 @@ function ReasoningPart({
   part,
   isStreaming,
 }: AiMessageContentReasoningPartProps) {
-  const [isOpen, setIsOpen] = useState(false);
+  // Start collapsed if reasoning is already done.
+  const [isOpen, setIsOpen] = useState(isStreaming);
   const $ = useOverrides();
+
+  // Auto-collapse when reasoning is done, while still allowing the user to
+  // open/collapse it manually during and after it's done.
+  useEffect(() => {
+    if (!isStreaming) {
+      setIsOpen(false);
+    }
+  }, [isStreaming]);
 
   return (
     <Collapsible.Root
--- a/docs/pages/api-reference/liveblocks-react-ui.mdx
+++ b/docs/pages/api-reference/liveblocks-react-ui.mdx
@@ -536,6 +536,10 @@ render: () => (
     Callback fired when the collapsed state changes. Use this to control the
     collapsed state externally.
   </PropertiesListItem>
+  <PropertiesListItem name="collapsible" type="boolean">
+    Whether the tool content can be collapsed. If set to `false`, clicking on it
+    or changing the `collapsed` prop will have no effect.
+  </PropertiesListItem>
   <PropertiesListItem name="className" type="string">
     CSS class name to apply to the tool container.
   </PropertiesListItem>
--- a/docs/pages/api-reference/liveblocks-react-ui.mdx
+++ b/docs/pages/api-reference/liveblocks-react-ui.mdx
@@ -538,7 +538,8 @@ render: () => (
   </PropertiesListItem>
   <PropertiesListItem name="collapsible" type="boolean">
     Whether the tool content can be collapsed. If set to `false`, clicking on it
-    or changing the `collapsed` prop will have no effect.
+    or changing the `collapsed` prop will have no effect. If there's no content,
+    this prop has no effect.
   </PropertiesListItem>
   <PropertiesListItem name="className" type="string">
     CSS class name to apply to the tool container.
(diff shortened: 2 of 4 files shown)


packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
```diff
@@ -26,6 +27,7 @@ import type {
 import * as Collapsible from "../../primitives/Collapsible";
 import { classNames } from "../../utils/class-names";
 import { ErrorBoundary } from "../../utils/ErrorBoundary";
+import { useSemiControllableState } from "../../utils/use-controllable-state";
```
nvie: Unused import.

nvie:
> Now that 3.0 is launched, are we going back to regular smaller releases? I added docs changes in this PR but that means they'll be published as soon as this PR lands. I can move them to their own PR but over many small PRs that seems like a recipe to forget about docs changes.

Yeah, it's a bit annoying. For reference docs, it'd be ideal to point the docs location for liveblocks.io to a specific (the latest) tag/branch instead of `main`. But that's probably only true for reference docs and not for other docs.

Another idea could be to add a `@since 3.0.1` tag to the new docs. That way it would be fine to merge these to `main` as-is, and at worst someone wanting to use the API would try to upgrade to 3.0.1 and see that it doesn't exist yet. I think this solution would have my preference.

nvie (approved):
Nice updates! 👌

packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
```diff
@@ -146,9 +147,18 @@ function ReasoningPart({
...
+  const [isOpen, setIsOpen] = useState(isStreaming);
   const $ = useOverrides();
 
+  // Auto-collapse when reasoning is done, while still allowing the user to
+  // open/collapse it manually during and after it's done.
+  useEffect(() => {
+    if (!isStreaming) {
+      setIsOpen(false);
+    }
+  }, [isStreaming]);
```
nimeshnayaju, line 160: **Nitpick:** Haven't tested myself but I'd do this to avoid an extra re-render:

```tsx
// If the collapsible isn't already open and reasoning is complete, we collapse the reasoning collapsible 
if (!isOpen && !isStreaming) {
    setIsOpen(false)
}
```

nimeshnayaju (approved):

packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
marcbouchenoire: > If the collapsible isn't already open and reasoning is complete

I'm not sure I see what you mean by this, if it's closed calling `setIsOpen(false)` would have no effect, did you mean the other way around?

About the extra re-render: I didn't call `setIsOpen` in the render phase because that could mean auto-collapsing "randomly" because of unrelated re-renders, e.g. a message is done reasoning and you open it manually but it closes on its own while you're reading it because of a re-render coming from something else. We could follow the route of `useSemiControllableState` with a "last change wins" setup, I wasn't sure what to choose here between the complexity and the extra re-render.

nimeshnayaju: Sorry, the code should have been:

```tsx
// If the collapsible is open and reasoning is complete, we collapse the reasoning collapsible 
if (isOpen && !isStreaming) {
    setIsOpen(false)
}
```


> I didn't call setIsOpen in the render phase because that could mean auto-collapsing "randomly" because of unrelated re-renders, e.g. a message is done reasoning and you open it manually but it closes on its own while you're reading it because of a re-render coming from something else. 

Isn't that the case with the current code too? If you open it manually, it'll automatically close on its own?

nvie: > _"I didn't call setIsOpen in the render phase"_

Doing a state update during render is a React anti-pattern anyway, right? You should only ever do state updates inside effects.

marcbouchenoire: > Isn't that the case with the current code too? If you open it manually, it'll automatically close on its own?

It will only close when streaming is done because of the `[isStreaming]` dependency array, in the render phase any re-render could trigger the change, like a Context higher in the app updating for example.

> Doing a state update during render is a React anti-pattern anyway, right?

In controlled cases like this one it could be fine, but here be dragons yeah.

marcbouchenoire:
> Another idea could be to add a `@since 3.0.1` tag to the new docs. That way it would be fine to merge these to main as-is, and at worst someone wanting to use the API would try to upgrade to 3.0.1 and see that it doesn't exist yet. I think this solution would have my preference.

@nvie The hardest thing would be to figure out how granular this could be, e.g. could it be used to add a single new prop in an existing prop table? Replace/update a code snippet? etc. I chose simplicity for now and opened [a 3.1 branch](https://github.com/liveblocks/liveblocks/pull/2499).

packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
nimeshnayaju: >Doing a state update during render is a React anti-pattern anyway, right? You should only ever do state updates inside effects.

@nvie It is generally an anti-pattern but in some cases, it is okay and better to do (and actually saves a render)  (See [docs](https://react.dev/reference/react/useState#storing-information-from-previous-renders)).

>This pattern can be hard to understand and is usually best avoided. However, it’s better than updating state in an effect. When you call the set function during render, React will re-render that component immediately after your component exits with a return statement, and before rendering the children. This way, children don’t need to render twice. The rest of your component function will still execute (and the result will be thrown away). If your condition is below all the Hook calls, you may add an early return; to restart rendering earlier.

I talked to @marcbouchenoire earlier today and because of some extra requirements, we need to have the `useEffect` in place to avoid unintended collapsing of the tool collapsible.

Issue #2500, opened 2025-06-27
@liveblocks/redux uses outdated redux version


ceduch:
**Describe the bug**

I'm looking to upgrade my app to next.js 15 and React 19, but I can't update my react-redux package to support react 19 because @liveblocks/redux uses an outdated Redux version.

I did not try your guides to set up liveblocks with redux recently but they are probably not working either.

**To Reproduce**

1. Update react-redux to 9.2.0

**Expected behavior**

@liveblocks/redux package should support the latest Redux version.

marcbouchenoire:
Hi! This was fixed by https://github.com/liveblocks/liveblocks/pull/2504 and released in [v3.1.0](https://github.com/liveblocks/liveblocks/releases/tag/v3.1.0).

PR #2504, opened 2025-06-30
Add support for Redux 5

SugarDarius:
### What's going on?

This PR add support for Redux v5 following this reported issue: https://github.com/liveblocks/liveblocks/issues/2500
There are no big changes at our level excepts the declaration of the enhancer in our users' code 👇🏻 

Old
```tsx
export function makeStore() {
  return configureStore({
    reducer: slice.reducer,
    enhancers: [
      liveblocksEnhancer<State>({
        client,
        storageMapping: { items: true },
        presenceMapping: { name: true, counter: true },
      }),
    ],
  });
}
```

New
```tsx
export function makeStore() {
  return configureStore({
    reducer: slice.reducer,
    enhancers: (getDefaultEnhancers) =>
      getDefaultEnhancers().concat( // 👈🏻 We need to call now `getDefaultEnhancers()` or `new Tuple` 
        liveblocksEnhancer<State>({
          client,
          storageMapping: { items: true },
        presenceMapping: { name: true, counter: true },
        })
      ),
  });
}
```

Using `getDefaultEnhancers()` is required by Redux as described in the [documentation](https://redux-toolkit.js.org/api/configureStore#enhancers).
> Typescript users are required to use a Tuple instance (if not using a getDefaultEnhancer result, which is already a Tuple), for better inference.  

You can read more about `getDefaultEnhancers()` [here](https://redux-toolkit.js.org/api/getDefaultEnhancers).

Documentation and guides are updated as well. 
Thanks 🙏🏻

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - `defineAiTool()()` now takes an optional `enabled` property. When set to
   `false`, the tool will not be made available to the AI copilot for new/future
-  chat messages, but stil l allow existing tool invocations to be rendered that
+  chat messages, but still allow existing tool invocations to be rendered that
   are part of the historic chat record.
 
 ### `@liveblocks/react`


marcbouchenoire (approved):
I'm not familiar with Redux but this makes sense.

1. The Redux docs use `concat` but imo it would be more consistent/familiar for us to use the spread syntax here instead (especially in the docs/examples):
	```
	const store = configureStore({
    	reducer,
  		enhancers: (getDefaultEnhancers) => [
    		...getDefaultEnhancers(),
    		liveblocksEnhancer({
      			client,
    			storageMapping: {},
      			presenceMapping: {},
    		})
  		),
	});
	``` 
3. Since this change needs a minor release, you could merge it to https://github.com/liveblocks/liveblocks/pull/2499

SugarDarius:
> The Redux docs use concat but imo it would be more consistent/familiar for us to use the spread syntax here instead (especially in the docs/examples):

@marcbouchenoire that's the thing with Redux v5. The `.concat()` isn't js native. 
The following syntax using the spread operator won't work
```tsx
const store = configureStore({
	reducer,
	enhancers: (getDefaultEnhancers) => [
		...getDefaultEnhancers(),
		liveblocksEnhancer({
  			client,
			storageMapping: {},
  			presenceMapping: {},
		})
	),
});
```
because the `.concat()` method comes from the object `Tuple` now required in Redux v5 as you can see in the declarations:
```tsx
declare class Tuple<Items extends ReadonlyArray<unknown> = []> extends Array<Items[number]> {
    constructor(length: number);
    constructor(...items: Items);
    static get [Symbol.species](): any;
    concat<AdditionalItems extends ReadonlyArray<unknown>>(items: Tuple<AdditionalItems>): Tuple<[...Items, ...AdditionalItems]>;
    concat<AdditionalItems extends ReadonlyArray<unknown>>(items: AdditionalItems): Tuple<[...Items, ...AdditionalItems]>;
    concat<AdditionalItems extends ReadonlyArray<unknown>>(...items: AdditionalItems): Tuple<[...Items, ...AdditionalItems]>;
    prepend<AdditionalItems extends ReadonlyArray<unknown>>(items: Tuple<AdditionalItems>): Tuple<[...AdditionalItems, ...Items]>;
    prepend<AdditionalItems extends ReadonlyArray<unknown>>(items: AdditionalItems): Tuple<[...AdditionalItems, ...Items]>;
    prepend<AdditionalItems extends ReadonlyArray<unknown>>(...items: AdditionalItems): Tuple<[...AdditionalItems, ...Items]>;
}

interface ConfigureStoreOptions<
    S = any,
    A extends Action = UnknownAction, 
    M extends Tuple<Middlewares<S>> = Tuple<Middlewares<S>>, 
    E extends Tuple<Enhancers> = Tuple<Enhancers>
    /* ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */
    P = S,
> {
    /**
     * The store enhancers to apply. See Redux's `createStore()`.
     * All enhancers will be included before the DevTools Extension enhancer.
     * If you need to customize the order of enhancers, supply a callback
     * function that will receive a `getDefaultEnhancers` function that returns a Tuple,
     * and should return a Tuple of enhancers (such as `getDefaultEnhancers().concat(offline)`).
     * If you only need to add middleware, you can use the `middleware` parameter instead.
     */
    enhancers?: (getDefaultEnhancers: GetDefaultEnhancers<M>) => E;
}
```
So we're forced to use this syntax in our examples as I updated them in Redux v5 to be up-to-date and same in our guides.

I will also let @CTNicholas and @nvie check this PR before merging it. 

Thanks 🙏🏻

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 - `defineAiTool()()` now takes an optional `enabled` property. When set to
   `false`, the tool will not be made available to the AI copilot for new/future
-  chat messages, but still allow existing tool invocations to be rendered that
+  chat messages, but stil l allow existing tool invocations to be rendered that
```
nvie:
```suggestion
  chat messages, but still allow existing tool invocations to be rendered that
```

nvie (approved):
Nice! I like how you made it compatible with both V4 and V5 🫶

PR #2511, opened 2025-07-04
Add global flag on LiveObjects to immediately throw when exceeding max object size

nvie:
Adds experimental setting `LiveObject.detectLargeObjects`, which can be enabled globally using `LiveObject.detectLargeObjects = true` (default is false). With this setting enabled, calls to `LiveObject.set()` or `LiveObject.update()` will throw as soon as you add a value that would make the total size of the LiveObject exceed the platform limit of 128 kB. The benefit is that you get an early error instead of a silent failure, but the downside is that this adds significant runtime overhead if your application makes many LiveObject mutations.

### DIFF
--- a/packages/liveblocks-core/src/crdts/LiveObject.ts
+++ b/packages/liveblocks-core/src/crdts/LiveObject.ts
@@ -578,11 +578,19 @@ export class LiveObject<O extends LsonObject> extends AbstractCrdt {
         }
       }
 
-      const newSize = new TextEncoder().encode(JSON.stringify(data)).length;
-      if (newSize > MAX_LIVE_OBJECT_SIZE) {
-        throw new Error(
-          `LiveObject size exceeded limit: ${newSize} bytes > ${MAX_LIVE_OBJECT_SIZE} bytes. See https://liveblocks.io/docs/platform/limits#Liveblocks-Storage-limits`
-        );
+      // Fast upper-bound check: multiply JSON string length by 4 (worst-case UTF-8)
+      // This is much faster than TextEncoder and gives us an upper bound
+      const jsonString = JSON.stringify(data);
+      const upperBoundSize = jsonString.length * 4;
+
+      // Only do the precise calculation if the fast check suggests we might be close
+      if (upperBoundSize > MAX_LIVE_OBJECT_SIZE) {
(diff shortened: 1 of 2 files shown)


packages/liveblocks-core/src/crdts/LiveObject.ts
```diff
@@ -544,6 +560,32 @@ export class LiveObject<O extends LsonObject> extends AbstractCrdt {
...
+          data[key] = value;
+        }
+      }
+
+      const newSize = new TextEncoder().encode(JSON.stringify(data)).length;
+      if (newSize > MAX_LIVE_OBJECT_SIZE) {
+        throw new Error(
+          `LiveObject size exceeded limit: ${newSize} bytes > ${MAX_LIVE_OBJECT_SIZE} bytes. See https://liveblocks.io/docs/platform/limits#Liveblocks-Storage-limits`
+        );
+      }
```
nimeshnayaju: For performance reasons, could we do a quick approximate check by multiplying the string's length by 4 (which returns an upper bound assuming that every character might take up to 4 bytes - as in worst-case UTF-8 encoding) and comparing it to the maximum allowed size? Then, we could perform a more precise check (the one you have currently).

packages/liveblocks-core/src/crdts/LiveObject.ts
```diff
@@ -544,6 +560,32 @@ export class LiveObject<O extends LsonObject> extends AbstractCrdt {
...
+      for (const key of Object.keys(patch)) {
+        const value = patch[key];
+        if (value === undefined) continue;
+        if (!isLiveNode(value)) {
+          data[key] = value;
+        }
+      }
+
+      const newSize = new TextEncoder().encode(JSON.stringify(data)).length;
```
nimeshnayaju: We're reading the number of bytes here but I guess `.length` returns the same value as `.byteLength` on a `Uint8Array`, so this is okay!

packages/liveblocks-core/src/crdts/LiveObject.ts
nvie: Ah yes, I think that's a great idea!

nvie: Implemented in cfa6542eb21df9f064f8d9a20ecd6afce62196f4 👍

PR #2513, opened 2025-07-07
Generalize and deduplicate URL sanitization

marcbouchenoire:
I started this PR to properly sanitize URLs in comments but I noticed that we had this issue in many places and each had its own duplicated/similar solution, so I tried to generalize as much as possible.

**Backend PR:** https://github.com/liveblocks/liveblocks-backend/pull/1121


packages/liveblocks-core/src/lib/url.ts
```diff
@@ -71,3 +75,88 @@ export function url(
...
+    return sanitizedUrl !== "" ? sanitizedUrl : null;
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Construct a URL with optional parameters and hash.
+ */
+export function generateUrl(
```
marcbouchenoire, line 138: There's probably more de-duplication possible between this and other utils above (`urljoin`, `toURLSearchParams`, etc) but I wanted to stay conservative in this PR to avoid doing too many things at once.

packages/liveblocks-core/src/lib/url.ts
```diff
@@ -71,3 +75,88 @@ export function url(
...
+ *
+ * Accepted URLs:
+ * - Absolute URLs with an http or https protocol (e.g. https://liveblocks.io)
+ * - Absolute URLs with a `www` prefix (e.g. www.liveblocks.io)
+ * - Relative URLs (e.g. /path/to/page)
+ *
+ * The presence/absence of trailing slashes is preserved.
+ * Rejected URLs are returned as `null`.
+ */
+export function sanitizeUrl(url: string): string | null {
```
marcbouchenoire, line 90: This utility doesn't use `window.location` so it works in all environments.

packages/liveblocks-emails/src/comment-body.tsx
```diff
@@ -139,9 +139,23 @@ export async function convertCommentBody<T, U extends BaseUserMeta = DU>(
...
+
+                // If the URL is invalid, its text/URL are used as plain text.
+                if (href === null) {
+                  return options.elements.text(
+                    {
+                      element: { text: inline.text ?? inline.url },
+                    },
+                    inlineIndex
+                  );
+                }
```
SugarDarius, line 152: Can you add a test to validate this case please?

SugarDarius (approved):
Thanks 🙏🏻

PR #2514, opened 2025-07-07
Fix tool call signal not being created correctly

nimeshnayaju:
This PR updates the tool call signal reading logic so that if a tool call signal is read before it's registered, we still create a signal.


nvie (approved):
Perfect! 👌

PR #2517, opened 2025-07-09
Add tests for auto `www.` links in comments

marcbouchenoire:
This PR adds new tests to make sure auto `www.` links are kept as is and not converted to HTTPS URLs.


SugarDarius (approved):
🚀

PR #2518, opened 2025-07-09
Expose Markdown components in `AiChat`

marcbouchenoire:
This PR does two main things:
- Finalize, test, and refactor the `Markdown` primitive (but still keeping it private for now)
- Expose Markdown components in `AiChat`’s `components` prop based on the changes above

The `Markdown.tsx` changes are best ready commit-by-commit.

**Question:** Currently in `AiChat` only assistant messages use Markdown rendering, should we apply that to user messages too? It could be our form of rich text rendering (and we could make the composer a bit more Markdown-friendly over time) which is great for LLMs, and more importantly it would enable customizing paragraphs or links for example in both messages, not just the assistant ones.


packages/liveblocks-react-ui/src/styles/index.css
```diff
@@ -626,6 +626,18 @@
...
+      margin-block: 0.25em;
+    }
+
+    > :where(p:first-child) {
+      margin-block-start: 0;
+    }
+
+    > :where(p:last-child) {
+      margin-block-end: 0;
+    }
```
marcbouchenoire, line 640: Fixing an existing problem where list items with paragraphs (`<li><p>...</p></li>`) had double margins.

docs/pages/api-reference/liveblocks-react-ui.mdx
```diff
@@ -425,6 +451,31 @@ function Chat() {
...
+  </PropertiesListItem>
+  <PropertiesListItem
+    name="Loading"
+    type="ComponentType<AiChatComponentsLoadingProps>"
+  >
+    The component used to render the loading state of the chat. Defaults to a
+    loading spinner.
+  </PropertiesListItem>
+  <PropertiesListItem name="markdown" type="Partial<MarkdownComponents>">
+    The components used to render Markdown content: `Paragraph`, `Heading`,
```
CTNicholas: Could you list each one here?

I'll do a pass on this PR later

CTNicholas (approved):
Nice one, thanks for doing the docs! I don't think markdown on the user side is important.

packages/liveblocks-react-ui/src/primitives/__tests__/Markdown.test.tsx
```diff
@@ -0,0 +1,992 @@
+import "@testing-library/jest-dom";
+
+import { render } from "@testing-library/react";
+
+import { Markdown, type MarkdownComponents } from "../Markdown";
+import { dedent } from "./_utils";
+
+describe("Markdown", () => {
```
nimeshnayaju, line 8: Thanks for adding these tests. I originally wanted to copy tests from Marked.js to see if we have parity with their renderer but didn't get a chance, so this is great!!

PR #2519, opened 2025-07-09
large message strategy throw

jrowny:
Add a new message strategy that throws when the message is too large

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -1256,7 +1256,8 @@ export type RoomDelegates = Omit<Delegates<AuthValue>, "canZombie">;
 export type LargeMessageStrategy =
   | "default"
   | "split"
-  | "experimental-fallback-to-http";
+  | "experimental-fallback-to-http"
+  | "throw";
 
 /** @internal */
 export type RoomConfig<M extends BaseMetadata> = {
@@ -1760,6 +1761,10 @@ export function createRoom<
             }
           });
         return;
+      }
+
+      case "throw": {
+        throw new Error("Message is too large for websockets, not sending.");
       }
     }
   }


SugarDarius (approved):

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -199,6 +199,7 @@ const client = createClient({
     - `"split"` Break the message up into chunks each of which is smaller than the maximum message size.
       Beware that using `"split"` will sacrifice atomicity of changes! Depending on your use case, this may or may not be problematic.
     - `"experimental-fallback-to-http"` Try sending the update over HTTP instead of WebSockets (experimental).
+    - `"throw"` Will throw an error when encountering a message that's too large
```
CTNicholas:
```suggestion
    - `"throw"` Will throw an error when encountering a message that’s too large.
```

docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -400,6 +400,7 @@ function App() {
     - `"split"` Break the message up into chunks each of which is smaller than the maximum message size.
       Beware that using `"split"` will sacrifice atomicity of changes! Depending on your use case, this may or may not be problematic.
     - `"experimental-fallback-to-http"` Try sending the update over HTTP instead of WebSockets (experimental).
+    - `"throw"` Will throw an error when encountering a message that's too large
```
CTNicholas:
```suggestion
    - `"throw"` Will throw an error when encountering a message that’s too large.
```

docs/pages/api-reference/liveblocks-client.mdx
CTNicholas: Can we add a sentence at the end that explains how to handle this? e.g is it a `try/catch` around `createClient`?

docs/pages/api-reference/liveblocks-react.mdx
CTNicholas: Same as above, is this an `ErrorBoundary` around a`LiveblocksProvider`?

jrowny:
@adam-subframe I ran into a catch 22. I was wrong about the default message strategy actually preventing future messages, it just ignores that message.

Unfortunately If we throw, it is difficult to handle as it won't trigger the error boundary. If we use our `useErrorListener`, it will not actually stop future messages either, as that's a "handled" error.

What I ended up doing is changing the default behavior to notify to the `useErrorListener` so that you can handle it yourself. 
This will happen synchronously, so you can decide to simply disconnect() to prevent further updates. That code would look like this:

```
  useErrorListener((error) => {
    setError(error); // show the user a message, toast, or whatever
    if (error.context.type === "LARGE_MESSAGE_ERROR") {
      room.disconnect(); // Prevent future messages by explicitly disconnecting from the room
    }
  });
```

docs/pages/api-reference/liveblocks-client.mdx
jrowny: I ended up going in a different direction, as throw is not easily caught.

docs/pages/api-reference/liveblocks-react.mdx
jrowny: Unfortunately ErrorBoundary doesn't catch things like this inside the core, so I ended up sending a message to useErrorListener instead.

PR #2520, opened 2025-07-10
Extend and improve `AiChat` scroll behaviors 

marcbouchenoire:
- [ ] Add an option for the behavior when sending a new message
  - Scroll to new message while putting it at the viewport's top (new default)
  - Scroll to new message, keeping it at the viewport's bottom
- [ ] Add an option for the behavior when scrolled at the bottom
  - Nothing (still the default)
  - Stick to the bottom

### DIFF
--- a/packages/liveblocks-react-ui/src/components/AiChat.tsx
+++ b/packages/liveblocks-react-ui/src/components/AiChat.tsx
@@ -209,14 +209,20 @@ export const AiChat = forwardRef<HTMLDivElement, AiChatProps>(
 
       const lastSentMessageHeight = (lastSentMessage as HTMLElement)
         .offsetHeight;
+      const lastSentMessageScrollMarginTop = Number.parseFloat(
+        getComputedStyle(lastSentMessage as HTMLElement).scrollMarginTop
+      );
       const containerHeight = containerRef.current.clientHeight;
       const footerHeight = footerRef.current?.offsetHeight ?? 0;
 
       // Compute the additional height needed if we want the new message to be positioned
       // at the top of the scrollable area's viewport.
       const trailingSpacerHeight =
-        containerHeight - lastSentMessageHeight - footerHeight;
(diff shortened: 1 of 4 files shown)


SugarDarius (approved):
:chef-kiss: 🚀

packages/liveblocks-react-ui/src/components/AiChat.tsx
```diff
@@ -133,49 +414,63 @@ export const AiChat = forwardRef<HTMLDivElement, AiChatProps>(
...
+          });
+        } else {
+          const messagesHeight = messagesRef.current?.offsetHeight ?? 0;
+          const footerHeight = footerRef?.current?.offsetHeight ?? 0;
+
+          // Scroll to the bottom of the messages to exclude the trailing space.
+          container.scrollTo({
+            top: messagesHeight + footerHeight - container.clientHeight,
+            behavior,
+          });
```
nimeshnayaju: I see - `includeTrailingSpace` is `true` when a new user message is sent.

In some of my past experiments, I've tried doing something around the lines of (because we already have those elements ready to be scrolled into view):

```ts
if (includeTrailingSpace) {
  footerRef.current?.scrollIntoView({ behavior: "smooth", block: "end", inline: "nearest" })
} else {
  scrollBottomRef.current?.scrollIntoView({ behavior: "smooth" })
}
```

Have you considered something like this?

marcbouchenoire: I tried it in the past but on the messages list and I wasn't satisfied with the lack of control, but now that we have two 0px "`IntersectionObserver` markers" (one before the trailing space, one after) you're right we can definitely avoid the manual measurement with `scrollTo`, done in https://github.com/liveblocks/liveblocks/pull/2520/commits/4250c2adbd6c8fe42d14647be0e98ef4acf598b2.

PR #2527, opened 2025-07-14
`AiChat`: Instantly scroll to bottom on initial render when messages are already in cache

nimeshnayaju:
The `AiChat` component would scroll instantly to bottom once chat messages were loaded - this was thanks to the `AutoScrollHandler` component that would only be mounted when the messages were rendered, which would normally happen after `AiChat` container was mounted because messages would took a short while to be fetched.

But in case when the messages are already in cache or loaded, the `AutoScrollHandler` would mount even before the `AiChat` container, so the following code logic would not run as expected because `containerRef.current` wasn't populated.

```ts
const container = containerRef.current;
if (container === null) return;

container.scrollTo({
  top: container.scrollHeight,
  behavior,
});
```

This PR fixes that by running a `useEffect` that runs when `AiChat` is mounted and only scrolls to bottom on initial render if messages wasn't empty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 ## v3.1.2
 
-### `@liveblocks/react-ui` and `@liveblocks/emails`
+### `@liveblocks/react-ui`
+
+- Improve URL sanitization in comments.
+- Fix `AiChat` component not scrolling instantly to the bottom on render when
+  messages are already loaded.
+
+### `@liveblocks/emails`
 
 - Improve URL sanitization in comments.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,15 @@
 ## vNEXT (not yet published)
 
-## v3.1.2
+## v3.1.3
 
 ### `@liveblocks/react-ui`
 
-- Improve URL sanitization in comments.
 - Fix `AiChat` component not scrolling instantly to the bottom on render when
   messages are already loaded.
 
-### `@liveblocks/emails`
+## v3.1.2
+
+### `@liveblocks/react-ui` and `@liveblocks/emails`
 
 - Improve URL sanitization in comments.
 
--- a/packages/liveblocks-react-ui/src/components/AiChat.tsx
+++ b/packages/liveblocks-react-ui/src/components/AiChat.tsx
@@ -178,6 +178,13 @@ export const AiChat = forwardRef<HTMLDivElement, AiChatProps>(
     }
     const scrollToBottom = scrollToBottomCallbackRef.current;
 
+    useEffect(() => {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,13 @@
 
 ## v3.1.2
 
-### `@liveblocks/react-ui` and `@liveblocks/emails`
+### `@liveblocks/react-ui`
```
marcbouchenoire, line 5: [v3.1.2](https://github.com/liveblocks/liveblocks/releases/tag/v3.1.2) is already out, v3.2 might be far away so feel free to start v3.1.3.

packages/liveblocks-react-ui/src/components/AiChat.tsx
```diff
@@ -178,6 +178,13 @@ export const AiChat = forwardRef<HTMLDivElement, AiChatProps>(
     }
     const scrollToBottom = scrollToBottomCallbackRef.current;
 
+    useEffect(() => {
```
marcbouchenoire: It feels a bit complex to have this here and keep the same `scrollToBottom("instant")` call in `AutoScrollHandler`, maybe we should remove `AutoScrollHandler` and bring everything here.

This effect would get a dependency of `Boolean(messages)` and would handle both `scrollToBottom("instant")` cases.

marcbouchenoire (approved):
I'm bit confused by why an effect would run before refs are assigned (even if it's in a nested component) 🤔

Also I'm changing this side of things heavily in https://github.com/liveblocks/liveblocks/pull/2520 so I'll make sure to replicate the fix there.

nimeshnayaju:
>I'm bit confused by why an effect would run before refs are assigned (even if it's in a nested component) 🤔

I did more digging after reading your comment. It was because, we were using `useLayoutEffect` and it turned out that `containerRef.current` wasn't populated in the layout effect of `AutoScrollHandler`. Changing it to `useEffect` worked. I couldn't find documentation around this, but this is surprising to me too.

```tsx
function AiChat() {
  const refA = useRef(null);

  return (
    <div ref={refA}>
      <A1 />
    </div>
  )
}

function A1() {
    useLayoutEffect(() => {
      // refA.current is null??
    }, [])

   return <></>
}
```

packages/liveblocks-react-ui/src/components/AiChat.tsx
nimeshnayaju: I ended up reverting the change and using `useEffect` instead of `useLayoutEffect` inside of `AutoScrollHandler`.

CHANGELOG.md
nimeshnayaju: Thanks! Just changed the changelog

PR #2531, opened 2025-07-18
Allow keeping Comments composers focused on submit

marcbouchenoire:
This PR adds a way to keep Comments composers focused on submit:
- as `blurOnSubmit={false}` on `Composer`
- as `blurComposerOnSubmit={false}` on `Thread`
- as `blurOnSubmit={false}` on `Composer.Form` primitive
  - It might look a bit weird to have this on the form rather than the editor, the reason for this is that the form holds most of the logic for the editor including the `blur` method so we would need some awkward child → parent communication

https://github.com/user-attachments/assets/84ce9c90-2f14-4ab5-bded-5f1289fea20e

Can you think of a better API/name? I wasn't satisfied with my other ideas for this.

We could later do the same for `AiChatComposer` if needed, but defaulting to `false` this time.


LuminaX-alt:
authorization and passkey can be added i9f it is used by any organization for their employee

PR #2533, opened 2025-07-18
Pass 'copilotId' to `set-tool-call-result` command

nimeshnayaju:
Fix copilot id not being passed to 'set-tool-call-result' command that is dispatched when a tool call is responded to. Previously, we were using the default copilot to generate messages from the tool call result.

See `generationOptions` in the screenshot:

Screenshot 2025-07-18 at 3 20 49 PM

### DIFF
--- a/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
+++ b/packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
@@ -49,11 +49,20 @@ export interface AiChatAssistantMessageProps extends ComponentProps<"div"> {
    * Override the component's components.
    */
   components?: Partial<GlobalComponents>;
+
+  /**
+   * @internal
+   * The id of the copilot to use to set tool call result.
+   */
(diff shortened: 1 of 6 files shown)


packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
```diff
@@ -49,11 +49,20 @@ export interface AiChatAssistantMessageProps extends ComponentProps<"div"> {
    * Override the component's components.
    */
   components?: Partial<GlobalComponents>;
+
+  /**
+   * @internal
+   * The id of the copilot to use to set tool call result.
+   */
+  copilotId?: string;
```
nimeshnayaju, line 57: I added an internal `copilotId` property to `AiChatAssistantMessage` component so that we can use this copilot id when setting tool call result. When using `AiChat` component, we pass the copilot id passed to the `AiChat` component to this component, so no change is required by the user.

jrowny (approved):

PR #2534, opened 2025-07-21
Prepare `AiComposer` primitive

marcbouchenoire:
- [x] Rename `AiChatComposer` to `AiComposer`
  - It's still mostly coupled to the concept of chats (e.g. being abortable, etc) but imo this name makes it easier to have a standalone version that's not part of `AiChat` (`AiComposer` feels clear enough in the same way that `Composer` wasn't called `CommentsComposer`)
- [x] Bring to parity with Comments' `Composer`
  - [x] Supporting async logic in `onComposerSubmit`
  - [x] A public hook to get access to some the composer's state/methods
  - [x] Accessibility-related attributes
  - [x] Focus-related attributes
- [x] Add missing public APIs to make it usable as a primitive
  - [x] Add `AiComposer.Abort` button and `abort` method in `useAiComposer`
  - [x] Add `canAbort` state in `useAiComposer` to swap between the submit and abort buttons
  - [x] Allow using `useSendAiMessage` with `onComposerSubmit` to handle message submission
    - `useSendAiMessage` should return the optimistically-created user message just created (like the Comments hooks), I'm exploring this in https://github.com/liveblocks/liveblocks/pull/2556. After an exploration in that other PR, I brought this specific change here in https://github.com/liveblocks/liveblocks/pull/2534/commits/cf822e5d26ee872a81c5c98295696a86776ba68b.

### Changes visible to the public for now

There are some breaking changes that I feel okay releasing given their nicheness, does anyone feel differently?

- (Breaking) Composer overrides are renamed:
  - `AI_CHAT_COMPOSER_PLACEHOLDER` → `AI_COMPOSER_PLACEHOLDER`
  - `AI_CHAT_COMPOSER_SEND` → `AI_COMPOSER_SEND`
  - `AI_CHAT_COMPOSER_ABORT` → `AI_COMPOSER_ABORT`
- (Breaking) Classes within the default composer changing:
  - `lb-ai-chat-composer-form` → `lb-ai-composer-form`, `lb-ai-chat-composer-editor` → `lb-ai-composer-editor`, etc.
  - But currently the composer only exists within `AiChat` and its top-level classes still contain `lb-ai-chat-composer` (`lb-root lb-ai-composer lb-ai-chat-composer`), it's an `lb-ai-composer` within `lb-ai-chat` so it can keep `lb-ai-chat-composer`.
- `useSendAiMessage` returns the optimistically-created user message instead of `void`

### Main areas for feedback

It's a noisy PR and the file name changes don't help so here are the most important areas for questions/feedback imo:
- The name change to `AiComposer`
- The APIs to allow building an abort button
- `useSendAiMessage` returning its optimistically-created message

### To do

- [ ] Update documentation
- [x] Update changelog
- [x] Check if examples use those classes

### DIFF
--- a/e2e/next-ai-kitchen-sink/app/styles/page.tsx
+++ b/e2e/next-ai-kitchen-sink/app/styles/page.tsx
@@ -591,7 +591,7 @@ export default function Home() {
           </div>
         </div>
         <div className="lb-ai-chat-footer">
-          <div className="lb-root lb-ai-chat-composer lb-ai-chat-composer-form lb-elevation lb-elevation-moderate">
+          <div className="lb-root lb-ai-chat-composer lb-ai-composer-form lb-elevation lb-elevation-moderate">
             {/* Debug panel */}
             <div
               style={{
--- a/e2e/next-ai-kitchen-sink/test/knowledge-isolation.test.ts
+++ b/e2e/next-ai-kitchen-sink/test/knowledge-isolation.test.ts
@@ -3,10 +3,10 @@ import type { Page } from "@playwright/test";
 
 async function setupChatA(page: Page) {
   const textInput = page.locator(
-    '[data-testid="chat-a"] .lb-ai-chat-composer-editor'
+    '[data-testid="chat-a"] .lb-ai-composer-editor'
   );
   const sendButton = page.locator(
-    '[data-testid="chat-a"] .lb-ai-chat-composer-action'
+    '[data-testid="chat-a"] .lb-ai-composer-action'
   );
 
   // If there's an ongoing operation (abort button is enabled), click it first to clear state
@@ -24,10 +24,10 @@ async function setupChatA(page: Page) {
 
 async function setupChatB(page: Page) {
   const textInput = page.locator(
-    '[data-testid="chat-b"] .lb-ai-chat-composer-editor'
+    '[data-testid="chat-b"] .lb-ai-composer-editor'
   );
   const sendButton = page.locator(
-    '[data-testid="chat-b"] .lb-ai-chat-composer-action'
+    '[data-testid="chat-b"] .lb-ai-composer-action'
   );
 
   // If there's an ongoing operation (abort button is enabled), click it first to clear state
@@ -133,12 +133,12 @@ test.describe("AiChat Knowledge Isolation", () => {
 
     // Wait for the AiChat components to be fully loaded
     await expect(
-      page.locator('[data-testid="chat-a"] .lb-ai-chat-composer')
+      page.locator('[data-testid="chat-a"] .lb-ai-composer')
     ).toBeVisible({
       timeout: 10000,
     });
     await expect(
-      page.locator('[data-testid="chat-b"] .lb-ai-chat-composer')
+      page.locator('[data-testid="chat-b"] .lb-ai-composer')
     ).toBeVisible({
       timeout: 10000,
     });
@@ -172,12 +172,12 @@ test.describe("AiChat Knowledge Isolation", () => {
 
     // Wait for components to load
     await expect(
-      page.locator('[data-testid="chat-a"] .lb-ai-chat-composer')
+      page.locator('[data-testid="chat-a"] .lb-ai-composer')
     ).toBeVisible({
       timeout: 10000,
     });
     await expect(
-      page.locator('[data-testid="chat-b"] .lb-ai-chat-composer')
+      page.locator('[data-testid="chat-b"] .lb-ai-composer')
     ).toBeVisible({
       timeout: 10000,
     });
@@ -230,12 +230,12 @@ test.describe("AiChat Knowledge Isolation", () => {
 
     // Wait for the AiChat components to be fully loaded
     await expect(
-      page.locator('[data-testid="chat-a"] .lb-ai-chat-composer')
+      page.locator('[data-testid="chat-a"] .lb-ai-composer')
     ).toBeVisible({
       timeout: 10000,
     });
     await expect(
-      page.locator('[data-testid="chat-b"] .lb-ai-chat-composer')
+      page.locator('[data-testid="chat-b"] .lb-ai-composer')
     ).toBeVisible({
       timeout: 10000,
     });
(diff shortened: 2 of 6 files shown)


packages/liveblocks-react/src/types/index.ts
```diff
@@ -1279,7 +1279,7 @@ type LiveblocksContextBundleCommon<M extends BaseMetadata> = {
   useSendAiMessage(
     chatId: string,
     options?: UseSendAiMessageOptions
-  ): (message: string) => void;
+  ): (message: string) => AiChatMessage;
```
nvie: I think it might be more obvious to return an `AiUserMessage` instead.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1185,6 +1186,10 @@ function useSendAiMessage(
         lastMessageId,
         content
       );
+      const newMessage =
+        client[kInternal].ai[kInternal].context.messagesStore.getMessageById(
+          newMessageId
+        )!;
```
nvie:
```suggestion
      const newMessage =
        client[kInternal].ai[kInternal].context.messagesStore.getMessageById(
          newMessageId
        ) as AiUserMessage;
```

packages/liveblocks-react-ui/src/components/internal/AiComposer.tsx
nvie, line 1: I assume these are only renames? Kind of hard to see if any code logic was updated in here or not.

packages/liveblocks-react-ui/src/components/AiChat.tsx
```diff
@@ -567,18 +567,19 @@ export const AiChat = forwardRef<HTMLDivElement, AiChatProps>(
...
-          <AiChatComposer
+          <AiComposer
             key={chatId}
             chatId={chatId}
             copilotId={copilotId as CopilotId}
             overrides={overrides}
             autoFocus={autoFocus}
-            onUserMessageCreate={({ id }) => setLastSentMessageId(id)}
-            className={
+            onComposerSubmitted={({ id }) => setLastSentMessageId(id)}
```
nvie, line 571: Better prop name 👍

packages/liveblocks-react-ui/src/components/internal/AiComposer.tsx
```diff
@@ -0,0 +1,212 @@
...
+          variant="primary"
+          aria-label={$.AI_COMPOSER_SEND}
+          icon={<SendIcon />}
+        />
+      </AiComposerPrimitive.Submit>
+    </ShortcutTooltip>
+  );
+}
+
+export const AiComposer = forwardRef<HTMLFormElement, AiComposerProps>(
```
nvie, line 148: Should we consider renaming this to `InternalAiComposer` to make it more explicit? The fact that you can now use an `AiComposer` from either `src/components/internal/` and `src/components/primitives/` is a bit confusing to me, and it's not obvious to me if you're reading code that is pretty far away from the import statement.

nvie (approved):
This is all looking good to me — thanks for polishing all this! 👍

packages/liveblocks-react/src/liveblocks.tsx
marcbouchenoire: Good catch, I'll do this in https://github.com/liveblocks/liveblocks/pull/2556.

packages/liveblocks-react-ui/src/components/internal/AiComposer.tsx
marcbouchenoire: For now we could, but one possibility is that we'll make `"Internal"AiComposer` publicly usable as a standalone component from `@liveblocks/react-ui` like Comment's `Composer`, and at the same time we'll also offer `AiComposer` from `@liveblocks/react-ui/primitives` but they're two different things. We technically have this issue already with Comments: `Composer` from `@liveblocks/react-ui` isn't the same as `Composer` from `@liveblocks/react-ui/primitives`, one is a pre-built component and the other is a few unstyled components grouped as a "primitive", same for `Comment`, etc.

packages/liveblocks-react-ui/src/components/internal/AiComposer.tsx
```diff
@@ -0,0 +1,212 @@
...
+  onComposerSubmit?: (
+    message: AiComposerSubmitMessage,
+    event: FormEvent<HTMLFormElement>
+  ) => void;
+
+  /**
+   * @internal
+   * The event handler called after the composer is submitted.
+   */
+  onComposerSubmitted?: (message: AiChatMessage) => void;
```
nimeshnayaju: If we end up making both props public in future, I think we still lack some clarity. To me, just by looking at the names, it's not clear enough how `onComposerSubmit` is different from `onComposerSubmitted`. I wonder if it's possible to only have one of these props to avoid confusion 🤔

packages/liveblocks-react-ui/src/components/internal/AiComposer.tsx
marcbouchenoire: There are some logic changes as well, but yeah I wish the diff would be better than this.

- [x] Bring to parity with Comments' `Composer`
  - [x] Supporting async logic in `onComposerSubmit`
  - [x] A public hook to get access to some the composer's state/methods
  - [x] Accessibility-related attributes
  - [x] Focus-related attributes
- [x] Add missing public APIs to make it usable as a primitive
  - [x] Add `AiComposer.Abort` button and `abort` method in `useAiComposer`
  - [x] Add `canAbort` state in `useAiComposer` to swap between the submit and abort buttons
  - [x] Allow using `useSendAiMessage` with `onComposerSubmit` to handle message submission

For the parity changes, they're "porting" logic from the Comments `Composer`. For the "abort button" changes, they're still private since the `AiComposer` primitive is staying private for now.

packages/liveblocks-react-ui/src/components/internal/AiComposer.tsx
marcbouchenoire: I agree, I kept it internal like the previous `onUserMessageCreate` because I don't think we should release it.

To replicate the "storing the last ID" behavior in user-land you could use `onComposerSubmit` to override the default behavior:

```tsx
onComposerSubmit={
  (message, event) => {
    event.preventDefault();

    const message = sendAiMessage(message.text);
    setLastSentMessageId(message.id)
  }
}
```

But there are gotchas like the fact that it would not be using `AiChat`'s `knowledge`/`chatId`/etc. imo it's fine for now to not have a great solution for this because I don't think storing the last message's ID (or something similar) is a common use-case with `AiChat`, it's more something you'd need if you work with just a composer.

PR #2535, opened 2025-07-21
Fix sync status not returning correct value for Y.js provider

nimeshnayaju:
### Description

This PR fixes a bug that was causing sync status to not return the correct value in certain situations (like deletions). The server now sends its own snapshot (state vector and delete set) on each update message, which the client uses to compare with its local snapshot. If the snapshots are identical, we say the document is synced. We do this for any loaded subdocuments too. 

An alternative approach I looked into when working on this fix is for the server to send hash of its snapshot instead of the total snapshot and compare with the hash computed from client's snapshot. It would mean slightly smaller message payload (although delete sets are quite small) but an extra step of logic, which could affect performance only slightly. The code would look something like:

```ts
const snapshot = Y.encodeSnapshot(Y.snapshot(document));
const hash = await crypto.subtle.digest('SHA-1', snapshot);
```

### What we were doing previously and why it didn't work

On each local update, we were generating a unique update id obtained by parsing the update's metadata (which included document's state vector), but state vectors alone aren't enough to determine if two clients are in sync as state vectors do not include all information to represent the state of a Y.js document (deletions are not tracked by the state vector).


packages/liveblocks-yjs/src/doc.ts
```diff
@@ -38,19 +45,42 @@ export default class yDocHandler extends Observable<unknown> {
...
     stateVector: string | null;
     readOnly: boolean;
     v2?: boolean;
+    remoteSnapshot: Uint8Array;
   }): void => {
+    this.remoteSnapshotΣ.set(
+      v2
+        ? Y.decodeSnapshotV2(_remoteSnapshot)
+        : Y.decodeSnapshot(_remoteSnapshot)
+    );
```
nimeshnayaju: When a remote update is received, we update the remote snapshot value. Similarly, when a  local update is made, we update the local snapshot value. When either value changes, we calculate the sync status and store it in the `syncStatusSignalΣ` signal. 


```ts
this.syncStatusSignalΣ = DerivedSignal.from(
  this.localSnapshotΣ,
  this.remoteSnapshotΣ,
  (localSnapshot, remoteSnapshot) => {
    if (remoteSnapshot === null) {
      return "loading";
    }
    const isEqual = Y.equalSnapshots(localSnapshot, remoteSnapshot);
    return isEqual ? "synchronized" : "synchronizing";
  }
);
```

packages/liveblocks-yjs/src/provider.ts
```diff
@@ -115,15 +132,42 @@ export class LiveblocksYjsProvider
...
+      ).map((handler) => handler.syncStatusSignalΣ.get());
+      const allSubdocsSynchronized = subdocStatuses.every(
+        (status) => status === "synchronized"
+      );
+      return allSubdocsSynchronized ? "synchronized" : "synchronizing";
+    });
+
+    this.overallSyncStatusΣ.subscribe(() => {
+      this.emit("status", [this.overallSyncStatusΣ.get()]);
+    });
```
nimeshnayaju: The overall signal is calculated by comparing sync statuses of root document and subdocuments. 

1. If the root document is loading, the overall status is also 'loading'
2. If the root document is synchronizing, the overall status is also 'synchronizing' 
3. If the root document has synchronized, we check the status for each subdocument to see if they are also synchronized - if all of them are, the overall status is 'synchronized', otherwise, it is 'synchronizing'

packages/liveblocks-yjs/src/provider.ts
```diff
@@ -30,7 +36,10 @@ export class LiveblocksYjsProvider
   public readonly awareness: Awareness;
 
   public readonly rootDocHandler: yDocHandler;
-  public readonly subdocHandlers: Map<string, yDocHandler> = new Map();
+  private readonly subdocHandlersΣ = new MutableSignal<
+    Map<string, yDocHandler>
+  >(new Map());
+  private readonly overallSyncStatusΣ: DerivedSignal<YjsSyncStatus>;
```
nimeshnayaju: I made `subdocHandlersΣ` a signal so that we can recalculate the overall synchronization status whenever a new subdocument is added or removed. The older `subdocHandlers` property in this class was public, so I had to introduce a getter and setter to maintain backwards compatibility. If you scroll all the way down in this file, you'll see:

```ts
get subdocHandlers(): Map<string, yDocHandler> {
  return this.subdocHandlersΣ.get();
}

set subdocHandlers(value: Map<string, yDocHandler>) {
  this.subdocHandlersΣ.mutate((map) => {
    map.clear();
    for (const [key, handler] of value) {
      map.set(key, handler);
    }
  });
}
```

packages/liveblocks-core/src/protocol/ServerMsg.ts
```diff
@@ -223,6 +223,7 @@ export type YDocUpdateServerMsg = {
   readonly stateVector: string | null; // server's state vector, sent in response to fetch
   readonly guid?: string; // an optional guid to identify which subdoc this update to
   readonly v2?: boolean; // whether this is a v2 update
+  readonly remoteSnapshot: string; // The snapshot of server's document. Used to detect if the client has the latest version of the document.
```
nimeshnayaju: This is a new property in `ServerMsgCode.UPDATE_YDOC` type messages that the server sends on each update. A snapshot includes state vector and delete set, so in cases when this message already includes the state vector (initial sync step), we'll have duplication, but state vectors and delete sets are both minimal values, so it shouldn't affect the message size by a lot.

packages/liveblocks-core/src/room.ts
```diff
@@ -3037,7 +3037,9 @@ export function createRoom<
 
   function yjsStatusDidChange(status: YjsSyncStatus) {
     return syncSourceForYjs.setSyncStatus(
-      status === "synchronizing" ? "synchronizing" : "synchronized"
+      status === "synchronizing" || status === "loading"
+        ? "synchronizing"
+        : "synchronized"
```
nimeshnayaju, line 3050: This is another bug fix - sync status for Y.js is either `loading`, `synchronizing` or `synchronized`. According to the previous logic, we were displaying `synchronized` for `loading` states.

nimeshnayaju:
> [!NOTE]
>  Tests are expected to fail until we merge the relevant backend changes. Will update this comment as we merge the backend code.

packages/liveblocks-core/src/protocol/ServerMsg.ts
jrowny: I'm worried about how big this may get. Have you tested it with a large document?

packages/liveblocks-core/src/room.ts
jrowny: this was an intentional decision, but I can't remember exactly why. There was a lot of discussion about it. I'll see if I can dig up that conversation.

packages/liveblocks-yjs/src/doc.ts
jrowny: I'm worried about the performance of this on old documents.

This will potentially run per key-stroke, and users are potentially sensitive to typing delays. 

I'm wondering if maybe a better approach is to debounce this.

packages/liveblocks-yjs/src/doc.ts
```diff
@@ -105,6 +135,8 @@ export default class yDocHandler extends Observable<unknown> {
     update: Uint8Array,
     origin: string | IndexeddbPersistence
   ) => {
+    this.localSnapshotΣ.set(Y.snapshot(this.doc));
```
jrowny: again, I'm worried about the perf, especially with large documents. Doing this on every single update seems rather brute force.

jrowny (changes requested):
Overall I'm not sure this is the right approach. Doing both a snapshot and a state vector potentially on every single update seems a bit heavy handed. 

It may be fast for smaller documents, or when running locally, but both the server and clients will have to continually do this work as a user types. 

State vectors don't include enough to represent the state of a document, but they DO include enough to know if there is a difference that matters. 

Comparing the remote vector to the local vector should always tell us if we're out of sync. 

The uniqueId hack we were doing before did not include an entire state vector, it only included the from/to. A state vector has ALL of the tuples, not just the fromt/to.

nimeshnayaju:
@jrowny 

>Doing both a snapshot and a state vector potentially on every single update seems a bit heavy handed.

I couldn't think of another way to reliably check if the client is in sync with server. Also, we already compute the server snapshot on each update even before this change. What's changed/added is that we now send that snapshot to the client and the client also computes its snapshot on each update + compares it with server's. This also does mean that websocket messages will now be larger, but **we could change the implementation to only send a hash** and compare hashes (client still has to compute its snapshot, hash it and compare the hash with server's hash).

I tested on a fairly large document and didn't notice any performance issues. I have reached out to Kevin, author of Y.js for his opinion on this approach. 

>State vectors don't include enough to represent the state of a document, but they DO include enough to know if there is a difference that matters.

The `isSnapshotEqual` method does perform an optimistic check first on state vector before checking equality on delete sets. But, if I understand correctly, we need both state vectors and delete sets to check if two documents are in sync.

>The uniqueId hack we were doing before did not include an entire state vector, it only included the from/to. A state vector has ALL of the tuples, not just the fromt/to.

That is true, it would also include from/to mapping from clients within a specific update. But, I don't think this necessarily tells us about server's state. Another approach I could think of is to assume client is in sync after the two step sync process and store the from/to mapping and delete set included in a local update and carefully updating it based on remote updates received and deriving the sync status based on the state of from/to mapping and delete set. But, I don't know if the approach can actually work.

>this was an intentional decision, but I can't remember exactly why. There was a lot of discussion about it. I'll see if I can dig up that conversation.

Would love to know! This is causing a bug where the sync status shows 'synchronized' during loading state.

>I'm worried about the performance of this on old documents. This will potentially run per key-stroke, and users are potentially sensitive to typing delays. I'm wondering if maybe a better approach is to debounce this.

Good point! I can add debouncing to the logic - I think it's a good idea.

nimeshnayaju:
cc: @jrowny 

From Kevin (Y.js author):

> Your approach sounds good. This is what I have been recommending for displaying a "sync status". However, I have a few notes:

> Computing a snapshot is not expensive, but it is not free. Yjs has to iterate through the whole document to find all deleted items. This probably shouldn't happen on every keystroke (or when drawing in a drawing app at 60 frames per second). I recommend computing the snapshot after a debounce.

I updated the code to update the local snapshot at a debounced rate.

> If you want to know whether the client is in-sync with the server, you don't even need to send the full snapshot - even if it is fairly small. You can simply compare the hash of the snapshot. If two Yjs documents are in-sync, they should have identical snapshots. 

I tested with sending hash of the remote snapshot instead but hash comparison in browser requires async APIs, which introduce race conditions and needs further handing 😩. Kevin mentioned, snapshots are fairly small, so I'm not too concerned about sending server's snapshot on each message. Also, Cloudflare has a hard cap on incoming websocket messages (1 MiB) but not on outgoing messages, so this definitely helps!

>FYI: Yjs v14 (currently WIP) doesn't need to iterate through the whole Yjs document to compute a snapshot. So your approach is future proof and will require less computation in the future. In Yjs v14 it is also easy to verify whether one snapshot contains more information than the other. This would allow you to take further steps (e.g., if the server has more content than the local document, then you should initiate a full sync with the server).

🙌

packages/liveblocks-core/src/room.ts
jrowny: I found the [discussion](https://github.com/liveblocks/liveblocks/pull/1784#discussion_r1658490537), and it was actually a discussion on the getStatus in yjs package, not here.

This function was actually added in https://github.com/liveblocks/liveblocks/pull/2043 and as far as I can tell, this is indeed bug. Potentially also why we see a "flash" of empty content.

nimeshnayaju: Does that mean it is safe to keep that code change?

nimeshnayaju:
After some discussion with Kevin and @jrowny, we've decided on comparing hash of server and local snapshot. We calculate local snapshot after a debounce. We feel good about the solution and will be releasing this change on Monday!

jrowny (approved):

PR #2537, opened 2025-07-21
Add tenantId to Identity type and update identify-user request body

FlowFlorent:
**@liveblocks/node**
- Adds `tenantId` as an optional parameter to the identifyUser client method

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -72,6 +72,7 @@ import { asyncConsume, runConcurrently } from "./lib/itertools";
 import { LineStream, NdJsonStream } from "./lib/ndjson";
 import { Session } from "./Session";
 import {
+  assertNonEmpty,
   assertSecretKey,
   fetchPolyfill,
   getBaseUrl,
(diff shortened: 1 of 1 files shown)


packages/liveblocks-node/src/client.ts
```diff
@@ -531,21 +531,21 @@ export class Liveblocks {
     const options = rest[0];
 
     const path = url`/v2/identify-user`;
-    const userId = typeof identity === "string" ? identity : identity.userId;
-    const groupIds =
-      typeof identity === "string" ? undefined : identity.groupIds;
 
-    assertNonEmpty(userId, "userId"); // TODO: Check if this is a legal userId value too
```
marcbouchenoire, line 538: Why do we don't check that anymore?

marcbouchenoire (commented):
If this is ready for a release later today, you can merge this PR into https://github.com/liveblocks/liveblocks/pull/2529

packages/liveblocks-node/src/client.ts
FlowFlorent: Good catch, removed by mistake

FlowFlorent: Added [here](https://github.com/liveblocks/liveblocks/pull/2537/files#diff-b7ffaad9c487753169585109f7bec10f73f1ee571f9f424f894f5966f0cac4b8R541)!

PR #2539, opened 2025-07-21
Group mentions node client

FlowFlorent:
**@liveblocks/node**

Adds group methods:
- create group
- delete group
- add members
- remove members
- get groups


SugarDarius (approved):
🚀

marcbouchenoire (approved):

PR #2542, opened 2025-07-23
Perform synchronization at a debounced interval to update synchronization status for a Y.js document

nimeshnayaju:
This PR fixes a bug that was causing sync status to not return the correct value in certain situations (like deletions). We now perform the two step sync at a debounced rate after each local update.

1. Initial status: `Idle`
2. Client performs update. Status → `LocalDirty`
3. Immediately after, another update is made. (no change in status)
5. No updates occur for 200ms; client initiates the two step sync by sending its state vector to server. Status → `Syncing`
6. Client A receives diff from server. Status → `Synced`.

If other clients make update to their document and if those updates are received while syncing is in progress, no change is made to the status.

This is an alternative solution to this [other solution](https://github.com/liveblocks/liveblocks/pull/2535).

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -3045,7 +3045,9 @@ export function createRoom<
 
   function yjsStatusDidChange(status: YjsSyncStatus) {
     return syncSourceForYjs.setSyncStatus(
-      status === "synchronizing" ? "synchronizing" : "synchronized"
+      status === "synchronizing" || status === "loading"
+        ? "synchronizing"
+        : "synchronized"
     );
   }
 
--- a/packages/liveblocks-yjs/src/doc.ts
+++ b/packages/liveblocks-yjs/src/doc.ts
@@ -1,16 +1,29 @@
+import { Signal } from "@liveblocks/core";
 import { Base64 } from "js-base64";
 import { Observable } from "lib0/observable";
 import { IndexeddbPersistence } from "y-indexeddb";
 import * as Y from "yjs";
 
+export enum SyncState {
+  Idle = "Idle",
+  LocalDirty = "LocalDirty",
+  Syncing = "Syncing",
+  Synced = "Synced",
+}
+
 export default class yDocHandler extends Observable<unknown> {
   private unsubscribers: Array<() => void> = [];
 
-  private _synced = false;
   private doc: Y.Doc;
   private updateRoomDoc: (update: Uint8Array) => void;
   private fetchRoomDoc: (vector: string) => void;
   private useV2Encoding: boolean;
+
+  public readonly _isInitialSyncCompleteΣ = new Signal(false);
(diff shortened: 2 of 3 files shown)


packages/liveblocks-core/src/room.ts
```diff
@@ -3045,7 +3045,9 @@ export function createRoom<
 
   function yjsStatusDidChange(status: YjsSyncStatus) {
     return syncSourceForYjs.setSyncStatus(
-      status === "synchronizing" ? "synchronizing" : "synchronized"
+      status === "synchronizing" || status === "loading"
+        ? "synchronizing"
+        : "synchronized"
```
nimeshnayaju, line 3050: @jrowny You mentioned the original code was intentional. Do you recall if there was a particular reason why?

packages/liveblocks-yjs/src/doc.ts
```diff
@@ -76,43 +89,63 @@ export default class yDocHandler extends Observable<unknown> {
...
+
   private updateHandler = (
     update: Uint8Array,
     origin: string | IndexeddbPersistence
   ) => {
     // don't send updates from indexedb, those will get handled by sync
     const isFromLocal = origin instanceof IndexeddbPersistence;
     if (origin !== "backend" && !isFromLocal) {
+      this._syncStateΣ.set(SyncState.LocalDirty);
       this.updateRoomDoc(update);
```
nimeshnayaju, line 142: On each local update, we set the sync status to `LocalDirty`.

packages/liveblocks-yjs/src/doc.ts
```diff
@@ -76,43 +89,63 @@ export default class yDocHandler extends Observable<unknown> {
...
+      // If there were edits made whiel we were syncing, this remote update may not necessarily have synced the document
+      if (this._syncStateΣ.get() !== SyncState.LocalDirty) {
+        this._syncStateΣ.set(SyncState.Synced);
+      }
     }
   };
 
   public syncDoc = (): void => {
-    this.synced = false;
+    this._syncStateΣ.set(SyncState.Syncing);
```
nimeshnayaju, line 101: When a sync request is made by the client, we set its sync status to `Syncing`.

packages/liveblocks-yjs/src/doc.ts
```diff
@@ -76,43 +89,63 @@ export default class yDocHandler extends Observable<unknown> {
       // now that we've sent our local and received from server, we're in sync
       // calling `syncDoc` again will sync up the documents
       this.synced = true;
+
+      // If there were edits made whiel we were syncing, this remote update may not necessarily have synced the document
+      if (this._syncStateΣ.get() !== SyncState.LocalDirty) {
+        this._syncStateΣ.set(SyncState.Synced);
```
nimeshnayaju, line 95: We set sync state to `Synced` after the server replies with its own state vector message. It is possible that some local updates are made while syncing was in flight, in which case we don't transition to `Synced` but wait for the next syncing to update the status to `Synced`.

PR #2555, opened 2025-07-30
Add prop on `Thread` to never show more than `n` comments

marcbouchenoire:
This PR introduces a new `maxVisibleComments` prop on `Thread`.

It's not set by default. It can be set as a number or for a more advanced API as an object to control which comments should become hidden.

### DIFF
--- a/packages/liveblocks-react-ui/src/components/Thread.tsx
+++ b/packages/liveblocks-react-ui/src/components/Thread.tsx
@@ -218,70 +218,61 @@ export const Thread = forwardRef(
         : findLastIndex(thread.comments, (comment) => comment.body);
     }, [showDeletedComments, thread.comments]);
     const hiddenCommentsIndices = useMemo(() => {
-      // If the "show more" button has been clicked, we don't want to hide any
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react-ui/src/components/Thread.tsx
```diff
@@ -210,6 +233,92 @@ export const Thread = forwardRef(
...
+      const maxVisibleCommentsCount =
+        typeof maxVisibleComments === "number"
+          ? maxVisibleComments
+          : maxVisibleComments?.max;
+      const visibleCommentsShow =
+        (typeof maxVisibleComments === "object"
+          ? maxVisibleComments?.show
+          : undefined) ?? "newest";
+
+      // If we explicitely want to show all comments or there's no limit set,
```
CTNicholas:
```suggestion
      // If we explicitly want to show all comments or there's no limit set,
```

CTNicholas (approved):
Very nice! I like your API here, with the three behaviours. Can't think of a better prop name.

PR #2556, opened 2025-07-30
Extend `useSendAiMessage` API (and `useCreateAiChat`'s a bit too)

marcbouchenoire:
Mutation hooks for Comments all take required parameters in their created functions not their hooks, which makes them easy to use in dynamic scenarios where the required parameters like a thread ID might not be static where the hooks are called.

```tsx
// `useCreateComment` doesn't ask for the thread ID in advance
const createComment = useCreateComment();
const comment = createComment({
  threadId: "th_xxx",
  body: {},
  attachments: [],
});

// `useCreateComment` doesn't ask for the thread and comment IDs in advance
const editComment = useEditComment();
editComment({
  threadId: "th_xxx",
  commentId: "cm_xxx",
  body: {},
  attachments: [],
});
```

`useSendAiMessage` on the other hand currently does require a chat ID in the hook. In this PR I explored an option where the `sendAiMessage` function would support an object with options instead of just the message string to create, and to go even further it would enable a scenario where `useSendAiMessage()` accepts to not have any arguments but when that's the case then `sendAiMessage` is required to be an object with a chat ID: `useSendAiMessage({ text: "Hello", chatId: "mandatory-id-here" })`.

```tsx
// ✅ The current basic API
const sendAiMessage = useSendAiMessage("this-id-is-required");
sendAiMessage("Hello, Liveblocks AI!");

// ✅ The current API with options
const sendAiMessage = useSendAiMessage("this-id-is-required", { copilotId: "co_xxx" });
sendAiMessage("Hello, Liveblocks AI!");

// 🆕 The function can override/complete options set in the hook, a `text` property is required to replace the string argument from above
const sendAiMessage = useSendAiMessage("this-id-is-required", { copilotId: "co_xxx" });
sendAiMessage({ copilotId: "co_xxx", text: "Hello, Liveblocks AI!" })

// 🆕 The hook is allowed to be empty but when that's the case, the function cannot be just a message and it needs to have a `chatId` (`text` is also still required like above)
const sendAiMessage = useSendAiMessage();
sendAiMessage({ chatId: "this-id-is-required", text: "Hello, Liveblocks AI!" })
```

Another thing I worked on is that currently the `sendAiMessage` function doesn't return anything, unlike Comments' `createComment`/`editComment`/etc which all return their optimistic change (since they're synchronous, like `sendAiMessage`). So in this branch `sendAiMessage` now returns the optimistically-created user message, it could be used for logging for example but also it would make it possible for us to only use `sendAiMessage` in `onComposerSubmit` callbacks while still keeping track of sent messages for scroll behaviors on submit (and it would allow user-land to do the same). I also did this in https://github.com/liveblocks/liveblocks/pull/2534 because I needed that change there.

```tsx
const sendAiMessage = useSendAiMessage("chat-id");
const [lastSentMessageId, setLastSentMessageId] = useState<MessageId | null>(null)

<AiComposer.Form
  onComposerSubmit={(message) => {
    const submittedMessage = sendAiMessage(message.text)

    // 🆕 This would now be possible in user-land
    setLastSentMessageId(submittedMessage.id)
  }}
>
  ...
</AiComposer.Form>
```

---

While I was there I also extended `useCreateAiChat` to allow just a chat ID.

```tsx
// ✅ The current basic API (no `title` or `metadata`)
const createAiChat = useCreateAiChat();
createAiChat({ id: "my-ai-chat" });

// 🆕 Another way to do the same as above
const createAiChat = useCreateAiChat();
createAiChat("my-ai-chat");

// Which would match `useDeleteAiChat()`
const deleteAiChat = useDeleteAiChat();
deleteAiChat("my-ai-chat");
```

---

### To do

- [x] If the API is validated, export `SendAiMessageOptions` type
- [x] Update changelog
- [x] Update docs/JSDoc

### DIFF
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -58,6 +58,7 @@ import type {
   AiChatMessagesAsyncSuccess,
   AiChatsAsyncResult,
   AiChatsAsyncSuccess,
+  CreateAiChatOptions,
   InboxNotificationsAsyncResult,
   LiveblocksContextBundle,
   NotificationSettingsAsyncResult,
@@ -1116,19 +1117,14 @@ function useAiChatSuspense(chatId: string): AiChatAsyncSuccess {
  * // Create a chat with a custom title
  * createAiChat({ id: "ai-chat-id", title: "My AI chat" });
  */
-function useCreateAiChat() {
+function useCreateAiChat(): {
+  (chatId: string): void;
+  (options: CreateAiChatOptions): void;
+} {
   const client = useClient();
 
   return useCallback(
-    (
-      options:
-        | string
-        | {
-            id: string;
-            title?: string;
-            metadata?: Record<string, string | string[]>;
-          }
-    ) => {
+    (options: string | CreateAiChatOptions) => {
       if (typeof options === "string") {
         options = { id: options };
       }
@@ -1190,7 +1186,10 @@ function useDeleteAiChat() {
 function useSendAiMessage(
   chatId: string,
   options?: UseSendAiMessageOptions
-): (message: string | SendAiMessageOptions) => AiUserMessage;
+): {
+  (text: string): AiUserMessage;
+  (options: SendAiMessageOptions): AiUserMessage;
+};
 
 /**
  * Returns a function to send a message in an AI chat.
@@ -1206,7 +1205,7 @@ function useSendAiMessage(
  * sendAiMessage({ chatId: "chat-id", text: "Hello, Liveblocks AI!", copilotId: "co_xxx" });
  */
 function useSendAiMessage(): (
-  message: WithRequired<SendAiMessageOptions, "chatId">
+  options: WithRequired<SendAiMessageOptions, "chatId">
 ) => AiUserMessage;
 
 /**
@@ -1235,9 +1234,11 @@ function useSendAiMessage(): (
 function useSendAiMessage(
   chatId?: string,
   options?: UseSendAiMessageOptions
-):
-  | ((message: string | SendAiMessageOptions) => AiUserMessage)
-  | ((message: WithRequired<SendAiMessageOptions, "chatId">) => AiUserMessage) {
+): {
+  (text: string): AiUserMessage;
+  (options: SendAiMessageOptions): AiUserMessage;
+  (options: WithRequired<SendAiMessageOptions, "chatId">): AiUserMessage;
+} {
   const client = useClient();
 
   return useCallback(
--- a/packages/liveblocks-react/src/types/index.ts
+++ b/packages/liveblocks-react/src/types/index.ts
@@ -16,6 +16,7 @@ import type {
 import type {
   AiChat,
   AiChatMessage,
+  AiUserMessage,
   AsyncError,
   AsyncLoading,
   AsyncResult,
@@ -91,6 +92,12 @@ export type SendAiMessageOptions = UseSendAiMessageOptions & {
    * The text of the message to send.
    */
   text: string;
+};
+
+export type CreateAiChatOptions = {
+  id: string;
+  title?: string;
+  metadata?: Record<string, string | string[]>;
 };
 
 export type ThreadsQuery<M extends BaseMetadata> = {
@@ -1276,15 +1283,10 @@ type LiveblocksContextBundleCommon<M extends BaseMetadata> = {
    * // Create a chat with a custom title
    * createAiChat({ id: "ai-chat-id", title: "My AI chat" });
    */
-  useCreateAiChat(): (
-    options:
-      | string
-      | {
-          id: string;
-          title?: string;
-          metadata?: Record<string, string | string[]>;
-        }
-  ) => void;
+  useCreateAiChat(): {
+    (chatId: string): void;
+    (options: CreateAiChatOptions): void;
+  };
 
   /**
    * Returns a function that deletes the AI chat with the specified id.
@@ -1315,7 +1317,10 @@ type LiveblocksContextBundleCommon<M extends BaseMetadata> = {
   useSendAiMessage(
     chatId: string,
     options?: UseSendAiMessageOptions
-  ): (message: string | SendAiMessageOptions) => AiChatMessage;
+  ): {
+    (text: string): AiUserMessage;
+    (options: SendAiMessageOptions): AiUserMessage;
+  };
 
   /**
    * Returns a function to send a message in an AI chat.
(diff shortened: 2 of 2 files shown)


packages/liveblocks-react/src/types/index.ts
```diff
@@ -1279,7 +1298,19 @@ type LiveblocksContextBundleCommon<M extends BaseMetadata> = {
   useSendAiMessage(
     chatId: string,
     options?: UseSendAiMessageOptions
-  ): (message: string) => void;
+  ): (message: string | SendAiMessageOptions) => AiChatMessage;
+  useSendAiMessage(
+    chatId?: never,
+    options?: never
+  ): (message: WithRequired<SendAiMessageOptions, "chatId">) => AiChatMessage;
```
nvie: I'd prefer to make it more obvious that this secondary way of invoking it would be without any args instead of first suggesting it takes 2 only to find out that they must only be `undefined`.

```suggestion
  useSendAiMessage(): (
    message: WithRequired<SendAiMessageOptions, "chatId">
) => AiChatMessage;
```

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1158,54 +1174,92 @@ function useDeleteAiChat() {
...
+ * const sendAiMessage = useSendAiMessage();
+ * sendAiMessage({ chatId: "chat-id", text: "Hello, Liveblocks AI!" })
+ *
+ * @example
+ * const sendAiMessage = useSendAiMessage(chatId, { copilotId: "co_xxx" });
+ * sendAiMessage("Hello, Liveblocks AI!");
+ *
+ * @example
+ * const sendAiMessage = useSendAiMessage(chatId);
+ * sendAiMessage({ copilotId: "co_xxx", text: "Hello, Liveblocks AI!" })
```
nvie: I really like all these variants. They are super intuitive to me 👍

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1158,54 +1174,92 @@ function useDeleteAiChat() {
...
  */
 function useSendAiMessage(
   chatId: string,
   options?: UseSendAiMessageOptions
-): (message: string) => void {
+): (message: string | SendAiMessageOptions) => AiChatMessage;
+function useSendAiMessage(
+  chatId?: never,
+  options?: never
```
nvie: Same here.


```suggestion
```

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1158,54 +1174,92 @@ function useDeleteAiChat() {
...
-    (message: string) => {
+    (message: string | SendAiMessageOptions) => {
+      const {
+        text: messageText,
+        chatId: messageOptionsChatId,
+        ...messageOptions
+      } = typeof message === "string" ? { text: message } : message;
+      // The `useSendAiMessage` overloads make it impossible (at the type level at least)
+      // to have no chat ID passed at all, one of the two places should have a value.
+      const resolvedChatId = messageOptionsChatId ?? chatId!;
```
nvie: I'd prefer to make this a runtime error if TS is bypassed and no chat ID is provided.

```suggestion
      const resolvedChatId = messageOptionsChatId ?? chatId ?? raise("chatId must be provided");
```

CTNicholas:
APIs looking good, great work! Could you show the other uses in the docs too?

packages/liveblocks-react/src/types/index.ts
marcbouchenoire: Done in https://github.com/liveblocks/liveblocks/pull/2556/commits/846476b69e55750d8106cbed491ee59a39c5a62e.

packages/liveblocks-react/src/liveblocks.tsx
marcbouchenoire: Done in https://github.com/liveblocks/liveblocks/pull/2556/commits/846476b69e55750d8106cbed491ee59a39c5a62e.

packages/liveblocks-react/src/liveblocks.tsx
marcbouchenoire: Done in https://github.com/liveblocks/liveblocks/pull/2556/commits/dfdad48146b620b044fd304babb757f8aa220caf.

marcbouchenoire:
> Could you show the other uses in the docs too?

@CTNicholas I wanted to have the API validated first, now done in https://github.com/liveblocks/liveblocks/pull/2556/commits/f044b1db2333d62920e8212959d027e4be5ee691.

packages/liveblocks-react/src/types/index.ts
```diff
@@ -1252,13 +1269,22 @@ type LiveblocksContextBundleCommon<M extends BaseMetadata> = {
...
-    metadata?: Record<string, string | string[]>;
-  }) => void;
+  useCreateAiChat(): (
+    options:
+      | string
+      | {
+          id: string;
+          title?: string;
+          metadata?: Record<string, string | string[]>;
+        }
```
nimeshnayaju: Nitpick: When using autocomplete, intellisense will show the term "options", so it might not be very clear that options can be equal to chat id. Could overloading this function solve this?

Also, I understand passing in 'id' only is maybe the most popular case but I do not mind keeping the arguments object only. Not strongly opinioniated though.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1158,54 +1174,131 @@ function useDeleteAiChat() {
...
-          timeout: options?.timeout,
+          stream: messageOptions.stream ?? options?.stream,
+          copilotId: (messageOptions.copilotId ?? options?.copilotId) as
+            | CopilotId
+            | undefined,
+          timeout: messageOptions.timeout ?? options?.timeout,
         }
       );
+
+      return newMessage;
```
nimeshnayaju, line 1300: I guess we are kind of required to return the message for the AiComposer PR. Also, this would match `useCreateComment`'s signature too, so it makes sense, right?

packages/liveblocks-react/src/types/index.ts
marcbouchenoire: I forgot about how often the argument names would be visible, thanks for reminding me! I improved this in https://github.com/liveblocks/liveblocks/pull/2556/commits/98800b776ec539099edccec18772526eac2a52ec, it does mean using overloads and increasing complexity slightly but imo it's not too bad and the user-facing improvements are worth it

packages/liveblocks-react/src/liveblocks.tsx
marcbouchenoire: More than being required by the other PR, it would match the Comments hooks yes and imo it makes sense even on its own

nimeshnayaju (approved):

PR #2562, opened 2025-07-31
Documentation SAML SSO & Directory Sync

ofoucherot:
### Description

Add documentation for SAML SSO and Directory Sync.


SugarDarius (commented):
Maybe we could add a paragraph about OIDC ones support?
They are supported by WorkOS but only through the dashboard. If a customer want an OIDC one we have to do it for now by hand - I think if one customer want on OIDC we could tell him to contact us to do it and we would get $$$$ for realizing the op. 

WDYT @ofoucherot?

PR #2569, opened 2025-08-04
Add E2E tests for our AI APIs

nvie:
This PR adds e2e tests, like we do for all of our other libraries, by programmatically using Playwright to interact with the AI kitchen sink application, and verifying its behavior.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -124,6 +124,47 @@ jobs:
           path: e2e/next-sandbox/playwright-report
           retention-days: 14
 
+  e2e-ai-tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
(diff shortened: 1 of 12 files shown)


marcbouchenoire (approved):
I wouldn't be against using the existing E2E app for this instead of the kitchen sink (which ideally would slowly disappear over time to be replaced by examples like Storage/Comments/etc imo), but it's a very loosely held opinion

nvie:
> I wouldn't be against using the existing E2E app for this instead of the kitchen sink (which ideally would slowly disappear over time to be replaced by examples like Storage/Comments/etc imo), but it's a very loosely held opinion

I'm not opposed to it either, but I didn't want to take on that work right now as part of that PR. But agreed, we can always merge them later.

PR #2570, opened 2025-08-05
Fixes knowledge leaking bug in AiChat

nvie:
When multiple `<AiChat />`s are mounted on screen, any knowledge passed to one would also be available in the others. This PR fixes that leakage.

(Knowledge registered via `<RegisterAiKnowledge />` will still be made available globally to any on-screen `<AiChat />` instance of course. This hasn't changed.)

Fixes LB-2236.

### DIFF
--- a/packages/liveblocks-react-ui/src/components/internal/AiChatComposer.tsx
+++ b/packages/liveblocks-react-ui/src/components/internal/AiChatComposer.tsx
@@ -1,5 +1,6 @@
 import type {
   AiChatMessage,
+  AiKnowledgeSource,
   CopilotId,
   MessageId,
   WithNavigation,
@@ -80,6 +81,13 @@ export interface AiChatComposerProps extends ComponentProps<"form"> {
    */
   copilotId?: CopilotId;
   /**
+   * The contextual knowledge to include in the chat. May be used by the
+   * assistant when generating responses. In addition to the knowledge passed
+   * in via this prop, the AiChat instance will also have access to any
+   * globally registered knowledge via <RegisterAiKnowledge />.
+   */
+  knowledge?: AiKnowledgeSource[];
+  /**
    * @internal
    */
   branchId?: MessageId;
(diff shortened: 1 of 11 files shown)


packages/liveblocks-react-ui/src/components/internal/AiChatComposer.tsx
```diff
@@ -168,6 +177,7 @@ export const AiChatComposer = forwardRef<HTMLFormElement, AiChatComposerProps>(
           {
             stream,
             copilotId,
+            knowledge: localKnowledge,
```
marcbouchenoire, line 180: With https://github.com/liveblocks/liveblocks/pull/2534, I would love for us to use the public `useSendAiMessage` API in the composer (instead of `createOptimistically`, `askUserMessageInChat`, etc),  to allow future primitives users to do the same things (and to have a single submission code path), but this option seems niche enough that we could keep it as a private option in `useSendAiMessage`.

nvie: That sounds like a great idea for sure 👍

nimeshnayaju:
@nvie I realize the PR is already merged, but I had a quick question: is knowledge scoping different from tool scoping? I know that with tools, one could call `<RegisterAiTool />` and pass in a `chatId` prop to make the tool only available for a particular chat. Is this the case for knowledge as well?

packages/liveblocks-react-ui/src/components/internal/AiChatComposer.tsx
```diff
@@ -79,6 +80,13 @@ export interface AiChatComposerProps extends ComponentProps<"form"> {
    * The id of the copilot to use to send the message.
    */
   copilotId?: CopilotId;
+  /**
+   * The contextual knowledge to include in the chat. May be used by the
+   * assistant when generating responses. In addition to the knowledge passed
+   * in via this prop, the AiChat instance will also have access to any
+   * globally registered knowledge via <RegisterAiKnowledge />.
+   */
+  knowledge?: AiKnowledgeSource[];
```
nimeshnayaju, line 89: To me, it's a little strange that we don't have to pass tools to the composer this way but we do for knowledge. Relates to this my other [comment](https://github.com/liveblocks/liveblocks/pull/2570#issuecomment-3154904726).

PR #2572, opened 2025-08-05
new e2e for node-sandbox

jrowny:
Adds a new e2e test for the liveblocks-node package


nvie (commented):
Fantastic — thanks for adding this. I just ran the e2e test locally and it works indeed.

While at this, I'm going to merge the existing (very old) e2e/node-sandbox test in here too and then rename this package back to e2e/node-sandbox, because we already have a CI job for that. A great excuse to upgrade that old setup and start adding more E2E tests for our Node APIs 😇

nvie (approved):
LGTM! I've gone ahead and merged this new project with `e2e/node-sandbox` (which we are already running on CI) as a place to test all Node related logic E2E. While at it, I've converted the project from Jest → Vitest. A great excuse to make the setup a bit more modern here.

Thanks for doing this! 👍

PR #2576, opened 2025-08-07
Add query options to `useAiChats` and `useAiChatsSuspense`

nimeshnayaju:
This PR adds query options to `useAiChats` and `useAiChatsSuspense`. AI chats can now be queried based on metadata.

#### Definition
```tsx
useAiChats(options?: { query?: { metadata?: Record<string, string | string[] | null> } })
```

Filter chats by the absence of metadata fields is also possible by passing `null` as the value. For example, to get all chats that don't have a `category` metadata field:

```tsx
useAiChats({ query: { metadata: { category: null } } })
```

The filtering works as follows:

1. **String value**: Filters for chats where the metadata key exists with that exact value
2. **Array value**: Filters for chats where the metadata key contains all specified values
3. **Null value**: Filters for chats where the metadata key is absent or null

### DIFF
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -66,13 +66,18 @@ import type {
   ThreadsAsyncResult,
   ThreadsAsyncSuccess,
   UnreadInboxNotificationsCountAsyncResult,
+  UseAiChatsOptions,
   UserAsyncResult,
   UserAsyncSuccess,
   UseSendAiMessageOptions,
   UseSyncStatusOptions,
   UseUserThreadsOptions,
 } from "./types";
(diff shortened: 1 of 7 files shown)


packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -991,9 +998,13 @@ function useAiChatsSuspense(): AiChatsAsyncSuccess {
 
   useEnsureAiConnection(client);
 
-  use(store.outputs.aiChats.waitUntilLoaded());
+  const queryKey = makeAiChatsQueryKey(options?.query);
+
+  console.log(queryKey)
```
marcbouchenoire: 👀

nimeshnayaju: Nice catch 🫢

PR #2577, opened 2025-08-07
Replace Jest by Vitest

nvie:
**Edit:** The `@liveblocks/react` one is significantly harder to convert than all the others, so I'll finish that transition in another PR focused on that, i.e. all the unchecked items below will be tackled in a follow-up.

---

These workspaces have been converted from Jest to Vitest now:

- [x] @liveblocks/core
- [x] @liveblocks/client
- [x] @liveblocks/node
- [x] @liveblocks/node-lexical
- [x] @liveblocks/node-prosemirror
- [ ] @liveblocks/react
- [x] @liveblocks/react-ui
- [x] @liveblocks/react-blocknote
- [x] @liveblocks/react-lexical
- [x] @liveblocks/react-tiptap
- [x] @liveblocks/redux
- [x] @liveblocks/zustand
- [x] @liveblocks/emails
- [x] @liveblocks/yjs
- [ ] e2e/next-sandbox
- [ ] tools/liveblocks-codemod

These workspaces have been upgraded from MSW v1 to v2:

- [x] @liveblocks/core
- [x] @liveblocks/node
- [x] @liveblocks/node-lexical
- [x] @liveblocks/node-prosemirror
- [ ] @liveblocks/react
- [x] @liveblocks/react-ui
- [x] @liveblocks/react-blocknote
- [x] @liveblocks/react-lexical
- [x] @liveblocks/react-tiptap
- [x] @liveblocks/redux
- [x] @liveblocks/zustand
- [x] @liveblocks/emails
- [x] @liveblocks/yjs

Final cleanup:

- [ ] Make sure MSW 1 is not used anywhere anymore
- [ ] Upgrade React testing library to the latest version in `@liveblocks/react`
- [ ] Remove Jest dependency from everywhere
- [ ] Remove `shared/jest-config` workspace

Post-approval, purely syntactical changes to avoid massive/unreadable diff noise:

- [x] Remove `globals: true` and use explicit imports of Vi test functions
- [ ] Move all files from `src/__tests__/*` to `test/*`

### DIFF
--- a/packages/liveblocks-emails/src/__tests__/comment-with-body.test.ts
+++ b/packages/liveblocks-emails/src/__tests__/comment-with-body.test.ts
@@ -1,10 +1,10 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, test } from "vitest";
 
 import { filterCommentsWithBody } from "../comment-with-body";
 import { commentBody1, generateThreadId, makeComment } from "./_helpers";
 
 describe("Comment with body", () => {
-  it("should filter comments with defined bodies", () => {
+  test("should filter comments with defined bodies", () => {
     const threadId = generateThreadId();
     const comment1 = makeComment({ userId: "user-0", threadId });
     const comment2 = makeComment({


nvie:
This PR is now ready for review.
@marcbouchenoire @nimeshnayaju I think you will be most interested here.

I've rewritten most of the projects. Unfortunately MSW 1 and Vitest are no good match so I could not keep those migrations fully decoupled and upgraded them to MSW 2 while at it.

The only one left is now `@liveblocks/react`. However, this one is a bit of a beast with lots and lots of timing-related failures, so I'm tempted to move this to another PR after all. At the very least we've converted _most_ of the projects to Vitest.

Once `@liveblocks/react` is converted, we should be able to remove all Jest dependencies, finally! 💆‍♂️
(And once _that_ happens, the end result should be a net negative diff, or at least a lot smaller than this one.)

packages/liveblocks-client/package.json
```diff
@@ -29,16 +29,16 @@
     "format": "(eslint --fix src/ || true) && prettier --write src/",
     "lint": "eslint src/",
     "lint:package": "publint --strict && attw --pack",
-    "test": "jest --silent --verbose --color=always --passWithNoTests",
+    "test": "NODE_OPTIONS=\"--no-deprecation\" vitest run --passWithNoTests",
```
marcbouchenoire, line 32: Is `NODE_OPTIONS="--no-deprecation"` here purely to avoid noise or do deprecation warnings get in Vitest's way?

packages/liveblocks-core/e2e/utils.ts
```diff
@@ -227,8 +227,8 @@ export function prepareSingleClientTest<S extends LsonObject>(
 
     actor.ws.pauseSend();
 
-    actor.ws.addEventListener("message", (message) => {
-      console.log(message.data);
+    actor.ws.addEventListener("message", (_message) => {
+      // console.log(message.data);
```
marcbouchenoire, line 233: Should we remove the event listener entirely?

packages/liveblocks-core/tsconfig.json
```diff
@@ -4,7 +4,8 @@
 
   "compilerOptions": {
     // TODO: Try to get rid of these overrides
-    "noUncheckedIndexedAccess": false // OVERWRITTEN to relax a bit
+    "noUncheckedIndexedAccess": false, // OVERWRITTEN to relax a bit
+    "types": ["vitest/globals"]
```
marcbouchenoire: It's not intuitive but setting `types` will stop TS from auto-importing types from `@types/*` packages: https://www.typescriptlang.org/tsconfig/#types. Maybe we could have a `.d.ts` file with `/// <reference types="vitest/globals" />` instead?

I would also be fine with the option to not use globals at all.

**Edit:** Ha I spoke too soon and missed this in the PR's description, I'm all for it 👍 

> - [ ] Remove `globals: true` and use explicit imports of Vi test functions

marcbouchenoire (approved):
Thanks for doing this! Can't wait for part II to be done ❤️

SugarDarius (approved):

nimeshnayaju (approved):
Thank you for doing this 🙏🏼

packages/liveblocks-core/e2e/utils.ts
nvie: The registration of the event listener itself is relevant for the test. Without it, the test fails.

packages/liveblocks-client/package.json
nvie: It's not related to Vitest, but is to reduce lots of console.log warnings about `punycode` being deprecated (whatever that is). I think in Jest it was only emitted once per test run, but in Vitest it's once per test file/worker so a lot more annoying.

packages/liveblocks-core/tsconfig.json
nvie: I've just removed all magic globals indeed 👍

PR #2579, opened 2025-08-08
Support custom composers in `FloatingComposer`

marcbouchenoire:
This PR adds `components={{ Composer }}` to `FloatingComposer` components for Tiptap and Lexical, to match `components={{ Thread }}` on `FloatingThreads`/`AnchoredThreads`.

Props passed to `FloatingComposer` are applied to the composer itself, not the floating/portal `div` around it, unlike `FloatingThreads`. We can't change this without a breaking change.

When calling `createThread` manually with `onComposerSubmit` in the passed custom composer, you need to tell the editor that it was created. `@liveblocks/react-tiptap` already had `editor.commands.addComment(thread.id)` and this PR adds `editor.dispatchCommand(ATTACH_THREAD_COMMAND, thread.id)` to do the same in `@liveblocks/react-lexical`. The commands in both package are already different so I didn't use the same name.

```tsx
<FloatingComposer
  components={{
    Composer: () => {
      return (
        <ComposerPrimitive.Form
          onComposerSubmit={(message, event) => {
            event.preventDefault();

            const thread = createThread({
              body: comment.body,
              attachments: comment.attachments,
              metadata: ...,
            });

            // Tiptap
            editor.commands.addComment(thread.id);

            // Lexical
            editor.dispatchCommand(ATTACH_THREAD_COMMAND, thread.id);
          }}
        >
          <ComposerPrimitive.Editor />
          {/* ... */}
        </ComposerPrimitive.Form>
      );
    },
  }}
/>;
```

The above is only needed if you use your own `onComposerSubmit` logic, because even if you use a custom composer built with primitives, you case use the incoming props' `onComposerSubmit` and the submission/attachment will work.

```tsx
<FloatingComposer
  components={{
    Composer: ({ onComposerSubmit }) => {
      return (
        <ComposerPrimitive.Form
          onComposerSubmit={onComposerSubmit}
        >
          <ComposerPrimitive.Editor />
          {/* ... */}
        </ComposerPrimitive.Form>
      );
    },
  }}
/>;
```

### DIFF
--- a/docs/pages/api-reference/liveblocks-react-lexical.mdx
+++ b/docs/pages/api-reference/liveblocks-react-lexical.mdx
@@ -1061,6 +1061,19 @@ function Toolbar() {
   >
     Override the component’s strings.
   </PropertiesListItem>
+  <PropertiesListItem
+    name="components"
+    type="Partial<FloatingComposerComponents>"
+  >
+    Override the component’s components.
+  </PropertiesListItem>
+  <PropertiesListItem
+    name="components.Composer"
+    type="(props: ComposerProps) => ReactNode"
+  >
+    Override the [`Composer`](/docs/api-reference/liveblocks-react-ui#Composer)
+    component.
+  </PropertiesListItem>
 </PropertiesList>
 
 ### FloatingThreads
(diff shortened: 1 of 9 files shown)


docs/pages/api-reference/liveblocks-react-lexical.mdx
```diff
@@ -1061,6 +1061,19 @@ function Toolbar() {
   >
     Override the component’s strings.
   </PropertiesListItem>
+  <PropertiesListItem
```
marcbouchenoire, line 1153: @CTNicholas I don't have time today (before the 3.3 release) to add detailed sections about this, feel free to do it or we can do it later.

CTNicholas (approved):

PR #2584, opened 2025-08-11
Add `filterThreads_experimental` field to TipTap extension to allow filtering of thread annotations

nimeshnayaju:
This PR adds an experimental option to `useLiveblocksExtension`, called `filterThreads_experimental` to allow users to filter thread annotations that are displayed on the TipTap editor.

https://github.com/user-attachments/assets/6c94840a-a648-49f7-a0bc-b12f446c4d1d

### DIFF
--- a/packages/liveblocks-react-tiptap/src/comments/CommentsExtension.ts
+++ b/packages/liveblocks-react-tiptap/src/comments/CommentsExtension.ts
@@ -213,6 +213,15 @@ const Comment = Mark.create({
               return;
             }
             const threadId = commentMark?.attrs.threadId as string | undefined;
+
+            const filtered = FILTERED_THREADS_PLUGIN_KEY.getState(
+              view.state
+            )?.filteredThreads;
+            if (threadId && filtered && !filtered.has(threadId)) {
+              selectThread(null);
+              return;
+            }
+
             selectThread(threadId ?? null);
           },
         },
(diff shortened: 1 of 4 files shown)


packages/liveblocks-react-tiptap/src/comments/CommentsExtension.ts
```diff
@@ -294,6 +336,85 @@ export const CommentsExtension = Extension.create<
...
+        },
+        view: (view) => {
+          const syncDom = () => {
+            const filteredThreads = FILTERED_THREADS_PLUGIN_KEY.getState(
+              view.state
+            )?.filteredThreads;
+
+            // Toggle attribute for all comment-mark spans
+            const els = view.dom.querySelectorAll<HTMLElement>(
+              "span.lb-root.lb-tiptap-thread-mark[data-lb-thread-id]"
```
marcbouchenoire: `span.lb-tiptap-thread-mark` is already precise enough imo, and maybe one day we'll move `.lb-root` to the editor instead of all elements.

```suggestion
              "span.lb-tiptap-thread-mark[data-lb-thread-id]"
```

packages/liveblocks-react-tiptap/src/types.ts
```diff
@@ -82,12 +84,13 @@ export interface AiConfiguration {
...
 }
 
-export type LiveblocksExtensionOptions = {
+export type LiveblocksExtensionOptions<M extends BaseMetadata> = {
   field?: string;
   comments?: boolean; // | CommentsConfiguration
   mentions?: boolean; // | MentionsConfiguration
   ai?: boolean | AiConfiguration;
   offlineSupport_experimental?: boolean;
+  threads_experimental?: ThreadData<M>[];
```
marcbouchenoire: Do we need to add `M` since we don't use it (as opposed to accepting `ThreadData[]`)? I'd prefer to not add `<M extends BaseMetadata>` or even `<M extends BaseMetadata = BaseMetadata>` to public types/APIs if we don't use/resurface it.

nimeshnayaju: Good idea. I've removed it as it doesn't really serve any purpose.

PR #2593, opened 2025-08-12
Complete Vitest migration for `@liveblocks/react` and remaining workspaces

nvie:
This PR is a follow-up to #2577, and will complete the Jest to Vitest migration that was started in the previous PR. The focus is on converting the more complex `@liveblocks/react` package and finishing the remaining workspaces and cleanup tasks.

## Remaining migrations

These workspaces still need to be converted from Jest to Vitest:
- [x] `@liveblocks/react`
- [x] #2597

This workspace should not need its dependency on `jest-config`:
- [x] `e2e/next-sandbox` (uses Playwright instead of Jest/Vitest)

## MSW v2 upgrade

- [x] Upgrade `@liveblocks/react` to use MSW v2 instead of v1

## Final cleanup tasks

- [x] ~Upgrade React Testing Library to the latest version in `@liveblocks/react`~  
      _I tried, but hit various issues around `@types/react` that I couldn't figure out._
- [x] Remove MSW v1 dependencies from all workspaces
- [x] Remove Jest dependencies from all workspaces
- [x] Remove the `shared/jest-config` workspace entirely
- [ ] Move test files from `src/__tests__/*` to `test/*` directory structure


packages/liveblocks-react/src/__tests__/useCreateComment.test.tsx
```diff
@@ -190,25 +185,26 @@ describe("useCreateComment", () => {
...
     expect(result.current.subscription.status).toEqual("not-subscribed");
 
-    await waitFor(() =>
+    await vi.waitFor(() =>
       expect(result.current.subscription.unreadSince).toBeNull()
     );
 
-    const comment = await act(() =>
-      result.current.createComment({
+    let comment!: CommentData;
```
marcbouchenoire, line 192: The timing differences brought by moving to Vitest/MSW2 prevent testing instant optimistic updates when using `await act()`, adding a delay in the mock works and so does this change of using `act()` synchronously. (it returns a Promise when returning a value but is synchronous when returning `void`)

I want to find a better fix though.

packages/liveblocks-react/src/__tests__/useDeleteInboxNotification.test.tsx
```diff
@@ -421,8 +436,16 @@ describe("useDeleteInboxNotification", () => {
   test("should support deleting a notification and its related thread", async () => {
     const now = new Date();
     const roomId = nanoid();
-    const thread1 = dummyThreadData({ roomId, createdAt: now, updatedAt: now });
-    const thread2 = dummyThreadData({ roomId });
```
marcbouchenoire, line 425: `dummyThreadData` has its own default comment but using a hard-coded user ID not present here, it might be the same `"user-id"` by chance but setting it is more robust.

packages/liveblocks-react/src/__tests__/useInboxNotifications.test.tsx
```diff
@@ -975,37 +953,55 @@ describe("useInboxNotificationsSuspense: error", () => {
...
-    expect(screen.getByText("Loading, yo")).toBeInTheDocument();
+    expect(screen.getByText("Loading")).toBeInTheDocument();
 
     // Wait until all fetch attempts have been done
-    await act(() => jest.advanceTimersToNextTimerAsync()); // fetch attempt 1
-    await act(() => jest.advanceTimersByTimeAsync(5_000)); // fetch attempt 2
-    await act(() => jest.advanceTimersByTimeAsync(5_000)); // fetch attempt 3
-    await act(() => jest.advanceTimersByTimeAsync(10_000)); // fetch attempt 4
-    await act(() => jest.advanceTimersByTimeAsync(15_000)); // fetch attempt 5
+    await vi.waitFor(() => expect(getInboxNotificationsReqCount).toBe(1));
```
marcbouchenoire, line 967: Doing `vi.advanceTimersToNextTimerAsync` and only relying on timers wasn't robust enough anymore.

e2e/next-sandbox/package.json
```diff
@@ -35,9 +35,7 @@
     "@eslint/compat": "^2.0.1",
     "@eslint/eslintrc": "^3.3.3",
     "@liveblocks/eslint-config": "*",
-    "@liveblocks/jest-config": "*",
```
nvie, line 38: 😍

e2e/next-sandbox/package.json
```diff
@@ -35,9 +35,7 @@
...
     "@playwright/test": "^1.55.0",
-    "@types/jest": "^29.5.14",
```
nvie, line 40: 😍

packages/liveblocks-react-flow/src/__tests__/flow.test.tsx
```diff
@@ -47,7 +47,7 @@ describe("useLiveblocksFlow", () => {
       })
     );
 
-    await waitFor(() => expect(result.current.isLoading).toBe(false));
+    await vi.waitFor(() => expect(result.current.isLoading).toBe(false));
```
nvie, line 50: While at this, I think this could be simplified even further, to:

```suggestion
    await vi.waitUntil(() => !result.current.isLoading);
```

packages/liveblocks-react-flow/vitest.config.ts
```diff
@@ -1,9 +1,16 @@
...
+  defaultLiveblocksVitestConfig({
+    test: {
+      environment: "jsdom",
+      setupFiles: ["vitest.setup.ts"],
+    },
+  }),
+  {
+    resolve: {
+      dedupe: ["react", "react-dom"],
+    },
```
nvie: Do you think we can remove this in favor of pnpm's `overrides` config once we make the switch? I.e. `{ "overrides": { "react": "$react", "react-dom": "$react-dom" } }`? This feels like a bit of a (necessary but undesirable) hack, or am I missing something?

packages/liveblocks-react-ui/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "3.18.1",
   "description": "A set of React pre-built components for the Liveblocks products. Liveblocks is the all-in-one toolkit to build collaborative products like Figma, Notion, and more.",
   "license": "Apache-2.0",
-  "author": "Liveblocks Inc.",
```
nvie, line 6: Was this accidentally removed?

packages/liveblocks-react/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "3.18.1",
   "description": "A set of React hooks and providers to use Liveblocks declaratively. Liveblocks is the all-in-one toolkit to build collaborative products like Figma, Notion, and more.",
   "license": "Apache-2.0",
-  "author": "Liveblocks Inc.",
```
nvie, line 6: Same here.

packages/liveblocks-react-flow/vitest.config.ts
marcbouchenoire: That was actually hiding an issue where `package-lock.json` had multiple React versions after some merges, but it's still very brittle (all packages use React 18.2.0 locally and this version comes from `liveblocks-devtools`'s `package.json`, super implicit) so the switch will still improve this! 🤞

packages/liveblocks-react-flow/src/__tests__/flow.test.tsx
marcbouchenoire: Since this can be applied to many places, I'm keeping it for a subsequent PR.

marcbouchenoire (approved):

PR #2594, opened 2025-08-13
Add minimal variant to `AiTool`

marcbouchenoire:
This PR introduces a new `variant` prop on `AiTool` with two options:
- `"block"` (default) is the current appearance
- `"minimal"` has a few differences (no border/padding, no status icon, shimmer when executing, etc) that make it look like the "Thinking…"/"Reasoning…" placeholders

While I was there I also tweaked a few related things like matching the "Thinking…"/"Reasoning…" placeholders' focus styles and improving the shimmer animation.

https://github.com/user-attachments/assets/bfb894e8-a511-4bc0-b4c6-c89adb2aea46

https://github.com/user-attachments/assets/5deab7a5-4469-4cb8-8174-d298a5599687

@CTNicholas I only added the new prop to the docs, no separate section about it for now.

I'll also wait for https://github.com/liveblocks/liveblocks/pull/2604 to be merged to avoid merge conflicts:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## v3.4.0
 
-### `@liveblocks/react-ui`
-
-- Add a minimal appearance to `AiTool` via a new `variant` prop.
-
 ### `@liveblocks/react`
 
 Tool calls will now stream in while under construction. This means that tools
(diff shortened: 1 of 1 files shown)


marcbouchenoire:
@CTNicholas I tried this value below yesterday and found it too visible but I was working on the shimmer itself so I was seeing it at a slower speed, maybe it's fine?

https://github.com/user-attachments/assets/4c6be94e-cbaa-47f1-923f-fb8fa16c7e19

PR #2597, opened 2025-08-14
Move `@liveblocks/codemod` to Vitest

marcbouchenoire:
We were using a built-in test helper from `jscodeshift` which was meant to use globals/Jest. The reimplementation is a bit different and imo simpler, there's now less boilerplate and it's more flexible.


tools/liveblocks-codemod/vitest.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from "vitest/config";
+
+// `@liveblocks/codemod` cannot move to ESM so we can't use
+// `defaultLiveblocksVitestConfig` from `@liveblocks/vitest-config`.
```
nvie, line 4: 😭

nvie (approved):
Amazing, thanks! 🙏

PR #2598, opened 2025-08-14
Add MFA doc page

SugarDarius:
### What's going on? 

This PR adds a small documentation page about enabling MFA in the dashboard for enterprise customers.
Thanks 🙏🏻


SugarDarius:
@CTNicholas I don't think having a "How to enable MFA" section makes sense for now as all the operations are made manually by us for now. You check it more in Notion documentation I wrote. 

I think it will be worth to add it once we will add self-serve MFA.

CTNicholas (approved):
Makes sense, thanks! Added a little more info, just in case people miss the banner.

PR #2600, opened 2025-08-14
Add `tenantId` params to all supported methods in `@liveblocks/node`

marcbouchenoire:
This PR adds `tenantId` params to all supported methods in `@liveblocks/node`.

The diff includes unrelated methods because the internal `#delete` didn't support query params, it now mirrors `#get`.


marcbouchenoire:
Moved to https://github.com/liveblocks/liveblocks/pull/2601 because of how I rewrote https://github.com/liveblocks/liveblocks/pull/2489's history.

PR #2601, opened 2025-08-14
Add `tenantId` params to all supported methods in `@liveblocks/node`

marcbouchenoire:
This PR adds `tenantId` params to all supported methods in `@liveblocks/node`.

The diff includes unrelated methods because the internal `#delete` didn't support query params, it now mirrors `#get`.

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -2052,6 +2052,15 @@ export class Liveblocks {
     return inflateRoomData(data);
   }
 
+  /**
+   * Triggers an inbox notification for a user.
+   * @param params.userId The user ID to trigger the inbox notification for.
+   * @param params.kind The kind of inbox notification to trigger.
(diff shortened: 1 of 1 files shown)


packages/liveblocks-node/src/client.ts
```diff
@@ -299,13 +299,15 @@ export type GetRoomsOptions =
 // prettier-ignore
 export type GetInboxNotificationsOptions =
   & InboxNotificationsQueryCriteria
-  & PaginationOptions;
+  & PaginationOptions
+  & { tenantId?: string };
```
nvie: Would it be more natural to add this field to the `InboxNotificationsQueryCriteria` type  instead?

marcbouchenoire: Ah, yep! I was a bit puzzled by the name (same as with `RoomsQueryCriteria`) but it seems to make sense.

PR #2604, opened 2025-08-15
Implement tool call streaming for AI protocol V5

nvie:
## Summary
- Upgraded AI protocol from V4 to V5 to support tool call streaming
- Implemented real-time streaming of tool invocations as they're being built
- Added comprehensive delta handling for text, reasoning, and tool invocation updates
- Created test page in e2e/next-ai-kitchen-sink for testing tool call streaming functionality

## Key Changes
- **Protocol upgrade**: Bumped AI WebSocket endpoint from `/ai/v4` to `/ai/v5`
- **Delta processing**: Implemented `patchContentWithDelta` function to handle streaming updates
- **Tool streaming**: Added support for `tool-stream`, `tool-delta`, and progressive tool invocation updates
- **JSON parsing**: Added `parsePartialJsonObject` utility for handling incomplete JSON during streaming
- **UI improvements**: Fixed spinner behavior during tool call receiving state
- **Test infrastructure**: Added comprehensive test suite with 30+ test cases covering all streaming scenarios

## Technical Implementation
- Tool invocations now progress through stages: `receiving` → `executing` → `executed`
- Partial JSON arguments are built progressively as deltas arrive
- Content updates are applied in-place to maintain React state consistency
- Backwards compatible with existing non-streaming tool implementations

## Test plan
- [x] All existing tests pass
- [x] New comprehensive test suite covers delta handling edge cases
- [x] Manual testing with tool call streaming test page
- [x] Verified streaming works with complex nested JSON tool arguments

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/liveblocks-core/src/lib/parsePartialJsonObject.ts
+++ b/packages/liveblocks-core/src/lib/parsePartialJsonObject.ts
@@ -1,0 +1,142 @@
+import type { JsonObject } from "./Json";
+import { tryParseJson } from "./utils";
+
+export function parsePartialJsonObject(partial: string): JsonObject {
+  partial = partial.trimStart();
+
+  if (partial.charAt(0) !== "{") {
+    // Not an object, don't even try to parse it
+    return {};
+  }
+
+  // If it's already valid JSON, return as-is
+  {
+    const quickCheck = tryParseJson(partial);
+    if (quickCheck) {
+      // Due to the '{' check above, we can safely assume it's an object
+      return quickCheck as JsonObject;
+    }
+  }
+
+  let result = partial;
+
+  // 1. Fix unclosed strings by appending a '"' if needed
+  let quoteCount = 0;
+  let escaped = false;
+  for (let i = 0; i < result.length; i++) {
+    const char = result[i];
+    if (escaped) {
+      escaped = false;
+    } else if (char === "\\") {
+      escaped = true;
+    } else if (char === '"') {
+      quoteCount++;
+    }
+  }
+
+  if (quoteCount % 2 === 1) {
+    result += '"';
+  }
+
+  // 2. Trim whitespace at the end
+  result = result.trimEnd();
+
+  // 3. If the last char is a ',', strip it
+  if (result.endsWith(",")) {
+    result = result.slice(0, -1);
+  }
+
+  // 4. If the last char is a '.', strip it
+  if (result.endsWith(".")) {
+    result = result.slice(0, -1);
+  }
+
+  // 5. Go from left to right and collect '[' and '{' in a stack
+  const stack: string[] = [];
+  let inString = false;
+  escaped = false;
+
+  for (let i = 0; i < result.length; i++) {
+    const char = result[i];
+
+    if (inString) {
+      if (escaped) {
+        escaped = false;
+      } else if (char === "\\") {
+        escaped = true;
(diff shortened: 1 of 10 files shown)


packages/liveblocks-core/src/types/ai.ts
```diff
@@ -484,7 +512,22 @@ export type AiKnowledgeSource = {
 
 // --------------------------------------------------------------------------------------------------
 
-export function appendDelta(
+/**
+ * Polyfill for Array.prototype.findLastIndex()
+ */
+export function findLastIndex<T>(
```
marcbouchenoire: We have [another of these in `@liveblocks/react-ui`](https://github.com/liveblocks/liveblocks/blob/main/packages/liveblocks-react-ui/src/utils/find-last-index.ts) but [browser support for it is really good these days](https://caniuse.com/mdn-javascript_builtins_array_findlastindex), should we bump our bundler configs to ES2023?

nvie: Ah yes, that would indeed be nice!

packages/liveblocks-core/src/lib/parsePartialJsonObject.ts
```diff
@@ -0,0 +1,142 @@
...
+    result += '"';
+  }
+
+  // 2. Trim whitespace at the end
+  result = result.trimEnd();
+
+  // 3. If the last char is a ',', strip it
+  if (result.endsWith(",")) {
+    result = result.slice(0, -1);
+  }
```
nimeshnayaju, line 47: Does this only strip the comma if it's the last character? Does that mean `{ "key": "value", }'` (and similar other partial jsons) will be parsed as `{}`?

nimeshnayaju (approved):

packages/liveblocks-core/src/types/ai.ts
nvie: I tried "just" enabling es2023, but it became a bit of a rabbit hole of having to upgrade esbuild/tsup/etc to support es2023, so I might do this in a later/dedicated PR instead 😬 For now, I've just DRYed up the polyfills.

packages/liveblocks-core/src/lib/parsePartialJsonObject.ts
nvie: Yes, but that would not be a valid JSON prefix string. This will not happen in practice.

| Input                | Output             |
| -------------------- | ------------------ |
| `{`                  | `{}`               |
| `{"`                 | `{}`               |
| `{"k`                | `{}`               |
| `{"key"`             | `{}`               |
| `{"key": `           | `{}`               |
| `{"key": "`          | `{"key": ""}`      |
| `{"key": "v`         | `{"key": "v"}`     |
| `{"key": "value"`    | `{"key": "value"}` |
| `{"key": "value", `  | `{"key": "value"}` |
| `{"key": "value", }` | `{}` (parse error) |

nvie: I've added an extensive test suite for this exact behavior, too.
**Update**: Now with [100% test coverage](https://github.com/liveblocks/liveblocks/blob/186379e90787aed6e08e8aaea375be821f9cbdc4/packages/liveblocks-core/src/lib/__tests__/parsePartialJsonObject.test.ts) even.

PR #2608, opened 2025-08-19
Re-evaluate `initialStorage` prop whenever room ID changes

nvie:
This PR changes the behavior of the "freezing" of the `initialPresence` and `initialStorage` props in `RoomProvider`.

Previously, these props would always stay frozen beyond the value they had during their first render. This behavior remains true, _except_ that they will now get re-evaluated if the room ID prop changes.

Cases like this should therefore now behave more natural:

```tsx
<RoomProvider id={roomId} initialStorage={{ roomId }} />
```

Previously the only way to work around this was to forcefully unmount/remount the RoomProvider by adding a `key={}` prop:

```tsx
<RoomProvider key={roomId} id={roomId} initialStorage={{ roomId }} />
```

This was [recognized as something we could have fixed](https://github.com/liveblocks/liveblocks/issues/560#issuecomment-1318785311) a long time ago already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## v3.4.0
 
+### `@liveblocks/react`
+
+- In `RoomProvider`, `initialPresence` and `initialStorage` now get re-evaluated
+  whenever the room ID (the `id` prop) changes.
+
 ## v3.3.4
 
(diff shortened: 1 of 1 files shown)


marcbouchenoire (approved):
Makes total sense 👌

PR #2612, opened 2025-08-20
Docs for tool call streaming

nvie:


CTNicholas (approved):
Thanks for working on this! Do the `params` stream in in the order they're defined? Or is it up to the AI to decide which order. We could make a note of this.

nvie:
> Thanks for working on this! Do the `params` stream in in the order they're defined? Or is it up to the AI to decide which order. We could make a note of this.

Args are streamed in left-to-right always.

This order **will never** happen:
```ts
{ "name": "John", "age": 42 }
{ "name": "John Doe", "age": 42 }
```

But this can:
```ts
{ "name": "John" }
{ "name": "John Doe" }
{ "name": "John Doe", "age": 4 }
{ "name": "John Doe", "age": 42 }
{ "name": "John Doe", "age": 42, "hobbies": [] }
{ "name": "John Doe", "age": 42, "hobbies": ["ch"] }
{ "name": "John Doe", "age": 42, "hobbies": ["chess"] }
```

There is no guarantee which field will show up first (it could be `age`, or `name` or `hobbies`)[^1] but once it starts filling in another field, all previous fields are considered "complete" and won't be touched anymore.

[^1]: That said, in my experience most LLMs will emit the fields in the order they are defined in the schema. But there isn't a guarantee for that.

nvie:
@CTNicholas Do you think I should add this explanation in there somehow? If so... could I take my comment from here and just put it in there in that same section, or do you have a better suggestion?

AbdulQadeerDeveloper:
> _ No description provided. _

Great Ideal

PR #2613, opened 2025-08-20
Implement node methods for managing AI copilots and knowledge source + Documentation

nimeshnayaju:
This PR implements node methods (`@liveblocks/node`) for managing AI copilots and knowledge source, including the following methods:

1. `getAiCopilots`
2. `createAiCopilot` 
3. `getAiCopilot`
4. `updateAiCopilot`
5. `deleteAiCopilot`
6. `createWebKnowledgeSource`
7. `createFileKnowledgeSource`
8. `deleteFileKnowledgeSource`
9. `deleteWebKnowledgeSource`
10. `getKnowledgeSources`
11. `getKnowledgeSource`
12. `getFileKnowledgeSourceMarkdown`
13. `getWebKnowledgeSourceLinks`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## vNEXT (not yet published)
 
-## v3.4.3
+## v3.5.0
 
 ### `@liveblocks/node`
 


docs/pages/api-reference/liveblocks-node.mdx
```diff
@@ -2400,6 +2400,379 @@ await liveblocks.deleteNotificationSettings({
...
+
+Pagination options are available to control the number of results returned.
+
+```ts
+const { data: copilots, nextCursor } = await liveblocks.getAiCopilots({
+  // Optional, the amount of copilots to load, between 1 and 100, defaults to 20
+  limit: 20,
+
+  // Optional, cursor used for pagination, use `nextCursor` from the previous page's response
+  startingAfter: "L3YyL3Jvb21z...",
```
jrowny, line 2432: shouldn't this just be cursor? 

In the comments for room pagination, it supports both and says to use cursor

packages/liveblocks-node/src/client.ts
```diff
@@ -2376,6 +2551,316 @@ export class Liveblocks {
...
+   */
+  public async getAiCopilots(
+    params: PaginationOptions = {},
+    options?: RequestOptions
+  ): Promise<Page<AiCopilot>> {
+    const res = await this.#get(
+      url`/v2/ai/copilots`,
+      {
+        limit: params.limit,
+        startingAfter: params.startingAfter,
```
jrowny, line 2570: i'm relatively sure it just accepts cursor, and not starting after. That's because there was this comment:

```// TODO: Remove this field as we're renaming it to `cursor`, which is more appropriate for the purpose```

in the getRooms api endpoint.

File: packages/liveblocks-node/src/client.ts [hunk 3]
```diff
@@ -323,6 +358,118 @@ export type UpsertRoomOptions = {
...
+} & PaginationOptions;
+
+type KnowledgeSourcePlain = DateToString<KnowledgeSource>;
+
+export type KnowledgeSource = (
+  | {
+      type: "ai-knowledge-web-source";
+      link: {
+        url: string;
+        type: "individual_link" | "crawl" | "sitemap";
```
jrowny, line 443: this enum is used quite a bit, do we have it on the frontend? maybe we can pull it into core?

packages/liveblocks-node/src/client.ts
```diff
@@ -416,7 +591,7 @@ export class Liveblocks {
     return await fetch(url, {
       method: "PUT",
       headers,
-      body,
+      body: new Uint8Array(body),
```
jrowny: is this required as the input type is already a Uint8Array? Does this affect yjs functions at all?

nimeshnayaju: I didn't mean to upload this change - it was because TypeScript was complaining; I'll revert this particular change. Thanks for noticing

docs/pages/api-reference/liveblocks-node.mdx
nimeshnayaju: Checked again, and the backend indeed accepts `cursor` instead of `startingAfter`. Do you think we should start using the term `cursor` starting from this endpoint?

jrowny: If the comment I saw on getRooms is true, yes, I think we only support cursor going forward as it's "marked for deprecation" and this is a new API. Maybe @nvie has an opinion?

PR #2616, opened 2025-08-22
Allow hiding `@` symbol in default components' mentions 

marcbouchenoire:
This PR wraps `@` symbols rendered in mentions within their own `span` so that they can be styled independently (to be hidden for example), and adds a few shared CSS classes across our packages.

**Context:** https://liveblocks.slack.com/archives/C03NG9R49HC/p1755702714607269

```css
/* Hide `@` in mentions everywhere (comments, composers, notifications, text editors, etc.) */
.lb-mention-symbol {
  display: none;
}

/* Only hide `@` in mentions within comments */
.lb-comment-mention .lb-mention-symbol {
  display: none;
}
```


CTNicholas (approved):
Good idea!

PR #2617, opened 2025-08-22
Optimize AI tool call streaming with incremental JSON parser

nvie:
This PR:

- Replaces O(n²) `parsePartialJsonObject` with O(n) stateful `IncrementalJsonParser`
- Eliminates full re-parsing on each streaming delta for large JSON tool calls
- Maintains 100% backward compatibility with existing behavior
- Even found a few more optimizations in edge cases, like parsing `{"a":1,"b":n` to `{"a":1,"b":null}` now (instead of `{"a":1}`)

This new implementation has significantly less parsing/scanning overhead, so we can expect some performance improvements, noticable only for very large tool call arguments (10KB+). I used the new stress test in the sink to get some anecdotal comparisons and it's a slight improvement and for _most_ tool calls it's not a huge difference, but for very large tool calls it can start to matter. I think the real perf benefits will actually happen from throttling and grouping incoming deltas. Each delta currently triggers a full rerender, which I would like to decouple a bit.

### Perf numbers

I ran some tests to measure the performance difference:

- **Small tool calls**: 1.6x faster than previous implementation
- **Medium tool calls**: 2.6x faster than previous implementation
- **Large tool calls**: 6.3x faster than previous implementation
- **Character-by-character streaming**: 2.2x faster than previous implementation


packages/liveblocks-core/src/types/ai.ts
```diff
@@ -603,3 +592,32 @@ export function patchContentWithDelta(
       return assertNever(delta, "Unhandled case");
   }
 }
+
+/**
+ * Creates a receiving tool invocation part for testing purposes.
+ * This helper eliminates the need to manually create fake tool invocation objects
+ * and provides a clean API for tests.
+ */
+export function hydrateReceivingToolInvocation(
```
marcbouchenoire: The term "hydrate" isn't obvious here to me, is that a common use of it? And since it's purely test-related, could it be a test utility (not in a source file) like other `createDummy*` utilities?

packages/liveblocks-core/src/lib/__tests__/IncrementalJsonParser.test.ts
```diff
@@ -0,0 +1,404 @@
+import fc from "fast-check";
+import { describe, expect, test } from "vitest";
+
+import { IncrementalJsonParser } from "../IncrementalJsonParser";
+
+// Helper to express texts more briefly
```
marcbouchenoire:
```suggestion
// Helper to express tests more briefly
```

packages/liveblocks-core/src/lib/__tests__/IncrementalJsonParser.test.ts
```diff
@@ -0,0 +1,404 @@
...
+    expect(parse('{"a":1,"b":n')).toEqual({ a: 1, b: null });
+  });
+
+  test("does not complete keywords in strings", () => {
+    // This should not complete "n" to "null" because it's inside a string
+    expect(parse('{"key":"n')).toEqual({ key: "n" });
+  });
+
+  test("fixes the original counterexample", () => {
+    // This was the counterexample that revealed the issue
```
marcbouchenoire: Could this be more specific? I'm not sure what the original counterexample or the issue are

packages/liveblocks-core/src/lib/IncrementalJsonParser.ts
```diff
@@ -0,0 +1,259 @@
...
+    if (this.#inString) {
+      return ""; // Don't complete anything if we're in an unterminated string
+    }
+
+    const lastChar = output.charAt(output.length - 1);
+
+    // Handle incomplete negative numbers
+    if (lastChar === "-") {
+      return "0"; // Complete to -0
+    }
```
marcbouchenoire, line 101: Should this be here in `#autocompleteKeywords`?

packages/liveblocks-core/src/lib/IncrementalJsonParser.ts
```diff
@@ -0,0 +1,259 @@
...
+
+    // Quick heuristic: only check partial keywords if last char could be part of one
+    if (NULL_KEYWORD_CHARS.has(lastChar)) {
+      if (output.endsWith("nul")) {
+        return "l";
+      } else if (output.endsWith("nu")) {
+        return "ll";
+      } else if (output.endsWith("n")) {
+        return "ull";
+      }
```
marcbouchenoire: I wonder if regexes start catching up to `endsWith`'s performance if we do a single regex match vs multiple `endsWith`

marcbouchenoire (approved):
Lovely ⚡️

packages/liveblocks-core/src/lib/IncrementalJsonParser.ts
nvie: It should, but I added it after naming the method. Maybe rename to just `#autocomplete` then?

packages/liveblocks-core/src/lib/__tests__/IncrementalJsonParser.test.ts
nvie: LOL I also don't know anymore.

packages/liveblocks-core/src/types/ai.ts
nvie: Maybe we should call this "deserialize"? That would also work for me. Or "inflate" or something. Problem is that all words already mean something in someones context 😐 

First I had `createReceivingToolInvocation()` but it felt like hiding an important detail. It's really subtle. We're basically turning a "static" object from the network into a "stateful" object here. It's extra subtle because the types look alike so much otherwise. (In fact, they are not different types, which bothers me, but the alternative is that we duplicate ALL types into a "network layer/static" vs a "runtime/living" type 😬)

I'm sometimes thinking we should make our network message types very different/smaller for this reason (and for bandwidth reasons), i.e. if we'd send this as `['H','name','{"hello":"worl']` (or something binary) over the wire, this transformation would be more obvious maybe. It would add a whole layer or almost-similar-but-slightly-different types. Our UiMessage vs AiMessage types are another example of this. At one point I think we have to bite the bullet and do it, but I didn't want to introduce that big change now just for this.

nvie: I've changed this back to `create` for lack of a better verb to express this concept. Fixed in d5c02994b2d6ab633fe25c993631d514739a1aef.

packages/liveblocks-core/src/lib/__tests__/IncrementalJsonParser.test.ts
nvie: Removed this comment 👍

packages/liveblocks-core/src/lib/IncrementalJsonParser.ts
nvie: Renamed to `autocompleteTail` in 914f8d8b46c93dadc3df2dda50a21bee41e57d66.

packages/liveblocks-core/src/lib/IncrementalJsonParser.ts
nvie: I gave it a try, and found these results:

- **Small tool calls:** 0.8x **slower** (!) than previous implementation (instead of 1.6x faster)
- **Medium tool calls:** 1.4x faster than previous implementation (instead of 2.6x faster)
- **Large tool calls:** 2.0x faster than previous implementation (instead of 6.3x faster)
- **Character-by-character streaming:** 2.1x faster than previous implementation (instead of 2.2x faster)

So using regexes here is significantly slower than using the `.endsWith()` approach.

marcbouchenoire: Thanks for making sure 🙏

PR #2618, opened 2025-08-25
Update `useSendAiMessage` to use the last used copilot id when no copilot id is provided

nimeshnayaju:
This PR updates `useSendAiMessage` to use the the last used copilot id when no copilot id is provided.


packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1274,6 +1275,12 @@ function useSendAiMessage(
         .getChatMessagesForBranchΣ(resolvedChatId)
         .get();
 
+      const resolvedCopilotId = (messageOptionsCopilotId ??
+        options?.copilotId ??
+        client[kInternal].ai.getLastUsedCopilotId(resolvedChatId)) as
+        | CopilotId
+        | undefined;
```
nimeshnayaju, line 1298: This is where we resolve the copilot id. If one is not provided in the hook or the `sendAiMessage` function, we retrieve the last used copilot id in the chat.

packages/liveblocks-core/src/ai.ts
```diff
@@ -752,10 +756,20 @@ function createStore_forChatMessages(
       .getOrCreate(branch || null);
   }
 
+  function getLastUsedCopilotId(chatId: string): CopilotId | undefined {
+    const pool = messagePoolByChatIdΣ.getOrCreate(chatId).get();
+    // Find the most recent non-deleted assistant message
+    const latest = pool.sorted.findRight(
+      (m) => m.role === "assistant" && !m.deletedAt
+    );
+    return latest?.copilotId;
```
nimeshnayaju, line 767: I've updated assistant messages's signature so that they also come with `copilotId` now. In order to get the last used copilot id in a chat, we traverse the chat tree starting from the most recent message until we find a non-deleted assistant message.

Note that this doesn't handle branches yet (because `useSendAiMessage` also doesn't support branch yet) to keep it simple, but adding branch support should be simple.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1274,6 +1275,12 @@ function useSendAiMessage(
         .getChatMessagesForBranchΣ(resolvedChatId)
         .get();
 
+      const resolvedCopilotId = (messageOptionsCopilotId ??
```
nimeshnayaju, line 1294: @marcbouchenoire Were you thinking of a warning when this function was called, if a copilot id wasn't passed in the messages options (either directly through this function argument or via the  hook argument)?

I was thinking it could be nice to display a warning as soon as the hook was mounted, but I realized one could pass copilot id in the `sendAiMessage` function returned by the hook too.

packages/liveblocks-core/src/ai.ts
```diff
@@ -552,8 +556,8 @@ function createStore_forChatMessages(
                   message.chatId,
                   message.id,
                   toolInvocation.invocationId,
-                  result ?? { data: {} }
-                  // TODO Pass in AiGenerationOptions here, or make the backend use the same options
+                  result ?? { data: {} },
+                  { copilotId: message.copilotId } // TODO: Should we pass the other generation options (tools, knowledge) as well?
```
nimeshnayaju, line 562: @nvie Was this the reason behind the bug that Chris reported? That we were not passing copilot id when setting tool call result if a tool call was defined with an `execute` callback?

packages/liveblocks-react/src/liveblocks.tsx
marcbouchenoire: Yes with the new signatures I don't think we have a choice but to warn when the function is called.

packages/liveblocks-core/src/ai.ts
marcbouchenoire, line 562: Should we update [`AiMessage.ToolInvocation`](https://github.com/liveblocks/liveblocks/blob/ce56343674cc687f7a53d9629392e92d69c41949/packages/liveblocks-react-ui/src/primitives/AiMessage/tool-invocation.tsx#L66) in the same PR?

nvie: Yes, I think so!

packages/liveblocks-core/src/ai.ts
nvie: Looks good to me, at least for now. Like I shared on our call just now, I think maybe we will later need `getLastUsedGenerationOptions` almost here…

packages/liveblocks-react/src/liveblocks.tsx
nvie: Yeah, I'd even be for making it a mandatory option now. Either `copilotId: "builtin"` or `copilotId: "co_abslsahlaskjl"` but no longer allow `copilotId: undefined` here. If we do, we should bump the protocol to V6, which is no problem.

nvie (approved):

packages/liveblocks-core/src/ai.ts
nimeshnayaju: I have updated the component to not require `copilotId` anymore. It uses the copilot id from the message it was part of ([commit](https://github.com/liveblocks/liveblocks/pull/2618/commits/af74a0e2081f95e8f70bca52124759a237df1a8f)).

As a result, I've removed `copilotId` internal field. Are you okay with that @marcbouchenoire - I don't reckon anyone was relying on it; also it was marked as internal.

packages/liveblocks-react/src/liveblocks.tsx
nimeshnayaju: @nvie Are you suggesting we pass `built-in` instead of `undefined` (or nothing)? One thing I like about not passing anything is that we can decide from server what we use for the built in term, so it's more flexible to changes. But, I'm not too attached to that idea.

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1274,6 +1275,17 @@ function useSendAiMessage(
         .getChatMessagesForBranchΣ(resolvedChatId)
         .get();
 
+      if (!messageOptionsCopilotId && !options?.copilotId) {
+        console.warn(
+          "No copilot ID was provided to `useSendAiMessage`. This will result in the last used copilot ID being used. This is not recommended, as it may result in unexpected behavior. Please provide a copilot ID to `useSendAiMessage` or `useSendAiMessage(chatId, options)`."
+        );
```
nimeshnayaju: @marcbouchenoire Do you have opinions on what blurb to use here? Should we include the message text being sent too for more context?

marcbouchenoire: I like the idea of including the text yeah, maybe something like:

> No copilot ID was provided to `useSendAiMessage` when sending "{theFirst20CharactersOfTheTextForThisCall}…". This will make the message use the chat's previous copilot ID which may result in unexpected behavior. Provide a copilot ID via the hook as `useSendAiMessage("{theChatIdUsedForThisCall}", { copilotId: "co_xxx" })` or via the function as `sendAiMessage({ text: "{theFirst20CharactersOfTheTextForThisCall}…", copilotId: "co_xxx" })` to ensure which copilot ID is used.

cc: @CTNicholas 

Regarding the warning, I think it should be only during development and not something end user-facing in production.

packages/liveblocks-core/src/ai.ts
marcbouchenoire: No that's perfect! It'll be easier to make it public later.

packages/liveblocks-react/src/liveblocks.tsx
marcbouchenoire: I think requiring `"builtin"` everywhere defeats its purpose of being for ease of setup 🤔 With the warning implemented below and the better fallback this PR provides I feel like we're already reducing the gotchas enough to still keep the basic setup easy to start with. I'm not counting the fact that it would be a breaking change on this hook and `AiChat` as an issue because we're early enough.

PR #2622, opened 2025-08-27
Fix partial Markdown not enabled in text parts

marcbouchenoire:
The new partial Markdown rendering (added in 3.4) was not enabled for text parts in assistant messages, only reasoning parts 🙈


CTNicholas (approved):
Thanks for fixing!

PR #2638, opened 2025-09-01
Throttle incoming deltas during streaming

nvie:
This PR adds delta throttling to our message streaming to prevent excessive re-renders during fast-streamed responses. Incoming deltas are now buffered and processed in batches every 25ms, providing a ceiling of max 40 rerenders/second, instead of the previous uncapped. I tried various throttles between 16-250ms, but found that going to 16ms only increased the number of renders without any more observable fluency for text based content, so I settled for 25ms.

Other events (settle, sync, errors, etc) flush any pending deltas before processing to maintain consistency.

For a large stream of data, this can reduce the number of rerenders significantly, while not sacrificing any fluency:

| Before | After |
|--------|--------|
| Screen Shot 2025-09-01 at 12 05 35@2x | Screen Shot 2025-09-01 at 12 04 59@2x |

### DIFF
--- a/packages/liveblocks-core/src/ai.ts
+++ b/packages/liveblocks-core/src/ai.ts
@@ -957,7 +957,7 @@ export function createAi(config: AiConfig): Ai {
   // a ceiling of max 40 rerenders/second during streaming.
   const DELTA_THROTTLE = 25;
   let pendingDeltas: { id: MessageId; delta: AiAssistantDeltaUpdate }[] = [];
-  let deltaBatchTimer: NodeJS.Timeout | number | null = null;
+  let deltaBatchTimer: ReturnType<typeof setTimeout> | null = null;
 
   function flushPendingDeltas() {
     const currentQueue = pendingDeltas;


packages/liveblocks-core/src/ai.ts
```diff
@@ -952,6 +952,42 @@ export function createAi(config: AiConfig): Ai {
...
+    if (deltaBatchTimer !== null) {
+      clearTimeout(deltaBatchTimer);
+      deltaBatchTimer = null;
+    }
+
+    // Process all pending deltas in a single batch
+    batch(() => {
+      console.warn(
+        `Processing ${currentQueue.length} pending deltas in this chunk`
+      );
```
marcbouchenoire: To be removed?

packages/liveblocks-core/src/ai.ts
```diff
@@ -952,6 +952,42 @@ export function createAi(config: AiConfig): Ai {
...
+      }
+    });
+  }
+
+  function enqueueDelta(id: MessageId, delta: AiAssistantDeltaUpdate) {
+    pendingDeltas.push({ id, delta });
+
+    // If no timer is running, start one to process the batch
+    if (deltaBatchTimer === null) {
+      deltaBatchTimer = setTimeout(flushPendingDeltas, DELTA_THROTTLE);
```
marcbouchenoire, line 984: Should we also `clearTimeout` this when WS disconnects (for any reason, including having no chats or `LiveblocksProvider` rendered anymore)?

marcbouchenoire (approved):
Great change!

packages/liveblocks-core/src/ai.ts
nvie: Thanks, good catch!

packages/liveblocks-core/src/ai.ts
nvie: Ah yes, great call. Added in f307a779d8bf6cc1c3817223e81b4c9f6463b41f 👍

packages/liveblocks-core/src/ai.ts
```diff
@@ -952,6 +952,39 @@ export function createAi(config: AiConfig): Ai {
     knowledge: new KnowledgeStack(),
   };
 
+  // Delta batch processing system to throttle incoming delta updates. Incoming
+  // deltas are buffered and only let through every every 25ms. This creates
+  // a ceiling of max 40 rerenders/second during streaming.
+  const DELTA_THROTTLE = 25;
+  let pendingDeltas: { id: MessageId; delta: AiAssistantDeltaUpdate }[] = [];
+  let deltaBatchTimer: NodeJS.Timeout | number | null = null;
```
SugarDarius:
```suggestion
  let deltaBatchTimer: ReturnType<typeof setTimeout> | null = null;
```
Just a nitpick typing here 😅

PR #2639, opened 2025-09-02
Auto-abort all my auto-executing messages when the page unloads

nvie:
This PR does what it says on the tin.

**Update:** The behavior has been made less aggressive now. The only messages that will get auto-aborted when the page unloads are the ones matching _all_ of these criteria:

- The message is an assistant message
- The message is owned by this client ("mine")
- The message is currently in `"awaiting-tool"` status
- The message has at least one tool invocation in `"executing"` stage
- The tool invocation has an `execute()` function defined

This means streaming text messages will keep generating, and HITL-style confirmation dialogs will still stay in executing stage and will not get aborted. Only auto-executed tools will get aborted this way.

Fixes LB-2384.


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### `@liveblocks/core`
 
+- Automatically abort all my AI messages (that are still generating or awaiting
```
marcbouchenoire: To be updated 👀

marcbouchenoire (approved):
Thanks for changing the behavior, it looks ideal to me 🙏

CHANGELOG.md
nvie: 🙏

PR #2641, opened 2025-09-03
Upgrade Turborepo to latest version

nvie:
Bumps Turborepo to 2.5.6. From what I could tell in the Turbo changelogs, no interesting new features nor breaking changes in this bump. So mostly bug fixes for us.


nvie:
> Any particular bug fix to do this now or just upkeep?

Just upkeep!

PR #2642, opened 2025-09-03
Upgrade Vitest to v3

nvie:
Not really sure why I didn't jump to V3 immediately when I migrated from Jest to Vitest, honestly, but fortunately this is a non-breaking change.


marcbouchenoire (approved):
3️⃣

PR #2645, opened 2025-09-04
Fix a bug in LiveLists where two clients would not reach eventual consistency

nvie:
This PR is quite large, but this is because I applied the boy scout rule to our E2E tests while diving deep to get this nasty race condition fixed. The actual fix is pretty small.

## 🐞 The actual bug fix (small change)

The most important part of this PR is this test that captured the core issue: this sequence of LiveList updates caused clients to get out of sync (not eventually-consistent):

- [The test that captured the bug](https://github.com/liveblocks/liveblocks/blob/4d9ed6a206667db4793e426673490ae3f31fe122/packages/liveblocks-core/e2e/list-consistency.test.ts#L12-L41)
- [The fix for that](https://github.com/liveblocks/liveblocks/commit/4d9ed6a206667db4793e426673490ae3f31fe122)

---

## 💆‍♂️ Improvements to the E2E setup (large change)

While digging in here, I improved the speed and correctness of our client-spec E2E tests setup:
- Replaced arbitrary `wait(600)` delays with deterministic beacon synchronization using broadcast events. Our tests were structured roughly like this: "client A does something, client B does something, client A sends messages to the server, client B sends messages to the server, now they see the same result". This worked, but the problem was that we need to decide how long to wait before testing whether they see the same result. We previously used an arbitrary delay of 600ms, which was actually quite slow. No need to wait that long normally. So I came up with a "beacon trick" where client A would now broadcast a "beacon" event after making its storage changes, and we wait until client B has seen that event. If we see that event, we know client B must have processed the storage updates. This happens way quicker than 600ms typically, and forms a clever sync primitive that doens't require any server-side support. This was long overdue!
- Added new LiveList conflict test sequence that captures the broken move/insert/undo scenario
- Split large test files for better parallelism and faster CI runs. The test suite now run ~3x faster due to the module splitting and beacon synchronization vs arbitrary waits

While working on this, I also spotted a bug where when calling `control.flushA()` (ie sending all buffered messages from Client A to B), it would "resume" the controlled socket (allowing the messages through), but it would not immediately "pause" the socket again, meaning that further messages sent in the test would no longer be fully controlled and could interfere with the test. I've fixed this by basically doing a ".resume() + broadcast beacon event + .pause()" here now when flushing the socket in the test. However, this brought to light another issue: because our clients would use a throttle, doing `root.set('a', 1); root.set('b', 2)` would actually send the first immediately, but then wait 100ms (default throttle) before the second change would get sent. This is great behavior normally, but a test where we explicitly want to fully control our timing, the throttle actually is working against us. Therefore I've added an internal way to disable client throttling entirely in these unit tests that use explicit message timing.

Turns out the combination of accidentally not locking the messaging gates after the first `control.flushA()` call and the throttling sending messages later made our tests pass accidentally but occasionally fail.

Furthermore, I've:
- Upgraded `fast-check` to 4.x
- Added a model-based test for "eventual consistency" of LiveLists. I've ran this test many many times and was not able to find a counter example thus far. (It generates randomized sequences [like this](https://share.cleanshot.com/VZn9cGGqKGSbN2p9WLn6), and validates the final result converges to the same value on all clients.)


packages/liveblocks-core/src/crdts/LiveList.ts
```diff
@@ -765,15 +765,17 @@ export class LiveList<TItem extends Lson> extends AbstractCrdt {
...
 
-    // Assign a temporary position until we get the fix from the backend
+    // If position is occupied, find a free position for item being moved
+    let actualNewKey = newKey;
     if (existingItemIndex !== -1) {
-      this.#items[existingItemIndex]._setParentLink(
-        this,
-        makePosition(newKey, this.#items[existingItemIndex + 1]?._parentPos)
+      // Find a free position near the desired position
+      actualNewKey = makePosition(
```
ofoucherot, line 772: 🤯

nvie: Before: https://share.cleanshot.com/HvmdRk4X529DlQccM8ZC

After: https://share.cleanshot.com/vQHZNyhswZH9mx4qNTCF

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 ### `@liveblocks/client`
 
 - Auto-abort this client's tool calls on page unload to prevent hanging chats.
+- Fix bug where a specific combination of concurrent LiveList mutations could
```
SugarDarius: `Fix a bug` or `Bug fix`? I'm asking because as a french speaking person it feels strange to read. But if it's valid in English please ignore this comment.

marcbouchenoire: "Fix" is a verb in "fix bug" and a noun in "bug fix", and we (almost) always use the verb form in changelog entries.

@nvie But the entry should be moved to 3.6.1 since 3.6.0 is already out.

nvie: I've moved this to the new unreleased section, and changed it to "Fixes a bug where...". That's more consistent with what we've done before anyway — thanks 👍

PR #2658, opened 2025-09-09
Fix AiReasoningDelta type definition

nvie:
The signature is never visible publicly.


SugarDarius (approved):

marcbouchenoire (approved):
I don't think this type is public either so we might not need to mention it in the changelog.

nvie:
> I don't think this type is public either so we might not need to mention it in the changelog.

Oh yeah of course, I confused it with the part, this is only the delta 👍

PR #2663, opened 2025-09-10
inbox notifications add query filter on kind

ofoucherot:
### Description
This PR adds query filters on the get inbox notifications method and react hook.
The first filter supported is roomId.

#### @liveblocks/client
```
const inboxNotifications = await client.getInboxNotifications({
  query: {
    kind:"thread",
  },
});
```

#### @liveblocks/react
```
const { inboxNotifications } = useInboxNotifications({
  query: {
     kind:"thread",
  }
});
```

### DIFF
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -3695,7 +3695,7 @@ function Inbox() {
 
 It’s possible to return inbox notifications that match a certain query with the
 `query` option. You can filter inbox notifications based on their associated
-roomId.
+roomId or kind.
 
 ```tsx
 // Returns threads that match the entire `query`, e.g. { roomId: "room1", ... }
@@ -3703,6 +3703,8 @@ const { inboxNotifications } = useInboxNotifications({
   query: {
     // Filter for roomId
     roomId: "room1",
+    // Filter for kind
+    kind: "thread",
   },
 });
 ```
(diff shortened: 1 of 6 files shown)


docs/pages/api-reference/liveblocks-react.mdx
```diff
@@ -3695,14 +3695,16 @@ function Inbox() {
 
 It’s possible to return inbox notifications that match a certain query with the
 `query` option. You can filter inbox notifications based on their associated
-roomId.
+roomId or kind.
```
marcbouchenoire:
```suggestion
room ID or kind.
```

packages/liveblocks-react/src/types/index.ts
```diff
@@ -159,6 +159,12 @@ export type InboxNotificationsQuery = {
    * all the user's inbox notifications will be returned.
    */
   roomId?: string;
+
+  /**
+   * Whether to only return inbox notifications for a specific kind. If not provided,
+   * all the user's inbox notifications will be returned.
```
marcbouchenoire: "all the user's inbox notifications will be returned" sounds like `roomId` would be ignored to me, maybe "all inbox notification kinds will be returned"?

marcbouchenoire (approved):
Added a couple of comments but they're just docs-related 👍

packages/liveblocks-react/src/types/index.ts
ofoucherot: I don't think the second sentence adds anything useful. I'm removing it.

PR #2678, opened 2025-09-15
Use right endpoint unread in count

ofoucherot:

### DIFF
--- a/packages/liveblocks-react/src/liveblocks.tsx
+++ b/packages/liveblocks-react/src/liveblocks.tsx
@@ -24,6 +24,7 @@ import {
   assert,
   console,
   createClient,
+  DefaultMap,
   HttpError,
   kInternal,
   makePoller,
@@ -48,7 +49,6 @@ import {
   useIsInsideRoom,
 } from "./contexts";
 import { ASYNC_OK } from "./lib/AsyncResult";
-import { count } from "./lib/itertools";
 import { ensureNotServerSide } from "./lib/ssr";
 import { useInitial, useInitialUnlessFunction } from "./lib/use-initial";
 import { useLatest } from "./lib/use-latest";
@@ -125,20 +125,14 @@ const _bundles = new WeakMap<
 >();
 
 function selectorFor_useUnreadInboxNotificationsCount(
-  result: InboxNotificationsAsyncResult
+  result: UnreadInboxNotificationsCountAsyncResult
 ): UnreadInboxNotificationsCountAsyncResult {
-  if (!result.inboxNotifications) {
+  if (!result.count) {
     // Can be loading or error states
     return result;
   }
 
-  return ASYNC_OK(
-    "count",
-    count(
-      result.inboxNotifications,
-      (n) => n.readAt === null || n.readAt < n.notifiedAt
-    )
-  );
+  return ASYNC_OK("count", result.count);
 }
 
 function selectorFor_useUser<U extends BaseUserMeta>(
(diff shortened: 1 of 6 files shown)


nvie:
Just a friendly reminder that if you intend to include this into release 3.8, you might want to rebase this against that release branch.

marcbouchenoire (approved):

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -708,6 +775,12 @@ function useDeleteAllInboxNotifications_withClient(client: OpaqueClient) {
       () => {
         // Replace the optimistic update by the real thing
         store.deleteAllInboxNotifications(optimisticId);
+
+        // Force a re-fetch of the unread notifications count
+        for (const poller of unreadNotificationsCountPollersByQueryKey.values()) {
+          poller.markAsStale();
+          poller.pollNowIfStale();
```
nvie, line 782: Does this warrant a `poller.forcePoll()`?

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -746,6 +752,27 @@ function createStore_forNotifications() {
...
+
+  function update(queryKey: InboxNotificationsQueryKey, count: number): void {
+    baseSignal.mutate((lut) => {
+      lut.set(queryKey, count);
+    });
+  }
+
+  return {
+    signal: DerivedSignal.from(baseSignal, (c) =>
+      Object.fromEntries([...c].map(([queryKey, count]) => [queryKey, count]))
```
nvie: Isn't this the same as just `Object.fromEntries(c)`?

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -620,6 +666,12 @@ function useMarkInboxNotificationAsRead_withClient(client: OpaqueClient) {
             readAt,
             optimisticId
           );
+
+          // Force a re-fetch of the unread notifications count
+          for (const poller of unreadNotificationsCountPollersByQueryKey.values()) {
+            poller.markAsStale();
+            poller.pollNowIfStale();
+          }
```
nimeshnayaju, line 674: I guess we can't really avoid doing a full refetch in these cases, so it's okay.

packages/liveblocks-react/src/liveblocks.tsx
ofoucherot: Maybe, I don't know if it's going to be used in other places! What do you think?

marcbouchenoire: Déjà vu: https://liveblocks.slack.com/archives/C08DGF1N7EG/p1744889343110589?thread_ts=1744884626.140219&cid=C08DGF1N7EG

nvie: Oh yeah, don't open that can of worms, let's keep it this way 👍

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -126,20 +126,14 @@ const _bundles = new WeakMap<
...
   }
 
-  return ASYNC_OK(
-    "count",
-    count(
-      result.inboxNotifications,
-      (n) => n.readAt === null || n.readAt < n.notifiedAt
-    )
-  );
+  return ASYNC_OK("count", result.count);
```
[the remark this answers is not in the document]
nvie: This is actually a great catch! 👌

PR #2690, opened 2025-09-22
Explain `"experimental-fallback-to-http"` should not be used with Storage

nvie:
We could certainly make this work, but currently it doesn't.


jrowny:
> We could certainly make this work, but currently it doesn't.


I don't think it works with Yjs either

SugarDarius (approved):

PR #2696, opened 2025-09-23
useThreads metadata number comparison search

ofoucherot:
#### liveblocks/client

- Support numerical operators `gt`, `lt`, `gte`,
  and `lte` in `room.getThreads` metadata query filters.

```
// Creating a thread with `{ posX: 87, level: 5 } metadata
await room.createThread({
  body: {
    // ...
  },
  metadata: { posX: 87, level: 5 },
});

// Filtering for threads with `posX` greater than 50 and lower than 100, and level greater than or equal to 5
const threads = await room.getThreads({
  query: {
    metadata: {
      posX: {
        gt: 50,
        lt: 100,
      },
      level: {
        gte: 5,
      },
    },
  },
});
```

#### liveblocks/react

- Support numerical operators `gt`, `lt`, `gte`,
  and `lte` in `useThreads` metadata query filters.

```
const { threads } = useThreads({
  query: {
    metadata: {
 
      // Filter for threads with number metadata that is greater than 50 and lower than 100
      posX: {
        gt: 50,
        lt: 100,
      },

      // Filter for threads with number metadata that is greater than or equal to 5
      level: {
        gte: 5,
      },
    },
  },
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@
 - Fixes an issue where `useUnreadInboxNotificationsCount` wasn't returning the
   proper count if there were more than a page of unread notifications.
 
+- Support numerical operators `gt`, `lt`, `gte`, and `lte` in `useThreads`
+  metadata query filters.
+
+### `@liveblocks/client`
+
+- Support numerical operators `gt`, `lt`, `gte`, and `lte` in `room.getThreads`
+  metadata query filters.
+
 ### `@liveblocks/react-ui`
 
 - Add `responseTimeout` property to `AiChat` to allow customization of the


packages/liveblocks-core/src/lib/objectToQuery.ts
```diff
@@ -12,23 +16,39 @@ import { isPlainObject, isStartsWithOperator } from "./guards";
...
 type SimpleFilterValue = string | number | boolean | null;
-type OperatorFilterValue = { startsWith: string };
+type OperatorFilterValue =
+  | { startsWith: string; lowerThan?: never; greaterThan?: never }
+  | {
+      lowerThan?: number;
+      greaterThan?: number;
+      greaterThanOrEqual?: number;
+      lowerThanOrEqual?: number;
+      startsWith?: never;
```
SugarDarius: What do you think about proposing more concise names like ` lt, gt, lte, gte`?
IMO it sounds more DX friendly and more classic pattern?

nvie: I'd also be a fan of these shorter notations, good suggestion 👍

ofoucherot: Yes, I agree it's better, but it's not consistent with `startsWith`, which already exists.

What do you think @nvie

nvie: Mongo also isn't consistent, but it makes writing queries a lot nicer anyway:
https://www.mongodb.com/docs/manual/reference/mql/query-predicates/#alphabetical-list-of-operators

Not saying Mongo is a great example to copy from _generally_, but I think in this case it isn't something to worry about.

nvie (approved):

packages/liveblocks-core/src/lib/objectToQuery.ts
ofoucherot: Good point! I've updated them.

PR #2701, opened 2025-09-26
Stop relying on `ai` feature flag in JWT token

nvie:
We're enabling this feature flag for everyone, and want to ultimately remove it from the token, so let's start ignoring it, starting with 3.8.


SugarDarius:
Duplicate of #2697

PR #2706, opened 2025-09-30
support threads query filter "subscribed"

ofoucherot:
#### liveblocks/client

Support query filter `subscribed` in `room.getThreads` method:

```
// Filtering for threads that the user is subscribed to
const threads = await room.getThreads({
  query: {
    subscribed: true,
  },
});
```

#### liveblocks/react

Support query filter `subscribed` in `useThreads ` hook:

```
const { threads } = useThreads({
  query: {
    // Filter for threads that the user is subscribed to
    subscribed: true,
  }
})
```

### DIFF
--- a/packages/liveblocks-react/src/ThreadDB.ts
+++ b/packages/liveblocks-react/src/ThreadDB.ts
@@ -1,5 +1,7 @@
 import type {
   BaseMetadata,
+  SubscriptionData,
+  SubscriptionKey,
   ThreadData,
   ThreadDataWithDeleteInfo,
   ThreadDeleteInfo,
@@ -8,7 +10,6 @@ import { batch, MutableSignal, SortedList } from "@liveblocks/core";
 
 import { makeThreadsFilter } from "./lib/querying";
 import type { ThreadsQuery } from "./types";
-import type { SubscriptionsByKey } from "./umbrella-store";
 
 function sanitizeThread<M extends BaseMetadata>(
   thread: ThreadDataWithDeleteInfo<M>
@@ -187,7 +188,7 @@ export class ThreadDB<M extends BaseMetadata> {
     roomId: string | undefined,
     query: ThreadsQuery<M> | undefined,
     direction: "asc" | "desc",
-    subscriptions: SubscriptionsByKey | undefined
+    subscriptions: Record<SubscriptionKey, SubscriptionData> | undefined
   ): ThreadData<M>[] {
     const index = direction === "desc" ? this.#desc : this.#asc;
     const crit: ((thread: ThreadData<M>) => boolean)[] = [];
--- a/packages/liveblocks-react/src/lib/querying.ts
+++ b/packages/liveblocks-react/src/lib/querying.ts
@@ -7,10 +7,11 @@ import {
   getSubscriptionKey,
   isNumberOperator,
   isStartsWithOperator,
+  type SubscriptionData,
+  type SubscriptionKey,
 } from "@liveblocks/core";
 
 import type { InboxNotificationsQuery, ThreadsQuery } from "../types";
-import type { SubscriptionsByKey } from "../umbrella-store";
 
 /**
  * Creates a predicate function that will filter all ThreadData instances that
@@ -18,7 +19,7 @@ import type { SubscriptionsByKey } from "../umbrella-store";
  */
 export function makeThreadsFilter<M extends BaseMetadata>(
   query: ThreadsQuery<M>,
-  subscriptions: SubscriptionsByKey | undefined
+  subscriptions: Record<SubscriptionKey, SubscriptionData> | undefined
 ): (thread: ThreadData<M>) => boolean {
   return (thread: ThreadData<M>) =>
     matchesThreadsQuery(thread, query, subscriptions) &&
@@ -28,7 +29,7 @@ export function makeThreadsFilter<M extends BaseMetadata>(
 function matchesThreadsQuery(
   thread: ThreadData<BaseMetadata>,
   q: ThreadsQuery<BaseMetadata>,
-  subscriptions: SubscriptionsByKey | undefined
+  subscriptions: Record<SubscriptionKey, SubscriptionData> | undefined
 ) {
   let subscription = undefined;
   if (subscriptions) {


FlowFlorent (approved):

packages/liveblocks-react/src/types/index.ts
```diff
@@ -124,6 +128,13 @@ export type ThreadsQuery<M extends BaseMetadata> = {
    * all threads will be returned.
    */
   resolved?: boolean;
+
+  /**
+   * Whether to only return threads that the user is subscribed to or not. If not provided,
+   * all threads will be returned.
+   */
+  subscribed?: boolean;
```
SugarDarius, line 130: The only possible value is `true` right? Should we not only authorize from a type level standpoint `subscribed?: true` as the property is `undefined` by default? 

Do we have use cases where users would need to explicitly set `false`? Maybe the QueryParser will prefer `boolean` instead of `true` right?

packages/liveblocks-react/src/lib/querying.ts
```diff
@@ -3,27 +3,42 @@ import type {
...
 
 function matchesThreadsQuery(
   thread: ThreadData<BaseMetadata>,
-  q: ThreadsQuery<BaseMetadata>
+  q: ThreadsQuery<BaseMetadata>,
+  subscriptions: SubscriptionsByKey | undefined
 ) {
   // Boolean logic: query.resolved? => q.resolved === t.resolved
-  return q.resolved === undefined || thread.resolved === q.resolved;
+  const subscription = subscriptions?.[getSubscriptionKey("thread", thread.id)];
```
marcbouchenoire: `getSubscriptionKey` is cheap but it's still extra work so maybe we could only call it if `subscriptions` exist?

And we can update/move the comment above about `query.resolved`

docs/pages/api-reference/liveblocks-client.mdx
```diff
@@ -2829,6 +2832,22 @@ const threads = await room.getThreads({
 });
 ```
 
+#### Filtering subscribed [#filtering-subscribed-status]
```
marcbouchenoire: I think I would prefer `#### Filtering subscribed status` or `#### Filtering subscription status`

packages/liveblocks-react/src/__tests__/useThreads.test.tsx
```diff
@@ -450,8 +450,8 @@ describe("useThreads", () => {
     });
     const subscriptions = [
       dummySubscriptionData({ subjectId: bluePinnedThread.id }),
-      dummySubscriptionData({ subjectId: blueUnpinnedThread.id }),
-      dummySubscriptionData({ subjectId: redPinnedThread.id }),
+      // dummySubscriptionData({ subjectId: blueUnpinnedThread.id }),
+      // dummySubscriptionData({ subjectId: redPinnedThread.id }),
```
marcbouchenoire: Should we remove them?

packages/liveblocks-react/src/ThreadDB.ts
```diff
@@ -178,20 +179,23 @@ export class ThreadDB<M extends BaseMetadata> {
...
+   *
+   * Subscriptions are needed to filter threads based on the user's subscriptions.
    */
   public findMany(
     // TODO: Implement caching here
     roomId: string | undefined,
     query: ThreadsQuery<M> | undefined,
-    direction: "asc" | "desc"
+    direction: "asc" | "desc",
+    subscriptions: SubscriptionsByKey | undefined
```
marcbouchenoire: What's your opinion on making this optional instead of asking `undefined`?

packages/liveblocks-react/src/types/index.ts
ofoucherot: No, we also support `subscribed: false`.

packages/liveblocks-react/src/umbrella-store.ts
```diff
@@ -1260,9 +1263,11 @@ export class UmbrellaStore<M extends BaseMetadata> {
             return result;
           }
 
+          const subscriptions = threadSubscriptions.get().subscriptions;
+
           const threads = this.outputs.threads
             .get()
-            .findMany(roomId, query ?? {}, "asc");
+            .findMany(roomId, query ?? {}, "asc", subscriptions);
```
marcbouchenoire, line 1270: Do you know if this will update/re-render if the subscriptions change but not the threads (e.g. when unsubscribing from a thread)? I can't fully remember how "threadifications" and subscriptions signals derive/link themselves

marcbouchenoire (commented):
Just leaving a few questions for now but I'm 90% sure I could already approve 👌

packages/liveblocks-react/src/types/index.ts
SugarDarius: Okay perfect then! 🙂

packages/liveblocks-react/src/lib/querying.ts
ofoucherot: Good point! I've updated it.

I removed the comment about boolean logic, I don't think it's adding anything.

docs/pages/api-reference/liveblocks-client.mdx
ofoucherot: I was thinking of `subscribed status` also. Let's go with that!

packages/liveblocks-react/src/__tests__/useThreads.test.tsx
ofoucherot: It helps to understand which threads are subscribed to and which ones are not.
But I updated the text so it's clearer.

packages/liveblocks-react/src/ThreadDB.ts
ofoucherot: I prefer to have `undefined` so that it forces you to check if you need to pass it or not. But I don't have a strong opinion on it.

packages/liveblocks-react/src/umbrella-store.ts
ofoucherot: Good point! And yes, I've added a test for that : https://github.com/liveblocks/liveblocks/pull/2706/files#diff-c9a686dd0722b4a5f2eadaea40bdeb74f30315075d1644762ccd656da18f2fccR2138

nimeshnayaju: I noticed a `threadSubscriptions.get()` call in this callback, and since `threadSubscriptions` is a signal too, the derived signal will recompute whenever `threadSubscriptions` changes. So, same conclusion as Olivier - it should re-compute and hence re-render.

nimeshnayaju (approved):

PR #2709, opened 2025-10-01
Use OpenRouter copilot for all AI E2E tests

nvie:
In order to make this update, I needed to update the E2E AI kitchen sink to pass an explicitly configured copilot ID in (if configured in the env var).

### DIFF
--- a/e2e/next-ai-kitchen-sink/test/multi-chat.test.ts
+++ b/e2e/next-ai-kitchen-sink/test/multi-chat.test.ts
@@ -129,7 +129,8 @@ test.describe("Multi Chat", () => {
     ).not.toBeVisible();
   });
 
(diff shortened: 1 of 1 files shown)


e2e/next-ai-kitchen-sink/test/concurrent-operations.test.ts
```diff
@@ -52,14 +52,14 @@ test.describe("Concurrent Operations", () => {
     const chatId = chatUrl.split("/").pop();
 
     // Send first message
-    await sendMessage(page, "What is 2+2?");
+    await sendMessage(page, "What is 2+2? Be brief.");
```
SugarDarius, line 55: FYI it will depend on which LLM you use but it's more effective to use "Be concise" instead of "Be brief". Usually the word "concise" is more direct and strict for LLMs.

PR #2717, opened 2025-10-06
Add URL metadata APIs

marcbouchenoire:


marcbouchenoire:
@SugarDarius Yes we read all of them in a specific order (since we only want one title, one description, etc.)

SugarDarius:
I have just one concern though.
While the name `useUrlMetadata` seems more neutral and future-proof does this hook is meant to be used for other things than creating preview links/cards?

If this hook purpose is to let users build UIs with preview (hover) cards does it feel better to call it `useLinkPreview` as a more oriented / opinionated name?

I'm just asking because `useUrlMetadata` seems a more technical related name than an UI related name.

marcbouchenoire:
@SugarDarius I personally prefer the explicitness of `useUrlMetadata` compared to something like `useLinkPreview` which doesn't feel immediately obvious to me what it would return, also I feel like "meta tags"/"metadata" are known enough to not be too obscure or technical.

SugarDarius:
@marcbouchenoire okay thanks 👍🏻

PR #2720, opened 2025-10-07
Add support for web retrievals and citations

nimeshnayaju:
This PR adds support for web search on the `<AiChat />` component. As a result, a few props are introduced :

1. `showRetrievals`: Controls how to show or hide retrievals. Can be set to:
    - `true`: Displays both knowledge and web retrievals
    - `false`: Hides both knowledge and web retrievals
    - `during`: Displays both knowledge and web retrievals when message is generating and hides when messages has fully generated.
    - `{ web: true | false | "during"; knowledge: true | false | "during" }`: Controls how to show or hide retrieval for web and knowledge individually.

2. `showSources`: Controls whether to show or hide sources. Can be set to `true` or `false`.
3. `showReasoning`: Controls whether to show or hide sources. Can be set to `true`, `false`, or `during`. (Not related to web search but part of this PR)

We now process `source` event types (available since V7) and retrieval event of kind `web` and the default components are updated to display sources and web retrievals.

### DIFF
--- a/packages/liveblocks-react/test-d/augmentation.test-d.tsx
+++ b/packages/liveblocks-react/test-d/augmentation.test-d.tsx
@@ -1099,7 +1099,12 @@ declare global {
   if (status.status === "generating") {
     // The partType might not exist if there's no content yet
     expectType<
-      "text" | "reasoning" | "retrieval" | "tool-invocation" | undefined
+      | "text"
+      | "reasoning"
+      | "retrieval"
+      | "tool-invocation"
+      | "sources"
+      | undefined
     >(status.partType);
     if (status.partType === "tool-invocation") {
       expectType<string>(status.toolName);
@@ -1119,7 +1124,12 @@ declare global {
   if (status.status === "generating") {
     // The partType might not exist if there's no content yet
     expectType<
-      "text" | "reasoning" | "retrieval" | "tool-invocation" | undefined
+      | "text"
+      | "reasoning"
+      | "retrieval"
+      | "tool-invocation"
+      | "sources"
+      | undefined
     >(status.partType);
     if (status.partType === "tool-invocation") {
       expectType<string>(status.toolName);
@@ -1138,7 +1148,12 @@ declare global {
   if (status.status === "generating") {
     // The partType might not exist if there's no content yet
     expectType<
-      "text" | "reasoning" | "retrieval" | "tool-invocation" | undefined
+      | "text"
+      | "reasoning"
+      | "retrieval"
+      | "tool-invocation"
+      | "sources"
+      | undefined
     >(status.partType);
     if (status.partType === "tool-invocation") {
       expectType<string>(status.toolName);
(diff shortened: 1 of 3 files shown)


packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
```diff
@@ -82,6 +88,15 @@ const AiMessageContent = forwardRef<HTMLDivElement, AiMessageContentProps>(
                   message={message as AiAssistantMessage}
                 />
               );
+
+            case "sources":
+              // Sources are only shown after the message is fully generated.
+              if (isGenerating) {
+                return null;
+              }
```
nimeshnayaju: @marcbouchenoire I thought about the same approach but wasn't sure if we should also hide it when we're awaiting for a tool result

marcbouchenoire: Probably yes? I overlooked this since we take `"awaiting-tool"` into account in a few similar booleans

nimeshnayaju: I think so too but I am conflicted only on 'Human in the loop' style of tool calls. I am not fully sure but I feel it may be awkward to not see any sources until user has confirmed or denied an action on a tool call. Currently, I am leaning towards, it's okay to do so, but am curious of your thoughts.

packages/liveblocks-react-ui/src/components/internal/AiChatAssistantMessage.tsx
```diff
@@ -155,33 +274,83 @@ export const AiChatAssistantMessage = memo(
...
-      <ReasoningPart {...props} components={ref.current} />
-    ),
-    []
+    () => (props: ReasoningPartProps) => {
+      if (
+        !showReasoning ||
+        (showReasoning === "during" && !props.isStreaming)
+      ) {
+        return null;
+      }
```
nimeshnayaju, line 318: `isStreaming` is currently set to true only if the message is in `generating` state (and the part is the last part in the message), which means if it's `awaiting-tool`, it'll be set to false. In this case, if `showReasoning` or `showRetrievals` is set to `during`, they'll not be displayed. Is that expected behavior @marcbouchenoire?

nimeshnayaju: Thinking about my own question: I think this is fine. `during` means we want to show reasoning/retrievals if the message is generating and the part is the last part in the message. If a tool call arrives, the reasoning/retrieval part won't be the last part, so we don't want to display them by default. Feel free to ignore my question!

packages/liveblocks-react-ui/src/primitives/AiMessage/index.tsx
nimeshnayaju: Done in https://github.com/liveblocks/liveblocks/pull/2720/commits/26b9e7bb5a3277035b066502c92cc090f63968e3

marcbouchenoire: I think it would be weird to see sources during HITL and then see them disappear again. I like the change you made to make it clear that we want them only at the end 👍

PR #2721, opened 2025-10-07
Add Ai websocket status private hook

SugarDarius:
### What's going on?
This PR adds an a new private for internal use only hook named `useAiWebSocketStatus`. It's mean to be used to get the current websocket status connection for Ai:
```tsx
import { useAiWebSocketStatus } from "@liveblocks/react/_private" 

export function AiComponent() {
  const status = useAiWebSocketStatus();
  const isDisabled = status === "disconnected"
  
  return <AiComponentPrimitive.Action  disabled={disabled} />
}
```

It's primarily used in `<AiComposerPrimitive.Form />` to update the `isDisabled` props  in the `AiComposerContext`.
It's allowing us to disable all the `AiComposer` primitives related component in case the paywall is active in a middle of a chat. So it will avoid users to either write / send a message or triggering actions when paywall is active. 

The status comes from an event source derived from ` managedSocket.events.statusDidChange`.


packages/liveblocks-react-ui/src/primitives/AiComposer/index.tsx
```diff
@@ -115,7 +120,8 @@ export const AiComposerForm = forwardRef<HTMLFormElement, AiComposerFormProps>(
     const lastMessageId = useSignal(messagesΣ, getLastMessageId);
     const abortableMessageId = useSignal(messagesΣ, getAbortableMessageId);
 
-    const isDisabled = isSubmitting || disabled === true;
+    const isDisabled = status === "disconnected";
+    isSubmitting || disabled === true;
```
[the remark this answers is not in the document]
SugarDarius: Oupsy

packages/liveblocks-react-ui/src/primitives/AiComposer/index.tsx
```diff
@@ -115,7 +120,8 @@ export const AiComposerForm = forwardRef<HTMLFormElement, AiComposerFormProps>(
     const lastMessageId = useSignal(messagesΣ, getLastMessageId);
     const abortableMessageId = useSignal(messagesΣ, getAbortableMessageId);
 
-    const isDisabled = isSubmitting || disabled === true;
+    const isDisabled =
```
marcbouchenoire: I'm wondering if maybe the condition should be `status !== "connected"` and instead of disabling the entire form when not connected it would change `canSubmit` so that you can still start entering text as soon as possible but you wouldn't be able to submit yet (via the button or `enter`).

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## vNEXT (not yet published)
 
+### `@liveblocks/react-ui`
+
+- Disable Ai chat composer when websocket connection is closed.
```
marcbouchenoire:
```suggestion
- Disable AI chat composers when WebSocket connection is closed.
```

marcbouchenoire (commented):
I added a question about the final behavior but the internals look good to me.

packages/liveblocks-react-ui/src/primitives/AiComposer/index.tsx
SugarDarius: Using `status !== "connected"` makes sense 👍🏻 
I chosed to disable all the form for also disabling actions (e.g abort). I understand the behavior will be better if you can enter text as soon as possible. 

I'll think about a solution to prevent form submission and prevent actions trigger 👍🏻

marcbouchenoire: If you add the condition to `canSubmit` in `AiComposerForm` that should be enough, it's already used in the editor and the action via a context.

SugarDarius: Are you sure? I may miss something here but it looks like `<AiComposerAbort />` is actionable even if I pass `canSubmit` to `false`.

nvie:
> This PR adds an a new private for internal use only hook named `useAiWebSocketStatus`.

I don't particularly like exposing the WebSocket connection details here, as it's an implementation detail that we're using WebSockets for AI chats. In a way, I'd like to model/expose this more as a state on the chat itself: it can be disfunctional under certain conditions (i.e. when the account doesn't allow access to this feature and retrying won't make sense).

What about integrating this into the existing `useAiChatStatus` hook and exposing it there as a new status? I think that might actually be nicer, doesn't force us to add a new hook, and hides the WebSocket details. (Technically it's a breaking change, because clients using this hook should handle the new possible status, but I think it's fine in practice.)

WDYT?

nvie (commented):
> This PR adds an a new private for internal use only hook named `useAiWebSocketStatus`.

I don't particularly like exposing the WebSocket connection details here, as it's an implementation detail that we're using WebSockets for AI chats. In a way, I'd like to model/expose this more as a state on the chat itself: it can be disfunctional under certain conditions (i.e. when the account doesn't allow access to this feature and retrying won't make sense).

What about integrating this into the existing `useAiChatStatus` hook and exposing it there as a new status? I think that might actually be nicer, doesn't force us to add a new hook, and hides the WebSocket details. (Technically it's a breaking change, because clients using this hook should handle the new possible status, but I think it's fine in practice.)

WDYT?

packages/liveblocks-react-ui/src/primitives/AiComposer/index.tsx
marcbouchenoire: Not 100% since we didn't have this condition before, but if WebSocket isn't connected I don't think we should be in an "abortable" state (`canAbort` should be `false`) so in practice the abort action won't be rendered but just to be sure we could make sure it's disabled if it's still rendered.

marcbouchenoire:
@nvie Even privately? Regardless of how we name it internally, I think preventing message submissions when we know for a fact they won't work still makes sense, even if we also expose this info via `useAiChatStatus` to visualize it.

nvie:
> Even privately?

Privately it obviously doesn't matter that much, but I think we should also expose this state publicly, not? If we don't expose this publicly, how would users otherwise build UIs that don't feel broken?

I agree that obviously the submission should be disabled or show the error when this state is detected/active (whether that uses a private or public API).

marcbouchenoire:
Yes I'm also saying both can exist 👍 To me this PR was only tackling the broken submission part privately out-of-the-box and the publicly available opt-in API would be done at another time.

SugarDarius:
@nvie I'm struggling to visualize how we could add the WebSocket connection status to `useAiChatStatus` 🤔 
In my understanding it's two separate statuses. Or am I wrong? 
When reading the code the intention behind `useAiChatStatus` is to indicate what's the status of a chat (idle, loading, generating, ....) it's not to indicate the status of the WebSocket connection. Even if it makes sense to tell the chat status depends on the WebSocket connection status. 

Could you explain what you're thinking about more deeply? Thanks 🙏🏻

nvie:
> I'm struggling to visualize how we could add the WebSocket connection status to `useAiChatStatus` 🤔

I was thinking about it like this:

```diff
 /**
  * Simplified status for the requested chat.
  * This hook offers a convenient way to update the UI while an AI chat
  * generation is in progress.
  */
 export type AiChatStatus = Relax<
   | { status: "loading" }
+  | { status: "disconnected"; reason: string }
   | { status: "idle" }
   | { status: "generating" } // Still generating, but there is no content yet
   | {
       status: "generating";
       partType: Exclude<AiAssistantContentPart["type"], "tool-invocation">;
     }
   | { status: "generating"; partType: "tool-invocation"; toolName: string }
 >;
```

Reason could be optional, only if it's easy to add. But this reflects that the chat is essentially in an "unusable" / error state, without leaking that it's about/related to a websocket connection.

SugarDarius:
@nvie Thanks 🙏🏻 
It makes sense! 

I wasn't thinking about making it right away public though but yeah it's important to let customers handle the UI on a disconnected chat 👍🏻

SugarDarius:
@nvie I made a proposal on the last two commits. Can you check when you have time to validate I'm going to the right direction and I don't miss anything? thanks 🙏🏻

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## v3.9.0 (not yet published)
 
+### `@liveblocks/react-ui`
+
+- Disable AI chat composers when WebSocket connection is closed.
```
nvie: Should we say maybe "when the AI service is not available (i.e. when usage limit is reached)" instead of talking about the WebSocket?

packages/liveblocks-react/src/types/ai.ts
```diff
@@ -45,6 +45,7 @@ export type RegisterAiToolProps = {
  * generation is in progress.
  */
 export type AiChatStatus = Relax<
+  | { status: "disconnected" } // WebSocket connection is disconnected
```
nvie:
```suggestion
  | { status: "disconnected" } // WebSocket connection is permanently rejected
```

packages/liveblocks-core/src/ai.ts
```diff
@@ -947,6 +947,7 @@ export type Ai = {
       name: string,
       chatId?: string
     ): DerivedSignal<AiOpaqueToolDefinition | undefined>;
+    statusΣ: Signal<Status>;
```
nvie, line 950: At first I was a bit surprised this was needed, as I through you could use the `events.status.subscribe()` abstraction here, but I now realize this isn't available in `Ai`, only in `Room`, so we definitely need a signal like this in here 👍 

💡 **Aside:** Ideally, we would even expose this signal a layer deeper, because we also already offer the `getStatus()` getter which reads the current websocket status synchronously. It'd actually be nice if the ManagedSocket class itself would instead expose this signal instead directly, so we could either use it directly here, and are able to implement the `getStatus()` and `events.status` observables directly on top of it, both in Room as well as in AI here. (But definitely out of scope for this PR!)

packages/liveblocks-react-ui/src/primitives/AiComposer/index.tsx
```diff
@@ -114,10 +114,13 @@ export const AiComposerForm = forwardRef<HTMLFormElement, AiComposerFormProps>(
       : emptyMessagesΣ;
     const lastMessageId = useSignal(messagesΣ, getLastMessageId);
     const abortableMessageId = useSignal(messagesΣ, getAbortableMessageId);
-
+    const status = useSignal(client[kInternal].ai.signals.statusΣ);
```
nvie: Since we're only interested in re-rendering if the state changes to (or from, which never happens) `disconnected`, which means "permanently disconnected", you may use a selector here:

```suggestion
    const isAvailable = useSignal(client[kInternal].ai.signals.statusΣ, status => status !== "disconnected");
```

I'd also suggest using positive expressions when dealing with boolean logic (e.g. "enabled" instead of "disabled", or "connected" instead of "disconnected"). It makes all the boolean logic below a lot easier to read.

I'm using "isAvailable" here instead of "isConnected" to avoid confusing because also "initial" or "connecting" or "reconnecting" are fine states to have here. Only "disconnected" is the state where you will want to act on I guess.

packages/liveblocks-react-ui/src/primitives/AiComposer/index.tsx
```diff
@@ -114,10 +114,13 @@ export const AiComposerForm = forwardRef<HTMLFormElement, AiComposerFormProps>(
...
     const isDisabled = isSubmitting || disabled === true;
-    const canAbort = abortableMessageId !== undefined;
-    const canSubmit = !isEditorEmpty && !canAbort;
+
+    const isDisconnected = status !== "connected";
+
+    const canAbort = abortableMessageId !== undefined && !isDisconnected;
+    const canSubmit = !isEditorEmpty && !canAbort && !isDisconnected;
```
nvie: I think that the suggestion above might reduce this logic to:

```suggestion
    const canAbort = isAvailable && abortableMessageId !== undefined;
    const canSubmit = isAvailable && !isEditorEmpty && !canAbort;
```

packages/liveblocks-react/src/liveblocks.tsx
```diff
@@ -1451,26 +1452,32 @@ function useAiChatStatus(
         .waitUntilLoaded()
   );
 
-  return useSignal(
+  // Subscribe to connection status
+  const isDisconnected = useSignal(
+    client[kInternal].ai.signals.statusΣ,
+    (status) => status !== "connected"
+  );
```
nvie: I'd also flip this logic:


```suggestion
  const isAvailable = useSignal(
    client[kInternal].ai.signals.statusΣ,
    (status) => status !== "disconnected"
  );
```

This is because the WebSocket connection states "initial", "connecting", "reconnecting", and "connected" are all fine / expected states. Only "disconnected" is a final error status that we want to distinguish on here.

packages/liveblocks-react-ui/src/primitives/AiComposer/index.tsx
SugarDarius: Makes sense 👍🏻

SugarDarius:
@nvie PTAL thanks 🙏🏻

PR #2758, opened 2025-10-25
don't normalize 400 levels

jrowny:
Right now all 400 levels are being returned as 403, which means you can't see:

429 = rate limiting
409 = room already exists
404 = room not found
400 = bad request
422 = bad input

This normalize status is only used on authorizeUser and identifyUser, so that's probably why it was missed. 

A user ran into an issue sending more than 1024 bytes and did not receive a 422, as they should have.


jrowny:
@nvie could we just make it not retry on any 400 level? is there any 400 level that we would expect to want to retry?

jrowny:
I guess a 429 would be retryable

jrowny:
Okay, I changed the auth-manager to use a list of non-retryables.

There are some that could be retryable, like 429, 421, 410

Our backend only returns 429, 403, or 422, but because the auth endpoint is usually on a user's server, we can expect anything really, including things like 404/405 when they setup NextJS incorrectly or something like that, having authmanager not retry on those things would be useful, imo.

PR #2836, opened 2025-11-13
Lexical provider fix

jrowny:
Lexical expects a brand new provider each time it calls for one, we have have tried to keep one provider per-room so that we can keep them generally in sync. This PR allows lexical to get a brand new provider. 

It also updates to 0.35.0 and fixes a warning because of capitalization of `importDOM` on custom nodes.

This fixes LB-2596

---

> [!NOTE]
> Ensures Lexical uses a new Yjs provider when requested, upgrades all Lexical deps to 0.35, and updates custom nodes to the new importDOM API.
> 
> - **Lexical/Yjs Integration**:
>   - Add `forceNewProvider` option to `getYjsProviderForRoom` and destroy existing provider when requested.
>   - `LiveblocksPlugin` uses `getYjsProviderForRoom(room, {}, true)` to supply a fresh provider to Lexical.
> - **Custom Nodes**:
>   - Rename `importDom` to `importDOM` in `mention-node` and `group-mention-node` to match API.
> - **Dependencies**:
>   - Upgrade `lexical` and `@lexical/*` packages to `0.35.0` across example and packages; align peer deps.
>   - Bump `prismjs` to `1.30.0`; minor lockfile additions (e.g., `happy-dom`, related types).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 24b22346e0e8395f0298b46f02c7f703bb1ee5b6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/liveblocks-yjs/src/providerContext.ts
```diff
@@ -19,11 +19,17 @@ const providersMap = new WeakMap<OpaqueRoom, LiveblocksYjsProvider>();
  */
 const getYjsProviderForRoom = (
   room: OpaqueRoom,
-  options: ProviderOptions = {}
+  options: ProviderOptions = {},
+  forceNewProvider: boolean = false
```
nvie, line 23: Why/when would you need this option?

jrowny: Because lexical accepts a factory function that assumes a brand new uninitialized provider each time it executes the factory

packages/liveblocks-react-lexical/src/liveblocks-plugin-provider.tsx
```diff
@@ -141,9 +141,10 @@ export const LiveblocksPlugin = ({
 
   const providerFactory = useCallback(
```
jrowny, line 142: @nvie this is the provider factory that's given to Lexical, which requires a new provider each time it is executed.

packages/liveblocks-yjs/src/providerContext.ts
```diff
@@ -19,11 +19,17 @@ const providersMap = new WeakMap<OpaqueRoom, LiveblocksYjsProvider>();
  */
 const getYjsProviderForRoom = (
```
nimeshnayaju, line 20: This method is publicly available and I feel hesitant to expose a new experimental feeling parameter (`forceNewProvider`) through a public method. Could we create a private method and use it exclusively in Lexical for now? In the core, it feels like it's replacing the Y.js provider (if it exists) with a new one. If that feels like a lot of effort and we want to publish this quickly, could we rename this to say it's experimental to give ourselves some flexibility on the API?

PR #2866, opened 2025-11-27
Stricten assumption about storage root internals

nvie:
This PR consists of two parts:

1. First, **rewrite tests**. The storage based test cases are updated to match the reality of what actual storage docs look like in practice: with a clearly named root node (called `"root"`, not the legacy key `"0:0"`). This change does not break any tests.
2. Secondly, **remove legacy support code**. This part removes the logic to support reading "legacy rooms" that previously did not use `"root"` as the storage tree's root node (ac1870eeb4f44f9b6d7a0dc1d553ca8bfeb8fcf6). This was only the case for very old rooms (created before Oct 2022), and is no longer needed, as all old rooms have been migrated to the new storage format at this point.

### DIFF
--- a/packages/liveblocks-core/src/crdts/LiveObject.ts
+++ b/packages/liveblocks-core/src/crdts/LiveObject.ts
@@ -52,6 +52,10 @@ export type LiveObjectUpdates<TData extends LsonObject> = {
   updates: LiveObjectUpdateDelta<TData>;
 };
 
+function isRootCrdt(id: string, _: SerializedCrdt): _ is SerializedRootObject {
+  return id === "root";
+}
+
 /**
  * The LiveObject class is similar to a JavaScript object that is synchronized on all clients.
  * Keys should be a string, and values should be serializable to JSON.
@@ -79,16 +83,15 @@ export class LiveObject<O extends LsonObject> extends AbstractCrdt {
     let root: SerializedRootObject | null = null;
 
     for (const [id, crdt] of items) {
-      if (id === "root") {
-        root = crdt as SerializedRootObject;
+      if (isRootCrdt(id, crdt)) {
+        root = crdt;
       } else {
-        const childCrdt = crdt as SerializedChild;
-        const tuple: IdTuple<SerializedChild> = [id, childCrdt];
-        const children = parentToChildren.get(childCrdt.parentId);
+        const tuple: IdTuple<SerializedChild> = [id, crdt];
+        const children = parentToChildren.get(crdt.parentId);
         if (children !== undefined) {
           children.push(tuple);
         } else {
-          parentToChildren.set(childCrdt.parentId, [tuple]);
+          parentToChildren.set(crdt.parentId, [tuple]);
         }
       }
     }


packages/liveblocks-core/src/crdts/LiveObject.ts
```diff
@@ -74,20 +74,21 @@ export class LiveObject<O extends LsonObject> extends AbstractCrdt {
...
 
     for (const [id, crdt] of items) {
-      if (isRootCrdt(crdt)) {
-        root = [id, crdt];
+      if (id === "root") {
+        root = crdt as SerializedRootObject;
       } else {
-        const tuple: IdTuple<SerializedChild> = [id, crdt];
-        const children = parentToChildren.get(crdt.parentId);
+        const childCrdt = crdt as SerializedChild;
```
jrowny: I kind of liked the "isRoot" because we could avoid the "as." I don't mind this but you could keep `isRootCrdt` and just do `id === "root"` and kept the code as it was before? You could still get rid of the isChild by just inverting isRoot. 

i.e. 
```
export function isRootCrdt(_: SerializedCrdt, id: string): _ is SerializedRootObject {
  return id ===" root";
}

let child:SerializedChild;
let root:SerializedRootObject;
const foo = {parentKey: undefined, parentId: undefined, type: 1} as SerializedCrdt;
const id = "root";

if(isRootCrdt(foo, id)) {
   root = foo;
} else {
   child = foo;
}
```

I tested this in the TS playground here: https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBDAnmYcDKwoEsCGAbLAL2ABMBhKE+AXnU1wOJICUIIYB5AIwCtgBjeAB862fEVJkAFljwkA3AChQkWAmSoMYxqVbtufQXFoBvRXDhRgOEhAB2eROpQAuOAEYlFgPTe4e+AheARgAZzhbOwByeCkcADdUHDgwHCs7eCs8HBgse1CZMHNLa0jHFLTgDIBJEgB+NztgRKgvEpt7ctT0mABpYEQGuCaWpQBfJWVwaHgkFFEGCXIZOWM4MwsrDocnOeA3ACY2rbKnbqqYWrdQmGw7AHNj0s6zyoz+xGvbrAfxyZUZnAAGYAVzsgjydjgWFCAQoVAAFAB9NxaRZMeEwAA00JIXzu9wAlG4kdDwmjxEwAgYQutilYYCCoFCsCRjNRqAAiSxsGCc8aKRR4YDwfgrPEUnTLWQKIUinnsFySpbU4KCJT8fLwIFsNYmc7vAZuMEkYBAn6kHEGy54uAms0Wkg4vZudxjOA4cn0SmSSgwSZYIEImFwv0InUQHGcqC8zmEwl0iwKmjAtjjODAPChVAbCximVrCMCxRAA

nvie: Yeah, that's actually quite a good idea and a clever use of type guard to refine the type of one param using the other — never thought of that. Fixed in 47306557d51c90514906a006ec52586ec2ac40ba.

PR #2873, opened 2025-11-28
Fix typos/errors in docs

marcbouchenoire:
This PR contains multiple passes to fix various typos and errors throughout the docs.

For https://github.com/liveblocks/liveblocks/commit/5f4ca3bfea3b6e994648a170e1fdb22e4ff16ca3, the main issue is that we need to highlight both Suspense and non-Suspense specificities but in some places (snippets, properties tables, etc) we have to somehow choose one version. We want to prioritize Suspense imports but at the same time Suspense snippets are often more complex so we mixed the two.

### DIFF
--- a/docs/pages/api-reference/liveblocks-react.mdx
+++ b/docs/pages/api-reference/liveblocks-react.mdx
@@ -4209,17 +4209,25 @@ useErrorListener((error) => {
 
 ### useSearchComments [@badge=RoomProvider]
 
-Returns the result of searching comments by text within the current room. The
-result includes the ID and the plain text content of the matched comments along
-with the parent thread ID of the comment. The hook allows filtering by metadata
-and resolved status of the parent thread and presence and absence of attachments
-and mentions in the comment.
+Search comments in the current room using semantic search and various filters.
+Returns a list of results, including the ID and the plain text content of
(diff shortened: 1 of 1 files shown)


examples/nextjs-yjs-monaco/src/components/CollaborativeEditor.tsx
```diff
@@ -55,7 +55,7 @@ export function CollaborativeEditor() {
         <Editor
           onMount={handleOnMount}
           height="100%"
-          width="100hw"
+          width="100%"
```
SugarDarius, line 58: Expected changes here?

SugarDarius (approved):
Thank you 🙏🏻 ✅

examples/nextjs-yjs-monaco/src/components/CollaborativeEditor.tsx
marcbouchenoire: Yes, `hw` doesn't exist and `%` is better than `vw` in this case.

SugarDarius: Sorry my comment was very unclear. But you answered perfectly as I was thinking about using `vw` instead.

PR #2887, opened 2025-12-05
Comments metadata

marcbouchenoire:
This PR adds support for comment-level metadata.

The goal is to have zero breaking changes, including type-level ones.

📚 [Tech design](https://www.notion.so/liveblocks/Comments-metadata-2c082084c81280dbab4afc9ba33edfd1)
*️⃣ [Backend PR](https://github.com/liveblocks/liveblocks-backend/pull/1396)

### To do

- [x] Changelog
- [x] Documentation
- [x] Examples

---

> [!NOTE]
> **Overview**
> 
> Enables comment-level metadata end-to-end and generalizes thread/comment typing.
> 
> - Introduces `CommentMetadata` and propagates generics `<TM, CM>` across `Room`, `Client`, API types, and protocol (defaults via `DTM`/`DCM`)
> - REST/Room APIs: `createThread`/`createComment` accept comment metadata; new `editCommentMetadata`; typed responses (`CommentData<CM>`, `ThreadData<TM, CM>`); new server msg `COMMENT_METADATA_UPDATED`
> - Node SDK: typed `createThread`, `createComment`, `editComment`, `editCommentMetadata`; tests for comment metadata
> - React/React UI/Lexical/TipTap: components and hooks support comment metadata (`Composer` accepts `commentMetadata`; new `useEditCommentMetadata`; updates to `useCreateComment`/`useEditComment`; Thread/Anchored/Floating components genericized)
> - Webhooks: adds `commentMetadataUpdated` event
> - Widespread test updates and utilities adjusted to new generics; maintains backward compatibility via default types
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 453651ff00e431f6e6387a9f8fd557ef3a0f5780. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/liveblocks-react/test-d/augmentation.test-d.tsx
+++ b/packages/liveblocks-react/test-d/augmentation.test-d.tsx
@@ -786,29 +786,17 @@ declare global {
         version: 1,
         content: [{ type: "paragraph", children: [{ text: "hi" }] }],
       },
-    });
-
-    expectType<"comment">(comment.type);
-    expectType<string>(comment.id);
-    expectType<string>(comment.threadId);
-
-    const commentWithMetadata = createComment({
-      threadId: "th_xxx",
-      body: {
-        version: 1,
-        content: [{ type: "paragraph", children: [{ text: "hi" }] }],
-      },
       metadata: {
         priority: 2,
         reviewed: false,
       },
     });
 
-    expectType<"comment">(commentWithMetadata.type);
-    expectType<string>(commentWithMetadata.id);
-    expectType<number>(commentWithMetadata.metadata.priority);
-    expectType<boolean | undefined>(commentWithMetadata.metadata.reviewed);
-    expectError(commentWithMetadata.metadata.nonexisting);
+    expectType<"comment">(comment.type);
+    expectType<string>(comment.id);
+    expectType<number>(comment.metadata.priority);
+    expectType<boolean | undefined>(comment.metadata.reviewed);
+    expectError(comment.metadata.nonexisting);
   }
 }
 
@@ -823,24 +811,12 @@ declare global {
       version: 1,
       content: [{ type: "paragraph", children: [{ text: "hi" }] }],
     },
-  });
-
-  expectType<"comment">(comment.type);
-  expectType<string>(comment.id);
-  expectType<string>(comment.threadId);
-
-  const commentWithMetadata = createComment({
-    threadId: "th_xxx",
-    body: {
-      version: 1,
-      content: [{ type: "paragraph", children: [{ text: "hi" }] }],
-    },
     metadata: { priority: 1 },
   });
 
-  expectType<"comment">(commentWithMetadata.type);
-  expectType<number>(commentWithMetadata.metadata.priority);
-  expectError(commentWithMetadata.metadata.nonexisting);
+  expectType<"comment">(comment.type);
+  expectType<number>(comment.metadata.priority);
+  expectError(comment.metadata.nonexisting);
 }
 
 // ---------------------------------------------------------


packages/liveblocks-core/src/protocol/Comments.ts
```diff
@@ -163,28 +165,33 @@ export type SearchCommentsResult = {
 /**
  * Represents a thread of comments.
  */
-export type ThreadData<M extends BaseMetadata = DM> = {
+export type ThreadData<
+  TM extends BaseMetadata = DTM,
+  CM extends BaseMetadata = DCM,
```
marcbouchenoire, line 170: Not just about this line specifically but there's a question about `CommentMetadata` generics and augmentation types [in the tech design](https://www.notion.so/liveblocks/Comments-metadata-2c082084c81280dbab4afc9ba33edfd1?source=copy_link#2c082084c81281949b8ed9596c2b2e09).

packages/liveblocks-core/src/types/LiveblocksError.ts
```diff
@@ -41,6 +42,13 @@ type CommentsOrNotificationsErrorContext =
       threadId: string;
       metadata: Patchable<BaseMetadata>;
     }
+  | {
+      type: "EDIT_COMMENT_METADATA_ERROR";
+      roomId: string;
+      threadId: string;
+      commentId: string;
+      metadata: Patchable<BaseMetadata>;
```
marcbouchenoire, line 50: Should we use augmentation types (e.g. `DTM`) in Liveblocks errors instead?

packages/liveblocks-core/src/api-client.ts
```diff
@@ -136,18 +136,21 @@ export interface RoomHttpApi<M extends BaseMetadata> {
...
     threadId,
+    commentMetadata,
     attachmentIds,
   }: {
     roomId: string;
     threadId?: string;
     commentId?: string;
-    metadata: M | undefined;
+    metadata: TM | undefined;
+    // TODO: Finalize API design
```
marcbouchenoire: [Question about this in the tech design](https://www.notion.so/liveblocks/Comments-metadata-2c082084c81280dbab4afc9ba33edfd1?source=copy_link#2c082084c81280eda19fcd86ee3c0095).

packages/liveblocks-core/src/room.ts
```diff
@@ -2847,7 +2864,9 @@ export function createRoom<
     roomId: string;
     threadId?: string;
     commentId?: string;
-    metadata: M | undefined;
+    metadata: TM | undefined;
+    // TODO: Finalize API design
+    commentMetadata: CM | undefined;
```
marcbouchenoire, line 2899: Same [comment](https://github.com/liveblocks/liveblocks/pull/2887#discussion_r2606432329).

packages/liveblocks-react/src/room.tsx
```diff
@@ -2507,21 +2663,22 @@ function useStorageSuspense<S extends LsonObject, T>(
...
   const room = useRoom();
 
-  const { store } = getRoomExtrasForClient<M>(client);
+  const { store } = getRoomExtrasForClient<TM, CM>(client);
   const queryKey = makeRoomThreadsQueryKey(room.id, options.query);
 
   use(store.outputs.loadingRoomThreads.getOrCreate(queryKey).waitUntilLoaded());
 
-  const result = useThreads(options);
+  // TODO: Breaking change? CM can't be inferred from the params so the generics are now required
```
marcbouchenoire: `useThreads<TM>(...)` will trigger a type error and ask for a second generic, `useThreads()` is fine and will use `BaseMetadata` by default.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1190,18 +1237,20 @@ function useMutation<
   S extends LsonObject,
   U extends BaseUserMeta,
   E extends Json,
-  M extends BaseMetadata,
+  TM extends BaseMetadata,
+  // TODO: Investigate breaking change in generic ordering
```
marcbouchenoire: Was the last `F` generic supposed to be manually set? If so, we'd need to use the `TM, F, CM` ordering instead of `TM, CM, F`.

packages/liveblocks-react/src/__tests__/useEditCommentMetadata.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+          priority: 2,
+          reviewed: null,
+        },
+      })
+    );
+
+    expect(result.current.threads).toBeDefined();
+    expect(result.current.threads?.[0]?.comments[0]?.metadata).toEqual({
+      priority: 2,
+      reviewed: null,
```
marcbouchenoire, line 193: [Question about this in the tech design](https://www.notion.so/liveblocks/Comments-metadata-2c082084c81280dbab4afc9ba33edfd1?source=copy_link#2c382084c81280cb90c9d86b19a65f6b).

packages/liveblocks-core/src/protocol/Comments.ts
nvie, line 170: Note sure if you already did this, but I always find it useful to locally remove these default `= DCM` types, and then follow all TS errors to make sure all of our internal call sites at least populate it explicitly. Then, after you're done fixing all those, put back the `= DCM` here.

Put differently, internally we should never have a case of `ThreadData` (without type params) or `ThreadData<TM>` (one param).

packages/liveblocks-core/src/types/LiveblocksError.ts
nvie: I'd keep it generic in the error case.

packages/liveblocks-core/src/api-client.ts
nvie: Hmm. Good point. I don't see a simple way without a complex migration, so I'd just stick with what you went for here and maybe reconsider improving this API with a 4.x release later.

packages/liveblocks-core/src/api-client.ts
```diff
@@ -1842,11 +1891,12 @@ export function createApiClient<M extends BaseMetadata>({
     };
   }
 
-  async function getUserThreadsSince_experimental<
-    M extends BaseMetadata,
```
nvie, line 1846: I noticed the old function took a local `M` param, and you removed it. Was this just never fixed and you're fixing it now?

packages/liveblocks-react/src/types/index.ts
```diff
@@ -118,7 +119,7 @@ export type UseAiChatsOptions = {
   query?: AiChatsQuery;
 };
 
-export type ThreadsQuery<M extends BaseMetadata> = {
+export type ThreadsQuery<TM extends BaseMetadata = DTM> = {
```
nvie: You're introducing a default type here that wasn't here before. Are you sure you want to do this?

packages/liveblocks-react/src/room.tsx
```diff
@@ -328,9 +342,10 @@ type RoomLeavePair<
   S extends LsonObject,
   U extends BaseUserMeta,
   E extends Json,
-  M extends BaseMetadata,
+  TM extends BaseMetadata,
+  CM extends BaseMetadata,
```
nvie, line 345: **Optional side thought.** Seeing this here now, I'm tempted to rename `BaseMetadata` → `Metadata`. What do you think? I believe this is also exactly what we name this type in the backend.

packages/liveblocks-react/src/room.tsx
nvie: I don't think (and surely _hope_) anybody is setting this type manually. It's used only to infer the return type of this function correctly. So I think this change is safe to make in practice.

packages/liveblocks-react/src/room.tsx
```diff
@@ -1712,21 +1846,21 @@ function useDeleteRoomComment(roomId: string) {
   );
 }
 
-function useAddReaction<M extends BaseMetadata>() {
-  return useAddRoomCommentReaction<M>(useRoom().id);
```
nvie, line 1716: 🙌

packages/liveblocks-react/src/room.tsx
nvie: I think that's fine? I would hope that `useThreads()` would not be typed manually in user's code bases. Ideally they are typed at the factory level, or set via global typing anyway.

marcbouchenoire: I think so too, I wanted to surface it through the noise just in case 👍

packages/liveblocks-react/src/types/index.ts
marcbouchenoire: Great catch, fixed in https://github.com/liveblocks/liveblocks/pull/2887/commits/2cee59ad07bfe59969d81f28e9238252c3516a73.

packages/liveblocks-core/src/api-client.ts
marcbouchenoire: Yes it shouldn't have been there, even the type in `LiveblocksHttpApi` was not referencing it, only this implementation.

packages/liveblocks-react/src/room.tsx
marcbouchenoire: I think it'd make sense (yep the backend uses `Metadata` in even more places with [my PR](https://github.com/liveblocks/liveblocks-backend/pull/1396)), but also we have used that convention for many types now in this repo:
- `BaseMetadata` (publicly exported as is)
- `BaseUserMeta` (publicly exported as is)
- `BaseActivitiesData` (publicly exported as is)
- `BaseGroupInfo`
- `BaseRoomInfo`

So I think that's more of a 4.x question 😕

packages/liveblocks-core/src/protocol/Comments.ts
marcbouchenoire: I started without for these reasons but I just did a new pass of that and [found a few missing types](https://github.com/liveblocks/liveblocks/pull/2887/commits/9ad13a571e6c12f5bb9be96e3714ee3ad84041cc), thanks for the reminder!

PR #2898, opened 2025-12-12
Branding

jrowny:
Adds the ability to show branding if enabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v3.12.0
 
-### `@liveblocks/core`
+### `@liveblocks/client`
 
 - Liveblocks branding will appear for free plans.
 - Added badgeLocation option for the client library to control the branding


packages/liveblocks-core/src/types/Others.ts
```diff
@@ -33,3 +33,9 @@ export enum TextEditorType {
   TipTap = "tiptap",
   BlockNote = "blocknote",
 }
+
+export type BadgeLocation =
```
SugarDarius, line 37: Surely a nitpick here from me but I wonder if the naming `BadgePosition` wouldn't be better? 🤔 
Any thoughts on this?

packages/liveblocks-core/src/client.ts
```diff
@@ -525,6 +526,12 @@ export type ClientOptions<U extends BaseUserMeta = DU> = {
    */
   preventUnsavedChanges?: boolean;
 
+  /**
+   * The location where the brand badge should be displayed when using a free plan.
+   * Default is "bottom-right".
+   */
+  badgeLocation?: BadgeLocation;
```
SugarDarius, line 533: Ditto. Reflecting on this and as an end users my take here would be to position the badge not locate it.

packages/liveblocks-core/src/brand.ts
```diff
@@ -0,0 +1,193 @@
...
+  if (document.getElementById("liveblocks-badge")) {
+    return;
+  }
+
+  // Create and inject the badge HTML
+  const badgeDiv = document.createElement("div");
+  badgeDiv.id = "liveblocks-badge";
+  badgeDiv.style.position = "fixed";
+  badgeDiv.style.opacity = "0";
+  badgeDiv.style.transition = "opacity 300ms";
```
SugarDarius, line 23: Just a classic linear transition here? No customized easing?

SugarDarius, line 23: Do we want to add a `zIndex` here? To ensure the badge is always on top of very elements.

packages/liveblocks-core/src/brand.ts
```diff
@@ -0,0 +1,193 @@
...
+  badgeDiv.appendChild(hideButton);
+
+  document.body.appendChild(badgeDiv);
+
+  // trigger twice so it goes from 0 to 1
+  requestAnimationFrame(() => {
+    requestAnimationFrame(() => {
+      badgeDiv.style.opacity = "1";
+    });
+  });
```
SugarDarius, line 191: Can you explain more why it requires here two `requestAnimationframe`?

packages/liveblocks-core/src/brand.ts
```diff
@@ -0,0 +1,193 @@
...
+): void => {
+  // Only inject in browser environments
+  if (typeof document === "undefined") {
+    return;
+  }
+
+  // Check if badge already exists to avoid duplicates
+  if (document.getElementById("liveblocks-badge")) {
+    return;
+  }
```
SugarDarius, line 16: Overkill take here 😅 but we could use a `MutationObserver` (heavy though) to ensure to always inject if an external script try to remove the badge after displaying this script execution.

SugarDarius (commented):
Good job 💪🏻 🚀 
I left some comments to discuss, not real blockers 🙂

packages/liveblocks-core/src/brand.ts
jrowny: It's not very noticeable at 300ms but I can add if you want

packages/liveblocks-core/src/brand.ts
jrowny: I guess because otherwise it never renders it's zero opacity state, without two, it doesn't do the fade in.

packages/liveblocks-core/src/brand.ts
jrowny: I think we just let that script win

packages/liveblocks-core/src/types/Others.ts
jrowny: The ticket has badgeLocation and I think it sounds okay to my ear.

packages/liveblocks-core/src/brand.ts
jrowny: thanks for the reminder, I added it

SugarDarius: All right! Let's keep the linear on then 👍🏻

packages/liveblocks-core/src/brand.ts
SugarDarius: Ah I see! It's to avoid both states are rendered in the same frame 👍🏻 Can you please add a comment to explain it for legacy?

Is there any reasons where the Web Animation Api isn't an option using `Element:.animate()`? 
Such as
```tsx
const badgeDiv = document.createElement("div");
badgeDiv.style.opacity = "0";

/* ... */
badgeDiv.animate(
  [
    { opacity: 0 },
    { opacity: 1 }
  ],
  {
    duration: 300,
    fill: 'forwards' // Keeps final state rendered
  }
);
```
If application it avoids the double stacked `requestAnimationFrame` and is easier to read and understand the intent imo.

packages/liveblocks-core/src/types/Others.ts
SugarDarius: Okay 👍🏻 
I was just asking because from an API design naming standpoint it feels weird to me to use `location` over `position`. But I'm not an English native speaker to I trust you 👍🏻 🙂

FlowFlorent (approved):

packages/liveblocks-core/src/protocol/ServerMsg.ts
```diff
@@ -275,6 +275,11 @@ export type RoomStateServerMsg<U extends BaseUserMeta> = {
   readonly users: {
     readonly [otherActor: number]: U & { scopes: string[] };
   };
+
+  /**
+   * Optional metadata sent from the server to the client.
+   */
+  readonly meta?: Record<string, unknown>;
```
nvie: Could we make the `unknown` more specific? At the very least it should be `Record<string, Json>` (because it comes from a JSON message), but maybe we can be even more precise, like `Record<string, string | string[]>` if we know the server will only ever send that?

packages/liveblocks-core/src/room.ts
```diff
@@ -1133,6 +1135,7 @@ export type DynamicSessionInfo = {
   readonly actor: number;
   readonly nonce: string;
   readonly scopes: string[];
+  readonly meta?: Record<string, unknown>;
```
nvie: Same as my comment above.

CHANGELOG.md
```diff
@@ -4,14 +4,14 @@
 
 ### `@liveblocks/core`
```
marcbouchenoire: We should say `@liveblocks/client` here even if it's in core since it's an internal package.

PR #2930, opened 2025-12-18
Convert our numeric TS enums to extensible const objects

nvie:
This PR changes our TypeScript `enum` declarations in the protocol layer to use `Object.freeze()` const objects with derived types and namespaces.

Enums changed:
- `ClientMsgCode`
- `ServerMsgCode`
- `OpCode`
- `CrdtType`

## Why

TypeScript enums are "closed" - they cannot be extended or augmented from other codebases. Working on new protocol versions in the backend, historically means quite a bit of type juggling and local temporary duplication. This change aims to fix that, without the need for publishing a new verssion of the client first.

The new pattern (`const` + `Object.freeze()` + type + namespace) provides the same API surface while allowing the backend to define additional values as needed during protocol development. I'll open a backend PR soon that would use this new way of doing enums.

## Technical impact on our bundle

[In this gist, you can see the impact on our bundle.](https://gist.github.com/nvie/16409bc14fae273887a7521725f88a5a)
Basically a comparison of the old output of `turbo run build` with this new way.

## No breaking change

As far as I can tell, this is not a breaking change.

- The runtime values remain identical
- The type-level API is preserved (`ServerMsgCode.UPDATE_PRESENCE` works both as a value and as a type due to the similarly named type/namespace)
- No changes to public APIs or exports

### DIFF
--- a/packages/liveblocks-core/src/index.ts
+++ b/packages/liveblocks-core/src/index.ts
@@ -271,7 +271,8 @@ export {
   patchNotificationSettings,
 } from "./protocol/NotificationSettings";
 export type {
-  AckOp,
+  AckOpV7,
+  AckOpV8,
   CreateListOp,
   CreateMapOp,
   CreateObjectOp,
@@ -283,7 +284,7 @@ export type {
   SetParentKeyOp,
   UpdateObjectOp,
 } from "./protocol/Op";
-export { ackOp, OpCode } from "./protocol/Op";
+export { OpCode } from "./protocol/Op";
 export type {
   RoomSubscriptionSettings,
   UserRoomSubscriptionSettings,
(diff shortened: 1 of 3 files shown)


marcbouchenoire (approved):
I like the change! I wonder if we should have "enum"-named helpers/branding for the creation side (both runtime and type level), not for DRY reasons but to make it clearer what these are and explicitly label them as such? I think that would help prevent us adding new actual `enum`s later on by mistake.

nvie:
> _I wonder if we should have "enum"-named helpers/branding for the creation side (both runtime and type level), not for DRY reasons but to make it clearer what these are and explicitly label them as such?_

I tried using a helper like `const CrdtType = makeEnum({ ... })`, but unfortunately it's not possible (AFAIK) to infer the necessary `namespace`-based trick that allows us to define both a runtime value `CrdtType`-the-object, and the type-level `CrdtType` namespace that allows us to avoid having to use `typeof CrdtType.OBJECT` in favor of "just" being able to say `CrdtType.OBJECT` still in type positions.

PR #2932, opened 2025-12-19
Prep for new Proper Ack™

nvie:
Replaces our (H)AckOp by a Proper Ack™ that will be used in the next internal V8 protocol. This future ack will correct the previous hack, and will also serve to reduce the size of our future acks, by no longer repeating back the entire Op anymore.

### DIFF
--- a/packages/liveblocks-core/src/protocol/Op.ts
+++ b/packages/liveblocks-core/src/protocol/Op.ts
@@ -32,12 +32,13 @@ export namespace OpCode {
  * only.
  */
 export type Op =
-  | AckOp
   | CreateOp
   | UpdateObjectOp
   | DeleteCrdtOp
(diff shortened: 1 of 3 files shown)


packages/liveblocks-core/src/protocol/Op.ts
```diff
@@ -105,40 +108,23 @@ export type DeleteCrdtOp = {
...
- * old clients such Op is accepted, but will effectively be a no-op as that
- * node does not exist, but as a side-effect the Op will get acknowledged.
- */
-export function ackOp(opId: string): AckOp {
-  return {
-    type: OpCode.DELETE_CRDT,
-    id: "ACK", // (H)ACK
-    opId,
-  };
-}
```
nvie, line 138: Not needed on the client side, will move this to server side.

PR #2936, opened 2025-12-23
Bump to V8, add streaming support for initial storage

nvie:
Related backend PR: https://github.com/liveblocks/liveblocks-backend/pull/1437.

This introduces protocol V8, with a change in response to the `FETCH_STORAGE` client message.

- In V7, we returned `STORAGE_STATE_V7` (was already renamed on the target branch), containing all nodes at once. We keep using this format for all existing clients out there (backward compatible).
- In V8, the server side introduces two optimizations:
  1. First of all, the serialization format is compacted to fit more data in less space. We basically serialize a node like `["0:0",{"type":0,"parentId":"root","parentKey":"a","data":{"x":123}}]` in a more compact form now: `["0:0",0,"root","a",{"x":123}]`. Same data, less overhead. On large documents, this overhead can really add up.
  2. Secondly, if the room happens to be using the new storage engine (not GA yet), the server response will additionally get _streamed_. This allows the client to start processing the first nodes before the server is done sending, and on the server there will be less memory pressure.

An overview:

| Protocol   | Storage Engine 1 (old)           | Storage Engine 2 (new)             |
| ---------- | -------------------------------- | ---------------------------------- |
| V7         | `STORAGE_STATE_V7` (200)<br/>🟡 Classic serialization format<br/>🟡 All nodes at once | `STORAGE_STATE_V7` (200)<br/>🟡 Classic serialization format<br/>🟡 All nodes at once |
| V8         | `STORAGE_CHUNK` (210)<br/>✅ Compact format<br/>🟡 Still all nodes at once | `STORAGE_CHUNK` (210)<br/>✅ Compact format<br/>✅ Multiple chunks (streaming) |

(A prior iteration of this PR introduced a new `stream: true` field on `FETCH_STORAGE`, but this turned out not to be needed.)

**The E2E tests are all failing for the obvious reason that this backend has not yet been deployed.**

---

> [!NOTE]
> Introduces protocol v8 with a leaner, streamable initial storage load and compact serialization.
> 
> - Add `STORAGE_CHUNK` and `STORAGE_STREAM_END` server messages; deprecate `STORAGE_STATE_V7` for v8 paths
> - Implement compact node format and helpers (`nodeStreamToCompactNodes`, `compactNodesToNodeStream`); export new compact node types
> - Update room logic to buffer chunked nodes, build storage from streamed data, and diff/apply updates; switch WS path to `/v8`
> - Adjust tests and mocks across core/react/redux/zustand to emit/consume chunked messages and compact nodes
> - Update `CHANGELOG.md` and public exports to reflect v8 and new APIs
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6dcc6cefbd9d051567c5870c4f22e1f38bb2cf23. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/liveblocks-core/src/room.ts
+++ b/packages/liveblocks-core/src/room.ts
@@ -1853,7 +1853,16 @@ export function createRoom<
     }
 
     if (context.root !== undefined) {
-      updateRoot(nodes);
+      const currentItems: NodeMap = new Map();
+      for (const [id, crdt] of context.pool.nodes) {
+        currentItems.set(id, crdt._serialize());
+      }
+
+      // Get operations that represent the diff between 2 states.
+      const ops = getTreesDiffOperations(currentItems, nodes);
+
+      const result = applyRemoteOps(ops);
+      notify(result.updates);
     } else {
       context.root = LiveObject._fromItems<S>(
         nodes as NodeStream,
@@ -1880,23 +1889,6 @@ export function createRoom<
     // Initial storage is populated using normal "set" operations in the loop
     // above, those updates can end up in the undo stack, so let's prune it.
     context.undoStack.length = stackSizeBefore;
-  }
-
-  function updateRoot(nodes: NodeMap) {
-    if (context.root === undefined) {
-      return;
-    }
-
-    const currentItems: NodeMap = new Map();
-    for (const [id, crdt] of context.pool.nodes) {
-      currentItems.set(id, crdt._serialize());
-    }
-
-    // Get operations that represent the diff between 2 states.
-    const ops = getTreesDiffOperations(currentItems, nodes);
-
-    const result = applyRemoteOps(ops);
-    notify(result.updates);
   }
 
   function _addToRealUndoStack(frames: Stackframe<P>[]) {


packages/liveblocks-core/src/protocol/ServerMsg.ts
```diff
@@ -13,7 +13,8 @@ export const ServerMsgCode = Object.freeze({
   ROOM_STATE: 104,
 
   // For Storage
-  INITIAL_STORAGE_STATE: 200,
+  INITIAL_STORAGE_STATE_V7: 200, // Only sent in V7
+  INITIAL_STORAGE_CHUNK: 210, // Used in V8+
```
jrowny: Would it be okay to just name this `STORAGE_CHUNK` or `STORAGE_STATE_CHUNK`? 

`INITIAL_STORAGE_CHUNK` reads to me like "The first storage chunk." 

I don't feel like we need "initial" at all, because it's clear with the word "state" that we're getting the state, vs. an update.

packages/liveblocks-zustand/src/__tests__/index.test.ts
```diff
@@ -150,8 +156,9 @@ async function prepareWithStorage<TState>(
 
   socket.callbacks.message[0]!({
     data: JSON.stringify({
-      type: ServerMsgCode.INITIAL_STORAGE_STATE,
-      items: options.items,
+      type: ServerMsgCode.INITIAL_STORAGE_CHUNK,
```
jrowny: rather than changing the tests from v7 to v8, shouldn't we maintain both tests for awhile until v8 adoption happens?

packages/liveblocks-core/src/room.ts
```diff
@@ -2553,6 +2572,7 @@ export function createRoom<
       // Only add the fetch message to the outgoing message queue if it isn't
       // already there
       messages.push({ type: ClientMsgCode.FETCH_STORAGE });
+      partialNodes.clear(); // Reset any partial state from previous fetch
```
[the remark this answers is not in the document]
nvie: Correct, this can go. I removed them in 8eaad2cebf547f7c7be0dfb41abbeb85aaf366e5.

packages/liveblocks-core/src/room.ts
```diff
@@ -2538,8 +2558,8 @@ export function createRoom<
   async function streamStorage() {
     // TODO: Handle potential race conditions where the room get disconnected while the request is pending
     if (!managedSocket.authValue) return;
-    const items = await httpClient.streamStorage({ roomId });
-    processInitialStorage({ type: ServerMsgCode.INITIAL_STORAGE_STATE, items });
+    const nodes = new Map(await httpClient.streamStorage({ roomId }));
+    processInitialStorage(nodes);
```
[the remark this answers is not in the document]
nvie: Valid point, but… not something I changed on this PR. This is already an issue on `main`.
Now that streaming is available (and _actually_ is a stream of chunks, rather than just one giant HTTP response), I'd much prefer to drop support for `unstable_streamData` over time, and recommend users upgrade their clients and switch to the new storage engine for rooms. (We can keep `unstable_streamData` in here to support old rooms that are still on the old engine for now, until we have a good migration plan.)

packages/liveblocks-core/src/protocol/ServerMsg.ts
nvie: I like that! I'll remove the `INITIAL_` prefixes from both.

packages/liveblocks-zustand/src/__tests__/index.test.ts
nvie: These client-side tests simulate the server behavior. (The best solution here would be to use the _actual_ server implementation here once we open-source `@liveblocks/server`, so this simulation won't be needed at all.)

Since we're simulating here, and since the client now uses v8, the server behavior actually changes, so this is accurate. In the backend test suite, we do test both protocol responses obviously. Just not needed here in the client side.

packages/liveblocks-core/src/protocol/ServerMsg.ts
nvie: Fixed in 0a26e503c73d4226f6bf97cc537613bfa907f9dc!

PR #2940, opened 2026-01-01
Add organizations to Next.js Starter Kit

CTNicholas:
Moved to this PR because of a difficult to solve merge conflict:

- https://github.com/liveblocks/liveblocks/pull/3109

### DIFF
--- a/starter-kits/nextjs-starter-kit/lib/actions/deleteDocument.ts
+++ b/starter-kits/nextjs-starter-kit/lib/actions/deleteDocument.ts
@@ -1,6 +1,7 @@
 "use server";
 
 import { auth } from "@/auth";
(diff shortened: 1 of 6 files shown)


starter-kits/nextjs-starter-kit/lib/actions/deleteDocument.ts
```diff
@@ -51,8 +51,8 @@ export async function deleteDocument({ documentId }: Props) {
     !userAllowedInRoom({
       accessAllowed: "write",
       userId: session?.user.info.id ?? "",
-      groupIds: session?.user.info.groupIds ?? [],
       room,
+      organizationId: session?.user.currentOrganizationId ?? "default",
```
marcbouchenoire: Should we put `"default"` in the constants file?

PR #2956, opened 2026-01-13
Fix Tiptap `Toolbar`

marcbouchenoire:
This PR fixes `Toolbar` not updating based on the current editor state. The issue is (roughly) that [Tiptap v3 made re-renders opt-in](https://tiptap.dev/docs/guides/upgrade-tiptap-v2#introducing-shouldrerenderontransaction) and `Toolbar` relied on them to update (e.g. `<Toggle active={editor.isActive("bold")} />`), `FloatingToolbar` didn't have the issue becomes it subscribes to selection changes which keeps re-renders. This PR only opts-in `Toolbar` to re-renders to fix the issue, we could optimize this but since we didn't have selective re-renders before I kept them out of scope.

**Report:** https://liveblocks.slack.com/archives/C038S22AG3S/p1768236343519729

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ metadata to individual comments in the same way as thread metadata.
 - Add `commentMetadata` prop to `FloatingComposer` to attach and update comment
   metadata.
 
+### `@liveblocks/react-tiptap`
+
+- Fix `Toolbar` not reflecting the editorʼs current state.
+
 ### `@liveblocks/node`
 
 - Add new parameters to `createThread`, `createComment`, and `editComment` to


packages/liveblocks-react-tiptap/src/toolbar/Toolbar.tsx
```diff
@@ -763,6 +763,13 @@ export const Toolbar = Object.assign(
       },
       forwardedRef
     ) => {
+      // Re-render the toolbar when the editor content and selection change.
+      useEditorState({
+        editor,
+        equalityFn: Object.is,
+        selector: (ctx) => ctx.editor?.state,
```
mareksuscak, line 770: @marcbouchenoire this isn't working as expected and selection is completely ignored or lagging. I feel like tiptap's state is mutable and not immutable so when selection changes, the state object is not recreated fully. Then because you override the equalityFn above and using Object.is which does a reference check for objects rather than deep equality check, this fix doesn't really do what it's designed to do. I checked the source code of tiptap and their default equality fn is `fast-deep-equal` which, if used, would've probably work correctly but possibly introduce performance issue. I'd recommend you either remove the equalityFn or change this selector fn to: `selector: (ctx) => ctx.editor?.state.selection,` which I verified works in our codebase.

PR #2957, opened 2026-01-14
Clarify opId semantics with explicit wire op types

nvie:
This PR adds two distinct type definitions to make clear when operations should have opIds and when they shouldn't. The goal is to encode the opId invariants directly in the type system so they're impossible to violate.

## Wire op types

Two new types distinguish operations by their network direction:

- `ClientWireOp` — Operations sent from client to server. These always include an `opId` so the server can acknowledge receipt.
- `ServerWireOp` — Operations received from the server. Three variants exist: echoed ops (with `opId`), ACK ops, and forwarded ops from other clients (without `opId`).

The base `Op` type no longer includes `AckOp` in its union since that's server-only.

## Reverse ops no longer get opIds

Previously, reverse operations (stored for undo/redo) were assigned opIds at creation time. Now they're stored without opIds, and only get assigned fresh opIds when actually applied during undo/redo. This has no behavioral impact but clarifies that opIds are for wire acknowledgement, not for identity tracking in the undo stack. The `HistoryOp` type is renamed to `Stackframe` to reflect this and to make it more obvious this isn't an "Op" in the way we talk about it everywhere else.

## Split `_toOps` methods

The CRDT classes now have two distinct serialization methods:

- `_toOps(parentId, parentKey)` — Returns ops without opIds, used for reverse operations only
- `_toOpsWithOpId(parentId, parentKey, pool)` — Returns ops with opIds, used for forward operations going over the wire

The base implementation in `AbstractCrdt._toOpsWithOpId` wraps `_toOps` and assigns opIds.

## LiveList unacknowledgedSets fix

Fixed a bug where `LiveList._toOps` was incorrectly populating `#unacknowledgedSets` during serialization for reverse ops. The unacknowledgedSets tracking is now only done when creating forward ops that will actually be sent to the server.

### DIFF
--- a/packages/liveblocks-node/src/client.ts
+++ b/packages/liveblocks-node/src/client.ts
@@ -8,6 +8,7 @@ import type {
   BaseMetadata,
   BaseUserMeta,
   ClientMsg,
+  ClientWireOp,
   CommentBody,
   CommentData,
   CommentDataPlain,
@@ -2697,7 +2698,7 @@ export class Liveblocks {
     // there hasn't been an update to the buffered ops for a while. This
     // behavior is slightly different from the browser client, which will emit
     // ops as soon as they are available (= throttling)
-    let opsBuffer: Op[] = [];
+    let opsBuffer: ClientWireOp[] = [];
     let outstandingFlush$: Promise<void> | undefined = undefined;
     let lastFlush = performance.now();
 
@@ -2746,7 +2747,7 @@ export class Liveblocks {
       const pool = createManagedPool(roomId, {
         getCurrentConnectionId: () => actor,
         onDispatch: (
-          ops: Op[],
+          ops: ClientWireOp[],
           _reverse: Op[],
           _storageUpdates: Map<string, StorageUpdate>
         ) => {


packages/liveblocks-core/src/room.ts
```diff
@@ -2683,22 +2706,20 @@ export function createRoom<
       throw new Error("redo is not allowed during a batch");
     }
 
-    const historyOps = context.redoStack.pop();
-    if (historyOps === undefined) {
+    const frames = context.redoStack.pop();
```
marcbouchenoire, line 2709: The "frame"/`Stackframe` naming isn't super intuitive to me (what a "frame" is) but I do see it in the backend, I'm surprised we didn't use it already client-side?

nvie: > The "frame"/`Stackframe` naming isn't super intuitive to me (what a "frame" is) but I do see it in the backend, I'm surprised we didn't use it already client-side?

Yeah, I struggled with a good name here for a stack entry. I found "entry" a bit vague (but you can argue "frame" is just as unclear). More common names could be "stack command" or "stack operations". Using "stack operations" I would actually like, as it would match our documentation…

Screen Shot 2026-01-14 at 12 03 38@2x

...but it would be confusing as such an "operation" could be actually comprised of multiple "ops" and even contain "presence changes" too, all in a single "stack frame". Basically a "stack frame" is whatever changes if you click an undo/redo button once.

nvie: > I do see it in the backend, I'm surprised we didn't use it already client-side

This is a bit surprising, because undo/redo stacks are a client-only concept (each client has its own undo/redo stack). The server does not know about undo/redo at all. Where did you see it in the backend?

marcbouchenoire: https://github.com/liveblocks/liveblocks-backend/blob/ce4a0cce603e4118dbd0d936a3a4889babfc12bd/apps/cloudflare/test/api/_utils/index.ts#L308, and it's related to Storage but it's different yeah.

nvie: (Ah, yes! That's actually a dreaded internal Cloudflare error we have to work around, nothing Liveblocks related.)

PR #2958, opened 2026-01-14
Deprecate tenantId in favour of organizationId

FlowFlorent:
#### @liveblocks/node
- client supports organizationId
- session support organizationId
- set tenantId as deprecated

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes request payload/query parameter shaping for multiple REST endpoints and session authorization; mistakes could route data to the wrong org or break filtering. Backward-compat mapping from `tenantId` to `organizationId` reduces upgrade risk but needs verification across endpoints.
> 
> **Overview**
> Adds first-class `organizationId` support across the `@liveblocks/node` client/session APIs while marking `tenantId` as deprecated.
> 
> Updates `prepareSession`/`Session.authorize` and multiple REST methods (e.g. `identifyUser`, `getRooms`, `createRoom`, inbox notification APIs, and `createGroup`) to send/query `organizationId`, with backward-compatible fallback that maps a provided `tenantId` into `organizationId` rather than sending `tenantId`.
> 
> Extends protocol types (`GroupData`) and Node client types (`RoomData`, query/options types) to include `organizationId`, and expands tests to cover both the new `organizationId` behavior and legacy `tenantId` compatibility.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ccc7331267d9fd1ff5ceeb0070ab91a3c5fb8fe6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/liveblocks-core/src/convert-plain-data.ts
```diff
@@ -210,6 +210,7 @@ export function convertToGroupData(data: GroupDataPlain): GroupData {
 
   return {
     ...data,
+    organizationId: data.organizationId ?? data.tenantId,
```
marcbouchenoire: Do we need this if the API returns both?

packages/liveblocks-node/src/__tests__/Session.test.ts
```diff
@@ -228,6 +235,15 @@ describe("authorization (new API)", () => {
...
+  test("can set organizationId when creating session", () => {
+    const session = makeSession({ organizationId: "org-123" });
+    expect(session).toBeDefined();
+
+    // The session should have the organizationId set internally
+    // We can verify this by checking that the session can be used normally
+    expect(session.allow("room-1", [P1]).hasPermissions()).toEqual(true);
+  });
+
   test("tenantId is optional when creating session", () => {
```
marcbouchenoire, line 250: Could we update the `tenantId` tests to make them "deprecated" too (maybe only testing the fallback or when both are set)?

packages/liveblocks-node/src/__tests__/client.test.ts
```diff
@@ -2846,6 +2924,74 @@ describe("client", () => {
...
+
+          if (
+            (data as typeof createGroupParams)?.groupId ===
+            createGroupParams.groupId
+          ) {
+            return HttpResponse.json(
+              {
+                type: "group",
+                id: "group1",
+                tenantId: "org1",
```
marcbouchenoire, line 2965: Should the mock reflect that the API returns both `tenantId` and `organizationId`?

packages/liveblocks-core/src/convert-plain-data.ts
FlowFlorent: Yes! Done ✅

packages/liveblocks-node/src/__tests__/client.test.ts
FlowFlorent: Adjusted!

PR #2966, opened 2026-01-16
Fix an issue where `LiveList` subscriptions would not always trigger

nvie:
Fix issue where LiveLists would not fire their subscriptions correctly, causing them to potentially show stale data after a specific combination of concurrent move and set operations. Not very common, but possible. The seemingly reported state would stay stale until the LiveList would get mutated again.

### DIFF
--- a/e2e/next-ai-kitchen-sink/playwright.config.ts
+++ b/e2e/next-ai-kitchen-sink/playwright.config.ts
@@ -1,10 +1,11 @@
+import { config } from "dotenv";
 import { defineConfig, devices } from "@playwright/test";
 
-/**
- * Read environment variables from file.
- * https://github.com/motdotla/dotenv
- */
-// require('dotenv').config();
+config({ path: ".env.local" });
+
+const BASE_URL =
+  process.env.NEXT_PUBLIC_LIVEBLOCKS_BASE_URL ?? "https://api.liveblocks.io";
+console.log(`Running against Liveblocks base URL: ${BASE_URL}`);
 
 /**
  * See https://playwright.dev/docs/test-configuration.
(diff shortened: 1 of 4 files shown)


packages/liveblocks-core/e2e/utils.ts
```diff
@@ -19,6 +19,10 @@ import { mapValues, wait, withTimeout } from "../src/lib/utils";
 import type { BaseUserMeta } from "../src/protocol/BaseUserMeta";
 import type { Room, RoomEventMessage } from "../src/room";
 
+const BASE_URL =
+  process.env.NEXT_PUBLIC_LIVEBLOCKS_BASE_URL ?? "https://api.liveblocks.io";
+console.log(`Running against Liveblocks base URL: ${BASE_URL}`);
```
nvie, line 24: (**Unrelated,** but sneaking this in here too.)
Add this line while at it, which makes it much easier to detect which env your running the E2E tests against, especially as you frequently change this.

packages/liveblocks-core/src/crdts/liveblocks-helpers.ts
```diff
@@ -151,6 +151,29 @@ export function lsonToLiveNode(value: Lson): LiveNode {
...
+ *
+ * Example:
+ * - Current: { "root": { a: 1 },  "node1": { b: 2 } }
+ * - New:     { "root": { a: 99 }, "node2": { c: 3 } }
+ *
+ * Returns:
+ *  - DELETE_CRDT for "node1" (removed)
+ *  - UPDATE_OBJECT for "root" (data changed: a: 1 → 99)
+ *  - CREATE_OBJECT for "node2" (added)
+ */
```
nvie, line 176: (**Unrelated,** but sneaking this in here too.)
Noticed this semi-related internal helper function did not have a clear doc.

PR #2974, opened 2026-01-20
Fix Slate uncaught errors in composers

marcbouchenoire:
This PR re-implements Slate's `focus` method without throwing to avoid `Composer` and `AiComposer` throwing in some conditions (e.g. `autoFocus` and unmounting in certain setups). We previously added a `try...catch` around it but [it can use `setTimeout` internally](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-dom/src/plugin/dom-editor.ts#L442) so we would lose control in these situations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 - Add more detailed timing info to `enableDebugLogging` to better debug
   connection issues
 
+### @liveblocks/react-ui
+
+- Fix an issue where `Composer` and `AiComposer` would throw an error when
+  mounted/unmounted.
+
 ## v3.13.2
 
 ### @liveblocks/node


packages/liveblocks-react-ui/src/primitives/AiComposer/index.tsx
```diff
@@ -143,14 +144,30 @@ export const AiComposerForm = forwardRef<HTMLFormElement, AiComposerFormProps>(
     const focus = useCallback(
       (resetSelection = true) => {
         try {
+          // Slate's `ReactEditor.focus` method can use `setTimeout` internally
+          // which prevents us from catching errors, so this is a reimplementation.
+          // https://github.com/ianstormtaylor/slate/blob/main/packages/slate-dom/src/plugin/dom-editor.ts
```
nimeshnayaju, line 149: 😩

PR #2978, opened 2026-01-20
Remove legacy token (SECRET_LEGACY) support

nvie:
Legacy tokens are no longer supported by the backend in production. This removes all client-side support including:

- LegacySecretToken type and TokenKind.SECRET_LEGACY enum value
- Legacy token handling in auth-manager, client, room, and ai modules
- Test utilities and behaviors for legacy tokens
- E2E test pages for legacy token authentication

### DIFF
--- a/packages/liveblocks-react-ui/src/__tests__/index.test.tsx
+++ b/packages/liveblocks-react-ui/src/__tests__/index.test.tsx
@@ -117,12 +117,13 @@ class MockWebSocket {
 
 window.WebSocket = MockWebSocket as any;
 
+// Access token with perms: { "*": ["room:write"] }
+const accessToken =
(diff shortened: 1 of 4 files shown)


packages/liveblocks-core/src/__tests__/auth-manager.test.ts
```diff
@@ -118,115 +104,38 @@ describe("auth-manager - secret auth", () => {
...
     const results = await Promise.all([
       authManager.getAuthValue({
         requestedScope: "room:read",
-        roomId: "room1",
+        roomId: "org1.room1",
       }),
       authManager.getAuthValue({
         requestedScope: "room:read",
-        roomId: "room2",
+        roomId: "org1.room1",
```
SugarDarius, line 132: Shouldn't be `org1.room2` instead?

nvie: Sharply spotted! 🦅👀

However, in this case it's a correction of a test that was previously passing using the legacy tokens, but was testing the wrong thing (two different rooms instead of the intended same room twice), so this correction is intentional. The purpose of the test is to use the same room twice. In full context, the test looks like:

Screen Shot 2026-01-21 at 09 26 35@2x

 Thanks for making me double-check!

PR #2991, opened 2026-01-23
Optimize position distribution in LiveList initialization

nvie:
This PR adds `makeNPositions()`, a new helper that generates n positions with even distribution across the position space, and uses it when initializing a LiveList with items.

## Problem

Previously, LiveList initialization called `makePosition()` sequentially for each item, which produces positions like `!`, `!!`, `!!!`, etc. This causes position strings to grow linearly with the number of items. For large LiveLists, like `new LiveList([/* many items */])`, this could produce large parent key strings, like `"~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~#"` (harmless, but unnecessarily large).

## Solution

The new `makeNPositions()` generator distributes items evenly across the 94-character alphabet using a recursive bucketing approach. For n positions, each will have at most `ceil(log₉₄(n))` characters. This means 100,000 items can be represented with positions of just 3 (!) characters each, instead of positions that could grow much longer with sequential generation.

The algorithm:
- Divides items into 94 buckets (one per alphabet digit)
- Each bucket gets ~n/94 items
- Recursively subdivides buckets that contain more than 1 item

LiveList's constructor now uses `makeNPositions(n)` instead of calling `makePosition()` in a loop.

### DIFF
--- a/packages/liveblocks-core/src/crdts/LiveList.ts
+++ b/packages/liveblocks-core/src/crdts/LiveList.ts
@@ -1,7 +1,7 @@
 import { nn } from "../lib/assert";
 import { nanoid } from "../lib/nanoid";
 import type { Pos } from "../lib/position";
-import { asPos, makePosition } from "../lib/position";
+import { asPos, makeNPositions, makePosition } from "../lib/position";
 import { SortedList } from "../lib/SortedList";
 import type { ClientWireOp, CreateListOp, CreateOp, Op } from "../protocol/Op";
 import { OpCode } from "../protocol/Op";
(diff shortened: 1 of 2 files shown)


nvie:
Turns out with the improved algorithm from #2992, this isn't even needed as a separate optimization. Closing this one in favor of #2992 alone.

With this as a separate optimization, these two would be different:

```ts
new LiveList([1,2,3])                   // ["!", '"', "#"]
new LiveList().push(1).push(2).push(3)  // ["!", "!!", '!"']
```

Now, by just closing this one but keeping #2992, we'll have:

```ts
new LiveList([1,2,3])                   // ["!", "!!", '!"']
new LiveList().push(1).push(2).push(3)  // ["!", "!!", '!"']
```

Simpler.

PR #2993, opened 2026-01-24
Improve docs around using Liveblocks in shared package setups

nvie:


docs/pages/errors/dupes.mdx
```diff
@@ -53,3 +58,62 @@ supposed to happen. Please let us know about this by
...
+        SL1 -->|"3.2.0"| LB1
+
+        A2[App 2] --> SL1b[Shared Lib]
+        A2 -->|"3.13.3"| LB2["❌ @liveblocks<br/><small><i>conflicting versions</i></small>"]
+        SL1b -->|"3.2.0"| LB2
+    end
+
+    style LB1 fill:#6f6,stroke:#090
+    style LB2 fill:#fcc,stroke:#b33
+```
```
nvie: @CTNicholas @pierrelevaillant Actually not sure if we can use mermaid in docs this way? Is this supported?

marcbouchenoire: No we don't do anything with them so it would show up as a code block, we can make a static image version of it instead

docs/pages/errors/dupes.mdx
```diff
@@ -53,3 +58,62 @@ supposed to happen. Please let us know about this by
...
+
+The recommended approach is to declare Liveblocks as a `"peerDependency"` in
```
marcbouchenoire: I wonder if we should say "dependency"/"peer dependency" since the fields `"dependency"`/`"peerDependency"` (singular) don't exist?

nvie: Fixed in d4cb39a5c6fef1bfbca21e03fe4d83464f5460be.

PR #3000, opened 2026-01-27
Make engine flag public, officially supported, and documented

nvie:
### **User description**
Officially adds support for letting users opt-in to `engine: 2` rooms on a per-room basis.

When using:
- `.enterRoom("my-room", { /* not specified */ })` — you get whatever engine is the default for your project (currently still 1).
- `.enterRoom("my-room", { engine: 1 })` — explictly stay on the old engine. (Eventually™ we may want to stop supporting this, but definitely not very soon.)
- `.enterRoom("my-room", { engine: 2 })` — opt-in to the new and improved storage engine.

This PR also deprecates:
- `unstable_streamData` option. Projects that relied on this flag are better off switching to the new engine type for new rooms. The flag is deprecated but kept for right now, to still support using this flag for existing type-1 rooms that have no alternative.

### DIFF
--- a/docs/references/v2.openapi.json
+++ b/docs/references/v2.openapi.json
@@ -4969,7 +4969,7 @@
           "engine": {
             "type": "integer",
             "enum": [1, 2],
-            "description": "Preferred storage engine version to use when creating new rooms. Engine 2 supports larger documents, is more performant, has native streaming support, and will become the default in the future."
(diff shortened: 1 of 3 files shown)


nvie:
> I think I'd prefer "engine v1 rooms" over "engine 1 rooms" but I'm not entirely sure how strongly I think that so still approving ✅

Are you suggesting we'd make this flag `engine: "v2"` instead of `engine: 2`?

marcbouchenoire:
> Are you suggesting we'd make this flag engine: "v2" instead of engine: 2?

No sorry only in docs/prose!

PR #3011, opened 2026-01-29
Respect `portalContainer` option in text editor packages

marcbouchenoire:
This PR makes the portaled elements in the text editor packages respect [the `portalContainer` option from `@liveblocks/react-ui`’s `LiveblocksUiConfig`](https://liveblocks.io/docs/api-reference/liveblocks-react-ui#:~:text=portalContainer,the%20portal%20into.) instead of always using `document.body`.

Fixes LB-3217.


marcbouchenoire:
@nimeshnayaju I tested both Lexical and Tiptap and didn't see any issue 👍

PR #3331, opened 2026-04-09
Replace tsd by Vitest

marcbouchenoire:
This PR moves our type tests from `tsd` to [Vitest's built-in types testing](https://vitest.dev/guide/testing-types). Doing this allows us to:
- remove a dependency
- unify our test tooling around Vitest (same APIs and possibly later same configs/runs)
- simplify isolation testing (with `tsd` we had to run isolated tests with `ls test-d/* | xargs -n1 tsd --files`, now we're just using Vitest's [projects](https://vitest.dev/guide/projects))
- improve tests' speed (**~7.5×**, see comparisons table below)

Fixes LB-3702.

### Assertions

`tsd` | Vitest
-- | --
`expectType<T>(val)` | `expectTypeOf(val).toEqualTypeOf<T>()`
`expectAssignable<T>(val)` | `expectTypeOf(val).toExtend<T>()`
`expectNotAssignable<T>(val)` | `expectTypeOf(val).not.toExtend<T>()`
`expectError(expr)` | `// @ts-expect-error on the offending line`

### `@liveblocks/vitest-config`

Vitest' types testing requires both a Vitest config and a `tsconfig.json`, to avoid polluting our packages' root I added a new type-specific shared Vitest config to the existing `@liveblocks/vitest-config` which generates hidden (inside `node_modules/.cache`) `tsconfig.json` files.

```tsx
import { makeTypecheckTestConfig } from "@liveblocks/vitest-config/typecheck";
import { defineConfig } from "vitest/config";

// For single configs:
export default defineConfig({
  test: makeTypecheckTestConfig(import.meta, ["test-d/**/*.test-d.ts"]),
});


// And projects (multiple configs) ones:
export default defineConfig({
  test: {
    projects: [
      {
        test: makeTypecheckTestConfig(
          import.meta,
          ["test-d/augmentation.test-d.ts"],
          "augmentation"
        ),
      },
      {
        test: makeTypecheckTestConfig(
          import.meta,
          ["test-d/no-augmentation.test-d.ts"],
          "no-augmentation"
        ),
      },
    ],
  },
});
```

### Speed comparisons

Measured locally with `hyperfine --warmup 1 --runs 3 `.

Package | `tsd` | Vitest | × faster
-- | -- | -- | --
`@liveblocks/core` | 25.835s | 1.388s | **18.61×**
`@liveblocks/client` | 17.918s | 1.684s | **10.64×**
`@liveblocks/node` | 7.116s | 1.783s | **3.99×**
`@liveblocks/zustand` | 7.413s | 1.044s | **7.10×**
`@liveblocks/react-flow` | 7.590s | 1.140s | **6.66×**
`@liveblocks/react-ui` | 7.727s | 3.568s | **2.17×**
`@liveblocks/react` | 11.278s | 3.305s | **3.41×**


packages/liveblocks-core/test-d/CommentBody.test-d.ts
```diff
@@ -1,296 +1,313 @@
...
-          },
-        ],
-      },
-    ],
+
+  test("should accept valid comment bodies", () => {
+    expectTypeOf({
+      version: 1,
+      content: [],
+    } satisfies CommentBody).toExtend<CommentBody>();
```
marcbouchenoire, line 148: @nvie I found that `.toExtend<T>()` works with `Relax` unions if `satisfies` is added 🤷

packages/liveblocks-client/test-d/client.augmentation.test-d.ts
```diff
@@ -25,39 +25,51 @@ declare global {
...
+    });
+
+    test("should return typed presence and storage values", async () => {
+      const { room } = client.enterRoom("room", {
+        initialPresence: { cursor: { x: 1, y: 2 } },
+        initialStorage: { animals: new LiveList(["🦊"]) },
+      });
+
+      expectTypeOf(room.getPresence()?.cursor.x).toBeNumber();
+      expectTypeOf(room.getPresence()?.cursor.y).toBeNumber();
```
nvie: I assume this is a shorthand for `.toEqualTypeOf<number>()`, right? I think would have a small preference for sticking to that format, because... why would we make a special API for numbers? I don't really get the benefit, other than maybe trying to make this "look more like natural English"? Any other benefits to this shorthand?

packages/liveblocks-core/test-d/ToolResultResponse.test-d.ts
```diff
@@ -1,68 +1,94 @@
...
+    const _3: ToolResultResponse = { yo: [1, 2, 3] };
 
-  expectNotAssignable<ToolResultResponse>({ error: 123 });
-  expectNotAssignable<ToolResultResponse>({
-    error: { code: 403, message: "Not authorized" },
-  });
+    // @ts-expect-error - `data` must be an object as well
+    const _4: ToolResultResponse = { data: "hi" };
+    // @ts-expect-error - `data` must be an object as well
+    const _5: ToolResultResponse = { data: 123 };
```
nvie: I also ran into this limitation of not being able to use the IMHO nicer `expectNotAssignable` helper anymore. Claude did suggest to me to indeed also use these `const _5` fake assignments.

What would you think of expressing this as a nameless (no let or const required) expression with satisfies, like this?

```ts
    // @ts-expect-error - `data` must be an object as well
    ({ data: 123 }) satisfies ToolResultResponse;
```

This would perform exactly the same check without having to do the fake assignment I think.

nvie (approved):
OMG, I love it! 😍

packages/liveblocks-core/test-d/CommentBody.test-d.ts
nvie: Oh wow — super nice!

packages/liveblocks-client/test-d/client.augmentation.test-d.ts
marcbouchenoire: No benefits apart from being more specific and sounding like English, I changed my mind and reverted to only using `.toEqualTypeOf<T>()` in https://github.com/liveblocks/liveblocks/pull/3331/commits/1133f140360b1848d21ee537cd515a7564ff87bc.

packages/liveblocks-core/test-d/ToolResultResponse.test-d.ts
marcbouchenoire: I prefer it 👍 Changed in https://github.com/liveblocks/liveblocks/pull/3331/commits/e6146281a5789c188d8cb118f539d68ab00328e5.

PR #3335, opened 2026-04-10
Improve mentions in composers

marcbouchenoire:
This PR improves how mentions suggestions appear after typing `@`. We were really strict about it to support email addresses but this PR relaxes the rules a bit while keeping this behavior.


SugarDarius (approved):
LGTM ✅

nvie (approved):

PR #3353, opened 2026-04-14
Promote `room.history.disable()` to an experimental API

nvie:
This implementation already existed as a fully hidden/private API used internally, but it may actually be useful for generated content (e.g. AI generated updates) that should not be undo-able.

I've kept the API marked experimental, because I'm not 100% convinced yet this is the stable/final form factor.

```ts
// Using the room client directly
room.history.disable(() => {
  room.batch(() => {
    root.set("summary", generatedSummary);
    root.set("status", "done");
  });
});

// Or via React hook
const { disable } = useHistory();

disable(() => {
  // These mutations won't appear on the undo/redo stack
  root.set("summary", generatedSummary);
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 ## vNEXT (not yet released)
 
+## v3.18.2
+
+### `@liveblocks/client`
+
+- New experimental `room.history.disable(fn)` API that allows running storage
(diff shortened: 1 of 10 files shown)


CTNicholas (approved):
Love it!

marcbouchenoire (approved):
🥳

PR #3354, opened 2026-04-15
Fix `history.disable()` edge cases by defining weirdness out of existence

nvie:
This fixes edge cases in `history.disable()` by replacing the length-based save/restore approach with physically swapping the undo/redo stack arrays for fresh empty ones during the callback. This defines some of the discovered weirdnesses out of existence: all history operations naturally work on the temporary stacks, so edge cases like `clear()` inside `disable()`, cap eviction at 50 items, and `undo()`/`redo()` within the block—while rare!—still just work correctly without any special guards.

This PR:
- Swaps undo/redo stacks with fresh arrays during the `disable()` callback, restores originals in `finally`
- Suppresses `room.events.history` notifications during `disable()` since changes are discarded anyway
- Renames internal `withoutHistory` to `disableHistory`

### DIFF
--- a/packages/liveblocks-core/src/__tests__/room.devserver.test.ts
+++ b/packages/liveblocks-core/src/__tests__/room.devserver.test.ts
@@ -5,7 +5,7 @@
  * For connection state machine, auth, reconnection, and wire protocol tests,
  * see room.mockserver.test.ts.
  */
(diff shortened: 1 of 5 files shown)


marcbouchenoire (approved):
🙌

PR #3362, opened 2026-04-15
Prep for PNPM

nvie:
This PR performs part one (of probably quite a few PRs like this), in preparation of switching the repo to PNPM.

- [x] Pin GitHub Actions to SHAs, use `node-version-file: ".nvmrc"`
- [x] Add missing `y-prosemirror` dependency to `liveblocks-emails`
- [x] Allow running all devserver tests in parallel locally
- [x] Upgrade Node to 24.x everywhere (.nvmrc, engines, CI)
- [x] Switch from `jsdom` to `happy-dom` (to avoid Node 24 conflicts)
- [x] Upgrade various dev dependencies


shared/vitest-config/package.json
```diff
@@ -8,9 +8,9 @@
     "./typecheck": "./typecheck.js"
   },
   "dependencies": {
-    "fast-check": "^4.3.0",
-    "jsdom": "^25.0.1",
-    "vite-tsconfig-paths": "^5.1.4",
+    "fast-check": "^4.6.0",
+    "happy-dom": "^20.9.0",
```
marcbouchenoire, line 12: It's already a dependency in the root `package.json`, which made me notice things like `vitest` and `@vitest/coverage-istanbul` in the root too maybe we should centralize them all here?

marcbouchenoire (approved):

shared/vitest-config/package.json
nvie: I'll do this in the PNPM pass — there will be a lot more to duplicate and DRY up soon.

PR #3367, opened 2026-04-16
Prepare dependency declarations for pnpm migration

nvie:
This PR performs changes that would otherwise be hidden in the noise of the PNPM transition. I'm making these upgrades in NPM now first.

- Adds eslint, vitest, and rollup as direct `devDependencies` to all packages that consume them (previously resolved only via hoisting)
- Upgrades radix-ui to latest patches and align devtools transitive deps in the way `pnpm import` would otherwise bump them
- Fixes a small build issue in liveblocks-devtools


packages/liveblocks-server/package.json
```diff
@@ -52,7 +52,7 @@
     "@types/node": "^22.10.3",
     "@typescript-eslint/eslint-plugin": "^8.57.1",
     "@typescript-eslint/parser": "^8.57.1",
-    "eslint": "^8.56.0",
+    "eslint": "^9.39.4",
```
marcbouchenoire: Is that intentional? Since it's a mirror

marcbouchenoire (approved):

packages/liveblocks-server/package.json
nvie: Correct, I also just noticed this — will revert it!

nvie: (Even if we kept it, it would get "fixed" automatically soon.)

PR #3372, opened 2026-04-17
Re-export nanoid from client

nvie:
...so examples relying on this don't have to rely on core (or a separate nanoid dependency).

### DIFF
--- a/packages/liveblocks-client/src/index.ts
+++ b/packages/liveblocks-client/src/index.ts
@@ -88,6 +88,7 @@ export {
   LiveList,
   LiveMap,
   LiveObject,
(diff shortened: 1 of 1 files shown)


packages/liveblocks-client/src/index.ts
```diff
@@ -88,6 +88,7 @@ export {
   LiveList,
   LiveMap,
   LiveObject,
+  nanoid,
```
nvie, line 91: @CTNicholas After I publish this version, I'll rewrite the examples to use clean client imports here again.

CTNicholas (approved):

PR #3373, opened 2026-04-17
Post-pnpm cleanups: dependency hygiene fixes

nvie:


tools/liveblocks-codemod/package.json
```diff
@@ -20,15 +20,15 @@
...
+    "@types/jscodeshift": "^17.3.0",
     "typescript": "^5.9.3"
   },
   "dependencies": {
     "execa": "4.0.3",
     "globby": "11.0.1",
     "inquirer": "7.3.3",
     "is-git-clean": "1.1.0",
-    "jscodeshift": "0.13.1",
+    "jscodeshift": "^17.3.0",
```
nvie, line 31: @marcbouchenoire After many struggles, I decided to upgrade this forward, and just eat the consequences (see the code changes required). If you have any time anywhere soon, I'd love to sit down with you for another day and see if we can unpin all hard-pinned package (like all the others in this package) and move them to a version range again. Not only here, but also in other projects.

CTNicholas (approved):

PR #3377, opened 2026-04-20
Make yjs and y-prosemirror required peer deps of @liveblocks/emails

nvie:
These were marked as optional before, but in practice consumers could not actually use the package without these peers installed.

Bundlers typically statically resolve all imports at build time, which then fails with "Module not found" for any consumer missing them, regardless of whether that consumer's code path ever uses the yjs or y-prosemirror functionality.


SugarDarius (approved):
✅

PR #3397, opened 2026-05-04
Add `markdownToCommentBody` helper to `@liveblocks/node`

marcbouchenoire:
This PR adds a new `markdownToCommentBody` helper to `@liveblocks/node` which converts a Markdown string into a comment body to help when programmatically creating comments (e.g. from a webhook, AI agent, etc).

- `CommentBody` only supports paragraphs, inline text formatting (bold, italic, strikethrough, inline code), links, and `@mention` syntax, other unsupported Markdown features are kept as plain text.
- Mentions are parsed as `@username` and emitted as `{ type: "mention", kind: "user", id: "username" }` nodes.
- Parsing is done with [marked](https://marked.js.org/)'s `Lexer` like the `Markdown` primitive in `@liveblocks/react-ui`. I initially wanted this helper to live in `@liveblocks/core` and be re-exported from both `@liveblocks/node` and `@liveblocks/client` (like `stringifyCommentBody`) but `@liveblocks/core` has currently 0 dependencies and we want to keep it that way, and the main use-case we're targeting is server-centric too.

Fixes LB-2777 and LB-3678.

## To do

- [ ] Use it in https://github.com/liveblocks/liveblocks/tree/main/examples/nextjs-comments-ai
- [ ] Notify customers who requested this


nvie (approved):
Looks like one or two tests aren't passing, but otherwise LGTM — this API is super convenient! 👌

PR #3465, opened 2026-05-14
Add explicit error handling for network errors in `streamStorage`

jamesbvaughan:
We've observed the following error after a client receives a 500 response from `/storage`:
```
TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))
    at new Map (<anonymous>)
    at Ee (../../../../node_modules/@liveblocks/core/dist/index.js:9902:27)
```
That `new Map` is this one: https://github.com/liveblocks/liveblocks/blob/446f419534d8437ff881cdb8fbe4a96e795219e7/packages/liveblocks-core/src/room.ts#L2956-L2958

This PR doesn't attempt to get to the root cause of that 500. It's just meant to help the client handle the error a bit more gracefully.

### DIFF
--- a/packages/liveblocks-core/src/api-client.ts
+++ b/packages/liveblocks-core/src/api-client.ts
@@ -1605,17 +1605,14 @@ export function createApiClient<
   }
 
   async function streamStorage(options: { roomId: string }) {
-    const result = await httpClient.rawGet(
+    const result = await httpClient.get(
       url`/v2/c/rooms/${options.roomId}/storage`,
       await authManager.getAuthValue({
         requestedScope: "room:read",
         roomId: options.roomId,
       })
     );
-    if (!result.ok) {
-      throw await HttpError.fromResponse(result);
-    }
-    return (await result.json()) as StorageNode[];
+    return result as unknown as StorageNode[];
   }
 
   /* -------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


packages/liveblocks-core/src/api-client.ts
```diff
@@ -1605,14 +1605,14 @@ export function createApiClient<
   }
 
   async function streamStorage(options: { roomId: string }) {
-    const result = await httpClient.rawGet(
+    const result = await httpClient.get(
```
jamesbvaughan, line 1608: Looking at the `rawGet` and `get` implementations, the key difference seems to be the handling of `!response.ok`, which I believe to be relevant here.

packages/liveblocks-core/src/api-client.ts
```diff
@@ -1605,14 +1605,14 @@ export function createApiClient<
...
       url`/v2/c/rooms/${options.roomId}/storage`,
       await authManager.getAuthValue({
         requestedScope: "room:read",
         roomId: options.roomId,
       })
     );
-    return (await result.json()) as StorageNode[];
+    return result as unknown as StorageNode[];
```
jamesbvaughan, line 1615: I initially tried to use the `get` method's type parameter rather than `as` here, but the incompatibility between `StorageNode[]` and `JsonObject` made that difficult. Let me know if you have a different preferred way to handle this.

PR #3481, opened 2026-05-20
Treat any inbound message as heartbeat liveness signal

nvie:
Updates the state machine so that not only PONG responses from the server are counted as still-alive, but basically any incoming message (PONGs included) is enough proof the connection is still alive.

### DIFF
--- a/packages/liveblocks-core/src/connection.ts
+++ b/packages/liveblocks-core/src/connection.ts
@@ -4,7 +4,7 @@ import type { Observable } from "./lib/EventSource";
 import { makeBufferableEventSource, makeEventSource } from "./lib/EventSource";
 import * as console from "./lib/fancy-console";
 import type { BuiltinEvent, Patchable, Target } from "./lib/fsm";
-import { FSM } from "./lib/fsm";
+import { FSM, IGNORE } from "./lib/fsm";
 import type { Json } from "./lib/Json";
 import { tryParseJson, withTimeout } from "./lib/utils";
 import { ServerMsgCode } from "./protocol/ServerMsg";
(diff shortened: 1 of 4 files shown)


marcbouchenoire (approved):

jrowny (approved):

jrowny (approved):

packages/liveblocks-core/src/connection.ts
```diff
@@ -503,10 +503,23 @@ function createConnectionStateMachine<T extends BaseAuthResult>(
...
+    //
+    // Only emit PONG when the machine is actually waiting for one, to
+    // avoid spamming didIgnoreEvent in the @ok.connected state (which has
+    // no handler for PONG).
+    if (machine.currentStateOrNull === "@ok.awaiting-pong") {
+      machine.send({ type: "PONG" });
+    }
+    if (event.data !== "pong") {
+      onMessage.notify(event);
+    }
```
nimeshnayaju: Makes sense! This means, that we consider 'pong' not received if no message has been received for the specified amount of time?

nimeshnayaju (approved):

packages/liveblocks-core/src/connection.ts
nvie: No, previously there was a ternary: _either_ a PONG event would be sent to the FSM, or we'd forward to the message to the Liveblocks client (the FSM doesn't know about those).

Now, we still forward the message to the Liveblocks client, but in addition, we _also_ send a PONG event to the FSM. But obviously a fake one.

Actually saying this out loud now makes it obvious that this is too much of a hack, and I'll do another pass here to solve this more Properly™ before merging.

nvie: @nimeshnayaju This is now more strategically fixed in 87f5eb0503cfb0d24fcd88710333593fc9f79186 rather than taking this direct tactical approach.

PR #3490, opened 2026-05-27
Fix `LiveList.push()` for concurrent clients

nvie:
This PR is the client-side half of a two-repo fix for LB-3483 (reported by Subframe).

> [!IMPORTANT]
> The [matching server-side PR](https://github.com/liveblocks/liveblocks-backend/pull/1768) (append logic, decoder, storage driver changes) in `liveblocks-backend` _must be deployed before this ships_. Old servers will reject an unknown `intent` value and the push will fail.
> **Update:** Done!

## Why

`LiveList.push()` was implemented as `insert(element, this.length)`, which bakes a client-computed `parentKey` into the CREATE op. When two clients push near-simultaneously, both guess the same end-of-list position. The server resolves the collision by slotting the second op *between* existing items rather than at the true end, so the list settles in the wrong order. Worse, observers see a rendered item get shoved to a different index mid-render.

There's a second, subtler version of the same problem on the *pushing* client itself: even once the server places everything correctly, a client that just pushed an item could watch it briefly jump out of place while a concurrent push from another client was being reconciled — the incoming remote op was interleaved purely by its raw position key.

## What changed

**Server-authoritative append.** `.push()` now marks the create Op with a new `intent: "push"`, which tags the root CREATE op so the server appends it to the authoritative end of the list instead of resolving the optimistically guessed position. Because the room processes ops serially, concurrent pushes get monotonically increasing keys — no collision, no wrong order.

**Optimistic no-flip on the originator.** When a remote op arrives that would land at or before this client's still-unacknowledged pushed items, those items are now kept together *after* it locally, repositions in push order to the positions the server is about to assign them. The incoming item never wedges between them, and the acks land as no-ops — nothing visibly moves on the way to the converged order.

`.insert()` is unchanged in behavior.

## Reproduction

Two new e2e tests against the dev server add the reproduction and only pass now that the fix it in place.

- In `list-push.test.ts`, reproduces the exact Subframe scenario: client A pushes `a1` then `a2`; client B, unaware of A's ops, pushes `b1` to what it thinks is position 0. Without the fix the list settles as `[a1, b1, a2]`; with it, `[a1, a2, b1]`.
- In `list-push-optimistic.test.ts`, guards the no-flip: while a concurrent remote push reconciles, an item this client already rendered never moves backward.

**This is a genuine behavioral fix, not a pure refactoring.**

Best reviewed commit-by-commit.


jrowny (approved):

packages/liveblocks-core/package.json
```diff
@@ -37,11 +37,11 @@
...
     "lint": "eslint src/",
     "lint:package": "publint --strict && attw --pack",
-    "test": "pnpm dlx liveblocks dev -p 1160 -c 'vitest run --coverage'",
-    "test:ci": "vitest run",
+    "test": "pnpm dlx liveblocks@rc dev -P -c 'vitest run --coverage'",
+    "test:ci": "pnpm dlx liveblocks@rc dev -P -c 'vitest run' --",
     "test:types": "vitest run --config ./vitest.config.typecheck.ts",
     "test:watch": "vitest",
-    "test:e2e": "pnpm dlx liveblocks dev -p 1160 -c 'vitest run --config=./vitest.config.e2e.ts'",
+    "test:e2e": "pnpm dlx liveblocks@rc dev -P -c 'vitest run --config=./vitest.config.e2e.ts'",
```
marcbouchenoire: A reminder to change these back once 1.5.0 is out.

marcbouchenoire (approved):

packages/liveblocks-core/package.json
nvie: Thanks, just did in 46a5bdfbd9a63d5f12fe09671520dce66bd73a3c! 🙏

ofoucherot (approved):

packages/liveblocks-core/src/room.ts
```diff
@@ -1875,10 +1891,20 @@ export function createRoom<
...
+      // when the stream we are applying already IS the fix. The
+      // `fromSnapshot` flag below patches only the one leak that bit us (the
+      // bump); it does not address the others. The proper fix is
+      // a node-stream reconcile that updates the tree in place, unified with
+      // the `_fromItems` path used on initial load, so a node stream never
+      // enters the op path at all. Until then `fromSnapshot` is a stopgap.
       const ops = getTreesDiffOperations(currentItems, nodes);
-
-      const result = applyRemoteOps(ops);
+      const result = applyRemoteOps(ops, /* fromSnapshot */ true);
```
nvie, line 1907: After fixing the "Optimistic no-flip on the originator" part of this fix (see PR description for explanation), I realized this had introduced a regression elsewhere: when a client reconnects before its pushes were acked by the server. (Whack-a-mole.)

I was able to replicate this in a test, and fixed it here tactically via this new `fromSnapshot` boolean, which basically skips running the optimistic repositioning "bump" after a reconnect. But this has made it apparent that this will need to be properly addressed by a more strategic solution, because adding Yet Another Boolean™ here is definitely a smell.

PR #3493, opened 2026-05-29
New permissions

marcbouchenoire:
This PR reworks the permission system across the client-side packages and the SDKs.

**Backend PR:** https://github.com/liveblocks/liveblocks-backend/pull/1769

### New permissions

Add new scopes and a way to opt-out to allow more granularity. Start with a base of `*:read` or `*:write` (with legacy aliases `room:read` and `room:write` still supported), and granularly change features with e.g. `storage:write` or `comments:none`.

- `*:read` (alias `room:read`)
- `*:write` (alias `room:write`)
- 🆕 `storage:read`
- 🆕 `storage:write`
- 🆕 `storage:none` (opt-out)
- `comments:read`
- `comments:write`
- 🆕 `comments:none` (opt-out)
- 🆕 `feeds:read`
- `feeds:write`
- 🆕 `feeds:none` (opt-out)


packages/liveblocks-core/src/protocol/Permissions.ts
```diff
@@ -0,0 +1,229 @@
+export const Permission = {
```
marcbouchenoire: The source of truth for all packages in the monorepo and for the backend as well.

packages/liveblocks-python/liveblocks/models/roll_project_secret_api_key_request_body_new_key_expiration_in.py
```diff
@@ -0,0 +1,24 @@
+from enum import StrEnum
+
+
+class RollProjectSecretApiKeyRequestBodyNewKeyExpirationIn(StrEnum):
```
marcbouchenoire, line 4: Unrelated, missing after https://github.com/liveblocks/liveblocks/pull/3464.

packages/liveblocks-core/src/api-client.ts
```diff
@@ -325,21 +325,6 @@ export interface RoomHttpApi<TM extends BaseMetadata, CM extends BaseMetadata> {
 
   getOrCreateAttachmentUrlsStore(roomId: string): BatchStore<string, string>;
 
-  uploadChatAttachment({
```
marcbouchenoire, line 328: AI Copilots' chat attachment endpoints were never used, removing them until we need them.

packages/liveblocks-core/src/protocol/Permissions.ts
```diff
@@ -0,0 +1,229 @@
...
+  LegacyFeedsWrite: "feeds:write",
+} as const;
+
+export type Permission = (typeof Permission)[keyof typeof Permission];
+
+export type AccessLevel = "write" | "read" | "none";
+
+export type RequiredAccessLevel = "read" | "write";
+
+export type PermissionCapabilities = {
```
ofoucherot: I'm not 100% convinced with this name `PermissionCapabilities`.
Maybe @nvie has some ideas?

packages/liveblocks-core/src/client.ts
```diff
@@ -699,7 +699,9 @@ export function createClient<U extends BaseUserMeta = DU>(
       ),
       authenticate: async () => {
         const resp = await authManager.getAuthValue({
-          requestedScope: "room:read",
+          // TODO: Should we have permissions for AI Copilots?
```
ofoucherot: Yes but we should add it later.

packages/liveblocks-core/src/protocol/Permissions.ts
nvie: `PermissionMatrix`?

packages/liveblocks-core/src/protocol/Permissions.ts
```diff
@@ -0,0 +1,229 @@
...
+  creation: AccessLevel;
```
nvie: Without having read _anything else_ about this PR (so please 🧂), it's not immediately clear to me what `creation` means here. Creation of… what? I'd say creation of comments or feeds would be determined by `comments: "write"` or `feeds: "write"` respectively already. So, what does `creation` mean?

packages/liveblocks-core/src/protocol/Permissions.ts
```diff
@@ -0,0 +1,229 @@
...
+  presence: AccessLevel;
+  storage: AccessLevel;
+  comments: AccessLevel;
+  feeds: AccessLevel;
+  personal: "write";
```
nvie: Also... I assume `personal: "write"` is deliberately hardcoded? If it's always hardcoded and mandatory, does it still add anything?

packages/liveblocks-core/src/__tests__/permissions.test.ts
nvie, line 1: Worth some `fast-check` property tests here? Interesting properties that I can think of would be:

1. Resolving capabilities is idempotent. Ie if you resolve _any_ input, say `resolve(input) === resolve(resolve(input))`.
2. Personal is always write? ie `resolve(input).personal === "write"`, independent of input?
3. Adding a `*:none` scope to any input never produces a resolved capability that has more access than without it?

packages/liveblocks-core/src/__tests__/permissions.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    ).toBeUndefined();
+  });
+});
+
+describe("permission matrix helpers", () => {
+  test("merges permission matrices by taking the strongest access per resource", () => {
+    const matrix = mergePermissionMatrices([
+      permissionMatrixFromScopes([Permission.RoomWrite]),
+      permissionMatrixFromScopes([Permission.RoomStorageNone]),
+    ]);
```
nvie: Conceptually, rather than doing `merge(matrixFrom(scopes1), matrixFrom(scopes2))`, should we not instead do `matrixFromScopes(defaultScopes, userScopes, groupScopes)`? That seems safer to me, and less ambiguous, because it will prevent the issue I raised in the comment above about taking away permissions by letting a user scope overwrite a default scope?

packages/liveblocks-core/src/protocol/Permissions.ts
```diff
@@ -0,0 +1,232 @@
...
+
+export type RequiredAccessLevel = "read" | "write";
+
+export type PermissionMatrix = {
+  room: AccessLevel;
+  presence: AccessLevel;
+  storage: AccessLevel;
+  comments: AccessLevel;
+  feeds: AccessLevel;
+  personal: "write";
```
nvie: In a nutshell, how would you explain to an end-user what `room` (r/w/-), `presence` (r/w/-) and `personal` (w) are exactly? I think it matters what meaning we assign to these combinations precisely? In total there are 9 possible configurations for these. How does a user pick which they need?

packages/liveblocks-core/src/permissions.ts
```diff
@@ -0,0 +1,261 @@
...
+  Permission,
+  permissionMatrixFromScopes,
+  resolveFullPermissionMatrix,
+  resolvePermissionMatrix,
+} from "./protocol/Permissions";
+
+export type RoomPermissionGrant = {
+  resource: string;
+  scopes: readonly string[];
+};
```
nvie: I found that "resource" is a bit over-used in this module. Here, it's a room ID / wildcard pattern, but everywhere else in the same file `resource: PermissionResources` means a *feature* (`storage`, `comments`, …). Should we rename both?

- `resource` → `pattern`
- `resource` → `feature` (or `topic` or something?)

packages/liveblocks-core/src/permissions.ts
```diff
@@ -0,0 +1,261 @@
...
+
+export type RoomPermission = Permission[];
+
+export type RoomPermissionObject = {
+  default?: "read" | "write";
+  presence?: "read" | "none";
+  storage?: "read" | "write" | "none";
+  comments?: "read" | "write" | "none";
+  feeds?: "read" | "write" | "none";
+};
```
nvie: This looks a lot like a "matrix". I worry a bit that "matrix", "resolved matrix", this "RoomPermissionObject" here, and also [this `Permission` type](https://github.com/liveblocks/liveblocks/pull/3493#discussion_r3388628857) all look very similar but carry lightly different meaning that isn't carried over through the type system, and understanding how data flows and gets changed throughout the various layers becomes opaque.

Also, `presence` here can't be set to `write` in object notation, but it looks like the array form (and the legacy `room:presence:write` scope) still resolve to `presence: write`? Or am I misunderstanding? So `{ default: "write" }` gives you presence write as a side effect, but there's no way to opt *into* it explicitly?

Honestly, I still keep getting confused about presence being part of our permission system, and I'd still lean toward defining presence away completely from the matrix entirely and simply state that users ALWAYS have presence write access. Maybe I don't get what no-access or read-only access to presence would even mean for a real application… 🤔

packages/liveblocks-core/src/protocol/Permissions.ts
ofoucherot: `personal` is not visible to the end-user, so we don't really have to explain it to them.
`personal` is just an explicit way to say that the token is valid and identifies the user; it grants access to resources that belong to the user (like user notification settings, user inbox notifications, etc.).

packages/liveblocks-core/src/protocol/Permissions.ts
marcbouchenoire: We discussed this already but for clarity, we're going with `PermissionMatrix`.

packages/liveblocks-core/src/permissions.ts
marcbouchenoire: We discussed this already but for clarity, we're removing the object notation for now and `presence` is no longer present (🥁) in the permission system.

packages/liveblocks-core/src/protocol/Permissions.ts
marcbouchenoire: We discussed this already but for clarity, `creation` is now `room`.

PR #3498, opened 2026-06-01
Fix Storage notification-equivalence between normal online activity vs reconnect

nvie:
This PR adds a notification-equivalence e2e suite for the storage reconnect path, and fixes three real `LiveObject` reconnect bugs it surfaced along the way when I added this coverage.

I added these tests as groundwork/preparation for a larger refactor I'm planning on. Before doing that, I wanted to nail down exactly what notifications subscribers *should* see, so the refactor has stable test coverage.

## The property

The new test helper connects two clients, and runs the same mutation twice:

1. **online** — Client A stays connected; B mutates; A receives live ops.
2. **reconnect** — Client A disconnects; B mutates; A reconnects and receives a catch-up.

It captures the `StorageUpdate` batches A gets in each, and asserts they're equivalent. Both phases start from identical state and apply identical mutations, so *any* difference in the notifications is a bug. (With the exception of changes that cancel each other out, like client B adding-then-deleting a key.)

### DIFF
--- a/packages/liveblocks-core/e2e/storage-notification-reconnect.test.ts
+++ b/packages/liveblocks-core/e2e/storage-notification-reconnect.test.ts
@@ -133,11 +133,14 @@ const sameJson = <S extends LsonObject>(a: Actor<S>, b: Actor<S>): boolean =>
 // connection/handshake noise leaks into the batches.
 async function bothPhases<S extends LsonObject>(
   initialStorageFn: () => S,
-  mutate: (root: LiveObject<S>) => void
+  mutate: (root: LiveObject<S>) => void,
+  opts: { convergeTimeoutMs?: number } = {}
 ): Promise<{
   online: { batches: StorageUpdate[][]; root: LiveObject<S> };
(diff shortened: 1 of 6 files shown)


packages/liveblocks-core/e2e/storage-notification-reconnect.test.ts
```diff
@@ -646,24 +680,27 @@ const listShapes = (
...
+    // is one of the divergences the reconcile refactor must eliminate.
+    const expected = [{ type: "set", index: 1, item: "B" }];
+    expect(collectListDeltas(online.batches, online.root.get("list"))).toEqual(
+      expected
+    );
+    expect(
+      collectListDeltas(reconnect.batches, reconnect.root.get("list"))
+    ).toEqual(expected);
+  }
+);
```
nvie, line 703: Marked as `.fails()`. Ideally this test _should_ pass, but it doesn't and fixing it would require a larger refactoring, which I consider out of scope for now.

PR #3499, opened 2026-06-02
Auto-insert missing CHANGELOG headings during release

nvie:
This PR makes `release.sh` auto-insert a `## vX.Y.Z` heading into `CHANGELOG.md` when one doesn't exist yet for the version being released, so I stop forgetting to do it by hand and triggering an extra PR roundtrip right before publishing. 😇

The heading is added directly below the permanent `## vNEXT` section. It piggy-backs on the existing `Bump to X.Y.Z` commit which already makes file changes during publishing.

It only fires for real releases that don't already have a heading:

| When asked to release...       | Behavior                                |
| --------------- | --------------------------------------- |
| `v3.19.4`        | ✅ inserts `## v3.19.4`                  |
| `v3.20.0`        | ✅ inserts `## v3.20.0`                  |
| `v3.19.4-rc2`    | ❌ skipped (pre-release)                 |
| `v3.18.0`        | ❌ skipped (heading already exists)      |

If the `vNEXT` section can't be found, it warns and leaves the file untouched rather than guessing.

### DIFF
--- a/.github/scripts/release.sh
+++ b/.github/scripts/release.sh
@@ -124,7 +124,7 @@ inject_changelog_heading () {
     echo "==> Adding CHANGELOG heading for v$VERSION"
     tmp="$(mktemp)"
     if awk -v ver="$VERSION" '
-        !done && /^## vNEXT \(not yet released\)$/ {
+        !done && /^## vNEXT( .*)?$/ {
             print
             print ""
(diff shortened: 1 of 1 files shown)


marcbouchenoire (approved):
Niiice 🙌

PR #3517, opened 2026-06-11
Clean up deprecated options

nvie:
This PR is a batch of harmless internal cleanups ahead of v3.20.

This PR:
- Removes the `engine` option from all APIs that still had it. It already was a no-op.
- Remove the `unstable_streamData` option. Also a no-op.
- Rename `room.getStorageSnapshot()` (deprecated) → `room.getStorageOrNull()`. Pure name cleanup.
- Internal rename: `getTreesDiffOperations` → `diffNodeMap`.
- Drop the legacy `OldConnectionStatus` devtools type (connection statuses from pre-1.1 clients).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## v3.20.0
 
+- Deprecate `room.getStorageSnapshot()` in favor of `room.getStorageOrNull()`.
+
 ## v3.19.5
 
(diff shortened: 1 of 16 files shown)


marcbouchenoire (approved):
🧹

nimeshnayaju (approved):

PR #3533, opened 2026-06-18
Public/private comments

marcbouchenoire:
This PR adds support public/private threads.

Threads now have a `visibility` property (`"public"` by default) and it can be set everywhere threads can be created. Threads can also be queried by `visibility` to filter a specific visibility.

New scoped `comments:public`/`comments:private` permissions are introduced to decide what threads users have access to.

```
comments
├── comments:public
└── comments:private
```

**Documentation PR:** https://github.com/liveblocks/liveblocks/pull/3543


ofoucherot (approved):
🚀

PR #3546, opened 2026-06-24
Adjust type to new shape returned by the backend for clients >= 3.21

nvie:
Clean up of `HistoryVersion` type, prior to implementing version history for Storage. Starting with >=3.21, the backend will no longer return these fields.

For older clients, those fields will keep being returned, so their types don't lie, but they will always be hardcoded as `type: "historyVersion", kind: "yjs"` offering no real value.

Also this deprecates the `useHistoryVersionData()` hook in favor of `useHistoryVersionYjsData()`. (Users should rename this hook after upgrading to 3.21.) This makes room for a storage version of the hook. The initial idea of returning both Yjs and Storage in a single hook was inefficient and unpractical to implement and use.


ofoucherot (approved):
LGTM!

PR #3556, opened 2026-06-29
Update example READMEs

marcbouchenoire:
While I was updating the new `nextjs-comments-private` README I noticed some of the gallery links were broken.


CTNicholas (approved):
Whoops! Thanks for fixing

PR #3594, opened 2026-07-21
Livetext

jrowny:
Adds support for LiveText
Adds `collaborationMode: "liveblocks"` to liveblocks-react-tiptap and liveblocks-react-blocknote

new packages:
liveblocks/prosemirror
liveblocks/codemirror
liveblocks/lexical


.github/workflows/publish-packages.yml
marcbouchenoire, line 1: Missing `liveblocks-codemirror` and `liveblocks-lexical`

packages/liveblocks-codemirror/README.md
```diff
@@ -0,0 +1,3 @@
+# `@liveblocks/codemirror`
+
+CodeMirror collaboration plugins backed by Liveblocks.
```
marcbouchenoire: Package READMEs should follow the existing template

packages/liveblocks-codemirror/rollup.config.js
marcbouchenoire, line 1: Is there a reason to use Rollup instead of `tsup`? We've only used Rollup when we needed to ship stylesheets but otherwise `tsup` is preferred

e2e/next-codemirror-liveblocks/package.json
marcbouchenoire, line 1: Are all these new e2e projects meant to stay or were they here mostly for development? Could public-facing examples and `next-sandbox` be enough?

e2e/next-lexical-liveblocks/vitest.config.ts
```diff
@@ -0,0 +1,36 @@
...
+// Force a single @liveblocks/core instance so kStorageUpdateSource (a Symbol)
+// matches between the mock-server test helpers (core/src) and app imports (@liveblocks/core).
+const liveblocksCore = path.resolve(
+  __dirname,
+  "../../packages/liveblocks-core/src/index.ts"
+);
+const liveblocksClient = path.resolve(
+  __dirname,
+  "../../packages/liveblocks-client/src/index.ts"
+);
```
marcbouchenoire: Not sure why we didn't run into this before/in other places?

packages/liveblocks-codemirror/src/version.ts
```diff
@@ -0,0 +1,4 @@
+export const PKG_NAME = "@liveblocks/codemirror";
+export const PKG_VERSION = "3.20.0";
```
marcbouchenoire: These shouldn't be hardcoded, they should be dynamically set at build time, ideally by `tsup`

packages/liveblocks-lexical/package.json
```diff
@@ -0,0 +1,104 @@
+{
+  "name": "@liveblocks/lexical",
+  "version": "3.20.0",
```
marcbouchenoire: We might need to sync these new packages' version numbers since existing packages got bumped from `main` multiple times

packages/liveblocks-lexical/styles.css
marcbouchenoire, line 1: These files (this one, the `.d.(c)ts` ones, and the `.map` one) should be in `.gitignore`

packages/liveblocks-prosemirror/src/plugin.ts
```diff
@@ -0,0 +1,526 @@
...
+  type ProseMirrorJsonNode,
+  stringifyDocument,
+} from "./schema";
+import { applyIncrementalOperations, classifyTransaction } from "./steps";
+import type { LiveblocksProsemirrorRoom } from "./types";
+
+export const LIVEBLOCKS_COLLABORATION_PLUGIN_KEY = new PluginKey<{
+  isReady: boolean;
+}>("liveblocks-collaboration");
+export const LIVEBLOCKS_TIPTAP_DOCUMENTS_KEY = "_tiptap_docs";
```
marcbouchenoire, line 27: Would it make sense to keep this ProseMirror-related? Even if Tiptap is the biggest consumer indirectly

packages/liveblocks-server/src/interfaces/IStorageDriver.ts
marcbouchenoire, line 1: Should we remove these changes (`liveblocks-server` and `liveblocks-cli`) since they will be mirrored from `liveblocks-backend` after merging?

packages/liveblocks-client/src/index.ts
```diff
@@ -88,6 +96,7 @@ export {
   LiveList,
   LiveMap,
   LiveObject,
+  LiveText,
```
marcbouchenoire, line 98: We should also re-export (some of) these from `@liveblocks/node`

marcbouchenoire (approved):
Amazing behemoth work! I left a few comments but mostly about packaging setups, examples, etc.

packages/liveblocks-core/src/crdts/__tests__/LiveText.convergence.fuzz.test.ts
nvie, line 1: I think the index space here is raw Unicode code units (`.length`), and I couldn't find a guard against landing an index inside a multi-byte emoji. If I read it right:

|                            | currently               | ideally          |
| -------------------------- | ----------------------- | ---------------- |
| `insert(1, "X")` into `😀` | `"\ud83dX\ude00"`       | `"X😀"`          |
| `delete(1, 1)` on `a😀b`   | `"a\ude00b"`            | `"ab"`           |
| `format(0, 1)` on `😀`     | half bold, two segments | whole emoji bold |

The first two produce lone surrogates, so invalid UTF-8 gets persisted and fanned out to everyone, corrupting the document.

The convergence property-based tests are genuinely good, but I noticed they never generate anything outside the simple ASCII range, so this will never show up there. In 0fab22f47a5c1d894a55b49cb2417590518ed304, I've added a couple of now-failing tests so we can more deliberately decide how we want to handle these edge cases to avoid document corruption on concurrent environments.

nvie (commented):
This is all looking pretty solid so far — really great work 😍! Because of that, I started looking into the more gnarly little edge cases.

I've extended the property-based tests a bit to find more edge cases (larger docs, higher variety, more weird cases), and fixed a bug where I could corrupt the document by formatting with `{ toString: false }` as an attribute.

shared/vitest-config/index.js
```diff
@@ -4,10 +4,20 @@ import { fileURLToPath } from "url";
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
+// Keep one core module identity when tests import source-only test helpers.
+const liveblocksCore = path.resolve(
+  __dirname,
+  "../../packages/liveblocks-core/src/index.ts"
+);
```
nvie, line 11: @marcbouchenoire Thank you for consolidating this ugly hack a bit!

The real Proper™ solution for these would of course be to have all those other packages no longer use the ugly mock server, but the real dev server instead. I wonder if that's now possible, or is still needed for a reason. The mock server should only still be used for things that require super precise timing that cannot be expressed with the dev server, which I would expect is only for edge cases in core itself, not for any higher-level packages.

It would be nice if we could get rid of this surprising behavior in tests, which I'm sure someone will get bitten by in the future.

marcbouchenoire: Agreed! I wasn't sure how far I should take this, but I agree it'd be better to not have it at all.

PR #3642, opened 2026-08-03
Improve version history components

marcbouchenoire:
This PR fixes and slightly improve version history components:

- `HistoryVersionSummary` is now always start-aligned (left or right depending on LTR or RTL)
- `HistoryVersionSummary` and `HistoryVersionPreview` now show “Unattributed” and “Unattributed edits” when a version doesn't have authors, and the docs have been updated to explain this and teach how to customize it via overrides

### DIFF
--- a/packages/liveblocks-react-ui/src/overrides.tsx
+++ b/packages/liveblocks-react-ui/src/overrides.tsx
@@ -126,8 +126,16 @@ export interface InboxNotificationOverrides {
   ) => ReactNode;
 }
 
+export interface HistoryVersionSummaryOverrides {
+  HISTORY_VERSION_SUMMARY_AUTHORS_LIST: (
+    list: ReactNode | undefined
+  ) => ReactNode;
+}
+
 export interface HistoryVersionPreviewOverrides {
-  HISTORY_VERSION_PREVIEW_AUTHORS_LIST: (list: ReactNode) => ReactNode;
+  HISTORY_VERSION_PREVIEW_AUTHORS_LIST: (
+    list: ReactNode | undefined
+  ) => ReactNode;
   HISTORY_VERSION_PREVIEW_RESTORE: string;
   HISTORY_VERSION_PREVIEW_EMPTY: ReactNode;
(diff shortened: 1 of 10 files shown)


packages/liveblocks-react-ui/src/overrides.tsx
```diff
@@ -243,9 +252,10 @@ export const defaultOverrides: Overrides = {
       {user} mentioned you{room ? <> in {room}</> : null}
     </>
   ),
-  HISTORY_VERSION_PREVIEW_AUTHORS_LIST: (list: ReactNode) => (
-    <>Edits from {list}</>
-  ),
+  HISTORY_VERSION_SUMMARY_AUTHORS_LIST: (list: ReactNode | undefined) =>
+    list === undefined ? "Unattributed" : list,
```
CTNicholas, line 256: I wonder if there's a better word for this. It's not quite _unattributed_, more that _there's literally no way to add a name_, our APIs don't allow it.

I think "System" might work better. Or "Automated update". Something like that.

I wonder if we should expand `mutateStorage` to allow passing a `userId`.

CTNicholas (commented):
Thanks for fixing this!

packages/liveblocks-react-ui/src/overrides.tsx
marcbouchenoire: Yeah I struggled to find a word that wouldn't be too opinionated for the end users, for example some customers might be using server-side updates on behalf of users and seeing "System"/"Server" could be confusing? Customers can customize it based on their own usage that we don't know about.

Agreed that this should happen but since it will be optional I kept this as we'll still have to support this 0 authors case.

CTNicholas: I see where you're coming from! I probably wouldn't go for "Server" as it sounds too technical.

Okay, lets leave it as it is. Your note in the docs is helpful, people can change the text if they wish!

CTNicholas (approved):

PR #3649, opened 2026-08-06
Prevent a stale thread refetch from reverting a settled resolve

### LINKED ISSUE #3636: Tiptap: resolved comment mark re-activates on incoming 408 thread-notification (store transiently reports resolved thread as unresolved)
## Summary
With `@liveblocks/react-tiptap` (Yjs + Comments), resolving a comment thread flickers
and can revert to unresolved. The reverting event is an inbound `408` thread-
notification WebSocket frame: on receiving it, the threads store *transiently* reports
the just/already-resolved thread as `resolved: false`, and the mark reconciliation in
`LiveblocksExtension` (`store.outputs.threads.subscribe` → `removeMark`/`addMark`)
re-activates the mark. It affects the in-editor mark AND the `FloatingThreads` card.

## Versions
`@liveblocks/react-tiptap`, `@liveblocks/react`, `@liveblocks/yjs`,
`@liveblocks/client` all `3.22.0`. React, custom `authEndpoint` (access tokens,
`room:write`). Single client (not a cross-client sync issue).

## Repro
1. Tiptap editor with `useLiveblocksExtension({ comments: true })`, `FloatingComposer`
   + `FloatingThreads`, one client.
2. Create a comment on a selection (real mouse — synthetic/dispatched clicks do NOT
   reproduce).
3. Resolve the thread. Then create + resolve more comments.
4. Observe: a resolved thread's mark reverts to highlighted/active (and its card
   reappears) when a `408` for it arrives — including `408`s triggered by unrelated
   new-comment activity. `mark-as-resolved` returns 200; the resolve did persist.

## Captured trace
+0ms     mark-as-resolved (start)         user resolves
+4ms     MARK resolved/orphan (cleared)   optimistic ✓
+444ms   mark-as-resolved 200
+760ms   ws:recv 408  <thread>            <-- clobber frame
+867ms   ws:send 301                       client re-writes the Yjs mark
+875ms   MARK re-added (orphan=false)
+2390ms  MARK highlighted/unresolved       <-- REVERT (sticks)

Also: forcing a `threads` REST read reconciles a stuck mark back to resolved (i.e. the
store eventually corrects, but the mark had already re-activated).

## Expected
An optimistic resolve should survive the server's own `408` thread-notification — the
mark decoration should not re-activate a thread the client resolved.

## Related
Looks like a surviving variant of the v2.7.2 fix ("Fixed a bug where resolved threads
remained visible in the editor").

## Workaround (FYI)
Patched the `LiveblocksExtension` reconciliation to subtract a client-side
"known-resolved" set from the store's unresolved set, so a resolved thread is never
re-activated regardless of `408` timing. Happy to share the patch.

VihaanAgarwal:
Fixes #3636

### Problem

Resolving a comment thread can flicker back to unresolved. The sequence:

1. `useMarkThreadAsResolved` adds an optimistic update, the REST call succeeds, and `patchThread` replaces the optimistic update with `resolved: true` on the cached thread. The patch keeps the thread's old `updatedAt`.
2. The server broadcasts a `THREAD_UPDATED` (408) frame, which triggers a refetch in `handleCommentEvent`. If that response is a stale snapshot from before the resolve (same `updatedAt`, `resolved: false`), `ThreadDB.upsertIfNewer` accepts it because the timestamps tie (`>=`), and the cached thread reverts to unresolved.
3. In `@liveblocks/react-tiptap`, the mark reconciliation in `LiveblocksExtension` sees the thread as unresolved again and re-adds the mark, so the thread re-activates in the editor and in `FloatingThreads`. This matches the trace in #3636: the clobbering 408 arrives a few hundred ms after `mark-as-resolved` returns 200.

### Fix

`patchThread` now stamps the patched thread with the mutation's `updatedAt`, the same way the `edit-thread-metadata` optimistic update and `editCommentMetadata` already do. Once the resolve settles, the cached thread carries the resolve timestamp, so a stale refetch from before the resolve fails the `upsertIfNewer` check and can't overwrite it. Genuinely newer server updates still apply as before.

`useMarkThreadAsUnresolved` and `useEditThreadMetadata` go through the same code path, so they get the same protection.

### Tests

Two new tests in `patchThread.test.ts` reproduce the sequence at the store level: seed a thread, settle a resolve (or unresolve), then apply a stale pre-mutation snapshot via `updateThreadifications`. Both fail on main (the thread flips back) and pass with this change. The full `@liveblocks/react` suite passes (315 tests). The two pre-existing `tsc` errors in `src/__tests__/index.test.tsx` reproduce on a clean checkout of main and are unrelated.

### DIFF
--- a/packages/liveblocks-react/src/__tests__/umbrella-store/patchThread.test.ts
+++ b/packages/liveblocks-react/src/__tests__/umbrella-store/patchThread.test.ts
@@ -1,0 +1,99 @@
+import { kInternal } from "@liveblocks/core";
+import { describe, expect, test } from "vitest";
+
+import { UmbrellaStore } from "../../umbrella-store";
+import { createComment, createThread } from "./_dummies";
+
+function makeSyncSource() {
+  return {
+    setSyncStatus: () => {},
+    destroy: () => {},
+  };
+}
+
+const NO_CLIENT = {
+  [kInternal]: {
+    as() {
+      return NO_CLIENT;
+    },
+    createSyncSource: makeSyncSource,
+  },
+} as any;
+
+describe("patchThread", () => {
+  test("a settled resolve survives a concurrent stale server refetch", () => {
+    const store = new UmbrellaStore(NO_CLIENT);
+
+    const thread = createThread({
+      id: "th_1",
+      roomId: "room_1",
+      createdAt: new Date("2024-01-01T00:00:00Z"),
+      updatedAt: new Date("2024-01-01T00:00:00Z"),
+      resolved: false,
+      comments: [
+        createComment({
+          threadId: "th_1",
+          roomId: "room_1",
+          createdAt: new Date("2024-01-01T00:00:00Z"),
+        }),
+      ],
+    });
+    store.updateThreadifications([thread], [], []);
+
+    // User resolves the thread; the REST call succeeds and the optimistic
+    // update is replaced by the real thing
+    const resolvedAt = new Date("2024-01-01T00:01:00Z");
+    const optimisticId = store.optimisticUpdates.add({
+      type: "mark-thread-as-resolved",
+      threadId: thread.id,
+      updatedAt: resolvedAt,
+    });
+    store.patchThread(thread.id, optimisticId, { resolved: true }, resolvedAt);
+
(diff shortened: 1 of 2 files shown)


marcbouchenoire (approved):
Thank you @VihaanAgarwal!

